PR #23, opened 2020-07-26
Fixes for various problems

### LINKED ISSUE #10: Dispatching of messages to target applications has high runtime overhead
The current approach is nicely decoupled but will result in high and varying runtime overhead.

See https://github.com/Snowfork/polkadot-ethereum/pull/5/files#r453730843

vgeddes:
Changes:
* Use a separate account store for PolkaETH balances. This makes the token fully operational now. Minting and transfer has been tested to work.
* Dispatch messages to applications using a static app registry (Closes #10)
* The WASM build of the runtime now compiles cleanly. 
  * Import `sp_std::prelude::*` in modules where replacements for libstd are required (`Vec` for example);
  * Use WASM-compatible hashing interface in bridge module (It's still blake256).
  * Fixup feature flags in all the pallet `Cargo.toml` files.
* Renamed instantiated `Balances2` module to `BalancesPolkaETH`
* Renamed instantiated `PolkaETH` module to `AppPolkaETH`
* Removed obsolete DummyApp since we have a PolkaETH app.


parachain/common/src/registry.rs
```diff
@@ -0,0 +1,18 @@
...
+
+pub struct Entry {
+    pub symbol: AppName,
+    pub id: AppID,
+}
+
+pub static REGISTRY: &'static [Entry] = &[
+    Entry {
+        symbol: AppName::PolkaETH,
+        id: [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
```
denalimarsh, line 16: How will we be determining an application's ID on polkadot? Might be confusing to have `[0, 0, 0...]` as an app ID, maybe we could hash the string "PolkaEth" into a 20-byte hex string so it's more similar to the `common.Address` datatype over on Ethereum?

Open to ideas on this one.

vgeddes: I agree, that needs to be improved. I still need to figure out the right types to use here (byte strings vs byte arrays, etc).

parachain/pallets/broker/src/lib.rs
```diff
@@ -45,7 +47,7 @@ decl_module! {
 		#[weight = 0]
 		pub fn accept(origin, app_id: AppID, message: Message) -> DispatchResult {
 			// we'll check the origin here to ensure it originates from a verifier
-			ensure_root(origin)?;
+			//ensure_root(origin)?;
```
philipstanislaus, line 50: Why has this been commented out?

parachain/runtime/src/lib.rs
```diff
@@ -285,16 +284,19 @@ impl dummy_verifier::Trait for Runtime {
 	type Scheduler = scheduler::Module<Runtime>;
 }
 
-impl dummy_app::Trait for Runtime {
-	type Event = Event;
-}
-
 impl balances::Trait<balances::Instance2> for Runtime {
-	type Balance = u128; // We'll need to consider using a BigNum (u256) for PolkaETH
+	type Balance = u128; // We'll need to use a BigNum (U256) for PolkaETH
```
philipstanislaus, line 288: Should we add a TODO here?

musnit: ye or github issue

parachain/pallets/broker/src/lib.rs
vgeddes: Oops, I commented that while testing submission of extrinsics from the relayer. Will push a fix.

parachain/runtime/src/lib.rs
vgeddes: There already is one from the last milestone: https://github.com/Snowfork/polkadot-ethereum/issues/18  :)

PR #24, opened 2020-07-26
Ethereum watcher

denalimarsh:
This PR implements the Ethereum watcher. I've commented out the Substrate chain for now.
- [x] Registry: registration of Ethereum applications
- [x] Streamer: starts subscriptions on registered applications for named event topic "AppEvent"
- [X] Router: constructs messages containing relevant RLP-encoded transaction data and verification data and relays them to the bridge

We can update the Registry once we've decided what the process for registering applications looks like. It's also easy to modify message contents should we want to add/remove information.

Next PR will add Keybase to the watcher and implement SignatureProver for the creation of verification data.


bridgerelayer/cmd/chains/ethereum/router.go
```diff
@@ -2,50 +2,75 @@ package ethereum
...
+	var buff bytes.Buffer
+	err := eLog.EncodeRLP(&buff)
+	if err != nil {
+		return types.Packet{}, err
+	}
 
-	// Message
-	message := types.NewMessage(txData, []byte{})
+	// Generate a proof by signing a hash of the encoded data
+	proof, err := prover.GenerateProof(buff.Bytes(), er.keybase.PrivateKey())
```
philipstanislaus, line 56: A general comment on private key handling: It would be more secure to just accept a `Signer` interface as opposed to a private key and likewise wrap the private key in a struct that only exposes the signer interface. That way, the consumer could never access any secret data. That prevents accidental exposure and prepares our codebase for HSMs/ledger devices and remote signing.

bridgerelayer/cmd/types/events.go
```diff
@@ -0,0 +1,23 @@
+package types
+
+import (
+	"github.com/ethereum/go-ethereum/common"
+	ctypes "github.com/ethereum/go-ethereum/core/types"
+)
+
+// EventName is a common event name required by each application
+const EventName = "AppEvent"
```
philipstanislaus, line 9: Have we decided that target applications need to emit that event?

I still believe that it would be very useful for the bridge to be able to bridge arbitrary events, i. e. we would need to register the events that need to be watched through the app registry.

philipstanislaus: Any thoughts on that @musnit ?

musnit: @philipstanislaus yeh topic for call tomorrow

vgeddes: @philipstanislaus hey, yeah, we were discussing that today, starting in this comment thread: https://github.com/Snowfork/polkadot-ethereum/pull/30#issuecomment-666412154

I'm definitely +1 on being able to bridge arbitrary events based on the ideas in that thread

PR #30, opened 2020-07-29
Use abi.encode instead of abi.encodePacked

denalimarsh:
This PR replaces abi.encodePacked with abi.encode.

Here's a guide about how encode works: https://medium.com/@libertylocked/what-are-abi-encoding-functions-in-solidity-0-4-24-c1a90b5ddce8. The changes in `test/test_bank.js` will be helpful for parsing the individual event variables from data. 

Sample SendETH event data:
```
Bank.AppEvent(
      _targetAppID: hex'746172676574617070313233' (type: bytes),
      _name: 'sendETH' (type: string),
      _data: hex'000000000000000000000000f17f52151ebef6c7334fad080c5704d77216b73200000000000000000000000000000000000000000000000000000000000000a0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000003782dace9d900000000000000000000000000000000000000000000000000000000000000000001000000000000000000000000000000000000000000000000000000000000003033386a34644735477a734c3162773255324156676579416b3651547871343356377a50626458416d62564c6a7644434b00000000000000000000000000000000' (type: bytes)
    )
```

Sample SendERC20 event data:
```
Bank.AppEvent(
      _targetAppID: hex'746f6b656e646578617070393837' (type: bytes),
      _name: 'sendERC20' (type: string),
      _data: hex'000000000000000000000000f17f52151ebef6c7334fad080c5704d77216b73200000000000000000000000000000000000000000000000000000000000000a00000000000000000000000002c2b9c9a4a25e24b174f26114e8926a9f2128fe400000000000000000000000000000000000000000000000000000000000000640000000000000000000000000000000000000000000000000000000000000001000000000000000000000000000000000000000000000000000000000000003033386a34644735477a734c3162773255324156676579416b3651547871343356377a50626458416d62564c6a7644434b00000000000000000000000000000000' (type: bytes)
    )
```

### DIFF
--- a/ethereum/contracts/Bank.sol
+++ b/ethereum/contracts/Bank.sol
@@ -67,6 +67,6 @@ contract Bank {
         pure
         returns(bytes memory)
     {
-        return abi.encodePacked(_sender, _recipient, _tokenAddr, _amount, _nonce);
+        return abi.encode(_sender, _recipient, _tokenAddr, _amount, _nonce);
     }
 }
--- a/ethereum/test/test_bank.js
+++ b/ethereum/test/test_bank.js
@@ -20,9 +20,8 @@ contract("Bank", function (accounts) {
   // TODO: implement on contract base58 conversion as recipient's encoded bytes vary in length
   const POLKADOT_ADDRESS2 = "1FRMM8PEiWXYax7rpS6X4XZX1aAAxSWx1CrKTyrVYhV24fg"
 
-  const ADDRESS_BYTE_LENGTH = 40;
-  const RECIPIENT_BYTE_LENGTH = 96;
-  const UINT256_BYTE_LENGTH = 64;
+  const RECIPIENT_BYTE_LENGTH = 128;
+  const BYTES32_LENGTH = 64;
 
   describe("Bank contract deployment", function () {
     beforeEach(async function () {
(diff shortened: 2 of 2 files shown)


vgeddes:
Thanks Denali.

This PR got me thinking about the overall structuring of the events, and the implications of packing opaque data into event logs using using `abi.encode` or any its variants.

I realize the Bank contract's `AppEvent` was supposed to mirror an equivalent `AppEvent`  on the Substrate side. Though after discussions with @philipstanislaus, I kinda dropped the generic `AppEvent` idea on the substrate side. I should have followed through and communicated those design changes to you.

`abi.encode` also has the downside of requiring that the custom-written decoder on the substrate side be kept in sync every time we change the format of whatever is encoded.

For phase 1, we're really only supporting ETH and ERC20 transfers, so instead of emitting a generic `AppEvent`, we could emit the the following events for ETH and ERC20 transfers respectively:

```
event LockupETH(address _sender, bytes _recipient, uint256 _amount, uint256 _nonce);
event LockupERC20(address _sender, bytes _recipient, address _tokenAddr, uint256 _amount, uint256 _nonce);
````

Note that the I haven't including the AppID in the event parameters. I'm not sure its needed. The relayer is in a better position to provide that when it relays the events, since it has a registry of all the registered smart contracts.

@musnit @philipstanislaus your thoughts on this?

**Comment by musnit** [comment]
@vgeddes I think we need to be building in a way to support generic state in future. We do need some proof of ability to support generic state in Phase 1 too. Using ETH/ERC20-specific events may be a bit redundant/discarded and not bring us any closer to being able to do that, and seems like it shoves the risk of how we will support generic state in an ideal way under the rug. One of our priorities does need to be that we resolve this risk. What are you thinking along that line?

We should perhaps set up a call for tomorrow or early next week to discuss this further if it is still an issue?

**Comment by vgeddes** [comment]
@musnit well I'm really focusing on the point that a generic `AppEvent` event is an obsolete artifact from our early design process. In the same way that on the Substrate side we're no longer emitting an `AppEvent` either, and instead emitting distinct events like `PolkaETH.Minted`,  `PolkaETH.Burned`, etc. For example, see https://github.com/Snowfork/polkadot-ethereum/blob/main/parachain/pallets/polkaeth-app/src/lib.rs#L48.

For generic state transfers we could just have the bank contract emit a distinct event like:
```
SendGeneric(sender, recipient, data)
```

Emitting distinct events is also good for debuggability, transparency and visibility. Ethereum users interacting with the bank contract can see distinct events being emitted, rather than opaque `AppEvent` events.

**Comment by musnit** [comment]
Yeh, I saw that - its also a concern. The contrast is that now we've shifted to a design in which the relayer (and likely the receiving bridge) becomes dependent on application-specific types/interfaces/functionality. You can see this in the relayer too here:  your bridge now becomes dependent on application code in some way. Eg: https://github.com/Snowfork/polkadot-ethereum/tree/94ecb3b157105f788784459e35aea12e6de50420/bridgerelayer/cmd/chains/ethereum/registry/applications -> the relayer is dependent on the sendEth application. cc @denalimarsh 

The main concerns from this are:
 - adding new applications/integrations becomes nontrivial, and requires relayers and potentially lower layer code to be updated too
 - verifying application-specific stuff in our light client may become a significant challenge too, and so there may be things we discover in Phase 2 that result in us having to rework this design significantly.

Having said that, given that it seems we're already a bit down this route, it may make sense to continue and accept the above risk, but we should discuss further.

**Comment by vgeddes** [comment]
> Having said that, given that it seems we're already a bit down this route, it may make sense to continue and accept the above risk, but we should discuss further. It sounds like you and @philipstanislaus also discussed this a bunch, so maybe there's additional context I'm missing.

Philip and I discussed a rough idea  (I'll send you the links to the group chats) - The substrate chain would somehow signal to the relayer which events it should be watching for. This would reduce the need the for the relayer to have hardcoded application-specific types. Likewise, the bank contract could signal to relayer which events it would need to watch out for. How this signalling would work out in practice has yet to be explored.

**Comment by vgeddes** [comment]
Here's one idea for how the relayer can dynamically determine which Substrate events to listen for:

On startup, the relayer could peform a storage query on the substrate bridge module. The response would be a list of event descriptors. For example:
```
PolkaETH.Minted(addr, amount)
PolkaETH.Burned(addr, amount)
PolkaETH.Transfer(addr, addr, amount)
PolkaERC20.Minted(token, addr, amount)
PolkaERC20.Burned(token, addr, amount)
PolkaERC20.Transfer(token, addr, addr, amount)
```

For this to work, we'll need to application pallets to call into the bridge module and add their events to the registry.

I'm wondering now if we can have the same scheme on the Ethereum side.

denalimarsh:
> Yeh, I saw that - its also a concern. The contrast is that now we've shifted to a design in which the relayer (and likely the receiving bridge) becomes dependent on application-specific types/interfaces/functionality. You can see this in the relayer too here: your bridge now becomes dependent on application code in some way. Eg: https://github.com/Snowfork/polkadot-ethereum/tree/94ecb3b157105f788784459e35aea12e6de50420/bridgerelayer/cmd/chains/ethereum/registry/applications -> the relayer is dependent on the sendEth application. cc @denalimarsh

Yes, by streaming events as opposed to transactions the application layer bleeds into the bridge layer. Ideally, every transaction is relayed across the bridge with the recipient application watching for transactions from a specific address and using ABI stored at the Substrate application-layer to parse the event data. I'm in favor of this approach and it underpinned some of my initial design decisions for M2. A potential downside I'm worried about is the load/cost of relaying each and every network tx when >0.0001% of them will be relevant to a receiving application on the other side.

The current registration scheme in the relayer is a temporary solution pending a decision on how we want to proceed.

> I'm wondering now if we can have the same scheme on the Ethereum side.

On the Ethereum side, the relayer still requires each application's deployed contract address and ABI in order to subscribe to a set of events.

Forcing each application to support named event `AppEvent` is a solution that **somewhat** mitigates the problem of application registration. If we force each application to support the same event topic e.g. `AppEvent(bytes message, bytes signature)` then the bridge could hypothetically only be aware of each application's ID. If we deploy a registry contract which allows for on-chain registration of contracts (for a price) then the bridge could automatically/dynamically support new apps by subscribing to the registry.

PR #38, opened 2020-07-31
Verifier contract with integrated Prover

denalimarsh:
This PR implements the Verifier contract. The feature optimizes for Golang signature generation, allowing operator-signed messages from the relayer to be verified in Solidity.

In JavaScript, the Web3 library automatically prefixes signed messages with `"\x19Ethereum Signed Message:\n32"` and rehashes them such that prefixing the message again within the Verifier contract is required. As the verification process differs between libraries, I've favored Golang as the more relevant current choice. The difference in on-chain signature validation requirements can be seen in Verifier.sol vs. VerifierWeb3.sol. In the future, the bridge will support Web3 signature generation, but it's well outside the requirements of this milestone.


ethereum/contracts/Verifier.sol
```diff
@@ -0,0 +1,61 @@
...
+
+        // If the version is correct return the signer address
+        if (v != 27 && v != 28) {
+            return (address(0));
+        } else {
+            // solium-disable-next-line arg-overflow
+            return ecrecover(h, v, r, s);
+        }
+    }
+}
```
vgeddes, line 61: Can you add trailing newlines to all the files in this PR that are missing them?

If you're using VS code, I suggest add the following config for an improved whitespace editing experience. For example, `files.insertFinalNewline` will automatically insert a final newline on save.
```
    "files.trimTrailingWhitespace": true,
    "files.insertFinalNewline": true,
    "files.trimFinalNewlines": true,
```

ethereum/contracts/Verifier.sol
```diff
@@ -0,0 +1,61 @@
...
+        returns (bool)
+    {
+        address signer = verify(_message, _signature);
+        return operator == signer;
+    }
+
+  /**
+     * @dev Verify checks if the signer's address matches the operator's address
+     */
+    function verify(bytes32 h, bytes memory signature)
```
philipstanislaus, line 24: The doc above does not match the return type here, right? 

It would make more sense if we called the logic that returns the address of the signer `recover` and the function that returns a bool if that's the operator `verify`. Is there a reason you swapped those?

ethereum/contracts/Verifier.sol
```diff
@@ -0,0 +1,61 @@
...
+        }
+
+        // Version of signature should be 27 or 28, but 0 and 1 are also possible versions
+        if (v < 27) {
+            v += 27;
+        }
+
+        // If the version is correct return the signer address
+        if (v != 27 && v != 28) {
+            return (address(0));
```
philipstanislaus, line 55: Should we not err here in that case?

ethereum/contracts/test/VerifierWeb3.sol
```diff
@@ -0,0 +1,73 @@
...
+        returns (bool)
+    {
+        address signer = verify(ethMessageHash(_message), _signature);
+        return operator == signer;
+    }
+
+  /**
+     * @dev Verify checks if the signer's address matches the operator's address
+     */
+    function verify(bytes32 h, bytes memory signature)
```
philipstanislaus, line 24: Same as above regarding function names and docs

ethereum/test/test_verifier.js
```diff
@@ -0,0 +1,40 @@
...
+      this.verifier = await Verifier.new(operator);
+    });
+
+    it("should correctly validate signatures", async function () {
+      const message =  web3.utils.soliditySha3("test123xyz~~")
+      const signature = await web3.eth.sign(message, operator);
+
+      // Recover a boolean indicating if the signature originates from the operator
+      const isOperator = await this.verifier.recover(message, signature);
+      isOperator.should.be.equal(true);
```
philipstanislaus, line 37: Please also add a test for the `verify` function. Thanks!

ethereum/contracts/Verifier.sol
denalimarsh: Added newlines in https://github.com/Snowfork/polkadot-ethereum/pull/38/commits/54c1a9a59441c2867dc8a2a1dc55793b09473901.

Thanks for the updated config!

ethereum/contracts/Verifier.sol
denalimarsh: Added errors in https://github.com/Snowfork/polkadot-ethereum/pull/38/commits/76840b1d8a23e9375df43b3ce592d353c5271f3a

ethereum/contracts/Verifier.sol
denalimarsh: I agree. The swap was not intentional. Updated in https://github.com/Snowfork/polkadot-ethereum/pull/38/commits/f2fe2d84c111628cbc300984e643275cbcc0c8dd

ethereum/contracts/test/VerifierWeb3.sol
denalimarsh: https://github.com/Snowfork/polkadot-ethereum/commit/f2fe2d84c111628cbc300984e643275cbcc0c8dd

ethereum/test/test_verifier.js
denalimarsh: https://github.com/Snowfork/polkadot-ethereum/pull/38/commits/ed50ef2d1eb91727cc86d0aa8050a332012c0874

denalimarsh:
I'm going to resolve conflicts with a merge instead of a rebase, just a heads up in case git history gets messy.

PR #40, opened 2020-08-04
Implement 1-way state transfer (ETH and ERC20)

vgeddes:
### Parachain
* Implement various decoders and tests for relayed message data
* Implement very simplistic ERC20 pallet: `parachain/pallets/polkaerc20-app/`
* Wire up ERC20 pallet into the chain runtime and application registry
### Solidity
* Updated Bank contract to use `bytes32` for recipient address on the parachain.
* Tweaked a few other types in the Bank contract.
* Updated truffle scripts `sendEth.js` for above changes


bridgerelayer/substrate/client.go
```diff
@@ -0,0 +1,131 @@
...
+
+	client := Client{
+		api, metadata,
+	}
+
+	return &client, nil
+}
+
+// SubmitExtrinsic submits a packet
+func (client *Client) SubmitExtrinsic(appID [32]byte, packet etypes.PacketV2) (bool, error) {
```
philipstanislaus, line 42: Let's call this 
```suggestion
func (client *Client) SubmitPacket(appID [32]byte, packet etypes.PacketV2) (bool, error) {
```
to distinguish from the general notion of an extrinsic. There might be others in the future.

bridgerelayer/substrate/client.go
```diff
@@ -0,0 +1,131 @@
...
+
+	appid := types.NewBytes32(appID)
+
+	payload, err := types.EncodeToBytes(packet)
+	if err != nil {
+		return false, err
+	}
+
+	// used for debugging
+	_err := ioutil.WriteFile("/tmp/packet.scale", payload, 0644)
```
philipstanislaus, line 62:
```suggestion
	err := ioutil.WriteFile("/tmp/packet.scale", payload, 0644)
```

philipstanislaus: Can we add a debug flag to the cli to trigger debug mode?

bridgerelayer/substrate/client.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	client := Client{
+		api, metadata,
+	}
+
+	return &client, nil
+}
+
+// SubmitExtrinsic submits a packet
```
philipstanislaus, line 41: Please document what bool return value indicates

bridgerelayer/substrate/client.go
```diff
@@ -0,0 +1,131 @@
...
+		return false, err
+	}
+
+	ext := types.NewExtrinsic(c)
+
+	era := types.ExtrinsicEra{IsMortalEra: false}
+
+	genesisHash, err := client.api.RPC.Chain.GetBlockHash(0)
+	if err != nil {
+		panic(err)
```
philipstanislaus, line 80: Why do we panic here instead of returning the error?

bridgerelayer/substrate/client.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	rv, err := client.api.RPC.State.GetRuntimeVersionLatest()
+	if err != nil {
+		panic(err)
```
philipstanislaus, line 85: And here?

bridgerelayer/substrate/client.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	key, err := types.CreateStorageKey(meta, "System", "Account", from.PublicKey, nil)
+	if err != nil {
+		panic(err)
```
philipstanislaus, line 90: And here?

bridgerelayer/substrate/client.go
```diff
@@ -0,0 +1,131 @@
...
+		SpecVersion: rv.SpecVersion,
+		TxVersion:   1,
+		Tip:         types.NewUCompactFromUInt(0),
+	}
+
+	extI := ext
+
+	err = extI.Sign(from, o)
+	if err != nil {
+		panic(err)
```
philipstanislaus, line 115: And here? And below? ;)

parachain/pallets/polkaerc20-app/src/tests.rs
```diff
@@ -0,0 +1,20 @@
...
+#[test]
+fn it_mints() {
+	new_tester().execute_with(|| {
+		let token_addr = H160::zero();
+		let alice: AccountId = Keyring::Alice.into();
+		assert_ok!(PolkaERC20::do_mint(token_addr, &alice, 500));
+
+		assert_eq!(FreeBalance::<MockRuntime>::get(&token_addr, &alice), 500);
+
+	});
```
philipstanislaus, line 19: Can we add a test case for `handle`?

parachain/pallets/polkaerc20-app/src/lib.rs
```diff
@@ -0,0 +1,127 @@
...
+	/// in the ethereum event.
+	fn make_account_id(data: &[u8]) -> T::AccountId {
+		T::AccountId::decode(&mut &data[..]).unwrap_or_default()
+	}
+
+	fn u128_to_balance(input: u128) -> Option<T::Balance>  {
+		input.try_into().ok()
+	}
+
+	pub fn do_mint(token_addr: H160, to: &T::AccountId, amount: T::Balance) -> DispatchResult {
```
philipstanislaus, line 86: Should this really be public?

parachain/primitives/ethereum/src/event.rs
```diff
@@ -0,0 +1,221 @@
...
+
+
+static EVENT_ABI: &'static ABIEvent = &ABIEvent {
+	signature: "AppEvent(uint256,bytes)",
+	inputs: &[
+		Param { kind: ParamKind::Uint(256), indexed: false },
+		Param { kind: ParamKind::Bytes, indexed: false },
+	],
+	anonymous: false
+};
```
philipstanislaus, line 16: In a future version, this should probably go into the app registry, right?

parachain/primitives/ethereum/src/signature.rs
```diff
@@ -0,0 +1,6 @@
+
+use sp_std::prelude::*;
+
+pub fn verify(_message: Vec<u8>, _signature: Vec<u8>) -> bool {
+	true
```
philipstanislaus, line 5: Please add a `//TODO` here

bridgerelayer/chains/ethereum/router.go
```diff
@@ -40,37 +46,36 @@ func (er Router) Route(eventData types.EventData) error {
...
 
 // SendPacket sends a tx data packet to the bridge
-func (er Router) sendPacket(packet types.Packet) error {
+func (er Router) sendPacket(appID [32]byte, packet types.PacketV2) error {
 	log.Info("Sending packet:\n", packet)
 
-	// Bridge.Send(packet.AppID, packet.Message)
+	client, err := substrate.NewClient()
+	if err != nil {
+		panic(err)
```
denalimarsh, line 76: I think it makes more sense to route messages through a single Substrate client that's initialized once when the Relayer is started instead of setting up a new client for each request. If the Substrate client is initialized once then it makes sense to use panic, but here we should be returning an error.

Returning the error is sufficient for now, we can refactor the Substrate client later.

bridgerelayer/go.mod
```diff
@@ -3,14 +3,16 @@ module github.com/snowfork/polkadot-ethereum/bridgerelayer
 go 1.13
 
 require (
-	github.com/centrifuge/go-substrate-rpc-client v1.1.1-0.20200428140716-e238735648d6
+	github.com/Snowfork/go-substrate-rpc-client v2.0.0-alpha.5.0.20200730202536-d47ecb76adac+incompatible
```
denalimarsh, line 6: Looks like we've got 3x `go-substrate-rpc-client` deps.
1. We're using both `github.com/Snowfork/go-substrate-rpc-client` and `github.com/snowfork/go-substrate-rpc-client`, if possible let's just use one version of the forked library in the Relayer.
2. See if `go mod tidy` removes the centrifuge grpc indirect dependency. If not, we can take a look at removing it in the next milestone by updating dependency imports in `Snowfork/go-substrate-rpc-client`.

bridgerelayer/substrate/client.go
```diff
@@ -0,0 +1,131 @@
...
+	etypes "github.com/snowfork/polkadot-ethereum/bridgerelayer/types"
+)
+
+// Client struct
+type Client struct {
+	api      *gsrpc.SubstrateAPI
+	metadata *types.Metadata
+}
+
+// NewClient foo
```
denalimarsh, line 21:
```suggestion
// NewClient is a new Substrate client
```

bridgerelayer/substrate/client.go
```diff
@@ -0,0 +1,131 @@
...
+func NewClient() (*Client, error) {
+
+	api, err := gsrpc.NewSubstrateAPI(config.Default().RPCURL)
+	if err != nil {
+		return nil, err
```
denalimarsh, line 26: Here we return the error, but on line 31 we panic. Returning an error allows us to gracefully handle an expected situation while panic shuts down the program. Generally, panic should be used for initialization/set up and any error which results in an unrecoverable state. Both error and panic are potentially valid depending on where initialization of the Substrate client takes place, but in any given context the error handling strategy should be consistent e.g. both panic or both return error.

bridgerelayer/substrate/client.go
denalimarsh: I think here we'll need `err = ioutil.WriteFile("/tmp/packet.scale", payload, 0644)` as err has already been declared.

parachain/common/src/registry.rs
```diff
@@ -2,17 +2,24 @@
...
 }
 
 pub struct Entry {
-    pub symbol: AppName,
-    pub id: AppID,
+	pub symbol: AppName,
+	pub id: AppID,
 }
 
 pub static REGISTRY: &'static [Entry] = &[
```
denalimarsh, line 14: On Ethereum the Bank contract is one application with two distinct events/tags. If my understanding is correct and this is an application registry (as opposed to an event registry) then the relationship between applications is not 1:1. Does that introduce any potential issues for bi-directional communication between applications?

parachain/pallets/polkaeth-app/src/mock.rs
```diff
@@ -26,7 +28,10 @@ impl_outer_event! {
     }
 }
 
-pub type AccountId = u64;
+pub type Signature = MultiSignature;
+
+pub type AccountId = <<Signature as Verify>::Signer as IdentifyAccount>::AccountId;
+// pub type AccountId = u64;
```
denalimarsh, line 34:
```suggestion
```

Remove legacy comment

vgeddes:
> 
> 
> Does `parachain/common/tests/fixtures/packet.scale` need to be committed?

Yes we do, its a test fixture used by `common::types::test::decode_packet`. The alternative is to convert it into a static hex string inlined in the rust code, but that would require some more work in order to convert it into that form. I could create a script to do that in a future PR.

parachain/common/src/registry.rs
vgeddes: I definitely still want a 1-1 correspondence between AppId's and the ethereum event tags (SendETH or SendERC20). I didn't manage to wire up that routing in this milestone though.

parachain/primitives/ethereum/src/event.rs
vgeddes: yeh, correct :)

I'll add a comment about that.

parachain/pallets/polkaerc20-app/src/lib.rs
```diff
@@ -0,0 +1,127 @@
...
+
+		// TODO: Rather implement From<DecodeError> for DispatchError
+		let sm = match SignedMessage::decode(&mut message.as_slice()) {
+			Ok(sm) => sm,
+			Err(_) => return Err(DispatchError::Other("Failed to decode message"))
+		};
+		
+		let event = match EthEvent::decode_from_rlp(sm.data) {
+			Ok(event) => event,
+			Err(_) => return Err(DispatchError::Other("Failed to decode message"))
```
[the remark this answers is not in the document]
vgeddes: Fixed in latest upload.

parachain/pallets/polkaeth-app/src/mock.rs
```diff
@@ -92,18 +97,18 @@ impl Trait for MockRuntime {
 pub type PolkaETHModule = Module<MockRuntime>;
 pub type BalancesPolkaETH = balances::Module<MockRuntime>;
 
-pub const ALICE: AccountId = 1;
-pub const BOB: AccountId = 2;
-pub const CAROL: AccountId = 3;
+// pub const ALICE: AccountId = 1;
+// pub const BOB: AccountId = 2;
+// pub const CAROL: AccountId = 3;
```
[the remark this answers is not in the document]
vgeddes: Fixed in latest upload.

parachain/pallets/polkaeth-app/src/mock.rs
vgeddes: Fixed in latest comment

parachain/pallets/polkaerc20-app/src/lib.rs
```diff
@@ -0,0 +1,127 @@
...
+
+pub trait Trait: system::Trait {
+	type Event: From<Event<Self>> + Into<<Self as system::Trait>::Event>;
+
+	type Balance: Parameter + Member + AtLeast32BitUnsigned + Default + Copy + Debug + MaybeSerializeDeserialize;
+}
+
+decl_storage! {
+	trait Store for Module<T: Trait> as PolkaERC20Map {		
+		pub FreeBalance: double_map hasher(twox_64_concat) H160, hasher(blake2_128_concat) T::AccountId => T::Balance;
```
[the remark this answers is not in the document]
vgeddes: Added a comment about that in the latest upload:

```
		// Free balances are represent as a doublemap: (TokenAddr, AccountId) -> Balance
		//
		// The choice of hashers was influenced by pallet-generic-asset, where free balances
		// are also represented using a StorageDouble with twox_64_concat and blake2_128_concat
		// hashers. So I'm assuming its a safe choice.
```

parachain/pallets/polkaeth-app/src/lib.rs
```diff
@@ -85,11 +90,7 @@ decl_module! {
 		fn mint(origin, to: T::AccountId,  amount: PolkaETH<T>) -> DispatchResult {
```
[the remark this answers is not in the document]
vgeddes: fixed in latest upload.

parachain/pallets/polkaeth-app/src/lib.rs
```diff
@@ -120,10 +121,54 @@ decl_module! {
...
+			Err(_) => return Err(DispatchError::Other("Failed to decode message"))
+		};
+
+		match event {
+			EthEvent::SendETH { sender, recipient, amount, nonce} => {
+				let to = Self::make_account_id(&recipient);
+				let amt = Self::u128_to_balance(amount.as_u128());
+				if let Some(value) = amt {
+					Self::do_mint(to, value);
+				}
```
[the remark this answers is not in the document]
vgeddes: I've added error handling for the `None` case. The whole story around the parsing and conversion of numbers needs to looked at in a subsequent PR, and so not too keen to add events right now while the design is still half-baked.

parachain/pallets/polkaerc20-app/src/lib.rs
```diff
@@ -0,0 +1,127 @@
...
+			Err(_) => return Err(DispatchError::Other("Failed to decode message"))
+		};
+			
+		match event {
+			EthEvent::SendERC20 { sender, recipient, token, amount, nonce} => {
+				let to = Self::make_account_id(&recipient);
+				let amt = Self::u128_to_balance(amount.as_u128());
+				if let Some(value) = amt {
+					Self::do_mint(token, &to, value)?;
+				}
```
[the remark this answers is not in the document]
vgeddes: I've added error handling for the `None` case. The whole story around the parsing and conversion of numbers needs to looked at in a subsequent PR, and so not too keen to add events right now while the design is still half-baked.

I'll rather try and aid debugging through unit tests which I need to add for `handle`.

parachain/pallets/polkaerc20-app/src/lib.rs
vgeddes: fixed in latest upload

bridgerelayer/substrate/client.go
vgeddes: no more panics in latest upload.

bridgerelayer/substrate/client.go
vgeddes: no more panics in latest upload.

bridgerelayer/substrate/client.go
vgeddes: latest version just returns an error. no more bool.

bridgerelayer/substrate/client.go
vgeddes: I've removed the debug code for now.

bridgerelayer/substrate/client.go
vgeddes: fixed in latest upload.

bridgerelayer/substrate/client.go
vgeddes: Yeh, the client code was a bit half-baked in that regard. I've fixed it in latest upload and there is no more panicking.

bridgerelayer/chains/ethereum/router.go
vgeddes: fixed in latest upload.

parachain/pallets/polkaerc20-app/src/tests.rs
vgeddes: Done. I've factored out most of the code in `handle` to a new function `handle_event`, and have added tests for it.

bridgerelayer/go.mod
vgeddes: Nice catch. The `go mod tidy` command didn't work, so I've cut a ticket to track: https://github.com/Snowfork/polkadot-ethereum/issues/43.

denalimarsh:
Please resolve any conflicts and merge https://github.com/Snowfork/polkadot-ethereum/pull/44 before merging to master. Thanks!

PR #51, opened 2020-08-11
Decode SCALE encoded compact unsigned integers in Solidity

denalimarsh:
Implements Scale.sol, a contract that implements decoding mechanisms for SCALE encoded Substrate data.
- Decode compact unsigned integers according to the specification in https://github.com/centrifuge/go-substrate-rpc-client/blob/40b1b01ef45829256dfba2028e2e134dcf736f15/scale/codec.go.
- Currently successfully decodes all integers in the range [0, 1073741824], but have noticed slight discrepancies with decoding some integers 4294967296 and above.
- Run test with `truffle test test/test_scale.js`

**Gas Expenditure Report**
- Decoding [0, 63] average gas: 22,312
- Decoding [64, 16383] average gas: 22,550
- Decoding [16384, 1073741823] average gas: 22,918
- Decoding [1073741823, 4503599627370496] average gas: 25,737

### DIFF
--- a/ethereum/contracts/Scale.sol
+++ b/ethereum/contracts/Scale.sol
@@ -1,0 +1,40 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.6.2;
+
+contract Scale {
+
+    event LogData(bytes _data);
+
+    // decodeAddress
+    function decodeAddress(bytes memory data)
+        public
+        pure
+        returns (bytes memory)
+    {
+        // TODO: do something with data
+        bytes memory decodedData = data;
+         return decodedData;
+    }
+
+    function decodeBytes(bytes memory data)
+        public
+        returns (bytes memory)
+    {
+        bytes memory decodedData = data;
+
+        bytes8 prefix = data[0];
+        if(prefix == keccak256(abi.encodePacked("0x04"))) {
+            emit LogData("check valid");
+        }
+
+        return decodedData;
+    }
+
+    function decodeUint64(bytes memory data) public {
+        // uint256 amount;
+        // uint256 nonce;
+        // {val: []byte{0xa8}, output: int64(168)},
+        // {val: []byte{0x15, 0x01}, output: int64(277)},
+
+    }
+}
--- a/ethereum/contracts/Scale.sol
+++ b/ethereum/contracts/Scale.sol
@@ -3,6 +3,7 @@ pragma solidity >=0.6.2;
 
 contract Scale {
 
+    // Read a byte at a specific index and return it as type uint8
     function readByteAtIndex(bytes memory data, uint8 index)
         public
         pure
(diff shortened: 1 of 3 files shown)


ethereum/contracts/Scale.sol
```diff
@@ -0,0 +1,86 @@
...
+            uint32 x1 = uint32(b) | uint32(b2) << 8;  // convert to little endian
+            uint32 x2 = x1 | uint32(b3) << 16;
+            uint32 x3 = x2 | uint32(b4) << 24;
+
+            x3 >>= 2;                                 // remove the last 2 mode bits
+            return uint256(x3);
+        } else if(mode == 3) {
+            uint8 l = b >> 2;                         // remove mode bits
+            require(l <= 63,                          // Max upper bound of 536 is (67 - 4)
+                    "Not supported: l>63 encountered when decoding a compact-encoded uint");
```
philipstanislaus, line 74: The biggest number u256 can hold is 32 bytes. I’d propose to just drop support for bigger numbers here and throw if l > 32

ethereum/contracts/Scale.sol
```diff
@@ -0,0 +1,86 @@
...
+            uint8[] memory buf = new uint8[](l+4);
+            for(uint8 i = 1; i < l+4+1; i++) {        // read bytes from data[1:l+4]
+                uint8 bNext = readByteAtIndex(data, i);
+                buf[i-1] = bNext;
```
philipstanislaus, line 78: Instead of reversing below, we could also fill from back, right?

```Suggestion
                buf[l+4-i] = bNext;
```

philipstanislaus (approved):

philipstanislaus:
Thanks for your work on this 🙏

ethereum/contracts/Scale.sol
denalimarsh: That's definitely the easiest solution. Will dropping support for bigger numbers come back to bite us in the future?

Another potential option may be spreading the value across multiple memory slots i.e. 2+ uint256 variables.

ethereum/contracts/Scale.sol
denalimarsh: Good idea, this will definitely save some gas at runtime.

ethereum/contracts/Scale.sol
philipstanislaus: It will bite someone I guess, but every solution I can think of increases the gas cost for us, hence I would limit support for now.

The cleanest approach would likely be a big int type, which supports arithmetic operations, but that’s clearly out of scope.

vgeddes (approved):
LGTM :+1:

We can hopefully use fixed encoding for u256 as discussed in the call today

ethereum/contracts/Scale.sol
denalimarsh: By rejecting all compact uints that take up greater than 32 bytes, all numbers that hit case 3 will be rejected. The smallest possible case 3 number is 1073741824 (decoded), which previously passed the test but is now rejected.

Should we just remove the case 3 logic?

ethereum/contracts/Scale.sol
denalimarsh: If we remove the compact uint case 3 logic then both `uint8ArrayToUint256` and `reverse` can be removed. However, this suggestion still applies to the new method `reverseBytes`.

Before this PR is merged I'll either address it or make an issue for it.

denalimarsh:
Asking for another round of reviews due to new functionality.

I implemented uint256 decoding, added uint256 tests, and addressed revisions (see comments, have a couple of questions for @philipstanislaus).  @vgeddes could you plug some encoded uint256s generated by our substrate code into the `uint256` unit test and make sure it works for our use case?

ethereum/contracts/Scale.sol
```diff
@@ -0,0 +1,124 @@
...
+
+      // Convert bytes (big endian) to little endian format
+    function bytesToLittleEndian(bytes memory arr)
+        internal
+        pure
+        returns (uint256)
+    {
+        uint256 number;
+        for(uint i = 0; i < arr.length; i++){
+            number = number+uint(uint8(arr[i]))*(2**(8*(arr.length-(i+1))));
```
vgeddes, line 51: Could we have spaces between binary operators to make the code less dense and improve readability?

ethereum/contracts/Scale.sol
```diff
@@ -0,0 +1,124 @@
...
+    // Decodes a SCALE encoded compact unsigned integer
+    function decodeUintCompact(bytes memory data)
+        public
+        pure
+        returns (uint256)
+    {
+        uint8 b = readByteAtIndex(data, 0);           // read the first byte
+        uint8 mode = b & 3;                           // bitwise operation
+
+        if(mode == 0) {                               // [0, 63]
```
vgeddes, line 90: According to the recommended style guide at https://solidity.readthedocs.io/en/v0.5.3/style-guide.html there should be a space between the `if` and the first parenthesis.

ethereum/contracts/Scale.sol
```diff
@@ -0,0 +1,124 @@
...
+            return b >> 2;                            // right shift to remove mode bits
+        } else if(mode == 1) {                        // [64, 16383]
```
vgeddes, line 92: same here

ethereum/contracts/Scale.sol
```diff
@@ -0,0 +1,124 @@
...
+            uint8 bb = readByteAtIndex(data, 1);      // read the second byte
+            uint64 r = bb;                            // convert to uint64
+            r <<= 6;                                  // multiply by * 2^6
+            r += b >> 2;                              // right shift to remove mode bits
+            return r;
+        } else if(mode == 2) {                        // [16384, 1073741823]
```
vgeddes, line 98: same here

ethereum/contracts/Scale.sol
```diff
@@ -40,6 +40,44 @@ contract Scale {
...
     }
 
+      // Convert bytes (big endian) to little endian format
+    function bytesToLittleEndian(bytes memory arr)
+        internal
+        pure
+        returns (uint256)
+    {
+        uint256 number;
+        for(uint i = 0; i < arr.length; i++){
```
philipstanislaus, line 50: Again, to save gas, we could just iterate backwards over the array instead of reversing it, right?

philipstanislaus (changes requested):
Nice, thanks!

ethereum/contracts/Scale.sol
philipstanislaus: Is the logic working now as expected? If it is still buggy, I would remove it. If it works, let's keep it in there.

denalimarsh: Still buggy, removed in https://github.com/Snowfork/polkadot-ethereum/pull/51/commits/e9e7231e103fa771989a28187874db29300be165. We can bring it back and resolve the issues when applicable.

ethereum/contracts/Scale.sol
denalimarsh: https://github.com/Snowfork/polkadot-ethereum/pull/51/commits/d3d5a2bc7a79b6949357ec8082aae20ce6b9c1b6

ethereum/contracts/Scale.sol
denalimarsh: https://github.com/Snowfork/polkadot-ethereum/pull/51/commits/d3d5a2bc7a79b6949357ec8082aae20ce6b9c1b6

ethereum/contracts/Scale.sol
denalimarsh: Noted, fixed in https://github.com/Snowfork/polkadot-ethereum/pull/51/commits/d3d5a2bc7a79b6949357ec8082aae20ce6b9c1b6

ethereum/contracts/Scale.sol
denalimarsh: Fixed in https://github.com/Snowfork/polkadot-ethereum/commit/d3d5a2bc7a79b6949357ec8082aae20ce6b9c1b6

ethereum/package.json
```diff
@@ -1,14 +1,17 @@
...
     "chai-as-promised": "^7.1.1",
     "chai-bignumber": "^3.0.0",
     "solc": "^0.6.12",
     "truffle": "^5.1.32",
-    "@truffle/contract": "^4.0.31",
     "web3": "^1.2.9",
     "web3-utils": "^1.2.9"
+  },
+  "devDependencies": {
+    "@polkadot/types": "^1.28.1"
```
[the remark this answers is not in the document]
denalimarsh: Removed in https://github.com/Snowfork/polkadot-ethereum/pull/51/commits/42ee4943545e90ff2c99159e5140d4cbc5cbe716.

ethereum/contracts/Scale.sol
denalimarsh: Created https://github.com/Snowfork/polkadot-ethereum/issues/56 for this. Small, but want to get this milestone functional and submitted before addressing some of these smaller optimizations.

ethereum/contracts/Scale.sol
denalimarsh: Created https://github.com/Snowfork/polkadot-ethereum/issues/56.

PR #54, opened 2020-08-13
Generic asset pallet and relayed events

### LINKED ISSUE #18: Use 256-bit BigNum for representing PolkaETH balances
See https://github.com/Snowfork/polkadot-ethereum/pull/15/files#r454413008

vgeddes:
These changes implement the following requirements for Milestone 4:
1. Burning of assets
2. Relaying of application events in a suitable form for the bridge relayer and ethereum smart contracts.

To align the implementations of the PolkaETH and PolkaERC20 applications, and share common behavior, I've introduced a new pallet `generic_asset`. It stores U256 balances for arbitrary assets identified by a `H160` hash.

Closes #18

### DIFF
--- a/parachain/pallets/bridge/src/lib.rs
+++ b/parachain/pallets/bridge/src/lib.rs
@@ -5,9 +5,11 @@ use frame_system::{self as system, ensure_signed};
 
 use sp_std::prelude::*;
 
+use codec::Encode;
+
 use sp_runtime::traits::Hash;
 
-use artemis_core::{AppID, Message, Broker, Bridge};
+use artemis_core::{AppID, Message, Broker, RelayEventEmitter};
 
 pub trait Trait: system::Trait {
 
@@ -28,7 +30,8 @@ decl_event!(
 		AccountId = <T as system::Trait>::AccountId,
 		Hash = <T as frame_system::Trait>::Hash,
 	{
-		MessageReceived(AccountId, AppID, Hash),
+		Received(AccountId, AppID, Hash),
+		Relayed(AppID, Vec<u8>),
 	}
 );
 
@@ -50,14 +53,17 @@ decl_module! {
 			let who = ensure_signed(origin)?;
 			T::Broker::submit(app_id, message.clone())?;
 			
-			Self::deposit_event(RawEvent::MessageReceived(who, app_id, T::Hashing::hash(message.as_ref())));
+			Self::deposit_event(RawEvent::Received(who, app_id, T::Hashing::hash(message.as_ref())));
 			Ok(())
 		}
 
 	}
 }
 
-impl<T: Trait> Bridge for Module<T> {
-
-	fn dummy() {}
+impl<T: Trait, K> RelayEventEmitter<K> for Module<T>
+	where T: Trait, K: Encode
+{
+	fn emit(app_id: &AppID, data: K) {
+		Self::deposit_event(RawEvent::Relayed(*app_id, data.encode()));
+	}
 }
--- a/parachain/pallets/generic-asset/src/lib.rs
+++ b/parachain/pallets/generic-asset/src/lib.rs
@@ -3,21 +3,14 @@
 /// Implementation for PolkaERC20 token assets
 ///
 use sp_std::prelude::*;
-use sp_std::{fmt::Debug, convert::TryInto};
 use sp_core::{H160, U256, RuntimeDebug};
-use sp_runtime::traits::{
-	CheckedAdd, MaybeSerializeDeserialize, Member,  AtLeast32BitUnsigned};
-use frame_system::{self as system};
+use frame_system::{self as system, ensure_signed};
 use frame_support::{
-	decl_error, decl_event, decl_module, decl_storage, Parameter,
+	decl_error, decl_event, decl_module, decl_storage,
 	dispatch::{DispatchResult, DispatchError},
-	storage::StorageDoubleMap
 };
 
-use codec::{Decode};
-
-use artemis_core::{AppID, Application, Message};
-use artemis_ethereum::{self as ethereum, SignedMessage};
+use codec::{Encode, Decode};
 
 #[cfg(test)]
 mod mock;
@@ -33,12 +26,12 @@ pub struct AccountData {
 }
 
 pub trait Trait: system::Trait {
-	type Event: From<Event<Self, I>> + Into<<Self as system::Trait>::Event>;
+	type Event: From<Event<Self>> + Into<<Self as system::Trait>::Event>;
 }
 
 decl_storage! {
 	trait Store for Module<T: Trait> as GenericAsset {
-		pub TotalIssuance: double_map hasher(blake2_128_concat) AssetId, hasher(blake2_128_concat) T::AccountId => U256;
+		pub TotalIssuance: map        hasher(blake2_128_concat) AssetId => U256;
 		pub Account:       double_map hasher(blake2_128_concat) AssetId, hasher(blake2_128_concat) T::AccountId => AccountData;
 	}
 }
@@ -50,11 +43,14 @@ decl_event!(
 	{
 		Burned(AssetId, AccountId, U256),
 		Minted(AssetId, AccountId, U256),
+		Transferred(AssetId, AccountId, AccountId, U256),
 	}
 );
 
 decl_error! {
 	pub enum Error for Module<T: Trait> {
+		/// Free balance got overflowed after transfer.
+		FreeTransferOverflow,
 		/// Total issuance got overflowed after minting.
 		TotalMintingOverflow,
 		/// Free balance got overflowed after minting.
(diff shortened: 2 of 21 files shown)


parachain/pallets/polkaerc20-app/src/lib.rs
```diff
@@ -3,60 +3,47 @@
...
 
 #[cfg(test)]
 mod mock;
 
 #[cfg(test)]
 mod tests;
 
-pub trait Trait: system::Trait {
-	type Event: From<Event<Self>> + Into<<Self as system::Trait>::Event>;
+pub const APP_ID: &[u8; 32] = &[1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0];
```
philipstanislaus, line 25: I’d there a reason why we use 1 here? We could also hash „PolkaERC20“ (or similar) to prevent conflicting ids, right?

parachain/pallets/polkaerc20-app/src/lib.rs
```diff
@@ -67,6 +54,14 @@ decl_module! {
 
 		fn deposit_event() = default;
 
+		#[weight = 0]
```
philipstanislaus, line 57: Please add a TODO so we don’t forget to adjust the weight.

parachain/pallets/bridge/src/lib.rs
```diff
@@ -50,14 +53,17 @@ decl_module! {
...
 
-impl<T: Trait> Bridge for Module<T> {
-
-	fn dummy() {}
+impl<T: Trait, K> RelayEventEmitter<K> for Module<T>
+	where T: Trait, K: Encode
+{
+	fn emit(app_id: &AppID, data: K) {
+		Self::deposit_event(RawEvent::Relayed(*app_id, data.encode()));
+	}
```
philipstanislaus, line 68: This is great 👍

parachain/pallets/polkaeth-app/src/lib.rs
```diff
@@ -6,57 +6,44 @@ use frame_system::{self as system, ensure_signed};
...
 
 #[cfg(test)]
 mod mock;
 
 #[cfg(test)]
 mod tests;
 
-pub type PolkaETH<T> =
-	<<T as Trait>::Currency as Currency<<T as system::Trait>::AccountId>>::Balance;
+pub const APP_ID: &[u8; 32] = &[0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0];
```
philipstanislaus, line 25: As above

parachain/pallets/polkaeth-app/src/lib.rs
```diff
@@ -67,42 +54,11 @@ decl_module! {
...
-			Ok(())
-		}
-
-		///
-		/// To initiate a transfer of PolkaETH to ETH, account holders must burn PolkaETH.
-		///
-		#[weight = 10_000]
-		fn burn(origin, amount: PolkaETH<T>, allow_death: bool) -> DispatchResult {
-			// TODO: verify origin
+		#[weight = 0]
```
philipstanislaus, line 57: As above

parachain/pallets/polkaeth-app/src/tests.rs
```diff
@@ -13,45 +23,36 @@ fn to_account_id(hexaddr: &str) -> [u8; 32] {
...
-			recipient: to_account_id("8eaf04151687736326c9fea17e25fc5287613693c912909cb226aa4794f26a48"),
-			amount: 1000.into(),
-			nonce: 1
-		};
+		let relay_event: RelayEvent<AccountId> = RelayEvent::Burned(bob.clone(), 20.into());
 
+		assert_eq!(MockEvent::bridge(bridge::RawEvent::Relayed(*APP_ID, relay_event.encode())), last_event());
 
-		assert_ok!(PolkaETHModule::handle_event(event));
-		assert_eq!(BalancesPolkaETH::free_balance(bob.clone()), 2000);
```
philipstanislaus, line 55: Can we re-add the assertion for Bob‘s free balance?

parachain/pallets/generic-asset/src/lib.rs
```diff
@@ -0,0 +1,145 @@
...
+decl_module! {
+
+	pub struct Module<T: Trait> for enum Call where origin: T::Origin {
+
+		type Error = Error<T>;
+
+		fn deposit_event() = default;
+
+		/// Transfer some liquid free balance to another account.
+		#[weight = 0]
```
philipstanislaus, line 75: As above

parachain/pallets/generic-asset/src/lib.rs
```diff
@@ -0,0 +1,145 @@
...
+		from: &T::AccountId,
+		to: &T::AccountId,
+		amount: U256)
+	-> DispatchResult {
+		if amount.is_zero() || from == to {
+			return Ok(())
+		}
+		<Account<T>>::try_mutate(asset_id, from, |from_account| -> DispatchResult {
+			<Account<T>>::try_mutate(asset_id, to, |to_account| -> DispatchResult {
+				from_account.free = from_account.free.checked_sub(amount).ok_or(Error::<T>::InsufficientBalance)?;
```
philipstanislaus, line 136: Do we not need to throw for unterflow  here?

parachain/pallets/generic-asset/src/tests.rs
```diff
@@ -0,0 +1,66 @@
...
+fn burn_should_raise_event() {
+	new_tester().execute_with(|| {
+		let token_addr = H160::zero();
+		let alice: AccountId = Keyring::Alice.into();
+		GenericAsset::do_mint(token_addr, &alice, 500.into()).unwrap();
+
+		GenericAsset::do_burn(token_addr, &alice, 20.into()).unwrap();
+		assert_eq!(TestEvent::generic_asset(RawEvent::Burned(token_addr, alice, 20.into())), last_event());
+	});
+}
```
philipstanislaus, line 66: Nice! 

Please also add tests for:

1. All the errors the module can throw
2. Transfers

parachain/pallets/polkaerc20-app/src/lib.rs
vgeddes: No specific reason. Though while we're on the topic, I'm not sure why we still need AppID's to be 256 bits. We originally thought we needed the whole 256 bits so we could stuff a contract address into it. But that hasn't been necessary to make our PolkaERC20 app work. In fact currently AppIDs are not used at all in the current parachain.

vgeddes:
> 
> On a general note - is there a specific reason why events need to be emitted through the bridge? As on Ethereum, can’t the relayer listen to arbitrary events?
>

Well I guess it depends on what you mean by arbitrary events. Considering the following interpretations:
1. Each app emits its own generic `RelayEvent(Vec<u8>)` instead of going through the bridge.
2. Each app emits more specific events such as `Burned(Address, Amount)`, etc.

If you mean (1), is there a downside to emitting the generic event through the bridge? As doing so simplifies the architecture a fair bit (The relayer need only listen for a single event). I think you previously voiced a concern about proper accounting of storage costs. Though I'm not familiar with the economic implications of that.

If you mean (2), well in our 2nd last concall we discussed arbitrary events vs a single generic event, and we agreed on the latter. The problem with arbitrary events is that we'd need to ensure event ABIs are kept in-sync across all components of our system, including the bridge relay.

parachain/pallets/generic-asset/src/lib.rs
vgeddes: InsufficientBalance sounds more descriptive about what's actually happening. And similar modules such as Balances and GenericAsset throw InsufficientBalance too.

parachain/pallets/generic-asset/src/tests.rs
vgeddes: done

philipstanislaus:
> > On a general note - is there a specific reason why events need to be emitted through the bridge? As on Ethereum, can’t the relayer listen to arbitrary events?
> 
> Well I guess it depends on what you mean by arbitrary events. Considering the following interpretations:
> 
> 1. Each app emits its own generic `RelayEvent(Vec<u8>)` instead of going through the bridge.
> 2. Each app emits more specific events such as `Burned(Address, Amount)`, etc.
> 
> If you mean (1), is there a downside to emitting the generic event through the bridge? As doing so simplifies the architecture a fair bit (The relayer need only listen for a single event). I think you previously voiced a concern about proper accounting of storage costs. Though I'm not familiar with the economic implications of that.
> 
> If you mean (2), well in our 2nd last concall we discussed arbitrary events vs a single generic event, and we agreed on the latter. The problem with arbitrary events is that we'd need to ensure event ABIs are kept in-sync across all components of our system, including the bridge relay.

I meant (2), but might have misunderstood our conversation. My takeaway was that the relayer does not modify the original event (such that it can be light-client verified), so it would not need ABI information, but the ABI would be registered for decoding on both chains. Did I misunderstand that? (Cc @musnit @denalimarsh)

musnit:
@vgeddes yep, as philip said in (2) sounds correct to me. give me a call if you wanna chat further on the topic.

parachain/pallets/polkaerc20-app/src/lib.rs
philipstanislaus: Happy to shorten it, which length makes sense to you? 64 bits?

vgeddes: We'll tackle AppId's in M5.

parachain/pallets/polkaeth-app/src/lib.rs
vgeddes: Looking to tackle this in M5.

parachain/pallets/polkaerc20-app/src/lib.rs
musnit: @vgeddes please add github issue or trello task

parachain/pallets/polkaeth-app/src/lib.rs
musnit: @vgeddes please add github issue or trello task

PR #57, opened 2020-08-25
EthereumApp, ERC20App data decoding and asset unlocking

denalimarsh:
- [x] Refactor Bank contract to EthereumApp contract and ERC20App contract
- [x] Decode data packages for EthereumApp and ERC20App

Git history is a bit messy.


ethereum/contracts/EthereumApp.sol
```diff
@@ -0,0 +1,84 @@
...
+        address payable recipient = _data.sliceAddress(32);
+        // Deocde amount int256
+        bytes memory amountBytes = _data.slice(32 + 20, 32);
+        uint256 amount = amountBytes.decodeUint256();
+
+        sendETH(recipient, amount);
+        emit Unlock(sender, recipient, amount);
+    }
+
+    function sendETH(
```
vgeddes, line 72: I see 2 `sendETH` functions in this contract?

ethereum/contracts/ERC20App.sol
```diff
@@ -73,37 +56,27 @@ contract Bank {
...
-            unlockERC20(recipient, token, amount);
-        }
-    }
-
-    function unlockETH(
-        address payable _recipient,
-        uint256 _amount
-    )
-        internal
+    function unlockERC20(bytes memory _data)
```
vgeddes, line 59: We need to add a signature field as per the milestone requirements. I'm submitting a signature along with the data in my relayer PR: https://github.com/Snowfork/polkadot-ethereum/pull/58/files#diff-6f08a58ca4132c1dcf83cb21d2575b5eR129

vgeddes: Can we also rename the function to `submit` to match what the relayer would call?

ethereum/contracts/EthereumApp.sol
```diff
@@ -0,0 +1,84 @@
...
+    )
+        internal
+        pure
+        returns(bytes memory)
+    {
+        return abi.encode(_sender, _recipient, _tokenAddr, _amount, _nonce);
+    }
+
+
+    function unlockETH(bytes memory _data)
```
vgeddes, line 55: As with the ERC20 app we'll need to submit a signature too. And can we rename the function to `submit` to match what the relayer calls?

ethereum/contracts/ERC20App.sol
vgeddes: So the method signature should be `submit(bytes memory _data, bytes memory _signature)`. Thoughts?

ethereum/contracts/EthereumApp.sol
vgeddes: So the method signature should be `submit(bytes memory _data, bytes memory _signature)`. Thoughts?

ethereum/contracts/ERC20App.sol
vgeddes: I've made this change in the integration branch at https://github.com/Snowfork/polkadot-ethereum/pull/59

ethereum/contracts/EthereumApp.sol
vgeddes: I've made this change in the integration branch at https://github.com/Snowfork/polkadot-ethereum/pull/59

ethereum/contracts/Decoder.sol
```diff
@@ -0,0 +1,85 @@
...
+            uint256 otherIndex = arr.length - i - 1;
+            arr[i] = arr[otherIndex];
+            arr[otherIndex] = current;
+        }
+        return arr;
+    }
+
+    // Slice a section from a byte array.
+    // Inspired by github.com/GNSPS/solidity-bytes-utils/blob/master/contracts/BytesLib.sol.
+    function slice(bytes memory _bytes, uint256 _start, uint256 _length) internal pure returns (bytes memory) {
```
philipstanislaus, line 36: What is the reason for implementing such a low level function? Why not just use the slice literal? https://ethereum.stackexchange.com/questions/78559/how-can-i-slice-bytes-strings-and-arrays-in-solidity 

If we want to keep this logic, could you also copy all the comments from github.com/GNSPS/solidity-bytes-utils/blob/master/contracts/BytesLib.sol for context? Thanks!

ethereum/contracts/ERC20App.sol
philipstanislaus: No strong opinion, but to me `submit` is much less clear than `unlock...`. Just my 2 cents ;)

ethereum/contracts/ERC20App.sol
```diff
@@ -52,19 +56,27 @@ contract ERC20App {
         return abi.encode(_sender, _recipient, _tokenAddr, _amount, _nonce);
     }
 
-    function submit(bytes memory data)
+    function unlockERC20(bytes memory _data)
```
philipstanislaus, line 59: How do we handle auth for this function? Presently, anyone can call it, right?

ethereum/contracts/ERC20App.sol
```diff
@@ -52,19 +56,27 @@ contract ERC20App {
...
-         uint256 amount = 1;                     // placeholder
+        require(_data.length == 104, "Data must contain 104 bytes for a successful decoding");
+
+        // Decode sender bytes
+        bytes memory sender = _data.slice(0, 32);
+        // Decode recipient address
+        address recipient = _data.sliceAddress(32);
+        // Decode token address
+        address tokenAddr = _data.sliceAddress(32 + 20);
+        // Deocde amount int256
```
philipstanislaus, line 71:
```suggestion
        // Decode amount int256
```

ethereum/contracts/EthereumApp.sol
```diff
@@ -47,18 +51,25 @@ contract EthereumApp {
...
-         address payable recipient = address(0); // placeholder
-         uint256 amount = 1;                     // placeholder
+        require(_data.length == 84, "Data must contain 84 bytes for a successful decoding");
 
-        unlockETH(recipient, amount);
+        // Decode sender bytes
+        bytes memory sender = _data.slice(0, 32);
+        // Decode recipient address
+        address payable recipient = _data.sliceAddress(32);
+        // Deocde amount int256
```
philipstanislaus, line 64:
```suggestion
        // Decode amount int256
```

ethereum/contracts/EthereumApp.sol
philipstanislaus: See above

ethereum/contracts/EthereumApp.sol
```diff
@@ -47,18 +51,25 @@ contract EthereumApp {
         return abi.encode(_sender, _recipient, _tokenAddr, _amount, _nonce);
     }
 
-     function submit(bytes memory data)
+
+    function unlockETH(bytes memory _data)
```
philipstanislaus, line 55: Also see above regarding auth for this function

ethereum/contracts/ERC20App.sol
denalimarsh: Auth should be restricted to the Broker contract, which checks the signature before calling these functions. A downside of this approach is that the Broker would need to be registered on the app as well.

ethereum/contracts/Decoder.sol
denalimarsh: Because index range access is only supported for dynamic calldata arrays and copying calldata to memory is not implemented in the version of Solidity we're using. Once copying calldata to memory is updated we'll be able to use the slice literal. Open to suggestions on this one, this was primarily just a workable solution for M4.

Related: https://github.com/ethereum/solidity/issues/7929, https://github.com/ethereum/solidity/issues/8210

denalimarsh:
Re https://github.com/Snowfork/polkadot-ethereum/pull/57#discussion_r479502090, https://github.com/Snowfork/polkadot-ethereum/pull/57#discussion_r479505173 - the upside here is that 'submit' facilitates a generalized relayer as each transaction has the same function signature and only the receiving application's contract address changes. Using `unlock` as the standardized function name would accomplish the same thing, but in the future not all applications will necessarily be unlocking tokens.

ethereum/contracts/ERC20App.sol
denalimarsh: See https://github.com/Snowfork/polkadot-ethereum/pull/57#issuecomment-683683514

philipstanislaus:
> Re [#57 (comment)](https://github.com/Snowfork/polkadot-ethereum/pull/57#discussion_r479502090), [#57 (comment)](https://github.com/Snowfork/polkadot-ethereum/pull/57#discussion_r479505173) - the upside here is that 'submit' facilitates a generalized relayer as each transaction has the same function signature and only the receiving application's contract address changes. Using `unlock` as the standardized function name would accomplish the same thing, but in the future not all applications will necessarily be unlocking tokens.

Makes sense 👍

ethereum/contracts/Decoder.sol
philipstanislaus: Okay, could you add that comment in code? Thanks!

ethereum/contracts/ERC20App.sol
philipstanislaus: Okay, but the logic to restrict access to the broker is not here currently, is it? 

Could you add a TODO in code? Thanks!

vgeddes:
I'm closing this since its already been merged in via the integration PR #59. Task #66 tracks addressing the feedback.

PR #67, opened 2020-09-03
Add attribution for keybase code

denalimarsh:
The keybase module is heavily influenced by ChainSafe's ChainBridge keypair code. This PR adds attribution in the form of copyright information copied directly from the original source to the relevant files. The library is licensed under [LGPL-3.0-only](https://spdx.org/licenses/LGPL-3.0-only.html), meaning that it can be used commercially and does not need to be distributed under the LGPL-3.0-only license.

Copying the copyright/license directly may not be the best approach here, should we remove/add to some of the attribution sections?


vgeddes (approved):
LGTM!

philipstanislaus (approved):
Thanks 👍

PR #69, opened 2020-09-07
Refactor relayer for improved quality

vgeddes:
This work includes many different refactorings, including:
* Improved design/implementation of ethereum and substrate chains
* Uniform message-passing between chain implementations
* Better error-handling and goroutine management using `context.Context` and `errgroup` packages.

Each chain now has the following constructs, loosely based on the latest ChainBridge code:
* Connection: Encapsulates connection-level information and state. Used by listeners and writers.
* Listener: Reads events from the chain and forwards them to a message channel
* Writer: Reads events from a message channel and writes them to the chain via transactions

Other changes:
* To allow for better security, secrets are now loaded via environment variables instead of plaintext config files. In the future we should retrieve secrets from an encrypted vault (such as Hashicorp Vault or AWS Secrets Manager, but thats not critical right now.
  * `ARTEMIS_RELAY_ETHEREUM_KEY`
  * `ARTEMIS_RELAY_SUBSTRATE_KEY`
* The files in the `keybase` module have been renamed back to their original form as in the ChainBridge repo.
* Introduced a toplevel application object `core.Relay` which manages all the chains and their lifecycle and errors.
* Use [Mage](https://github.com/magefile/mage) instead of `Make` for build tasks. See [rationale](https://github.com/magefile/mage#why).
* Use [Revive](https://github.com/mgechev/revive) for linting. This a configurable linter that allows us to disable the annoying "Exported thing must have a comment or be unexported". I don't particularly like that rule as I don't believe it leads to better or more useful comments. Our packages are all used internally and are not being published, so in practice we don't really need to add filler comments just to pass the lint.

I've added a sprinkling of tests. Especially for the Connection and Writer components. However I didn't write tests for the listener components since we'll totally rewrite those from M6 onwards.


bridgerelayer/chain/ethereum/chain.go
```diff
@@ -0,0 +1,95 @@
...
+
+	"github.com/snowfork/polkadot-ethereum/bridgerelayer/chain"
+	"golang.org/x/sync/errgroup"
+
+	"github.com/sirupsen/logrus"
+	"github.com/snowfork/polkadot-ethereum/bridgerelayer/crypto/secp256k1"
+	"github.com/spf13/viper"
+)
+
+// EthChain streams the Ethereum blockchain and routes tx data packets
```
denalimarsh, line 15:
```suggestion
// Chain streams the Ethereum blockchain and routes tx data packets
```

bridgerelayer/chain/ethereum/listener.go
```diff
@@ -0,0 +1,96 @@
...
+func (li *Listener) pollEvents(ctx context.Context) error {
+	li.log.Info("Polling started")
+	events := make(chan gethTypes.Log)
+	for _, app := range li.apps {
+		query := makeQuery(app)
+		_, err := li.conn.client.SubscribeFilterLogs(ctx, query, events)
+		if err != nil {
+			li.log.WithFields(logrus.Fields{
+				"address": app.ID,
+			}).Error("Failed to subscribe to application events")
```
denalimarsh, line 54:
```suggestion
			}).Error("Failed to subscribe to application events")
			continue
```

bridgerelayer/chain/ethereum/listener.go
```diff
@@ -0,0 +1,96 @@
...
+		} else {
```
denalimarsh, line 55:
```suggestion
		}
```
Remove else

bridgerelayer/chain/substrate/chain.go
```diff
@@ -0,0 +1,102 @@
...
+
+	"golang.org/x/sync/errgroup"
+
+	"github.com/sirupsen/logrus"
+	"github.com/snowfork/polkadot-ethereum/bridgerelayer/chain"
+	"github.com/snowfork/polkadot-ethereum/bridgerelayer/crypto/sr25519"
+	"github.com/spf13/viper"
+)
+
+// Chain ...
```
denalimarsh, line 15: Makes sense to add proper method descriptions in this PR

bridgerelayer/chain/substrate/chain.go
```diff
@@ -0,0 +1,102 @@
...
+
+	// Generate keypair from secret
+	kp, err := sr25519.NewKeypairFromSeed(secret, "")
+	if err != nil {
+		return nil, err
+	}
+
+	conn := NewConnection(endpoint, kp.AsKeyringPair(), log)
+
+	listener := NewListener(
```
denalimarsh, line 54:
```suggestion
	listener := NewListener(conn, subMessages, blockRetyLimit, blockRetryInterval, log)
```
Nit: any reason this is broken up onto multiple lines? The line length seems manageable.

bridgerelayer/chain/substrate/connection.go
```diff
@@ -0,0 +1,65 @@
...
+	"context"
+
+	"github.com/sirupsen/logrus"
+
+	gsrpc "github.com/snowfork/go-substrate-rpc-client"
+	"github.com/snowfork/go-substrate-rpc-client/signature"
+	"github.com/snowfork/go-substrate-rpc-client/types"
+)
+
+// Connection ...
```
denalimarsh, line 13: ditto on method descriptions, the `...` was a placeholder meant to updated on the next significant refactor (i.e. this PR)

bridgerelayer/chain/substrate/listener.go
```diff
@@ -0,0 +1,162 @@
...
+		encoder := scale.NewEncoder(buf)
+		encoder.Encode(evt.AccountID)
+		encoder.Encode(evt.Recipient)
+		encoder.Encode(evt.TokenID)
+		encoder.Encode(evt.Amount)
+
+		li.messages <- chain.Message{AppID: chain.Erc20AppID, Payload: buf.Bytes()}
+	}
+
+	for _, evt := range events.ETH_Transfer {
```
denalimarsh, line 151: This means the bridge layer must be updated for any unique events in new applications, correct? The current structure is elegant for our use-case but down the road will face generalization issues. I suppose this isn't an issue as the listener will be refactored once we're listening for blocks instead of events.

bridgerelayer/chain/substrate/writer.go
```diff
@@ -0,0 +1,118 @@
...
+					"error": err,
+				}).Error("Failure submitting message to substrate")
+			}
+		}
+	}
+}
+
+// SubmitPacket submits a packet, it returns true
+func (wr *Writer) Write(msg *chain.Message) error {
+
```
denalimarsh, line 55: Opinion: standardize function newlines. Some methods do not have a newline and some do.

bridgerelayer/core/relay.go
```diff
@@ -0,0 +1,90 @@
...
+			}).Error("Failed to start chain")
+			return
+		}
+		log.WithField("name", chain.Name()).Info("Started chain")
+	}
+
+	// Wait until a fatal error or signal is raised
+	if err := eg.Wait(); err != nil {
+		if !errors.Is(err, context.Canceled) {
+			log.WithField("error", err).Error("Encountered an unrecoverable failure")
```
denalimarsh, line 82: The first letter of error strings shouldn't be capitalized, see https://github.com/golang/go/wiki/CodeReviewComments#error-strings. It may be the case that logrus is handling it for us, but I'm generally in favor of following well-established golang patterns.

bridgerelayer/magefile.go
```diff
@@ -0,0 +1,15 @@
+//+build mage
+
+package main
+
+import (
+	"github.com/magefile/mage/sh"
```
denalimarsh, line 6: Wondering why we've switched to Magefile instead of Makefile?

denalimarsh (commented):
Great work, this code is much cleaner. Did a first pass, will take another look once we've added some basic tests. Looks somewhat similar to ChainSafe's ChainBridge architecture, which makes sense as the Relayer also uses event-based subscription at the moment.

Concerned about how some decisions impact future generalization, but that's not really blocking as this refactor seems to be primarily targeted at polishing up the current codebase for project delivery.

bridgerelayer/magefile.go
vgeddes: yeah I'm not a fan of using Makefiles as a means of running arbitrary tasks. Along the lines of what the Mage author said, they can end up being a confusing tangle of Bash scriptlets with interspersed with Make syntax. At least that was my experience on previous projects.

Also, Make isn't trivially installable on Windows, which reduces cross-platform compatibility.

bridgerelayer/chain/substrate/listener.go
vgeddes: Yeah, this is indeed a problem in the current implementation, where events are transferred instead of blocks. Due to the way events are encoded in Substrate, the relayer must unfortunately have full knowledge of the parachain ABI (i.e the ABI of all the events that could possibly be emitted).

Hopefully when we transfer block headers instead of events this problem will go away.

bridgerelayer/core/relay.go
vgeddes: Thanks, fixed a few cases throughout the code. Though according to that doc its ok for logged messages to use a capital, so I left those alone.

bridgerelayer/chain/ethereum/listener.go
vgeddes: fixed in latest upload

bridgerelayer/chain/ethereum/listener.go
vgeddes: fixed in latest upload.

bridgerelayer/chain/substrate/chain.go
vgeddes: Well actually, what I've been working towards is actually not having comments like that. In the latest upload I've removed some redundant filler comments.

I'm now using [Revive](https://github.com/mgechev/revive) for linting. This a configurable linter that allows us to disable the annoying "Exported thing must have a comment or be unexported". I don't particularly like that rule as I don't believe it leads to better or more useful comments. Our packages are all being used internally and are not being published, so names are not really being exported in the full sense of the word.

Plus we're going to rewrite parts of the relay substantially, and having to add meaningless comments for exports will just add more work.

bridgerelayer/chain/substrate/writer.go
vgeddes: Fixed in latest upload.

bridgerelayer/chain/substrate/chain.go
denalimarsh: That's reasonable. In this context, I agree that the comments are somewhat meaningless. Will it raise any concerns during the client's code review? Can probably get around that if it comes up by explaining the thought process here.

denalimarsh (approved):
Changes LGTM

bridgerelayer/chain/substrate/chain.go
```diff
@@ -0,0 +1,101 @@
...
+
+type Chain struct {
+	listener *Listener
+	writer   *Writer
+	conn     *Connection
+}
+
+const Name = "Substrate"
+
+func NewChain(ethMessages chan chain.Message, subMessages chan chain.Message) (*Chain, error) {
```
philipstanislaus, line 26: Can we be more specific with argument names and channel types here? 

```suggestion
func NewChain(msgsToSend <-chan chain.Message, msgsObserved chan<- chain.Message) (*Chain, error) {
```

philipstanislaus (commented):
This is great, simplifies the architecture quite a bit 👍

bridgerelayer/chain/substrate/listener.go
```diff
@@ -0,0 +1,160 @@
...
+		li.log.Info("Handling transfer event")
+
+		buf := bytes.NewBuffer(nil)
+		encoder := scale.NewEncoder(buf)
+		encoder.Encode(evt.AccountID)
+		encoder.Encode(evt.Recipient)
+		encoder.Encode(evt.Amount)
+
+		li.messages <- chain.Message{AppID: chain.EthAppID, Payload: buf.Bytes()}
+	}
```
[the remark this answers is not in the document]
vgeddes: Good catch. I've cut this issue so we can track it: https://github.com/Snowfork/polkadot-ethereum/issues/80

bridgerelayer/chain/ethereum/listener.go
```diff
@@ -0,0 +1,100 @@
...
+			li.log.WithFields(logrus.Fields{
+				"address": app.ID,
+			}).Error("Failed to subscribe to application events")
+			continue
+		}
+
+		li.log.WithFields(logrus.Fields{
+			"address": app.ID,
+		}).Info("Subscribed to application events")
+	}
```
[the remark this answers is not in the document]
vgeddes: I've opened #80 to track this

bridgerelayer/chain/substrate/chain.go
vgeddes: Added #101 to track this.

PR #71, opened 2020-09-07
Fix segfault in relayer

vgeddes:
The fix was to make the dummy verifier not use the `scheduler` pallet for submitting the message back to the broker. This ensures that the scheduler events are not emitted from the parachain. The relayer was unable to read these events properly.


philipstanislaus (approved):
Could you add a comment in code to clarify why we do that? Thanks!

PR #72, opened 2020-09-07
Truffle scripts should connect to localhost:9545

vgeddes:
Truffle develop binds to `localhost:9545` by default so we should update our development truffle scripts accordingly.

### DIFF
--- a/ethereum/scripts/getERC20Balance.js
+++ b/ethereum/scripts/getERC20Balance.js
@@ -19,7 +19,7 @@ module.exports = async () => {
             return
         }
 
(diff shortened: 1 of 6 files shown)


vgeddes:
@denalimarsh what port does your `truffle develop` instance listen on? On Ubuntu, Truffle v5.1.X listens on 9545, so we should try and find a common solution.

philipstanislaus (approved):

denalimarsh (approved):
`truffle develop` starts at 9545 for me, these changes are correct

PR #74, opened 2020-09-08
Change relayer license to LGPL

vgeddes:
We can't incorporate LGPL-licensed [ChainBridge](https://github.com/ChainSafe/ChainBridge) code into an Apache-licensed project. The terms of the license state that a modified version of a work must also be made available under the LGPL.

As such we should relicense under the LGPL.


denalimarsh (approved):
Good catch. Wasn't aware that LGPL doesn't play nicely with other licenses 😢

musnit:
@vgeddes is apache not compatible/a superset of lgpl? anyway seems fine

musnit (approved):

vgeddes:
> 
> 
> @vgeddes is apache not compatible/a superset of lgpl? anyway seems fine

I did some reading and it seems Apache-licensed code can be incorporated into an LGPL codebase. However the reverse is not true, which is why we're having to relicense.

PR #75, opened 2020-09-09
Updated Solidity contracts for M5

denalimarsh:
- Application contract: an abstract contract containing methods which all Applications must implement.
- Broker contract: application registration + message routing (including signature verification via Verifier).
- Verifier contract: can now verify signatures against the message's original raw bytes. In the client, the operator signs a hashed, prefixed message (to mimic the behavior of web3.eth.sign); the changes allow the Verifier to recreate the signed hash for verification.

This PR was developed for the relayer architecture in https://github.com/Snowfork/polkadot-ethereum/pull/69, not the current relayer in master.

TODO: 
- [ ] Smart contract tests
- [ ] Integration with refactored relayer

Marking R4R as this may impact the design considerations of other components.

### DIFF
--- a/ethereum/contracts/Application.sol
+++ b/ethereum/contracts/Application.sol
@@ -3,5 +3,5 @@ pragma solidity >=0.6.2;
 
 // Application contains methods that all applications must implement
 abstract contract Application {
-    function submit(bytes memory _data) public virtual;
+    function handle(bytes memory _message) public virtual;
 }
--- a/ethereum/contracts/Application.sol
+++ b/ethereum/contracts/Application.sol
@@ -3,5 +3,5 @@ pragma solidity >=0.6.2;
 
 // Application contains methods that all applications must implement
 abstract contract Application {
-    function submit(bytes memory _data) public virtual;
+    function handle(bytes memory _data) public virtual;
 }
--- a/ethereum/contracts/Application.sol
+++ b/ethereum/contracts/Application.sol
@@ -1,0 +1,7 @@
+ // SPDX-License-Identifier: MIT
+pragma solidity >=0.6.2;
+
+// Application contains methods that all applications must implement
+abstract contract Application {
+    function submit(bytes memory _data) public virtual;
+}
--- a/ethereum/contracts/Application.sol
+++ b/ethereum/contracts/Application.sol
@@ -3,5 +3,5 @@ pragma solidity >=0.6.2;
 
 // Application contains methods that all applications must implement
 abstract contract Application {
-    function handle(bytes memory _message) public virtual;
+    function submit(bytes memory _data) public virtual;
 }
--- a/ethereum/contracts/Bridge.sol
+++ b/ethereum/contracts/Bridge.sol
@@ -7,6 +7,9 @@ import "./Application.sol";
 contract Bridge {
     using Decoder for bytes;
 
+    // 0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470 is returned for accounts without code
+    bytes32 NON_CONTRACT_ACCOUNT_HASH = 0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470;
+
     uint64 public mostRecentBlock;
     mapping(uint64 => uint64) public mostRecentBlockEvent;
     mapping(address => bool) public applications;
@@ -57,18 +60,12 @@ contract Bridge {
         view
         returns(bool)
     {
-        if(applications[_appID]){
-            return false;
-        }
+        require(!applications[_appID], "Application is already registered");
 
-        // 0x0 is the value returned for not-yet created accounts.
-        // 0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470 is returned for accounts without code.
-        bytes32 accountHash = 0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470;
         bytes32 codehash;
         assembly { codehash := extcodehash(_appID) }
-        if(codehash != 0x0 && codehash != accountHash) {
-            return false;
-        }
+        require(codehash != 0x0, "There's no account for this address on the network");
+        require(codehash != NON_CONTRACT_ACCOUNT_HASH, "Only contract accounts can be registered as applications");
 
         return true;
     }
--- a/ethereum/contracts/Bridge.sol
+++ b/ethereum/contracts/Bridge.sol
@@ -10,13 +10,10 @@ contract Bridge {
     // 0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470 is returned for accounts without code
     bytes32 NON_CONTRACT_ACCOUNT_HASH = 0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470;
 
-    uint64 public mostRecentBlock;
-    mapping(uint64 => uint64) public mostRecentBlockEvent;
+    mapping(bytes32 => bool) public eventHashes;
     mapping(address => bool) public applications;
 
-
     constructor(address[] memory apps) public {
-        mostRecentBlock = 0;
         for(uint256 i = 0; i < apps.length; i++) {
             if(verifyApp(apps[i])) {
                 applications[apps[i]] = true;
@@ -27,28 +24,20 @@ contract Bridge {
     /**
      * @dev routes the message to the specified application ID after verifying the operator's signature
      * @param _message address _message expected type: Message { AppID [32]byte, Payload []byte }
-     * @param _verification bytes _verification expected type: Verification { BlockNumber int64, EventID int64 }
+     * @param _hash bytes _hash is a unique hash of the event (app_id, payload, blockNum, eventId)
      */
-    function handle(bytes memory _message, bytes memory _verification)
+    function handle(bytes memory _message, bytes32 _hash)
         public
     {
-        uint64 blockNumber = uint64(_verification.slice(0, 8).decodeUint256());
-        require(blockNumber >= mostRecentBlock, "Blocks must be processed chronologically");
-
-        uint64 eventID = uint64(_verification.slice(8, 16).decodeUint256());
-        require(eventID > mostRecentBlockEvent[blockNumber], "Events must be processed chronologically");
-
         address appID = _message.sliceAddress(32);
         require(applications[appID], "App ID not found. Only registered application are supported");
+
+        require(!eventHashes[_hash], "Event has already been processed");
+        eventHashes[_hash] = true;
 
         bytes memory payload = _message.slice(32, _message.length-1);
         Application app = Application(appID);
         app.handle(payload);
-
-        mostRecentBlockEvent[blockNumber] = eventID;
-        if(blockNumber > mostRecentBlock) {
-            mostRecentBlock = blockNumber;
-        }
     }
 
     /**
--- a/ethereum/contracts/Bridge.sol
+++ b/ethereum/contracts/Bridge.sol
@@ -39,8 +39,8 @@ contract Bridge {
         require(!eventHashes[hashed], "Event has already been processed");
         eventHashes[hashed] = true;
 
-        // TODO: slice correct data
-        bytes memory payload = _message.slice(32, _message.length-1);
+        // Slice payload, dropping the 14-byte verification data
+        bytes memory payload = _message.slice(0, _message.length-15);
         Application app = Application(_appId);
         app.handle(payload);
     }
--- a/ethereum/contracts/Bridge.sol
+++ b/ethereum/contracts/Bridge.sol
@@ -26,7 +26,7 @@ contract Bridge {
      * @param _message address _message expected type: Message { AppID [32]byte, Payload []byte }
      * @param _hash bytes _hash is a unique hash of the event (app_id, payload, blockNum, eventId)
      */
-    function handle(bytes memory _message, bytes32 _hash)
+    function submit(bytes memory _message, bytes32 _hash)
         public
     {
         address appID = _message.sliceAddress(32);
@@ -37,7 +37,7 @@ contract Bridge {
 
         bytes memory payload = _message.slice(32, _message.length-1);
         Application app = Application(appID);
-        app.submit(payload);
+        app.handle(payload);
     }
 
     /**
--- a/ethereum/contracts/Bridge.sol
+++ b/ethereum/contracts/Bridge.sol
@@ -4,7 +4,7 @@ pragma solidity >=0.6.2;
 import "./Decoder.sol";
 import "./Application.sol";
 
-contract Broker {
+contract Bridge {
     using Decoder for bytes;
 
     uint64 public mostRecentBlock;
--- a/ethereum/contracts/Bridge.sol
+++ b/ethereum/contracts/Bridge.sol
@@ -37,7 +37,7 @@ contract Bridge {
 
         bytes memory payload = _message.slice(32, _message.length-1);
         Application app = Application(appID);
-        app.handle(payload);
+        app.submit(payload);
     }
 
     /**
--- a/ethereum/contracts/Bridge.sol
+++ b/ethereum/contracts/Bridge.sol
@@ -3,6 +3,7 @@ pragma solidity >=0.6.2;
 
 import "./Decoder.sol";
 import "./Application.sol";
+import "./Verifier.sol";
 
 contract Bridge {
     using Decoder for bytes;
@@ -12,31 +13,35 @@ contract Bridge {
 
     mapping(bytes32 => bool) public eventHashes;
     mapping(address => bool) public applications;
+    Verifier verifier;
 
-    constructor(address[] memory apps) public {
-        for(uint256 i = 0; i < apps.length; i++) {
-            if(verifyApp(apps[i])) {
-                applications[apps[i]] = true;
+    constructor(address _verfierAddr, address[] memory _apps) public {
+        verifier = Verifier(_verfierAddr);
+        for(uint256 i = 0; i < _apps.length; i++) {
+            if(verifyApp(_apps[i])) {
+                applications[_apps[i]] = true;
             }
         }
     }
 
     /**
      * @dev routes the message to the specified application ID after verifying the operator's signature
-     * @param _message address _message expected type: Message { AppID [32]byte, Payload []byte }
-     * @param _hash bytes _hash is a unique hash of the event (app_id, payload, blockNum, eventId)
+     * @param _appId address _appId
+     * @param _message bytes _message contains information to be passed to the application
      */
-    function submit(bytes memory _message, bytes32 _hash)
+    function submit(address _appId, bytes memory _message)
         public
     {
-        address appID = _message.sliceAddress(32);
-        require(applications[appID], "App ID not found. Only registered application are supported");
+        require(verifier.verifyOperator(), "Tx origin must be the Bridge operator");
+        require(applications[_appId], "App ID not found. Only registered application are supported");
 
-        require(!eventHashes[_hash], "Event has already been processed");
-        eventHashes[_hash] = true;
+        bytes32 hashed = keccak256(_message);
+        require(!eventHashes[hashed], "Event has already been processed");
+        eventHashes[hashed] = true;
 
+        // TODO: slice correct data
         bytes memory payload = _message.slice(32, _message.length-1);
-        Application app = Application(appID);
+        Application app = Application(_appId);
         app.handle(payload);
     }
 
--- a/ethereum/contracts/Bridge.sol
+++ b/ethereum/contracts/Bridge.sol
@@ -30,7 +30,7 @@ contract Bridge {
         public
     {
         uint64 blockNumber = uint64(_verification.slice(0, 8).decodeUint256());
-        require(blockNumber > mostRecentBlock, "Blocks must be processed chronologically");
+        require(blockNumber >= mostRecentBlock, "Blocks must be processed chronologically");
 
         uint64 eventID = uint64(_verification.slice(8, 16).decodeUint256());
         require(eventID > mostRecentBlockEvent[blockNumber], "Events must be processed chronologically");
--- a/ethereum/contracts/Bridge.sol
+++ b/ethereum/contracts/Bridge.sol
@@ -49,7 +49,7 @@ contract Bridge {
         view
         returns(bool)
     {
-        require(!applications[_appID], "Application is already registered");
+        require(applications[_appID], "Application is already registered");
 
         bytes32 codehash;
         assembly { codehash := extcodehash(_appID) }
--- a/ethereum/contracts/Broker.sol
+++ b/ethereum/contracts/Broker.sol
@@ -1,0 +1,59 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.6.2;
+
+import "./Verifier.sol";
+import "./Decoder.sol";
+import "./Application.sol";
+
+contract Broker {
+    using Decoder for bytes;
+
+    mapping(address => bool) public applications;
+    Verifier public verifier;
+
+    constructor(address verifierAddr) public {
+        verifier = new Verifier(verifierAddr);
+    }
+
+    /**
+     * @dev registers new applications
+     * @param _appID address _appID is the contract address to be registered as a supported application
+     */
+    function registerApp(address _appID)
+        public
+    {
+        require(!applications[_appID], "Application is already registered");
+
+        // 0x0 is the value returned for not-yet created accounts.
+        // 0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470 is returned for accounts without code.
+        bytes32 accountHash = 0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470;
+        bytes32 codehash;
+        assembly { codehash := extcodehash(_appID) }
+        require(codehash != 0x0 && codehash != accountHash, "Application must be a contract deployed to this network");
+
+        // TODO: Consider checking that the application implements the abstract Application method 'submit' according
+        //       to EIP-165 (incomplete implementation: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-165.md)
+
+        applications[_appID] = true;
+    }
+
+    /**
+     * @dev routes the message to the specified application ID after verifying the operator's signature
+     * @param _data address _data expected type: Message { AppID [32]byte, Payload []byte }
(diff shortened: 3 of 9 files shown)


ethereum/contracts/Application.sol
```diff
@@ -0,0 +1,7 @@
+ // SPDX-License-Identifier: MIT
+pragma solidity >=0.6.2;
+
+// Application contains methods that all applications must implement
+abstract contract Application {
+    function submit(bytes memory _data) public virtual;
```
vgeddes, line 6: To prevent devs and users confusing this method with the `submit` method on the Bridge/Broker contract, I'd rather call it `handle`.

```suggestion
    function handle(bytes memory _data) public virtual;
```
This is what we do on the Substrate side:
https://github.com/Snowfork/polkadot-ethereum/blob/main/parachain/pallets/polkaerc20-app/src/lib.rs#L110

ethereum/contracts/Broker.sol
```diff
@@ -0,0 +1,59 @@
...
+
+    constructor(address verifierAddr) public {
+        verifier = new Verifier(verifierAddr);
+    }
+
+    /**
+     * @dev registers new applications
+     * @param _appID address _appID is the contract address to be registered as a supported application
+     */
+    function registerApp(address _appID)
```
vgeddes, line 22: We shouldnt need this, since the broker should hardcode all the applications its knows about. Having a public `registerApp` function also implies that there is some trusted authority who is able to register apps. Our design at least tries to be trustless.

For example, on the Substrate side, the broker uses this hardcoded list of applications: https://github.com/Snowfork/polkadot-ethereum/blob/main/parachain/primitives/core/src/registry.rs

ethereum/contracts/ERC20App.sol
```diff
@@ -56,9 +57,9 @@ contract ERC20App {
         return abi.encode(_sender, _recipient, _tokenAddr, _amount, _nonce);
     }
 
-    function submit(bytes memory _data, bytes memory _signature)
+    function submit(bytes memory _data)
```
vgeddes, line 60: In line with my previous comment above:
```suggestion
    function handle(bytes memory _message)
```

ethereum/contracts/EthereumApp.sol
```diff
@@ -3,8 +3,9 @@ pragma solidity >=0.6.2;
 
 import "@openzeppelin/contracts/math/SafeMath.sol";
 import "./Decoder.sol";
+import "./Application.sol";
 
-contract EthereumApp {
+contract EthereumApp is Application {
```
vgeddes, line 8:
```suggestion
contract ETHApp is Application {
```

ethereum/contracts/EthereumApp.sol
```diff
@@ -51,9 +52,9 @@ contract EthereumApp {
         return abi.encode(_sender, _recipient, _tokenAddr, _amount, _nonce);
     }
 
-
-    function submit(bytes memory _data, bytes memory _signature)
+    function submit(bytes memory _data)
```
vgeddes, line 55:
```suggestion
    function handle(bytes memory _message)
```

ethereum/contracts/Broker.sol
```diff
@@ -0,0 +1,59 @@
...
+
+        applications[_appID] = true;
+    }
+
+    /**
+     * @dev routes the message to the specified application ID after verifying the operator's signature
+     * @param _data address _data expected type: Message { AppID [32]byte, Payload []byte }
+     * @param _signature address _signature expected type: Signature []byte
+     */
+    function submit(bytes memory _data, bytes memory _signature)
```
vgeddes, line 45: Looks like relayer will call this method to submit messages to the Ethereum side.

On the Substrate side, things are bit different. We have a Bridge module that accepts messages from the Relayer, and then forwards those messages to the Broker module. If we want to follow that, then I'd suggest the following structure:

Bridge Contract:
```
function send(address _app_id, bytes memory _message) {
  emit MessageReceived(keccak_hash(_message))
  broker.submit(address _app_id, bytes memory _message)
}
```

Broker Contract:
```
function submit(address _app_id, bytes memory _message) {
  // dispatch to target app
}
```

But whether we want a Bridge contract is up for debate. Though it would be nice to mirror the substrate side. @musnit any thoughts from your side?

vgeddes, line 45: Also the `_signature` field in `submit` was just a hack/convenience to get M4 out the door. On the substrate side, the signature is part of the data/message bytes:

In rough pseudocode:
```
message {
  []byte payload
  []byte signature
}
```

We'll still need to think of a suitable encoding scheme for Message, be it RLP or Scale, etc.

vgeddes (changes requested):
This is a good start. Added a few comments around these main concerns:
* Whether broker should have hardcoded registry of applications
* Whether we should mirror Substrate side and introduce a Bridge contract

ethereum/contracts/Broker.sol
musnit: i think for simplicity we should keep it mirrored for now. if we want to modify in future for other reasons, can do so, but would prefer for it to be symmetrical and clean at least initially if there are no side effects to that.

musnit: what is the signature here for?

vgeddes: Its to mirror the substrate side, and implement signature-based verification of messages. For example, on the Substrate side, this what the relayer sends to substrate:
https://github.com/Snowfork/polkadot-ethereum/blob/main/parachain/primitives/ethereum/src/message.rs#L15

I know @denalimarsh has implemented a verifier contract though I haven't had a proper chance to review it yet.

Of course post-M5 we'll be using light-client based verification, so we'll end up throwing this signature code away.

musnit: @vgeddes right but on the substrate side, is this SignedMessage used by the broker? or only by other modules?

vgeddes: On the Substrate side, things are still bit messy and ad-hoc. While I've implemented the primitives to verify signed messages, the verifier module isn't actually using those primitives yet. Currently, after no-op dummy verification, the broker submits an opaque Message to an application. The application then decodes that opaque message to a SignedMessage. Fixing all that on my TODO list for M5.

I suppose the ideal approach on both chains would be the following sequence:

* Bridge receives `OpaqueMessage`
* Bridge forwards `OpaqueMessage` to Broker
* Broker forwards `OpaqueMessage` to Verifier
* Verifier decodes `OpaqueMessage` into a ` SignedMessage { data, signature }`.
* Verifier verifies message, strips off the signature, and submits `VerifiedMessage { data }` back to the broker
* Broker dispatches `VerifiedMessage { data }` to target application

musnit:
yeh something like this sounds right

musnit:
same should apply for the ethereum side too ideally

ethereum/contracts/Broker.sol
denalimarsh: This feels like an arbitrary waste of gas.  Calling a function on another contract and passing it the payload increases total gas cost. Why not just emit the event on the Broker contract or include `function send` on the Broker contract?

As gas isn't a priority right now I'm happy to add the Bridge contract to mirror the substrate side, but we should be aware of the implicit consequences.

musnit: would it be worth considering the inverse, ie, is it up for debate whether we even want a Bridge **pallet**?

ethereum/contracts/EthereumApp.sol
denalimarsh: https://github.com/Snowfork/polkadot-ethereum/pull/75/commits/31a105e146a248e1e665be92dab4405052e072b7

ethereum/contracts/EthereumApp.sol
denalimarsh: https://github.com/Snowfork/polkadot-ethereum/pull/75/commits/31a105e146a248e1e665be92dab4405052e072b7

ethereum/contracts/ERC20App.sol
denalimarsh: https://github.com/Snowfork/polkadot-ethereum/pull/75/commits/31a105e146a248e1e665be92dab4405052e072b7

ethereum/contracts/Application.sol
denalimarsh: https://github.com/Snowfork/polkadot-ethereum/pull/75/commits/31a105e146a248e1e665be92dab4405052e072b7

ethereum/contracts/Broker.sol
denalimarsh: https://github.com/Snowfork/polkadot-ethereum/pull/75/commits/b7b0e9cd067a333b960ddca54ef0b5c88e1967c6

ethereum/contracts/Broker.sol
vgeddes: Come to think of it, all the bridge pallet really does is forward the message to the broker. So yeah I wouldn't mind removing it. Though we should then rename the Broker contract to Bridge.

I'm up for it.

vgeddes: Currently my the latest parachain refactor has the following message flow:
```
Bridge:
  send(msg):
    broker.submit(msg)
    
Broker:
  submit(msg):
    if verifier.verify(msg) == true
      self.dispatch_to_app(msg)
```

If we ditch the old Bridge contract, and rename things, then the flow would look like this:
```
Bridge:
  submit(msg):
    if verifier.verify(msg)
      self.dispatch_to_app(msg)
```

denalimarsh: Removed signature-based verification in the Broker contract and replaced with chronological-based verification using block number and event hash in https://github.com/Snowfork/polkadot-ethereum/pull/75/commits/58850b4f6d0450ebd23b31d7deabe839608c37d2, https://github.com/Snowfork/polkadot-ethereum/pull/75/commits/f9121d56afc6b7403efaffe93c6a2d3b9a7813e5

denalimarsh: In anticipation of this change on the Substrate side I've renamed Broker.sol to Bridge.sol in https://github.com/Snowfork/polkadot-ethereum/pull/75/commits/f5c731cfa11dfda704bca49a681da2e857c20d64

ethereum/contracts/Bridge.sol
```diff
@@ -0,0 +1,75 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.6.2;
+
+import "./Decoder.sol";
+import "./Application.sol";
+
+contract Bridge {
+    using Decoder for bytes;
+
+    uint64 public mostRecentBlock;
```
vgeddes, line 10: To support pluggable verification modules, we want the verifier contract to handle replay protection. This allows the bridge to support any kind of verification in future. So I'd move this code to the verifier contract.

ethereum/contracts/Bridge.sol
```diff
@@ -0,0 +1,75 @@
...
+            }
+        }
+    }
+
+    /**
+     * @dev routes the message to the specified application ID after verifying the operator's signature
+     * @param _message address _message expected type: Message { AppID [32]byte, Payload []byte }
+     * @param _verification bytes _verification expected type: Verification { BlockNumber int64, EventID int64 }
+     */
+    function handle(bytes memory _message, bytes memory _verification)
```
vgeddes, line 29: The substrate side doesn't embed the app_id within the message.

To better mirror the substrate side, the method signature would look like this:
```
submit(address _app_id, bytes memory _message, bytes memory _verification)
```

On the substrate side the signature looks like the following:
```
submit(app_id: AppID, message: Message)

struct Message {
	payload: Vec<u8>,
	verification: VerificationInput,
}
```

I don't think we need to worry about making the signatures exactly the same. So passing message and verification as separate parameters like you're doing is totally fine for this milestone.

ethereum/contracts/Bridge.sol
```diff
@@ -0,0 +1,75 @@
...
+        require(eventID > mostRecentBlockEvent[blockNumber], "Events must be processed chronologically");
+
+        address appID = _message.sliceAddress(32);
+        require(applications[appID], "App ID not found. Only registered application are supported");
+
+        bytes memory payload = _message.slice(32, _message.length-1);
+        Application app = Application(appID);
+        app.handle(payload);
+
+        mostRecentBlockEvent[blockNumber] = eventID;
```
vgeddes, line 45: As per my previous comment this should go into the verifier contract.

ethereum/contracts/Bridge.sol
```diff
@@ -0,0 +1,75 @@
...
+        view
+        returns(bool)
+    {
+        if(applications[_appID]){
+            return false;
+        }
+
+        // 0x0 is the value returned for not-yet created accounts.
+        // 0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470 is returned for accounts without code.
+        bytes32 accountHash = 0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470;
```
vgeddes, line 66: What is this hardcoded `accountHash` and whose account does it represent?

ethereum/contracts/ERC20App.sol
```diff
@@ -56,20 +57,20 @@ contract ERC20App {
...
     }
 
-    function submit(bytes memory _data, bytes memory _signature)
+    function handle(bytes memory _message)
         public
-        returns (uint256)
+        override
     {
-        require(_data.length == 104, "Data must contain 104 bytes for a successful decoding");
+        require(_message.length == 104, "Message must contain 104 bytes for a successful decoding");
```
vgeddes, line 64: I'd define `104` as a constant variable.

ethereum/contracts/Broker.sol
vgeddes: Same here. I've done the same on the Substrate side.

ethereum/contracts/Bridge.sol
```diff
@@ -0,0 +1,75 @@
...
+     * @param _verification bytes _verification expected type: Verification { BlockNumber int64, EventID int64 }
+     */
+    function handle(bytes memory _message, bytes memory _verification)
+        public
+    {
+        uint64 blockNumber = uint64(_verification.slice(0, 8).decodeUint256());
+        require(blockNumber >= mostRecentBlock, "Blocks must be processed chronologically");
+
+        uint64 eventID = uint64(_verification.slice(8, 16).decodeUint256());
+        require(eventID > mostRecentBlockEvent[blockNumber], "Events must be processed chronologically");
```
vgeddes, line 36: Do we need to guard against `mostRecentBlockEvent[blockNumber]` raising a NotFound error (or however Solidity signals such errors)?

vgeddes: Also we'll need to investigate what an Substrate event id represents. It may not be the same as an Ethereum event id which typically represents the numeric index of an event with a block.

I'll circle back to you on this.

ethereum/contracts/Bridge.sol
```diff
@@ -0,0 +1,75 @@
...
+    /**
+     * @dev routes the message to the specified application ID after verifying the operator's signature
+     * @param _message address _message expected type: Message { AppID [32]byte, Payload []byte }
+     * @param _verification bytes _verification expected type: Verification { BlockNumber int64, EventID int64 }
+     */
+    function handle(bytes memory _message, bytes memory _verification)
+        public
+    {
+        uint64 blockNumber = uint64(_verification.slice(0, 8).decodeUint256());
+        require(blockNumber >= mostRecentBlock, "Blocks must be processed chronologically");
```
vgeddes, line 33: Surely you mean `blockNumber < mostRecentBlock`?

ethereum/contracts/Bridge.sol
denalimarsh: See line 65:
```
// 0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470 is returned for accounts without code.
```

i.e. user addresses have this account hash while accounts with code (smart contracts) have different account hashes.

ethereum/contracts/Bridge.sol
denalimarsh: The null value of `uint` is 0, so an empty `mostRecentBlockEvent[blockNumber]` returns 0; this line is why I requested that the embedded event ID counter starts from 1 on the substrate side. The first event in a block will be event 1 and therefore pass this check.

ethereum/contracts/Bridge.sol
denalimarsh: Here `mostRecentBlock` refers to the block associated with the most recently processed event. If we've already processed event 1 from block 100, we shouldn't process any events from block 1-99. We only want to process additional events from block 100 or blocks greater than 100.

ethereum/contracts/Bridge.sol
vgeddes: Sounds good, was unfamiliar with that concept.

Small nitpick, can we hoist `accountHash` up and make it a constant variable like?:
```
bytes32 constant accountCodeHash = 0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470;
``

File: ethereum/contracts/Bridge.sol (hunk 12 above)
Comment (vgeddes) [reply ethereum/contracts/Bridge.sol:36]: Sounds good 👍

File: ethereum/contracts/Bridge.sol [hunk 14]
```diff
@@ -0,0 +1,75 @@
...
+     * @dev verifies new applications
+     * @param _appID address _appID is the application's contract address to be verified
+     */
+    function verifyApp(address _appID)
+        internal
+        view
+        returns(bool)
+    {
+        if(applications[_appID]){
+            return false;
```
philipstanislaus, line 61: Should we not error here instead of just ignore duplicates?

ethereum/contracts/Bridge.sol
philipstanislaus: Agree that a constant would be clearer here 👍

ethereum/contracts/Bridge.sol
```diff
@@ -0,0 +1,75 @@
...
+        }
+
+        // 0x0 is the value returned for not-yet created accounts.
+        // 0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470 is returned for accounts without code.
+        bytes32 accountHash = 0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470;
+        bytes32 codehash;
+        assembly { codehash := extcodehash(_appID) }
+        if(codehash != 0x0 && codehash != accountHash) {
+            return false;
```
philipstanislaus, line 70: Again, should we not error here?

philipstanislaus (commented):
Nice 👍

ethereum/contracts/Bridge.sol
denalimarsh: Updated in https://github.com/Snowfork/polkadot-ethereum/pull/75/commits/5eeaae24f7fcfe531fae0fc50d00e05dc333fd5b

ethereum/contracts/Bridge.sol
denalimarsh: Updated in https://github.com/Snowfork/polkadot-ethereum/pull/75/commits/5eeaae24f7fcfe531fae0fc50d00e05dc333fd5b

ethereum/contracts/Bridge.sol
denalimarsh: Updated in https://github.com/Snowfork/polkadot-ethereum/pull/75/commits/5eeaae24f7fcfe531fae0fc50d00e05dc333fd5b

ethereum/contracts/Bridge.sol
denalimarsh: Replaced event IDs and block numbers with just checking event hashes in https://github.com/Snowfork/polkadot-ethereum/pull/75/commits/0bf3adece367c08ce19bc50007eb975eb7de2e71. As it's a simple look-up map it's much easier to keep it in the Bridge contract for now. Given the several iterations on verification in the past few days, I'm hesitant to lock something in that we may end up changing as well. Can always move it to Verifier later on.

ethereum/contracts/Bridge.sol
denalimarsh: https://github.com/Snowfork/polkadot-ethereum/pull/75/commits/98f7e1c48e30847421a6a66fe2b4e8084eb51e64

ethereum/contracts/ERC20App.sol
denalimarsh: https://github.com/Snowfork/polkadot-ethereum/pull/75/commits/a651ad881e49d877b50dcff6db64065b63f52d3b

ethereum/contracts/Bridge.sol
```diff
@@ -37,7 +37,7 @@ contract Bridge {
 
         bytes memory payload = _message.slice(32, _message.length-1);
         Application app = Application(appID);
-        app.handle(payload);
+        app.submit(payload);
```
vgeddes, line 40: It should be `app.handle()` to mirror the substrate naming conventions. Following this language: The relayer _submits_ messages to the bridge, and the bridge asks an application to _handle_ a message.

I realize my Solidity changes made this confusing, they were stop gap measures which made applications look like a bridge.

ethereum/contracts/ETHApp.sol
```diff
@@ -36,46 +32,28 @@ contract ETHApp is Application {
...
-        internal
-        pure
-        returns(bytes memory)
-    {
-        return abi.encode(_sender, _recipient, _tokenAddr, _amount, _nonce);
+        emit Transfer(msg.sender, _recipient, msg.value);
     }
 
-    function handle(bytes memory _message)
+    function submit(bytes memory _data)
```
vgeddes, line 38: as per my previous comment, this should be called `handle`.

ethereum/contracts/Bridge.sol
denalimarsh: agreed, this makes much more sense.

ethereum/contracts/ETHApp.sol
denalimarsh: https://github.com/Snowfork/polkadot-ethereum/pull/75/commits/3ce50f6763d7cd67b4a9968a76b0de19118fad56

ethereum/contracts/Bridge.sol
denalimarsh: https://github.com/Snowfork/polkadot-ethereum/pull/75/commits/3ce50f6763d7cd67b4a9968a76b0de19118fad56

ethereum/contracts/Bridge.sol
```diff
@@ -0,0 +1,61 @@
...
+            }
+        }
+    }
+
+    /**
+     * @dev routes the message to the specified application ID after verifying the operator's signature
+     * @param _message address _message expected type: Message { AppID [32]byte, Payload []byte }
+     * @param _hash bytes _hash is a unique hash of the event (app_id, payload, blockNum, eventId)
+     */
+    function submit(bytes memory _message, bytes32 _hash)
```
vgeddes, line 29: The signature should be `submit(address _appId, bytes memory _message)`, as this what the relayer is currently expecting.

Is there a mechanism in Solidity to hash the message?

ethereum/contracts/Bridge.sol
```diff
@@ -0,0 +1,61 @@
...
+    {
+        address appID = _message.sliceAddress(32);
+        require(applications[appID], "App ID not found. Only registered application are supported");
+
+        require(!eventHashes[_hash], "Event has already been processed");
+        eventHashes[_hash] = true;
+
+        bytes memory payload = _message.slice(32, _message.length-1);
+        Application app = Application(appID);
+        app.handle(payload);
```
vgeddes, line 40: Don't we have to call into the verifier contract to verify that the sender's address is the one we trust.

For reference, here is the main entry point into the verifier on the Substrate side:
https://github.com/Snowfork/polkadot-ethereum/blob/app-identifiers/parachain/pallets/verifier/src/lib.rs#L52

vgeddes: So along those lines I'd suggest adding a method `Verifier.verify(address _sender, bytes _message)` to the verifier, have the bridge call that.

ethereum/contracts/Bridge.sol
denalimarsh: In Solidity we can hash with `keccak256(abi.encodePacked("one", "two", "three"));`

ethereum/contracts/Bridge.sol
denalimarsh: The Verifier contract expects the signature to be passed separately from the contents of the signed message.
1. `verify(bytes32 _hash, bytes memory _signature)`
2. `verifyBytes(bytes memory _rawMessage, bytes memory _signature)`

So we need to parse the signature from the `_message`. But in my understanding, the message currently doesn't contain a signature?

Side note: I think some of the confusion around naming conventions here stems from the fact that Ethereum generally uses "message" to refer to the data that an address applied their signature to.

denalimarsh: The main branch currently has the verification data as: 
```
type VerificationBasic struct {
	BlockNumber uint64
	EventIndex  uint32
}
```

Is the proposal here that we should
1. Parse `Message.VerificationBasic.BlockNumber` and `Message.VerificationBasic.EventIndex`
2. Create a hash via `hash(AppID, BlockNumber, EventIndex, Message.Data)`
3. Validate that we haven't processed this hash yet
4. Check Verifier contract to see if the signature is from the operator

I'm confused because there is no signature available, we've replaced it with the BlockNumber/EventIndex, correct?

denalimarsh: Or the Verifier contract just checks msg.sender and doesn't check the signature?

vgeddes: Just checking msg.sender and performing a hash of the entire message should be sufficient:

Basically this rough pseudocode shows what the bridge needs to do:
```
submit(app_id, message):
  verifySender(msg.sender)
  VerifyHash(hash(app_id, message))
```

Since the block number and event index are embedded in the message bytes, we can just hash the entire message for replay protection

Right now the relayer isn't sending the signature anymore. Also type `types.VerificationBasic` in the relayer isn't being used anymore.

ethereum/contracts/Bridge.sol
denalimarsh: hashed the message on contract in https://github.com/Snowfork/polkadot-ethereum/pull/75/commits/0f905430101dc05b1624c155753ac5169d4f6624

ethereum/contracts/Bridge.sol
denalimarsh: updated in https://github.com/Snowfork/polkadot-ethereum/pull/75/commits/0f905430101dc05b1624c155753ac5169d4f6624, https://github.com/Snowfork/polkadot-ethereum/pull/75/commits/c07a362125122d0bfc0f8030e84165872d3d9dd4

vgeddes (approved):
LGTM! 💯

Issue #61, opened 2020-08-31
Make use of App IDs to identify applications


vgeddes:
Our early design envisioned identifying applications on both chains using opaque 256-bit identifiers. During development however we never diverted from that approach, and used more ad-hoc means of identification in order to get a basic proof of work complete.

We'll need to figure out a final solution for M5.

Issue #16, opened 2020-07-14
Origins for PolkaETH dispatchables need to be verified


vgeddes:
* https://github.com/Snowfork/polkadot-ethereum/pull/15/files#r454417710

vgeddes:
Obsoleted by #79.

Issue #11, opened 2020-07-13
Naming of traits could be more descriptive


vgeddes:
For example, rename `Broker` to `BridgeIn`. Especially since the trait only covers a subset of the Broker module's behavior.

See review comments:
* https://github.com/Snowfork/polkadot-ethereum/pull/5#discussion_r453719793
* https://github.com/Snowfork/polkadot-ethereum/pull/5#discussion_r453720140

vgeddes:
Obsoleted by #79

PR #79, opened 2020-09-13
Replay protection + Use App identifiers for routing messages

vgeddes:
The parachain now routes messages to target applications using an `AppId` field sent along with the message by the relayer.The AppId is now a byte array of length 20. Typically it will contain an ethereum contract address. 

In the parachain, The function `artemis_core::registry::lookup_app` can be used to lookup a registered app via its AppId. The bridge uses this to implement message dispatch.

In the relayer, the config file `config.toml` provides the necessary information to allow the various event listeners and writers to match up messages with correct AppIds:

```toml
[ethereum]
endpoint = "ws://localhost:9545/"

[ethereum.apps]
[ethereum.apps.eth]
address = "0x0d27b0069241c03575669fED1BaDCBccDc0dd4d1"  // This is used as an AppId on both chains
abi = "~/.config/artemis-relay/ethereum/ETH.json"

[ethereum.apps.erc20]
address = "0x8FE1B1233F7032cEF8Cfc5eAAF411dfFaA77A07C"     // This is used as an AppId on both chains
abi = "~/.config/artemis-relay/ethereum/ERC20.json"

[substrate]
endpoint = "ws://127.0.0.1:9944/"
``` 

Other Changes:

* Parachain:
  * The verifier pallet now does replay protection, signature checking etc.
  * The verifier is now loosely coupled to the bridge via a trait. It no longer uses a dispatchable call to inform the broker that a message was verified. This greatly simplifies the architecture in the interim.
  * Each app now does its own payload decoding to make apps independent of each other. Previously the apps relied on the `artemis-ethereum` crate for this, which was the wrong approach in many ways.
* Bridge
  * Refactored the Ethereum contract and ABI- related modules.
  * Implemented an improved decoder for Substrate Events (No longer using the GSRPC implementation). The new version preserves event ordering among other requirements.
  * Much improved configuration management. Viper configuration is now unmarshaled into Config objects.
* Solidity
  * Solidity app contracts now emit `Transfer(sender, recipient, ..., amount)` instead of `AppEvent(data)`

The changes address or would obsolete the following issues: #9, #11, #16, #61


parachain/pallets/verifier/src/lib.rs
```diff
@@ -0,0 +1,87 @@
...
+#[cfg(test)]
+mod tests;
+
+pub trait Trait: system::Trait {
+	type Event: From<Event> + Into<<Self as system::Trait>::Event>;
+}
+
+decl_storage! {
+	trait Store for Module<T: Trait> as VerifierModule {
+		RelayKey get(fn key) config(): T::AccountId;
```
philipstanislaus, line 24: Please add a comment describing RelayKey. Thanks!

parachain/pallets/verifier/src/lib.rs
```diff
@@ -0,0 +1,87 @@
...
+		pub VerifiedPayloads: map hasher(blake2_128_concat) T::Hash => ();
```
philipstanislaus, line 25: Just as a note for the future to keep the storage size bounded: 

We could simply have a double map here, where the first key is the block height of the received message. We can the define a `maxTailLength`, e. g. 200 blocks, and just remove all storage keys where `height < currentHeight - 200`. At the same time, we decline verification for older blocks.

philipstanislaus: To support this, we could already start storing the double map here, but not do the tail length check and failures?

ethereum/contracts/ERC20App.sol
```diff
@@ -38,22 +36,7 @@ contract ERC20App {
...
-        address _tokenAddr,
-        uint256 _amount,
-        uint256 _nonce
-    )
-        internal
-        pure
-        returns(bytes memory)
-    {
-        return abi.encode(_sender, _recipient, _tokenAddr, _amount, _nonce);
+        emit Transfer(msg.sender, _recipient, _tokenAddr, _amount);
```
philipstanislaus, line 39: That's great, finally no more AppEvent ;)

ethereum/scripts/dumpABI.js
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env node
+
+const fs = require('fs');
+
+const args = process.argv.slice(2);
+const contract = JSON.parse(fs.readFileSync(args[0], 'utf8'));
```
philipstanislaus, line 6: Could you not just require the json file?

philipstanislaus (commented):
Just reviewed the `parachain` and `ethereum` parts so far – will review the rest in the afternoon.

I really like the simplification of the architecture by decoupling the verifier and the independent payload decoding. Great tests as well 👍

parachain/pallets/bridge/src/lib.rs
```diff
@@ -1,56 +1,73 @@
...
+	fn verify(sender: T::AccountId, app_id: AppId, message: &Message) -> DispatchResult {
+		T::Verifier::verify(sender, app_id, &message)
+	}
+
+	/// Dispatch `message` to application `app`.
+	fn dispatch(app: App, message: Message) -> DispatchResult {
+		match app {
+			App::ETH => T::AppETH::handle(message),
+			App::ERC20 => T::AppERC20::handle(message)
 		}
```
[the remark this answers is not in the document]
vgeddes: Would be difficult to achieve given constraints with how Substrate implements pallets and how it loosely couples them together via traits.

Firstly, as a primer, the `Application` type is a trait that is implemented by `pallet_eth::Module` and `pallet_erc20::Module` pallet modules. These modules are never instantiated. Indeed Substrate doesn't instantiate any of its pallets. It just calls static methods on the type.

That means a hypothetical registry lookup function would never be able to return a instantiated application _object_. And Rust doesn't support first-class types (i.e passing types as parameters into functions, or returning types from functions).

That limitation forces us to implement the hardcoded message dispatch in the bridge module, because that's the first place where `Application` types were injected by the runtime:
https://github.com/Snowfork/polkadot-ethereum/blob/main/parachain/runtime/src/lib.rs#L277

parachain/pallets/bridge/src/lib.rs
```diff
@@ -1,56 +1,73 @@
...
 
 pub trait Trait: system::Trait {
-
-	type Event: From<Event<Self>> + Into<<Self as system::Trait>::Event>;
-
-	type Broker: Broker;
+	type Event: From<Event> + Into<<Self as system::Trait>::Event>;
+	type Verifier: Verifier<<Self as system::Trait>::AccountId>;
+	type AppETH: Application;
+	type AppERC20: Application;
```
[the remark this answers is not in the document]
vgeddes: Added some thoughts about this in https://github.com/Snowfork/polkadot-ethereum/pull/79#discussion_r487834594

parachain/pallets/asset/src/lib.rs
```diff
@@ -39,9 +39,9 @@ decl_event!(
 	where
 		AccountId = <T as system::Trait>::AccountId,
 	{
-		Burned(AssetId, AccountId, U256),
-		Minted(AssetId, AccountId, U256),
-		Transferred(AssetId, AccountId, AccountId, U256),
+		Burned(H160, AccountId, U256),
+		Minted(H160, AccountId, U256),
+		Transferred(H160, AccountId, AccountId, U256),
```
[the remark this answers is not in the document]
vgeddes: Yeah, I had to make this change purely for UX purposes. The problem is that the Polkadot JS app already has a non-overridable type mapping for the `AssetId` parameter. The similarly named `assets` pallet in FRAME also emits an event which includes a parameter named `AssetId`, and Polkadot-JS It interpets it as an unsigned integer rather than the 20-byte hash we need.

What we really need to do here is find an alias for `H160` that is semantically correct. maybe `EthereumAddr` or `TokenID`. I dunno.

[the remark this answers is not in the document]
vgeddes: Added ticket to track: #81

parachain/pallets/verifier/src/lib.rs
vgeddes: Good point, is a bit obscure. Will add some more commentary.

bridgerelayer/tools/list_events.go
```diff
@@ -0,0 +1,106 @@
...
+	"github.com/snowfork/go-substrate-rpc-client/types"
+)
+
+type Events []Event
+
+func (e Events) Len() int {
+	return len(e)
+}
+
+func (e Events) Less(i, j int) bool {
```
philipstanislaus, line 18: Does it make sense to call this for clarity?

```suggestion
func (e Events) Lte(i, j int) bool {
```

bridgerelayer/tools/list_events.go
```diff
@@ -0,0 +1,106 @@
...
+			return true
+		}
+	}
+	return false
+}
+
+func (e Events) Swap(i, j int) {
+	tmp := e[i]
+	e[i] = e[j]
+	e[j] = tmp
```
philipstanislaus, line 33:
```suggestion
	e[i], e[j] = e[j], e[i]
```

bridgerelayer/tools/list_events.go
```diff
@@ -0,0 +1,106 @@
...
+		}
+	}
+
+	format := "tm[%#v] = reflect.TypeOf(%s%s{})\n"
+
+	for _, ev := range events {
+		fmt.Printf(format, [2]string{ev.ModuleName, ev.Name}, ev.ModuleName, ev.Name)
+	}
+
+}
```
philipstanislaus, line 106: Nice functionality 👍

philipstanislaus (approved):
Finished my review, LGTM 👍

bridgerelayer/chain/substrate/listener.go
```diff
@@ -55,6 +57,7 @@ func (li *Listener) pollBlocks(ctx context.Context) error {
 	}
 	currentBlock := uint64(block.Number)
 
+	retryInterval := time.Duration(10) * time.Second
```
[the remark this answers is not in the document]
vgeddes: Good point. I've added issue #82 to track for future implementation.

bridgerelayer/tools/list_events.go
vgeddes: Ah `Less` is an Interface method which I'm required to implement so that I can sort an array of events.

bridgerelayer/tools/list_events.go
vgeddes: Yeah, I need to implement that tool so I could discover all the events that could possibly be emitted from our parachain. Was useful in implementing the new event decoder.

ethereum/scripts/dumpABI.js
vgeddes: yeah I could have, though this is a once-off hacky script which will probably be discarded post-M5.

parachain/pallets/verifier/src/lib.rs
vgeddes: I've added a bunch of documentation across the parachain in the latest upload.

parachain/pallets/verifier/src/lib.rs
vgeddes: Added #83 to track.

parachain/pallets/bridge/src/lib.rs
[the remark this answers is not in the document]
philipstanislaus: Understood – is there another way around concrete types in the bridge though? You could just pass the registry itself as a type to the bridge pallet, and the registry gets filled with applications upon its instantiation? Then `handle` would be called on the registry, which determines the right application at runtime?

bridgerelayer/tools/list_events.go
philipstanislaus: Okay, got it. Less would semantically be `<` though, not `<=`, right?

PR #89, opened 2020-09-16
Miscellaneous Solidity fixes

denalimarsh:
Addresses issues:
- https://github.com/Snowfork/polkadot-ethereum/issues/42
- https://github.com/Snowfork/polkadot-ethereum/issues/56

### DIFF
--- a/ethereum/contracts/Bridge.sol
+++ b/ethereum/contracts/Bridge.sol
@@ -54,7 +54,7 @@ contract Bridge {
         view
         returns(bool)
     {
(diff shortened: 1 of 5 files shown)


philipstanislaus (approved):
That’s great! 

Just one question: Is the duplicate decodeUint256 necessary?

denalimarsh:
> That’s great!
> 
> Just one question: Is the duplicate decodeUint256 necessary?

Right now Scale.sol contains all the Scale decoding work, while Decode.sol is a library that only includes the functions required by the current codebase.

PR #90, opened 2020-09-18
Solidity test suite for M5

denalimarsh:
Fixes up the solidity test suite to reflect the latest changes

### DIFF
--- a/ethereum/contracts/ETHApp.sol
+++ b/ethereum/contracts/ETHApp.sol
@@ -14,7 +14,7 @@ contract ETHApp is Application {
     uint256 public nonce;
     uint256 public totalETH;
 
-    event Transfer(address _sender, bytes32 _recipient, uint256 _amount);
+    event AppTransfer(address _sender, bytes32 _recipient, uint256 _amount);
     event Unlock(bytes _sender, address _recipient, uint256 _amount);
 
     constructor() public {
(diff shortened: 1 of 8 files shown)


denalimarsh:
We need to update the event name `Transfer` to `AppTransfer` because Transfer conflicts with the ERC20 token transfer event.

ethereum/contracts/ETHApp.sol
```diff
@@ -14,7 +14,7 @@ contract ETHApp is Application {
     uint256 public nonce;
     uint256 public totalETH;
 
-    event Transfer(address _sender, bytes32 _recipient, uint256 _amount);
+    event AppTransfer(address _sender, bytes32 _recipient, uint256 _amount);
```
vgeddes, line 17: Ooh, we can't make this change without updating both the Relayer and Parachain. Is there a compelling reason why we should rename?

musnit: @vgeddes See above comment

vgeddes: Oh I see! Sorry I must have missed the other comment. @denalimarsh I'm gonna push the necessary parachain and relayer changes to this branch too.

vgeddes (approved):
LGTM :rocket:

PR #100, opened 2020-09-20
Update system for changes in Solidity contracts

vgeddes:
This was needed to get cross-chain transfers working in both directions after we updated the Solidity side.

Changes:
* Fixed payload decoding in Solidity (for the substrate -> ethereum flow)
* Removed obsolete nonce stuff in Solidity contracts
* Relayer now submits messages to Ethereum via the Bridge contract
* Use updated contract ABI's in parachain payload decoders
* Updated README's
* Misc improvements to various scripts (truffle scripts, etc)


denalimarsh (approved):
Looks good to me. Thanks for adding `.env` to gitignore 👍

ethereum/contracts/ERC20App.sol
```diff
@@ -42,16 +35,17 @@ contract ERC20App is Application {
         public
         override
     {
-        require(_data.length == MESSAGE_LENGTH, "Message must contain 104 bytes for a successful decoding");
+        require(_data.length >= PAYLOAD_LENGTH, "Invalid Payload");
```
philipstanislaus, line 38: Could you add a comment why we allow longer payloads? Thanks!

vgeddes: This was due to uncertainty as to whether the bridge contract was correctly removing the verification data from the end of a message. The verification data is used for replay protection and so is not required by target applications. At the time, I opted to loosen restrictions in order to get everything working together.

I've cut https://github.com/Snowfork/polkadot-ethereum/issues/129 to address this issue.

PR #105, opened 2020-09-21
Fix bug in ETHApp contract

vgeddes:


ethereum/contracts/ETHApp.sol
```diff
@@ -36,7 +36,7 @@ contract ETHApp is Application {
         public
         override
     {
-        require(_data.length == PAYLOAD_LENGTH, "Invalid payload");
+        require(_data.length >= PAYLOAD_LENGTH, "Invalid payload");
```
philipstanislaus, line 39: Again, please add a comment here why a longer payload might occur.

PR #108, opened 2020-09-23
JavaScript client for Bridge interaction

denalimarsh:
This PR implements a JavaScript client for interacting with the Bridge. Example usage is in the `example.js` file.


denalimarsh:
I've duplicated the built contracts from `ethereum/build/contracts` to keep the future BridgeClient npm package self-contained (and the built contracts currently aren't committed to git). Open to ideas for improvement.

bridgeclient/src/client/index.js
```diff
@@ -0,0 +1,150 @@
+const Web3 = require('web3');
+
+const config = require("../../config.js");
+const ETHApp = require('../contracts/ETHApp.json');
```
vgeddes, line 4: Can you add a bash script which copies over the necessary contracts from their original source (`ethereum/contracts`)? That will make it easier to maintain the copies. Please also note the script in the readme.

vgeddes (commented):
Thanks Denali, this looks good.

In terms of project organisation, can we rather place this new code under a toplevel directory `test`. For example `test/src` instead of `bridgeclient/src`.

In https://github.com/Snowfork/polkadot-ethereum/pull/109 I've also put a lot of E2E related stuff under `test`, so it would be good to be aligned on that. When I add the JS code for interfacing with substrate I could put it under `test/src` too.

Eventually we could organise our JS code into packages and use yarn workspaces, etc.

bridgeclient/config.js
```diff
@@ -0,0 +1,5 @@
+// Bridge applications
+module.exports = {
+    APP_ETH_CONTRACT_ADDRESS: '0x0823eFE0D0c6bd134a48cBd562fE4460aBE6e92c',
```
vgeddes, line 3: In the E2E test we'll likely shell out to a truffle script to retrieve the address. So I wouldn't include these hardcoded addresses.

Instead, can we update the `BridgeClient` constructor so that it accepts addresses? Something like:

```js
new BridgeClient(endpoint, ethAppAddr, erc20AppAddr)
```

bridgeclient/src/client/index.js
denalimarsh: Added in https://github.com/Snowfork/polkadot-ethereum/pull/108/commits/66eee87936fb9bdf7518112254a0fe7549e8c7b2

bridgeclient/config.js
denalimarsh: Added in https://github.com/Snowfork/polkadot-ethereum/pull/108/commits/66eee87936fb9bdf7518112254a0fe7549e8c7b2

vgeddes (approved):
LGTM 🚀

PR #110, opened 2020-09-23
Remove obsolete prover code

vgeddes:
Its no longer being used by anything and so could cause confusion for external reviewers.


denalimarsh (approved):
LGTM

PR #112, opened 2020-09-24
Fetch deployed contract addresses dynamically from E2E test

denalimarsh:
Fetches deployed contract addresses dynamically from E2E test


ethereum/scripts/fetchDeployedAddrs.js
```diff
@@ -0,0 +1,25 @@
...
+const ERC20App = artifacts.require("ERC20App")
+const TestToken = artifacts.require("TestToken")
+
+module.exports = async () => {
+    try {
+        const eth20AppInstance = await ETHApp.deployed()
+        const erc20AppInstance = await ERC20App.deployed()
+        const tokenInstance = await TestToken.deployed();
+
+        const jsonStr = "{" +
```
vgeddes, line 15: It would be cleaner to serialize an object using `console.log(JSON.stringify(obj, null, 2))`, where `obj = {  ethApp: ... }`

denalimarsh: Updated in https://github.com/Snowfork/polkadot-ethereum/pull/112/commits/659629678494987bc0f9c00e341cac161a3341b0

vgeddes:
No longer necessary. I created an alternative solution that fits in with the current test bootstap process as documented in `README.md`.

PR #115, opened 2020-09-28
Security fix: restrict application message handling to Bridge contract

denalimarsh:
This PR update applications with a simple `msg.sender` check so that only Bridge can call the handle method on applications. The Bridge contract registers its address on each application during initial on-contract app verification. Messages submitted to the Bridge by the relayer will be routed from Bridge contract -> application contract successfully and other actors won’t be able to send fabricated messages directly to the applications.

### DIFF
--- a/ethereum/contracts/Application.sol
+++ b/ethereum/contracts/Application.sol
@@ -4,4 +4,6 @@ pragma solidity >=0.6.2;
 // Application contains methods that all applications must implement
 abstract contract Application {
     function handle(bytes memory _data) public virtual;
+
+    function register(address _bridge) public virtual returns(bool);
 }
--- a/ethereum/contracts/Application.sol
+++ b/ethereum/contracts/Application.sol
@@ -3,7 +3,13 @@ pragma solidity >=0.6.2;
 
 // Application contains methods that all applications must implement
 abstract contract Application {
+    /**
+     * @dev Handles a SCALE encoded message from the Bridge
+     */
     function handle(bytes memory _data) public virtual;
 
+    /**
+     * @dev Registers the Bridge contract on the application
+     */
     function register(address _bridge) public virtual returns(bool);
 }
--- a/ethereum/contracts/Application.sol
+++ b/ethereum/contracts/Application.sol
@@ -11,5 +11,5 @@ abstract contract Application {
     /**
      * @dev Registers the Bridge contract on the application
      */
-    function register(address _bridge) public virtual returns(bool);
+    function register(address _bridge) public virtual;
 }
--- a/ethereum/contracts/Bridge.sol
+++ b/ethereum/contracts/Bridge.sol
@@ -18,9 +18,7 @@ contract Bridge {
     constructor(address _verfierAddr, address[] memory _apps) public {
         verifier = Verifier(_verfierAddr);
         for(uint256 i = 0; i < _apps.length; i++) {
-            if(verifyApp(_apps[i])) {
-                applications[_apps[i]] = true;
-            }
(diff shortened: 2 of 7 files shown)


ethereum/contracts/ETHApp.sol
```diff
@@ -20,6 +21,12 @@ contract ETHApp is Application {
         totalETH = 0;
     }
 
+    function register(address _bridge) public override returns(bool) {
+        require(bridge == address(0), "Bridge has already been registered");
```
vgeddes, line 25: Are you sure this shouldnt be `require(bridge != address(0)`?

ethereum/contracts/ETHApp.sol
```diff
@@ -20,6 +21,12 @@ contract ETHApp is Application {
...
+        bridge = _bridge;
+        return true;
```
vgeddes, line 27: Rather just make this a `void` method if it is going to return `true` under all expected conditions.

ethereum/contracts/Bridge.sol
```diff
@@ -61,6 +57,8 @@ contract Bridge {
         require(codehash != 0x0, "There's no account for this address on the network");
         require(codehash != NON_CONTRACT_ACCOUNT_HASH, "Only contract accounts can be registered as applications");
 
-        return true;
+        Application app = Application(_appID);
+        require(app.register(address(this)), "Application failed registration");
```
vgeddes, line 61: `app.register` always returns true, so the `require` seems unnecessary.

vgeddes (approved):
Looks good, though added a few minor comments.

ethereum/contracts/Application.sol
```diff
@@ -4,4 +4,6 @@ pragma solidity >=0.6.2;
 // Application contains methods that all applications must implement
 abstract contract Application {
     function handle(bytes memory _data) public virtual;
+
+    function register(address _bridge) public virtual returns(bool);
```
vgeddes, line 8: We'll need doc comments for both `handle` and `register` so we can generate nice Solidity docs.

ethereum/contracts/Bridge.sol
musnit: in our current apps yes, but may be useful for future apps to allow failing registration?

ethereum/contracts/ETHApp.sol
musnit: as above, think its true in the case of ETHApp but may not be true for every app

ethereum/contracts/Application.sol
denalimarsh: https://github.com/Snowfork/polkadot-ethereum/pull/115/commits/beb6cde3e8933817fca0f11e89e4af106c1bffb2

ethereum/contracts/Bridge.sol
denalimarsh: - Vincent: Already registered applications will throw an error here on `app.register(address(this))` due to the check `require(bridge == address(0), "Bridge has already been registered");` so it's redundant.
- Aidan: Future apps should fail registration if they're already registered, which is handled by the current code.

I've updated the contracts in https://github.com/Snowfork/polkadot-ethereum/pull/115/commits/ce891d0610f5a38410cd1ff327087dda266cc932

ethereum/contracts/ETHApp.sol
denalimarsh: Went with the void + exception approach in https://github.com/Snowfork/polkadot-ethereum/pull/115/commits/ce891d0610f5a38410cd1ff327087dda266cc932

PR #120, opened 2020-10-07
Update README.md

Noc2:
I had to create a parachain-state folder inside build otherwise "sudo docker-compose up -d -- parachain" failed.

### DIFF
--- a/test/README.md
+++ b/test/README.md
@@ -24,7 +24,7 @@ yarn install
 
 Run the following commands one by one:
 ```bash
-mkdir build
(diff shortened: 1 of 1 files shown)


vgeddes:
Thanks @Noc2 for catching that. The dev/build process is definitely fiddly and I want to streamline it over the next few weeks.

I've added https://github.com/Snowfork/polkadot-ethereum/pull/122 which includes your fix (slightly changed), as well as some other readme improvements.

Closing this one for now.

PR #124, opened 2020-10-11
Update substrate dependency to v2.0.0-rc6

vgeddes:
This is a prerequisite for Cumulus integration.

Changes:
* The parachain now depends on Substrate v2.0.0-rc6
* No longer need to use our forked version of GSRPC as https://github.com/centrifuge/go-substrate-rpc-client is now compatible with v2.0.0-rc6.

The E2E tests passed successfully.


Rizziepit:
For my own understanding: are most of the changes in 5883e0d simply the latest boilerplate from substrate-node-template?

vgeddes:
> For my own understanding: are most of the changes in [5883e0d](https://github.com/Snowfork/polkadot-ethereum/commit/5883e0d521ce1a388012b687a8f1e86d74695f79) simply the latest boilerplate from substrate-node-template?

Pretty much, yeah! 💯 

Our parachain was originally forked off from substrate-node-template, so we can use it as a reference when we need to upgrade our chain to the latest & greatest.

parachain/node/src/service.rs
```diff
@@ -174,95 +194,104 @@ pub fn new_full(config: Configuration) -> Result<impl AbstractService, ServiceEr
...
-
-			Ok((import_queue, finality_proof_request_builder))
-		})?
-		.with_finality_proof_provider(|client, backend| {
-			// GenesisAuthoritySetProvider is implemented for StorageAndProofProvider
-			let provider = client as Arc<dyn StorageAndProofProvider<_, _>>;
-			Ok(Arc::new(GrandpaFinalityProofProvider::new(backend, provider)) as _)
-		})?
-		.build_light()
+pub fn new_light(config: Configuration) -> Result<TaskManager, ServiceError> {
```
musnit, line 224: @vgeddes are you familiar with this code/function or do you know what it does in detail? Looks like it's creating a light client for the Relay chain or something along those lines? 

I'm curious because it seems like it could be relevant for either:
 - Creating the server to support the relayer/light client for the Polkadot-Light-Client-In-Solidity tasks
   or
 - Code reference for a Rust implementation to draw inspiration from for the Solidity work

cc @denalimarsh 

Any insights/thoughts?

musnit:
Overall looks fine - I don't think we can really review all the changes coming in from 2.0.0 in terms of security/details, but if our functionality all still works and e2e tests pass, that should be sufficient for testnet needs.

parachain/node/src/service.rs
vgeddes: > @vgeddes are you familiar with this code/function or do you know what it does in detail? Looks like it's creating a light client for the Relay chain or something along those lines?

Not really unfortunately. Its just boilerplate that is included in any chain forked from substrate-node-template. But yeah it does provide some pointers for further investigation.

I don't think it has any knowledge of a relay chain though. Its a light client purely for a substrate chain (with aura, etc).

musnit (approved):

philipstanislaus (approved):
LGTM!

PR #125, opened 2020-10-12
Fix ETH unlock edge case

denalimarsh:
Addresses https://github.com/Snowfork/polkadot-ethereum/issues/121 by catching an edge case when the available amount on the ETH Application is slightly less than the unlock request by setting the requested unlock amount to the maximum available amount.

### DIFF
--- a/ethereum/contracts/ETHApp.sol
+++ b/ethereum/contracts/ETHApp.sol
@@ -61,7 +61,11 @@ contract ETHApp is Application {
         internal
     {
         require(_amount > 0, "Must unlock a positive amount");
-        require(totalETH > _amount, "ETH token balances insufficient to fulfill the unlock request");
+
+        // If amount is greater than available balances, set amount to max available
+        if(totalETH < _amount) {
+            _amount = totalETH;
+        }
 
         totalETH = totalETH.sub(_amount);
         _recipient.transfer(_amount);
--- a/ethereum/contracts/ETHApp.sol
+++ b/ethereum/contracts/ETHApp.sol
@@ -61,11 +61,7 @@ contract ETHApp is Application {
         internal
     {
         require(_amount > 0, "Must unlock a positive amount");
-
-        // If amount is greater than available balances, set amount to max available
-        if(totalETH < _amount) {
-            _amount = totalETH;
-        }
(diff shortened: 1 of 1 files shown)


musnit:
lgtm

vgeddes (commented):
Am a bit worried that we are decreasing `amount` to `totalETH` in this particular edge case. It implies that there is an imbalance, potentially because the relayer or verifier contract are accepting invalid or exploitative messages.

In future it would be great if instead of `totalETH`, we could rather keep track of locked-up amounts by depositor. Though I don't know if that's possible or desirable for these kind of bridge contracts.

cc @musnit

musnit:
@vgeddes Tracking-by-depositor doesn't work - consider the following:
 - Aidan locks 5 ETH
 - Aidan sends 5 PolkaETH to Vincent on Polkadot
 - Vincent burns 5 PolkaETH and wants to unlock 5 ETH on Ethereum

The above example shows how the depositor is basically irrelevant as the locked up ETH needs to be completely fungible to support a fungible PolkaETH token on the Polkadot side.

I agree though that the edge case does imply an imbalance, and yes, if that edge case ever occurs there is an imbalance. It should only occur in the exploitative/invalid scenario you suggest. Maybe it would be better to just use 

```
require(totalETH >= _amount, "ETH token balances insufficient to fulfill the unlock request");
```

(using ```>=``` rather than ```>```)

instead of the decrease solution?

vgeddes:
> @vgeddes Tracking-by-depositor doesn't work - consider the following:

Gotcha :+1:

> I agree though that the edge case does imply an imbalance, and yes, if that edge case ever occurs there is an imbalance. It should only occur in the exploitative/invalid scenario you suggest. Maybe it would be better to just use
> 
> ```
> require(totalETH >= _amount, "ETH token balances insufficient to fulfill the unlock request");
> ```
> 
> (using `>=` rather than `>`)
> 
> instead of the decrease solution?

I kinda prefer this too.

denalimarsh:
> (using `>=` rather than `>`)

This is the correct answer. Updated in https://github.com/Snowfork/polkadot-ethereum/pull/125/commits/29745462cdb2bc779b8b80dd7172b0693228fe7c.

vgeddes (approved):
LGMT 🚀

PR #127, opened 2020-10-14
Add local usage guide

musnit:
Guide for how to setup and use the webapp locally

### DIFF
--- a/docs/local_usage.md
+++ b/docs/local_usage.md
@@ -1,0 +1,48 @@
+# Running the bridge locally and testing it end to end manually from UI yourself
+
+## Local extension Setup:
+ - The tests setup a ganache server with a preset mnemonic. Add the first key from this mnemonic to your Metamask as an extra account to use for testing. You can find this first key here: ARTEMIS_ETHEREUM_KEY in /polkadot-ethereum/test/docker-compose.yml
+ - Add Localhost:8545 to your Metamask and use it
+ - Setup your polkadot-js extension in your browser and create a test Play account
+
+## Local Server Setup:
+ - Run all test docker containers as per the test README
+ - Make sure you have pulled and setup our substrate-ui fork in a seperate folder to this repo (https://github.com/Snowfork/substrate-ui)
+ - Look at the deployed smart contract addresses in the test build config which should have been generated here: /test/build/address.json
+ - Update the bridge react package in your local substrate-ui fork's config with the deployed ETHApp and ERC20App address (In /substrate-ui/packages/page-ethereum-bridge/src/config.tsx)
+ - Add the TestToken address from /test/build/address.json to your metamask wallet assets so you can see your TestTokens. Some should have been automatically minted to your test metamask account
+ - Run our substrate-ui fork
+
+## Using the substrate-ui fork
+ - Go to localhost:3000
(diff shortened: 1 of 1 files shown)


musnit:
@Rizziepit @jakai7 This should guide how to actually use the system in practice as a human. Please both try this and make sure you can do it successfully.

vgeddes:
Just one thing to note, there is a bit of a schism in our development process, since the docker containers managed by `tests/docker-compose.yml` are really only meant for running E2E tests, or for third-parties to play with our system. I think the original thougt process on my part is that users would only only need to run a single command to bootstrap the entire system. Though as `test/README.md` shows its quite a bit more complicated.

For day-to-day development its more convenient to setup the system by following the component README's in order (ethereum, parachain, readme). Docker just gets in the way of things IMO.

Ultimately, we should try and converge on a solution that allows for day-to-day development, running of E2E tests on a single stack

docs/local_usage.md
```diff
@@ -0,0 +1,48 @@
+# Running the bridge locally and testing it end to end manually from UI yourself
```
vgeddes, line 1: You'll need to add header to this file for it to appear in the generated docs. See `docs/about.md` for an example.

docs/local_usage.md
```diff
@@ -0,0 +1,48 @@
...
+## Local extension Setup:
+ - The tests setup a ganache server with a preset mnemonic. Add the first key from this mnemonic to your Metamask as an extra account to use for testing. You can find this first key here: ARTEMIS_ETHEREUM_KEY in /polkadot-ethereum/test/docker-compose.yml
+ - Add Localhost:8545 to your Metamask and use it
+ - Setup your polkadot-js extension in your browser and create a test Play account
+
+## Local Server Setup:
+ - Run all test docker containers as per the test README
+ - Make sure you have pulled and setup our substrate-ui fork in a seperate folder to this repo (https://github.com/Snowfork/substrate-ui)
+ - Look at the deployed smart contract addresses in the test build config which should have been generated here: /test/build/address.json
+ - Update the bridge react package in your local substrate-ui fork's config with the deployed ETHApp and ERC20App address (In /substrate-ui/packages/page-ethereum-bridge/src/config.tsx)
```
vgeddes, line 12: My understanding is that the Javascript app in its current state was purely meant for us to demo functionality (in a scripted manner) in zoom calls, etc. I wouldn't really recommend third-parties use it right now as its relatively tedious to configure, setup, and use.

Its actually much easier to just call this 1-liner 
```
truffle exec scripts/sendEth.js <AMOUNT> <SUB_ACCOUNT_ID>
```

docs/local_usage.md
```diff
@@ -0,0 +1,48 @@
+# Running the bridge locally and testing it end to end manually from UI yourself
+
+## Local extension Setup:
+ - The tests setup a ganache server with a preset mnemonic. Add the first key from this mnemonic to your Metamask as an extra account to use for testing. You can find this first key here: ARTEMIS_ETHEREUM_KEY in /polkadot-ethereum/test/docker-compose.yml
```
jakai7, line 4: I think it will add more context if we add links to [ganachee](https://www.trufflesuite.com/ganache) and [Metask](https://docs.metamask.io/guide/#why-metamask) for users who are not familiar with them.

PR #128, opened 2020-10-19
Infrastructure automation with Ansible

vgeddes:
Added ansible playbooks to perform the following:
* Provision EC2 resources
* Deploy the relay chain
* Deploy the parachain

Details:
* Relay chain
  * Simple 2 node setup (alice & bob). On t3.medium instances
  * RPC url is wss://rococo-rpc.snowfork.network (SSL certs issued by LetsEncrypt)
* Parachain
  * 2 collator nodes, on t3.medium instances.
  * RPC url is wss://parachain-rpc.snowfork.network (SSL certs issued by LetsEncrypt)

Chain security: Not much. The relay chain supports Sudo access with the built-in Alice account. The parachain also accepts relayed messages from our builtin `//Relay` account seed. But this is just for testing, we're going to ditch our relay chain when we integrate with the actual Rococo chain.


musnit:
i don't know ansible so can't easily review, but given its just for our testnet, if it works, then that's review enough :)

Also, we discussed it a few months ago I think but it may be worth skimming through https://github.com/gantree-io/gantree-lib-nodejs again to see if it is useful for us, or if there is some ansible code we can copy/reuse. I'm not sure how active it is anymore but if it looks useful/relevant for us, i can reach out to their team to get updates.

musnit (approved):

vgeddes:
> i don't know ansible so can't easily review, but given its just for our testnet, if it works, then that's review enough :)
> 
> Also, we discussed it a few months ago I think but it may be worth skimming through https://github.com/gantree-io/gantree-lib-nodejs again to see if it is useful for us, or if there is some ansible code we can copy/reuse. I'm not sure how active it is anymore but if it looks useful/relevant for us, i can reach out to their team to get updates.

Ah yeah, I forgot about that. Looked over their Ansible code, and there is some stuff we could definitely reuse as we improve our testnet (hardening, robustness, security, etc).

PR #133, opened 2020-10-20
XCMP documentation

vgeddes:
First cut of our XCMP spec.


docs/specs/xcmp.md
```diff
@@ -0,0 +1,83 @@
...
+Effects:
+1. H will withdraw 42 DOT from Alice's local account.
+2. D will perform the exchange, depositing 21 PolkaETH into Alice's local account.
+3. H can take further action depending on the outcome of (2).
+
+## Asset Identification
+
+To support cross-parachain transfers, the consensus system needs to be able to identify the relative location of bridged ethereum assets held within our parachain.
+
+These assets are all stored in our custom multi-asset [pallet](https://sad-curie-a48c3f.netlify.app/artemis_asset/index.html), and are individually identified by 20-byte identifiers. These identifiers will usually but not always correspond to a contract address on the Ethereum side.
```
philipstanislaus, line 62:
```suggestion
These assets are all stored in our custom multi-asset [pallet](https://sad-curie-a48c3f.netlify.app/artemis_asset/index.html), and are individually identified by 20-byte identifiers. These identifiers can be used within [XCMP's MultiAssets](https://github.com/paritytech/xcm-format/blob/master/README.md#multiasset-universal-asset-identifiers). They will usually but not always correspond to a contract address on the Ethereum side.
```

philipstanislaus (commented):
Thanks, added a few questions :)

docs/specs/xcmp.md
```diff
@@ -0,0 +1,83 @@
...
+Parties:
+- H: Home chain (Polkadot-Ethereum Bridge)
+- D: Destination chain
+- R: Reserve chain
+
+Effects:
+1. H will withdraw 21 PolkaETH from Alice's local account.
+2. The sovereign account of H on R will be reduced by 21 PolkaETH.
+3. The sovereign account of D on R will be credited with 21 PolkaETH.
+4. D will mint 21 PolkaETH into Bob's account.
```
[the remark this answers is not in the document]
vgeddes: I think Gav explained reserve-backed transfers a bit better on riot chat:

> Reserve-backed derivatives are destroyed on the chain which is sending. On the reserve chain itself, assets are transferred from the sovereign account of the sending chain to the sovereign account of the receiving chain. Reserve backed derivatives are then minted on the receiving chain, but only because the reserve chain informed the receiving chain that a deposit had been made to its sovereign account.

Note that the scenario I described is only for transfers between parachains linked by a common relay chain. So this means that:

H: Our parachain (polkadot-ethereum bridge)
R: The relay chain which holds sovereign accounts for both H & D.
D: The destination parachain.

The actual ethereum blockchain is not involved at all.

The XMCP v0 spec at https://github.com/paritytech/xcm-format describes the concept of sovereign accounts in more detail.

docs/specs/xcmp.md
```diff
@@ -0,0 +1,83 @@
...
+
+### Alice attempts to exchange 42 DOT for 21 PolkaETH
```
musnit, line 43: this example makes sense but i dont think our chain will support exchanging assets, so may not be relevant to include

docs/specs/xcmp.md
```diff
@@ -0,0 +1,83 @@
...
+  - [Alice attempts to exchange 42 DOT for 21 PolkaETH](#alice-attempts-to-exchange-42-dot-for-21-polkaeth)
+- [Asset Identification](#asset-identification)
+- [Outstanding Issues](#outstanding-issues)
+  - [Numeric precision](#numeric-precision)
+
+## Introduction
+
+Our parachain will support cross-parachain token transfers using [XCMP](https://github.com/paritytech/xcm-format/blob/master/README.md).
+
+Since our parachain endeavours to be trustless, only reserve-backed transfers will be supported. Participating chains will need to agree on a parent chain to hold sovereign (reserve) accounts.
```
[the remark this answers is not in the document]
musnit: i'm not sure SPREE will help with this, but should look into it yeh

musnit, line 24: @vgeddes Parachains wanting to use our bridge need to trust us anyway, which implies that we can force them to use us as the effective chain to hold sovereign accounts. If this simplifies things then we should consider this design too instead of a design in which the relay chain holds the sovereign account. What do you think?

docs/specs/xcmp.md
vgeddes: yeah, good point. I was just looking for more possible use cases beyond transferring PolkaETH/ERC20 to other parachains (and vice-versa).

I remove the use case

musnit:
@vgeddes looks like a good start just for asset transfer into our parachain.

the important piece we still need is asset transfer across to ethereum, and being received from ethereum. this means asset transfer triggered by and and user and by a parachain.

for an asset coming in from ethereum, i suppose our chain will mint that new asset - should it mint into an account for that user on our parachain, or on the relay chain? similarily, for a parachain-owned account, will it work the same or do they need to be specified differently, ie, will the ethereum smart contract need to accept something special in order for the asset to be minted into a parachain-controlled account rather than a user-controlled account?

secondly, for a parachain, will they be able to watch that event/balance themself and so react to the new mint, or do we need to explicitly send them an XCMP to inform them that a new asset has been minted for them? whose side is the responsibility on?

then, for burning and returning an asset, i guess we'll need a new XCMP type, not just the basic asset transfer/withdraw type. can we use one of the existing standards for this or will we need a custom message?

vgeddes:
> the important piece we still need is asset transfer across to ethereum, and being received from ethereum. this means asset transfer triggered by and and user and by a parachain.

My understanding is that XCMP is primarily for horizontal message exchange between parachains in a Polkadot network. While the XCMP spec does seem to suggest that Parachain ⇄ Ethereum or Parachain ⇄ Bitcoin messaging could be a possible future use-case, I think its still very much in the prototype/idea stage.

Given that we already have working Parachain ⇄ Ethereum messaging, not sure we need to redefine it in terms of XCMP. I think its an interesting idea though.

> for an asset coming in from ethereum, i suppose our chain will mint that new asset - should it mint into an account for that user on our parachain, or on the relay chain?

So far with my current envisioned design, the asset will always be minted on the parachain. Users can then request an upward transfer of the asset to the relay chain if desired. That involves vertical message-passing (VMP) which strictly speaking is separate from XCMP.

Its still not clear to me how secondary assets such as PolkaETH and PolkaERC20 will be stored and managed in the relay chain, and even if VMP would support them transferring them or not. Our Rococo parachain has a primary asset called "DEV" (which is probably a testnet version of DOT), and has support for transferring it to and from the relay chain.

So that's something I'd need to look into.

musnit:
i wasnt thinking to redefine in terms of xcmp - just that it would be good
to be able to trigger a  Parachain ⇄ Ethereum also via sending an XCMP
message to our parachain rather than via sending an extrinsic.

and yeh looking into how secondary assets will/can be stored on relay chain
or on other parachains will be good - i guess it may be the case that we
just have secondary assets stored on our parachain under a "external
parachain abc" account and then that parachain is responsible for managing
it's ledger of how all it's eth maps to individual users on its chain.

vgeddes:
I have better understanding now of the of the relay chain with regard to reserve-backed assets. It can only act as the reserve chain for DOT simply because that's the only asset it it capable of storing. For other assets we'll need a parachain to act as a reserve (even our own chain as you suggested earlier).

docs/specs/xcmp.md
vgeddes: Great idea, I like it, especially as it means we don't have to trust another reserve chain and manage sovereign accounts.

PR #134, opened 2020-10-20
Eth header types & hash computation for verifier

Rizziepit:
This PR defines the `Header` and `HeaderId` structs that represent Ethereum block headers. The definition is copied from https://github.com/paritytech/parity-bridges-common/blob/master/primitives/ethereum-poa/src/lib.rs. I made two changes to the original though:
1. I renamed `AuraHeader` to `Header` because:
  a. It's not Aura-specific, and
  b. [This naming clash issue](https://github.com/paritytech/parity-bridges-common/pull/354) is resolved by importing it like `use artemis_ethereum::Header as EthHeader` in the pallet. This renaming is, as far as I'm concerned, part of the dispatchable API definition.
2. I added tests for the `compute_hash` function.

These generic Ethereum-Substrate definitions could prob go in their own crate but I don't think it's worth the effort right now.


parachain/primitives/ethereum/Cargo.toml
```diff
@@ -5,13 +5,16 @@ authors = ["Snowfork <contact@snowfork.com>"]
...
-rlp = { version = "0.4", default-features = false }
+ethbloom = { version = "0.8", default-features = false }
 ethereum-types = { version = "0.9.2", default-features = false }
+hex = { package = "rustc-hex", version = "2.1.0", default-features = false }
+hex-literal = { version = "0.3.1", default-features = false }
 libsecp256k1 = { version = "0.3", default-features = false }
+parity-bytes = { version = "0.1", default-features = false }
+rlp = { version = "0.4", default-features = false }
+serde = { version = "1.0", optional = true }
+serde-big-array = { version = "0.2", optional = true }
```
Rizziepit, line 17: I sorted the dependencies but let me know if you folks would prefer I revert this to keep blame clean

parachain/README.md
```diff
@@ -153,6 +153,25 @@ For interacting with our chain using the Polkadot-JS API, you'll need to supply
...
+    "ommersHash": "H256",
+    "extraData": "Vec<u8>",
+    "stateRoot": "H256",
+    "receiptsRoot": "H256",
+    "logBloom": "Bloom",
+    "gasUsed": "U256",
+    "gasLimit": "U256",
+    "difficulty": "U256",
+    "seal": "Vec<Vec<u8>>"
+  },
```
Rizziepit, line 172: This was intended to be in my next PR.... whoops... I think it's harmless here though

parachain/primitives/ethereum/Cargo.toml
philipstanislaus: LGTM

parachain/primitives/ethereum/src/lib.rs
```diff
@@ -27,3 +39,182 @@ impl From<ethabi::Error> for DecodeError {
...
+pub struct HeaderId {
+	/// Header number.
+	pub number: u64,
+	/// Header hash.
+	pub hash: H256,
+}
+
+/// An Ethereum block header.
+#[derive(Clone, Default, Encode, Decode, PartialEq, RuntimeDebug)]
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
```
philipstanislaus, line 54: Is using std features an issue here? Why do we need them?

philipstanislaus (commented):
LGTM

parachain/primitives/ethereum/src/lib.rs
Rizziepit: We don't strictly need `Serialize` and `Deserialize` afaik, but it might be convenient for development to load headers from e.g. JSON. It's mostly here because I copied the definition from [parity-bridges-common](https://github.com/paritytech/parity-bridges-common/blob/master/primitives/ethereum-poa/src/lib.rs) though.

W.r.t. the use of std features... are you familiar with Substrate's use of "std" and "no_std" to differentiate a normal build and WASM build? serde cannot be compiled to WASM. So we exclude its use in the WASM build. I'm not sure if you're familiar with how Substrate uses WASM. Let me know if you want more info

parachain/primitives/ethereum/Cargo.toml
Rizziepit: Makes sense. I'll refrain from sorting more stuff

parachain/primitives/ethereum/src/lib.rs
vgeddes: I suppose the `derive(Serialize, Deserialize)` could be useful for unit tests which definitely do have `std` available.

parachain/primitives/ethereum/src/lib.rs
```diff
@@ -27,3 +39,182 @@ impl From<ethabi::Error> for DecodeError {
...
+
+
+#[cfg(test)]
+mod tests {
+
+	use super::*;
+	use hex_literal::hex;
+
+	#[test]
+	fn header_compute_hash_poa() {
```
musnit, line 162: do we need this?

Rizziepit: Unclear... the test is intended to demonstrate that the `Header` definition is consensus-engine-agnostic. The `seal` field captures either `mixHash` + `nonce` for PoW, `step` + `signature` for PoA, etc. I kept the `seal` field because of this description: https://ethereum.stackexchange.com/questions/6093/what-does-it-mean-to-seal-a-block. Happy to change if you think it's better to make this PoW-specific

musnit: ah right, nah that's fine then. is there anything in this PR that you think we should try get changed back in parity-bridges-common?

Rizziepit: No. I made no significant changes except adding tests

musnit:
@Rizziepit feel free to merge your own prs once they're reviewed/approved

PR #147, opened 2020-11-06
Ethash verification for light client

Rizziepit:
This implements Ethash validation using merkle-proof-of-DAG-items instead of the standard "light client" implementation described in https://eth.wiki/en/concepts/ethash/ethash. Note that this adds complexity to the relayer because it has to provide the proof data. Fortunately there's already a Go implementation of this: https://github.com/talbaneth/ethashproof. 

Features:
* `artemis_ethereum::ethashproof::EthashProver` with `hashimoto_light` and `hashimoto_merkle` functions to check whether difficulty target has been met
  * This uses the `rust-ethash` package and copies logic from rainbow-bridge (https://github.com/near/rainbow-bridge/blob/master/contracts/near/eth-client/src/lib.rs)
* `artemis_ethereum::ethashdata::DAGS_MERKLE_ROOTS` contains merkle roots of first 512 epochs (gives us ~1.5 years runway)
* Struct `DoubleNodeWithMerkleProof` to work with proof data 
* Struct `BlockWithProofs` to deserialize proof data from relayer (right now for tests only)
* Tests using Ethereum mainnet blocks


parachain/pallets/verifier-lightclient/Cargo.toml
```diff
@@ -12,12 +12,21 @@ targets = ["x86_64-unknown-linux-gnu"]
...
 [dependencies.codec]
 default-features = false
 features = ["derive"]
 package = "parity-scale-codec"
 version = "1.3.4"
 
+[dependencies.ethash]
+default-features = false
+git = "https://github.com/nearprotocol/rust-ethash"
+branch = "upgrade-eth-types"
```
Rizziepit, line 28: Should I create an issue to pin this version?

vgeddes: yeah, I think go ahead and create an issue for that. I think it may be safer to fork their repo though, since in this case we're depending on a development branch of someone else's repo. That branch could be deleted or updated under our feet at any time.

parachain/pallets/verifier-lightclient/src/mock.rs
```diff
@@ -1,6 +1,8 @@
 // Mock runtime
+#![cfg(test)]
```
vgeddes, line 2: Any reason you added this here? It shouldn't be necessary in this context, since `mock.rs` is already marked to be conditionally compiled due to https://github.com/Snowfork/polkadot-ethereum/blob/eth-header-verification-3/parachain/pallets/verifier-lightclient/src/lib.rs#L20

parachain/pallets/verifier-lightclient/src/tests.rs
```diff
@@ -1,12 +1,15 @@
+#![cfg(test)]
```
vgeddes, line 1: ditto

parachain/primitives/testutils/src/lib.rs
```diff
@@ -0,0 +1,129 @@
...
+                .map(|e| e.into())
+                .collect(),
+        }
+    }
+}
+
+impl BlockWithProofs {
+    pub fn from_file(filename: &str) -> Self {
+        let raw: BlockWithProofsRaw =
+            serde_json::from_reader(std::fs::File::open(std::path::Path::new(filename)).unwrap()).unwrap();
```
vgeddes, line 92: No need to use `Path::new` here since `&str` implements `AsRef<Path>`: https://doc.rust-lang.org/std/convert/trait.AsRef.html#impl-AsRef%3CPath%3E

```suggestion
            serde_json::from_reader(std::fs::File::open(filename).unwrap()).unwrap();
```

parachain/pallets/verifier-lightclient/src/mock.rs
```diff
@@ -86,13 +88,31 @@ pub fn child_of_genesis_ethereum_header() -> EthereumHeader {
 	child	
 }
 
-pub fn new_tester() -> sp_io::TestExternalities {
-	let mut storage = system::GenesisConfig::default().build_storage::<MockRuntime>().unwrap();
+pub fn ethereum_header_from_file(block_num: u64) -> EthereumHeader {
+	let filename = format!("./src/testdata/{}.json", block_num);
+	serde_json::from_reader(
+		std::fs::File::open(std::path::Path::new(&filename)).unwrap()
```
vgeddes, line 94: No need to use Path::new here since `str` implements `AsRef<Path>`: https://doc.rust-lang.org/std/convert/trait.AsRef.html#impl-AsRef%3CPath%3E

Basically since `str` implements `AsRef<Path>`, `File::open` knows that it can call `filename.as_ref()` to create a `Path` object.

vgeddes (commented):
Looks great. Awesome test coverage! 😄

parachain/pallets/verifier-lightclient/src/mock.rs
Rizziepit: Ha, I didn't actually notice the imports and thought we were compiling these in the regular build. Good to know we're not :D

parachain/pallets/verifier-lightclient/src/lib.rs
```diff
@@ -124,7 +129,38 @@ impl<T: Trait> Module<T> {
...
 		);
 
-		// TODO check PoW
+		if !VerifyPoW::get() {
+			return Ok(());
+		}
+
+		// Adapted from https://github.com/near/rainbow-bridge/blob/3fcdfbc6c0011f0e1507956a81c820616fb963b4/contracts/near/eth-client/src/lib.rs#L363
+		// See YellowPaper formula (50) in section 4.3.4
+		let parent = Headers::<T>::get(header.parent_hash).unwrap().header;
```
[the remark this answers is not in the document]
Rizziepit: I'll fix it up. This reminds me of two related questions I had:
1. Can pallet logic be executed concurrently? I'm checking for the existence of the key earlier in the function. So unwrap should be safe if there's no concurrency. Might still be poor style to use `unwrap` though.
2. Is there a Substrate alternative for `std::panic::catch_unwind`? `EthashProver` can panic because the `ethash` API doesn't provide a way to return `Result` instead. I can create a PR to change the API but seems like overkill to please Substrate

parachain/pallets/verifier-lightclient/Cargo.toml
Rizziepit: Let me actually check if tests pass with the latest published version. It uses an old version of `ethereum-types` and `primitive-types`.

Rizziepit: Nope. I'll make an issue

Rizziepit: https://github.com/Snowfork/polkadot-ethereum/issues/148

parachain/pallets/verifier-lightclient/src/lib.rs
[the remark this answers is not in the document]
vgeddes: > * Can pallet logic be executed concurrently? I'm checking for the existence of the key earlier in the function. So unwrap should be safe if there's no concurrency. Might still be poor style to use `unwrap` though.

Pallet logic in the runtime is executed serially. However its important to never use `unwrap` in Substrate, as it can cause a panic and undefined behavior. This could be exploited to launch DOS attacks on the chain.

> * Is there a Substrate alternative for `std::panic::catch_unwind`? `EthashProver` can panic because the `ethash` API doesn't provide a way to return `Result` instead. I can create a PR to change the API but seems like overkill to please Substrate

I'm pretty sure there isn't an equivalent of `catch_unwind` in substrate. Even in normal Rust, `catch_unwind` won't catch all panics according to the docs.

If `EthashProver` and any of its dependencies can potentially panic we'll need to fix that. What particular `ethash` API was causing issues for `EthashProver`?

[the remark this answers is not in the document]
musnit: @Rizziepit do you have a clear idea of what circumstances would result in a panic?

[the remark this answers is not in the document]
Rizziepit: @musnit any incorrect or malformed proof will cause a panic. But I just thought of a way to avoid panic in Ethash and return `Result` from `EthashProver`. I'll update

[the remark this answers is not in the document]
Rizziepit: I've updated `EthashProver` and related code so there are no more panics. @vgeddes @musnit let me know if you're happy for me to merge this

parachain/primitives/ethereum/src/ethashproof.rs
```diff
@@ -0,0 +1,353 @@
...
+    pub fn hashimoto_light(
+        &mut self,
+        header_hash: H256,
+        nonce: H64,
+        header_number: u64,
+    ) -> (H256, H256) { 
+        let epoch = header_number / EPOCH_LENGTH;
+        let cache = match self.dags_cache {
+            Some(ref mut c) => c.get(epoch, header_number),
+            None => panic!("EthashProver wasn't configured with hashimoto light cache"),
```
vgeddes, line 225: One more panic here. I know the `hashimoto_light` function isn't being used anywhere yet, but lets fix it now so we don't accidently cause problems in the future.

parachain/primitives/ethereum/src/ethashproof.rs
```diff
@@ -0,0 +1,353 @@
...
+        &self,
+        header_hash: H256,
+        nonce: H64,
+        header_number: u64,
+        nodes: &[DoubleNodeWithMerkleProof],
+    ) -> Result<(H256, H256), &'static str> {
+        // Check that we have the expected number of nodes with proofs
+        const MIXHASHES: usize = MIX_BYTES / HASH_BYTES;
+        if nodes.len() != MIXHASHES * ACCESSES / 2 {
+            return Err("Incorrect number of nodes");
```
vgeddes, line 166: Rather than using strings as the error type, can we introduce an enum like this:

```rust
#[derive(Debug, Clone)]
enum Error {
  UnexpectedNumberOfNodes,
  InvalidProof,
}
```
And then return either `Err(Error::UnexpectedNumberOfNodes)` or `Err(Error::InvalidMerkleProof)` as appropriate.

This makes it easier to pattern the match on possible errors returned by `hashimoto_merkle`.

parachain/pallets/verifier-lightclient/src/lib.rs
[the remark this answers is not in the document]
vgeddes: Thanks Riz, looks much better. Though I added another comment about using strings vs enums as an error type.

parachain/pallets/verifier-lightclient/src/lib.rs
```diff
@@ -124,7 +129,42 @@ impl<T: Trait> Module<T> {
...
+
+		// Simplified difficulty check to conform adjusting difficulty bomb
+		let header_mix_hash = header.mix_hash().ok_or(Error::<T>::InvalidHeader)?;
+		let header_nonce = header.nonce().ok_or(Error::<T>::InvalidHeader)?;
+		let (mix_hash, result) = EthashProver::new().hashimoto_merkle(
+			header.compute_partial_hash(),
+			header_nonce,
+			header.number,
+			proof,
+		).map_err(|e| Error::<T>::InvalidHeader)?;
```
vgeddes, line 160: Linter is going to complain about `e` not being used.

```suggestion
		).map_err(|_| Error::<T>::InvalidHeader)?;
```

parachain/primitives/ethereum/src/ethashproof.rs
musnit: @vgeddes i'm totally guessing, but this panic looks like it would only occur if the cache is not set, which should only happen if the chain is not initialized correctly or if the cache expires after a few years, in which case the bridge would actually be broken and we likely do want a panic causing the parachain to halt?

@Rizziepit can you confirm?

Rizziepit: @musnit correct-ish. It simply means the struct wasn't initialized using the below:
```
let prover = EthashProver::with_hashimoto_light(10);
```

I coded it like this because I don't want the `hashimoto_light` function to work by default. An engineer needs to first decide what size cache they can tolerate. There's also the issue that the `EthashProver` needs to be a thread-safe singleton for the cache to be re-used. I didn't delve into how to implement that because I expect Substrate will complicate the matter.

Suffice to say, that panic indicates engineering error and cannot be caused by user input. The node is unusable. I'm not sure `Result` is appropriate. Thoughts, @vgeddes?

vgeddes: Sounds good. I understand now :+1:

Rizziepit:
Error enum in https://github.com/Snowfork/polkadot-ethereum/pull/147/commits/8f19d0f01f9553d43f0278574dbc3efc1b2e2d2a. Let me know if you're happy for me to merge @vgeddes

vgeddes (approved):
LGTM! 🚀 

You can merge it in.

parachain/primitives/ethereum/src/ethashproof.rs
```diff
@@ -0,0 +1,363 @@
...
+            Some(ref mut c) => c.get(epoch, header_number),
+            None => panic!("EthashProver wasn't configured with hashimoto light cache"),
+        };
+        let full_size = ethash::get_full_size(epoch as usize);
+        return ethash::hashimoto_light(header_hash, nonce, full_size, cache.as_slice());
+    }
+}
+
+// https://github.com/paritytech/substrate/commit/510e68b8d06a3d407eda0d4c1c330bd484140b65
+fn keccak_512(data: &[u8]) -> [u8; 64] {
```
tomusdrw, line 243: I think this should rather be a host function (`sp_io::hashing`) for performance reasons.

tomusdrw: I left a comment in the linked PR, cause I think it was an oversight to not include it there.

musnit: Cool, will add an issue to our repo here and please keep us updated when/if this host function is added for us to use

tomusdrw: Added a PR here https://github.com/paritytech/substrate/pull/7531

drewstone:
Curious why you rolled your own ethhash instead of using existing rust-ethereum? I've been building this over here https://github.com/hicommonwealth/substrate-eth-light-client, and had previously also used the rainbow bridge template before swapping it out directly for Wei's ethash implementation. The extrinsic is extremely heavy tho, is that why maybe?

parachain/primitives/ethereum/src/ethashdata.rs
```diff
@@ -0,0 +1,518 @@
...
+    hex!("09c8e0bdd1ba73602063d89d11ff5210"),
+    hex!("ecab8f890f1f3d6ff564eebab034ebb3"),
+    hex!("9823e2e8911c5a0fb2ff6feb52be2c0f"),
+    hex!("4fa03dbaabcf99e71903f3177444bfaa"),
+    hex!("a4c9667d00998d262c47f6d2c79f43c2"),
+    hex!("b3e65ee26384d25bc7750b3f37e72883"),
+    hex!("7dea46d1a183f3070760eee0bf5fce3c"),
+    hex!("ecb2963c22a757569fe659635f4b0243"),
+    hex!("9ed968203144981e6e697db052910c27"),
+    hex!("4aa6ca6ebef942d8766065b2e590fd32"),
```
philipstanislaus, line 517: Could you create an issue that we need to update these roots in around 1.5 years?

philipstanislaus (commented):
Bit late to the party – many thanks, this looks great!

Just one minor comments for a future PR.

parachain/primitives/ethereum/src/ethashdata.rs
Rizziepit: Will do

Rizziepit:
> Curious why you rolled your own ethhash instead of using existing rust-ethereum? I've been building this over here https://github.com/hicommonwealth/substrate-eth-light-client, and had previously also used the rainbow bridge template before swapping it out directly for Wei's ethash implementation. The extrinsic is extremely heavy tho, is that why maybe?

@drewstone see discussion at https://github.com/paritytech/parity-bridges-common/issues/324. We are actually using https://github.com/rust-ethereum/ethash. See `hashimoto_with_hasher`

parachain/primitives/ethereum/src/ethashdata.rs
Rizziepit: Done https://github.com/Snowfork/polkadot-ethereum/issues/158. I mentioned Q1 2022 as the deadline for this which should provide a few months' leeway before the node stops accepting headers

PR #150, opened 2020-11-10
Ethash configuration via runtime parameter

Rizziepit:
This diff changes the way we configure the Eth lightclient pallet. The intention is to use this approach for pruning / finalization configuration and whatever else we need in future.

**What I was doing before**: storing a `verify_pow` bool in storage. This is convenient but causes us to read from storage unnecessarily.

**What I'm doing now**: hardcoding it for the runtime in `impl verifier_lightclient::Trait for Runtime`. This mimics parity-bridges-common. The drawback is that multiple configurations (e.g. for tests) require more runtime implementation boilerplate.

### DIFF
--- a/parachain/pallets/verifier-lightclient/src/lib.rs
+++ b/parachain/pallets/verifier-lightclient/src/lib.rs
@@ -38,7 +38,7 @@ pub struct StoredHeader<Submitter> {
 pub trait Trait: system::Trait {
 	type Event: From<Event> + Into<<Self as system::Trait>::Event>;
 	// Determines whether Ethash PoW is verified for headers
-	// NOTE: true by default, should only be false for dev
+	// NOTE: Should only be false for dev
 	type VerifyPoW: Get<bool>;
 }
 
--- a/parachain/pallets/verifier-lightclient/src/lib.rs
+++ b/parachain/pallets/verifier-lightclient/src/lib.rs
@@ -23,19 +23,6 @@ mod mock;
 #[cfg(test)]
 mod tests;
 
-#[derive(Clone, PartialEq, RuntimeDebug)]
-pub struct EthashConfiguration {
-	// Determines whether Ethash PoW is verified for headers
-	// NOTE: true by default, should only be false for dev
-	pub verify_pow: bool,
-}
-
-impl Default for EthashConfiguration {
-	fn default() -> Self {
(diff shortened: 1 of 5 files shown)


parachain/pallets/verifier-lightclient/src/mock.rs
```diff
@@ -67,13 +72,49 @@ impl system::Trait for MockRuntime {
...
+	type AvailableBlockRatio = AvailableBlockRatio;
+	type Version = ();
+	type ModuleToIndex = ();
+	type AccountData = ();
+	type OnNewAccount = ();
+	type OnKilledAccount = ();
+	type SystemWeightInfo = ();
+}
+
+
```
vgeddes, line 103: Can we move the parameter types down here from above:

```suggestion
parameter_types! {
	pub EthashDisabledConfiguration: EthashConfiguration = EthashConfiguration { verify_pow: false };
	pub EthashEnabledConfiguration: EthashConfiguration = EthashConfiguration::default();
}
```

parachain/pallets/verifier-lightclient/src/lib.rs
```diff
@@ -23,6 +23,19 @@ mod mock;
 #[cfg(test)]
 mod tests;
 
+#[derive(Clone, PartialEq, RuntimeDebug)]
+pub struct EthashConfiguration {
```
vgeddes, line 27: Is this configuration struct really necessary? It only defines 1 field and we could define it on the pallet `Trait` instead (this is kinda what is Trait is meant to be used for).

```rust
pub trait Trait: system::Trait {
	type Event: From<Event> + Into<<Self as system::Trait>::Event>;
	/// Ethash PoW configuration
	type VerifyPoW: Get<bool>;
}
```

Rizziepit: You're right. I based this on parity-bridges-common: https://github.com/paritytech/parity-bridges-common/blob/master/modules/ethereum/src/lib.rs#L375.

Theoretically there are a number of params for Ethash that we might want to make configurable. But we can cross that bridge when we get there. I'll update

vgeddes (approved):
LGTM!

philipstanislaus (commented):
LGTM!

PR #156, opened 2020-11-12
Eth header finalization & pruning

Rizziepit:
This PR adds tracking of finalized Ethereum headers and pruning of old finalized headers. A header is considered final when it's in the highest difficulty chain and has at least 35 descendants (35 is the number of "confirmations" that Coinbase requires).

### DIFF
--- a/parachain/pallets/verifier-lightclient/src/lib.rs
+++ b/parachain/pallets/verifier-lightclient/src/lib.rs
@@ -37,6 +37,9 @@ pub struct StoredHeader<Submitter> {
 
 pub trait Trait: system::Trait {
 	type Event: From<Event> + Into<<Self as system::Trait>::Event>;
+	/// The number of descendents, in the highest difficulty chain, a block
+	/// needs to have in order to be considered final.
+	type DescendantsUntilFinalized: Get<u8>;
 	// Determines whether Ethash PoW is verified for headers
 	// NOTE: Should only be false for dev
 	type VerifyPoW: Get<bool>;
@@ -46,6 +49,8 @@ decl_storage! {
 	trait Store for Module<T: Trait> as VerifierModule {
 		/// Best known block.
 		BestBlock: (EthereumHeaderId, U256);
+		/// Best finalized block.
+		FinalizedBlock: EthereumHeaderId;
 		/// Map of imported headers by hash.
 		Headers: map hasher(identity) H256 => Option<StoredHeader<T::AccountId>>;
 		/// Map of imported header hashes by number.
@@ -59,14 +64,16 @@ decl_storage! {
 		build(|config| {
 			let initial_header = &config.initial_header;
 			let initial_hash = initial_header.compute_hash();
+			let initial_id = EthereumHeaderId {
+				number: initial_header.number,
+				hash: initial_hash,
+			};
 
 			BestBlock::put((
-				EthereumHeaderId {
-					number: initial_header.number,
-					hash: initial_hash,
-				},
+				initial_id,
 				config.initial_difficulty,
 			));
+			FinalizedBlock::put(initial_id);
 			Headers::<T>::insert(
 				initial_hash,
 				StoredHeader {
@@ -91,10 +98,14 @@ decl_event!(
 
 decl_error! {
 	pub enum Error for Module<T: Trait> {
+		/// Header is same height or older than finalized block (we don't support forks).
+		AncientHeader,
 		/// Header's parent has not been imported.
 		MissingParentHeader,
 		/// Header has already been imported.
 		DuplicateHeader,
+		/// Header is on a stale fork, i.e. it's not a descendant of the latest finalized block
+		HeaderOnStaleFork,
 		/// One or more header fields are invalid.
 		InvalidHeader,
 	}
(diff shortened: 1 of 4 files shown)


parachain/pallets/verifier-lightclient/src/lib.rs
```diff
@@ -23,6 +23,11 @@ mod mock;
 #[cfg(test)]
 mod tests;
 
+/// Max number of finalized headers to keep.
+const FINALIZED_HEADERS_TO_KEEP: u64 = 5_000;
+/// Max number of headers we're pruning in single import call.
+const HEADERS_TO_PRUNE_IN_SINGLE_IMPORT: u64 = 8;
```
Rizziepit, line 29: These two values come from parity-bridges-common

parachain/pallets/verifier-lightclient/src/lib.rs
```diff
@@ -197,17 +248,125 @@ impl<T: Trait> Module<T> {
...
+			if i < required_descendants {
+				continue;
+			}
+			if ancestor.number > finalized_block_id.number {
+				new_finalized_block_id = EthereumHeaderId {
+					hash: ancestor_hash,
+					number: ancestor.number,
+				};
+			} else {
+				assert!(ancestor_hash == finalized_block_id.hash);
```
vgeddes, line 301: I'm still a bit uncomfortable with using asserts in production code. Even if theoretically there is no chance of a panic occuring. External reviewers/auditors may also raise concerns and ask about their implications.

As I demonstrated in the comment above, maybe raise a `Error::<T>::Undefined` in all cases where we would panic?

parachain/pallets/verifier-lightclient/src/lib.rs
```diff
@@ -119,26 +152,44 @@ decl_module! {
...
+		let finalized_header_id = FinalizedBlock::get();
+		ensure!(
+			header.number > finalized_header_id.number,
+			Error::<T>::AncientHeader,
+		);
+
+		// TODO: limit N where N = header.number - finalized_header.number
+		// to avoid iterating over long chains here
+		for (ancestor_hash, ancestor) in ancestry::<T>(header.parent_hash) {
+			if ancestor.number > finalized_header_id.number {
```
vgeddes, line 174: This code is duplicated in `get_best_finalized_header`. Could use the `best_descendant` method I wrote in another comment to factor out this logic?

parachain/pallets/verifier-lightclient/src/lib.rs
```diff
@@ -119,26 +152,44 @@ decl_module! {
...
+
+		let finalized_header_id = FinalizedBlock::get();
+		ensure!(
+			header.number > finalized_header_id.number,
+			Error::<T>::AncientHeader,
+		);
+
+		// TODO: limit N where N = header.number - finalized_header.number
+		// to avoid iterating over long chains here
+		for (ancestor_hash, ancestor) in ancestry::<T>(header.parent_hash) {
```
vgeddes, line 173: Can simplify with `find_map` iterator function:
```rust
let foo = ancestry::<T>(header.parent_hash)
    .find_map(|(hash, ancestor)|
        if ancestor.number <= finalized_header_id.number {
            Some((hash, ancestor))
        } else {
            None
        }
    )

// handle result
match foo{
    Some(...) => ...
    None => ...
}
```

parachain/pallets/verifier-lightclient/src/lib.rs
```diff
@@ -197,17 +248,125 @@ impl<T: Trait> Module<T> {
...
+	// Return the latest block that can be finalized based on the given
+	// highest difficulty chain and previously finalized block.
+	fn get_best_finalized_header(
+		best_block_id: &EthereumHeaderId,
+		finalized_block_id: &EthereumHeaderId,
+	) -> EthereumHeaderId {
+		let mut new_finalized_block_id = finalized_block_id.clone();
+
+		let required_descendants = T::DescendantsUntilFinalized::get() as usize;
+		for (i, (ancestor_hash, ancestor)) in ancestry::<T>(best_block_id.hash).enumerate() {
```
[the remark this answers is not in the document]
Rizziepit: Let me see if I can make it work

parachain/pallets/verifier-lightclient/src/lib.rs
Rizziepit: This is only temporary while I test syncing with mainnet via the relayer. I'd prefer to create an issue to remove all assertions in a week or two. These assertions are saving me time right now. Thoughts?

vgeddes: Ok, that makes sense, sounds good 👍🙂

parachain/pallets/verifier-lightclient/src/lib.rs
```diff
@@ -197,17 +248,125 @@ impl<T: Trait> Module<T> {
...
+
+	// Return the latest block that can be finalized based on the given
+	// highest difficulty chain and previously finalized block.
+	fn get_best_finalized_header(
+		best_block_id: &EthereumHeaderId,
+		finalized_block_id: &EthereumHeaderId,
+	) -> EthereumHeaderId {
+		let mut new_finalized_block_id = finalized_block_id.clone();
+
+		let required_descendants = T::DescendantsUntilFinalized::get() as usize;
```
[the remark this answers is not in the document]
vgeddes: Actually I think I was wrong on this one. Since `u8` converts losslessly into `usize` with `as`, what you have it totally fine.

philipstanislaus (commented):
Nice, especially the iterator and tests 👍

parachain/pallets/verifier-lightclient/src/lib.rs
```diff
@@ -119,26 +152,44 @@ decl_module! {
...
+		);
+
+		// TODO: limit N where N = header.number - finalized_header.number
+		// to avoid iterating over long chains here
+		for (ancestor_hash, ancestor) in ancestry::<T>(header.parent_hash) {
+			if ancestor.number > finalized_header_id.number {
+				continue;
+			}
+			// This assertion holds because the AncientHeader check above ensures that
+			// iteration starts at or after the latest finalized block.
```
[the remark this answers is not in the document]
philipstanislaus: Nevermind, has been fixed already.

PR #159, opened 2020-11-17
Basic forwarding of Ethereum headers to Substrate

Rizziepit:
This PR implements very simple forwarding of Ethereum headers to the Substrate Eth Lightclient. This entails:
* New SCALE-encodable structs for headers and their proof data
* A new channel for headers and the receiving / sending logic
* Flags `--direction` and `--headers` to run the relay in one direction only, and forward headers only
* Some trace logging in Substrate

The command to forward only headers from Eth -> Sub:
```
build/artemis-relay run --config ./.config/relay-config.toml --headers --direction 1
```

NOTE: The Eth -> Sub direction is fully implemented. Sub -> Eth needs handlers for the header channel in the Sub `Listener` and Eth `Writer`


Rizziepit:
@vgeddes I'd like your thoughts on this refactor of the relay. I figured it's reasonable to provide the option of running the relay in one direction only, and to forward headers only. This avoids the need to configure ABI files / use Truffle. So this draft PR adds flags `--direction` and `--headers` and configures the chains with a channel for headers. Yay / nay?

musnit:
@Rizziepit haven't looked through the code but I think making our relayer configurable and capable of being started in different modes for either single/multiple directions and for relaying different kinds of proof data/events/etc is the right move

relayer/cmd/root.go
```diff
@@ -24,6 +24,12 @@ var rootCmd = &cobra.Command{
 func init() {
 	cobra.OnInitialize(initConfig)
 	rootCmd.PersistentFlags().StringVar(&configFile, "config", "", "config file")
+	rootCmd.PersistentFlags().Int(
```
vgeddes, line 27: Since these `direction` and ` headers` flags are only meaningful for the `run` subcommand, can we move them over to `cmd/run.go`?

I left the `config` flag on the root command just in case we add more subcommands that rely on a common config.

relayer/cmd/root.go
```diff
@@ -24,6 +24,12 @@ var rootCmd = &cobra.Command{
...
+		"direction",
+		0,
+		"Relay messages bi-directionally (0), from Eth to Sub (1), or from Sub to Eth (2)",
+	)
+	rootCmd.PersistentFlags().Bool("headers", false, "Only forward headers")
```
vgeddes, line 32: I'd rename this flag to `--headers-only`.

As an aside, in a future iteration we will probably want to convert this into a `mode` flag which allows one to specify headers, events, or both.

relayer/chain/chain.go
```diff
@@ -14,8 +14,15 @@ type Message struct {
...
+type Header struct {
+	HeaderData interface{}
+	ProofData  interface{}
+}
+
 type Chain interface {
 	Name() string
 	Start(ctx context.Context, eg *errgroup.Group) error
 	Stop()
+	WithReceiver(messages chan Message, headers chan Header) error
```
vgeddes, line 26: Would prefer to name these `SetReceiver` and `SetSender` respectively since they mutate state. In my view, `With` is often used in constructor/factory method names when an object is being created with some input state.

relayer/chain/ethereum/header.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+
+	header := Header{
+		ParentHash:       types.NewH256(gethheader.ParentHash.Bytes()),
+		Timestamp:        types.NewU64(gethheader.Time),
+		Number:           types.NewU64(gethheader.Number.Uint64()),
+		Author:           types.NewH160(gethheader.Coinbase.Bytes()),
+		TransactionsRoot: types.NewH256(gethheader.TxHash.Bytes()),
+		OmmersHash:       types.NewH256(gethheader.UncleHash.Bytes()),
+		ExtraData:        gethheader.Extra,
```
vgeddes, line 70: Just to be more explicit:
```suggestion
		ExtraData:        types.NewBytes(gethheader.Extra),
```

parachain/pallets/verifier-lightclient/src/lib.rs
```diff
@@ -143,8 +143,44 @@ decl_module! {
 		#[weight = 0]
 		pub fn import_header(origin, header: EthereumHeader, proof: Vec<EthashProofData>) -> DispatchResult {
 			let sender = ensure_signed(origin)?;
-			Self::validate_header_to_import(&header, &proof)?;
-			Self::import_validated_header(&sender, &header)
+
+			frame_support::debug::trace!(
```
vgeddes, line 147: Can we import `frame_support::debug` so that we don't have to specificy the full path everytime?

vgeddes (approved):
LGMT! 🚀

relayer/chain/chain.go
Rizziepit: Good catch :) This started off as factory methods until I realized I want it to return an error

relayer/cmd/root.go
Rizziepit: Something to consider: most events probably cannot be validated without the node being able to cross-reference the relevant header/block. So, if you run the relay with events only, you'll need to sync headers some other way

musnit: Yeh all the events stuff will be deprecated/modified to support verification with cross reference one we have transaction receipt proofs working so np for now

relayer/chain/ethereum/listener.go
```diff
@@ -20,47 +20,61 @@ type Listener struct {
...
+			}).Info("Subscribed to contract events")
 		}
+	}
 
-		li.log.WithFields(logrus.Fields{
-			"contractAddress": contract.Address.Hex(),
-			"contractName":    contract.Name,
-		}).Info("Subscribed to contract events")
+	headers := make(chan *gethTypes.Header)
+	_, err := li.conn.client.SubscribeNewHead(ctx, headers)
```
vgeddes, line 71: Oh I missed a problem during my previous review. See https://github.com/Snowfork/polkadot-ethereum/pull/163 for more context.

The problem is that background subscriptions created by `li.conn.client.SubscribeNewHead` can fail if the connection to Infura drops for any reason.

Can you add code like the following as a stop-gap measure (or anything else as you see fit)?
https://github.com/Snowfork/polkadot-ethereum/pull/163/files#diff-30d1bf781a4c9c5a1e0134c30fcdf147722200659b16c9cdd4a62b9390e22809R67

```golang
		case err := <-subscription.Err():
			li.log.WithError(err).Error("Subscription terminated")
			return err
```

It will make sure the relayer is terminated after a connection drop (so it can be restarted by Systemd).

relayer/chain/chain.go
```diff
@@ -14,8 +14,15 @@ type Message struct {
 	Payload interface{}
 }
 
+type Header struct {
```
musnit, line 17: We may want these types to be Ethereum specific, as for the other direction we're not really relaying headers. Can leave it as is for now and refactor once we get to implementing the other direction though.

relayer/chain/ethereum/header.go
```diff
@@ -0,0 +1,88 @@
...
+		return nil, err
+	}
+
+	header := Header{
+		ParentHash:       types.NewH256(gethheader.ParentHash.Bytes()),
+		Timestamp:        types.NewU64(gethheader.Time),
+		Number:           types.NewU64(gethheader.Number.Uint64()),
+		Author:           types.NewH160(gethheader.Coinbase.Bytes()),
+		TransactionsRoot: types.NewH256(gethheader.TxHash.Bytes()),
+		OmmersHash:       types.NewH256(gethheader.UncleHash.Bytes()),
```
musnit, line 69: Ommers :P

relayer/chain/ethereum/listener.go
Rizziepit: I'll update this today

relayer/chain/ethereum/header.go
Rizziepit: Not sure if you're familiar with this naming? It's from the yellow paper. Pity it isn't used everywhere. I'm going to keep using it for the sake of gender neutrality

musnit: Oh I thought it was just Afrikaans for uncle, not gender neutral 😅

Rizziepit: Lol def not Afrikaans

Rizziepit:
@vgeddes I've added subscription error handling for headers now. The code's a bit gross because the events subscription is optional, i.e. the events subscription error channel can be nil. But I think it's still manageable

vgeddes (approved):
Looks good!

relayer/chain/ethereum/listener.go
```diff
@@ -20,52 +20,74 @@ type Listener struct {
...
+			}).Debug("Subscribed to contract events")
+		}
 	}
 
-	for _, contract := range li.contracts {
-		li.log.WithFields(logrus.Fields{
-			"addresses":    contract.Address.Hex(),
-			"contractName": contract.Name,
-		}).Debug("Subscribed to contract events")
+	li.log.Info("Polling headers started")
```
philipstanislaus, line 73:
```suggestion
	li.log.Info("Polling events started")
	li.log.Info("Polling headers starting...")
```

relayer/chain/ethereum/listener.go
```diff
@@ -20,52 +20,74 @@ type Listener struct {
...
+	events := make(chan gethTypes.Log)
+	var eventsSubscriptionErr <-chan error
+	headers := make(chan *gethTypes.Header)
+	var headersSubscriptionErr <-chan error
 
-	query := makeFilterQuery(li.contracts)
+	if li.messages == nil {
+		li.log.Info("Not polling events since channel is nil")
+	} else {
+		li.log.Info("Polling events started")
```
philipstanislaus, line 54:
```suggestion
		li.log.Info("Polling events starting...")
```

relayer/chain/ethereum/listener.go
```diff
@@ -20,52 +20,74 @@ type Listener struct {
...
-			"contractName": contract.Name,
-		}).Debug("Subscribed to contract events")
+	li.log.Info("Polling headers started")
+
+	subscription, err := li.conn.client.SubscribeNewHead(ctx, headers)
+	if err != nil {
+		li.log.WithError(err).Error("Failed to subscribe to new headers")
+		return err
 	}
+	headersSubscriptionErr = subscription.Err()
```
philipstanislaus, line 81:
```suggestion
	li.log.Info("Polling headers started")
	
```

relayer/chain/substrate/chain.go
```diff
@@ -59,14 +67,18 @@ func (ch *Chain) Start(ctx context.Context, eg *errgroup.Group) error {
...
-	if err != nil {
-		return err
+	if ch.writer != nil {
+		err = ch.writer.Start(ctx, eg)
+		if err != nil {
+			return err
+		}
 	}
 
 	return nil
```
philipstanislaus, line 84: Again, error here?

philipstanislaus (approved):
LGTM, added a few minor notes. 🎉

PR #160, opened 2020-11-17
Add tool to generate MMR sample data

vgeddes:
Added a tool which generates mock MMR data. This is just the start and of course we can improve it further in various ways (MMR serialization, printing, etc)

The only caveat is that it only generates balanced/binary merkle trees, which of course are still valid MMRs. Generating real-world MMRs is a bit trickier since there don't seem to be any suitable MMR golang libraries.

*Note: Will probably rewrite this in Rust so we can match better match substrate things*

The MMR consists of 4 leaf nodes, and a single peak:
```
      R
   /    \
 / \    / \       
0   1  2   3
```

Merkle tree implementation uses https://godoc.org/github.com/cbergoon/merkletree.

To compile:
```
go build -o build/gen-mmr tools/gen-mmr.go
```

To run:
```
build/gen-mmr
```

### DIFF
--- a/relayer/tools/gen-mmr.go
+++ b/relayer/tools/gen-mmr.go
@@ -29,10 +29,14 @@ type H256 [32]byte
 
 // Leaf node for MMR
 type MMRLeaf struct {
-	blockHash        H256
-	paraHeadsRoot    H256
+	// Relay chain block hash
+	blockHash H256
+	// Merkle root for Parachain block header hashes
+	paraHeadsRoot H256
+	// Has validator merkle roor
 	hasValidatorRoot bool
-	validatorRoot    H256
+	// Merkle root for validator public keys
+	validatorRoot H256
 }
 
 // hashing strategy for Merkle trees
@@ -40,7 +44,7 @@ func hashStrategy() hash.Hash {
 	return sha3.NewLegacyKeccak256()
 }
 
-// Merkle Tree for ParaHeads
+// Hashing strategy for ParaHead leaf nodes
 func (head ParaBlockHash) CalculateHash() ([]byte, error) {
 	h := sha3.NewLegacyKeccak256()
 	if _, err := h.Write(head[:]); err != nil {
@@ -53,7 +57,7 @@ func (head ParaBlockHash) Equals(other merkletree.Content) (bool, error) {
 	return head == other, nil
 }
 
-// Merkle Tree for validator keys
+// Hashing strategy for ValidatorPublicKey leaf nodes
 func (key ValidatorPublicKey) CalculateHash() ([]byte, error) {
 	h := sha3.NewLegacyKeccak256()
 	if _, err := h.Write(key); err != nil {
@@ -66,7 +70,7 @@ func (key ValidatorPublicKey) Equals(other merkletree.Content) (bool, error) {
 	return reflect.DeepEqual(key, other), nil
 }
 
-// Merkle Tree for MMRLeaf
+// Hashing strategy for MMRLeaf leaf nodes
 func (leaf MMRLeaf) CalculateHash() ([]byte, error) {
 	h := sha3.NewLegacyKeccak256()
 	if _, err := h.Write(leaf.blockHash[:]); err != nil {
--- a/relayer/tools/gen-mmr.go
+++ b/relayer/tools/gen-mmr.go
@@ -101,6 +101,7 @@ func makeValidatorKeypair() *Keypair {
 	public := pk.Public().(*ecdsa.PublicKey)
 
 	return &Keypair{
+		// FIXME: Its not clear whether this marshalling is compatible with Solidity!
 		publicBytes: elliptic.Marshal(elliptic.P256(), public.X, public.Y),
 		public:      public,
 		private:     pk,
--- a/relayer/tools/gen-mmr.go
+++ b/relayer/tools/gen-mmr.go
@@ -39,7 +39,7 @@ type MMRLeaf struct {
 	validatorRoot H256
 }
 
-// hashing strategy for Merkle trees
+// hashing strategy for interior Merkle tree nodes
 func hashStrategy() hash.Hash {
 	return sha3.NewLegacyKeccak256()
 }
--- a/relayer/tools/gen-mmr.go
+++ b/relayer/tools/gen-mmr.go
@@ -1,0 +1,171 @@
+package main
+
+import (
+	"crypto/ecdsa"
+	"crypto/elliptic"
+	"fmt"
+	"hash"
+
+	"reflect"
+
+	"github.com/cbergoon/merkletree"
+	secp256k1 "github.com/ethereum/go-ethereum/crypto"
+	"golang.org/x/crypto/blake2b"
+	"golang.org/x/crypto/sha3"
+)
+
+// Validator keypair (secp256k1)
+type Keypair struct {
+	publicBytes []byte
+	public      *ecdsa.PublicKey
+	private     *ecdsa.PrivateKey
+}
+
+type ValidatorPublicKey []byte
+type ParaBlockHash [32]byte
+
+// Uninterpreted hash type
+type H256 [32]byte
+
+// Leaf node for MMR
+type MMRLeaf struct {
+	blockHash        H256
+	paraHeadsRoot    H256
+	hasValidatorRoot bool
+	validatorRoot    H256
+}
+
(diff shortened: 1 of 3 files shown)


relayer/tools/gen-mmr.go
```diff
@@ -0,0 +1,172 @@
...
+type ParaBlockHash [32]byte
+
+// Uninterpreted hash type
+type H256 [32]byte
+
+// Leaf node for MMR
+type MMRLeaf struct {
+	blockHash        H256
+	paraHeadsRoot    H256
+	hasValidatorRoot bool
```
denalimarsh, line 34: I believe we can remove this bool as the nil contents of `validatorRoot` can indicate if the MMR has a validator root, correct?

denalimarsh (commented):
Thanks for putting this together. 

So in the given 4-leaf example, the data structure would be:
```
      node2 
      /       \ 
node0 node1 node3
```
Where `node2` and `node3` are the peaks and the MMR root is the hash of those peaks.

Are we verifying through block hashes, parachain header merkle roots, or both? Here's a sample:

To verify an individual parachain header within an MMR leaf we'll need:
1. The parachain header's siblings (this is the other parachain header in the node)
2. The containing node's paraHeadsRoot.
If the two parachain header's hashed together is the same as the containing node's paraHeadsRoot, the parachain header is verified.

To verify an individual leaf/node of the MMR we'll need:
3. The path of leaf/node paraHeadsRoot from our leaf to the parent peak
4. The parent peak's paraHeadsRoot.
If each leaf/node's paraHeadsRoot hashed together is the same as the parent peak's paraHeadsRoot then we've verified our path to the peak.

To verify a peak within the MMR we need:
5. The peak's paraHeadsRoot.
6. All of the MMR's peaks paraHeadsRoot.
7. The root of the MMR
To verify we hash the peaks together and if they match the root of the MMR the peak is verified within the MMR.

Is that correct? I don't think everything matches up. In the above example, the block hashes are completely excluded. If someone could critique the above example in relation to the code contained in this PR it would be helpful for the MMR verification.

vgeddes:
> So in the given 4-leaf example, the data structure would be:
> 
> ```
>       node2 
>     /  \    \
> node0 node1 node3
> ```
> 

Actually it would look like this, where Root is the MMR peak:

```
        Root

Hash(1+2)  Hash(3+4)

1            2       3           4
```

musnit:
> To verify an individual parachain header within an MMR leaf we'll need:
> 
> 1. The parachain header's siblings (this is the other parachain header in the node)
> 2. The containing node's paraHeadsRoot.
>    If the two parachain header's hashed together is the same as the containing node's paraHeadsRoot, the parachain header is verified.

This sounds about right, and then (2) would be the actual content of the leaf node we verify that is part of the MMR that has been committed to and (1) would be proof data that is submitted seperately alongside the leaf node that allows us to verify our actual parachain header once the leaf node is verified. Let's not worry about perfecting this until we have a real payload to confirm with though - these 2 steps look sufficient for our next deliverable even if we need to tweak them once we get real data.

musnit:
> To verify an individual leaf/node of the MMR we'll need:
> 3. The path of leaf/node paraHeadsRoot from our leaf to the parent peak
> 4. The parent peak's paraHeadsRoot.
> If each leaf/node's paraHeadsRoot hashed together is the same as the parent peak's paraHeadsRoot then we've verified our path to the peak.

"The parent peak's paraHeadsRoot" - not 100% i'm following what you mean by this line but overall looks correct

musnit:
> To verify a peak within the MMR we need:
> 5. The peak's paraHeadsRoot.
> 6. All of the MMR's peaks paraHeadsRoot.
> 7. The root of the MMR
> To verify we hash the peaks together and if they match the root of the MMR the peak is verified within the MMR.

Also looks right

musnit:
> Is that correct? I don't think everything matches up. In the above example, the block hashes are completely excluded. If someone could critique the above example in relation to the code contained in this PR it would be helpful for the MMR verification.

Will do, though starting to think that this generated example is maybe a bad idea and we should really aim to generate using the Rust code that is gonna be running in practice cc @vgeddes lets discuss further on a call if needed

musnit:
> > Is that correct? I don't think everything matches up. In the above example, the block hashes are completely excluded. If someone could critique the above example in relation to the code contained in this PR it would be helpful for the MMR verification.
> 
> Will do, though starting to think that this generated example is maybe not ideal and we should really aim to generate using the Rust code that is gonna be running in practice cc @vgeddes lets discuss further on a call if needed

Having said that, I think if we get a full working light client integrated and have this MMR proof/verification in a library or second contract, then replacing/tweaking it to a slightly different data structure/process should be pretty quick and easy and so I think going ahead even if this is inaccurate is still a good idea

relayer/tools/gen-mmr.go
vgeddes: Well, `hasValidatorRoot` is still useful for SCALE-encoding the leaf. The encoder/decoder can use the flag to encode/decode appropriately.

relayer/tools/gen-mmr.go
```diff
@@ -0,0 +1,176 @@
...
+type MMRLeaf struct {
+	// Relay chain block hash
+	blockHash H256
+	// Merkle root for Parachain block header hashes
+	paraHeadsRoot H256
+	// Has validator merkle roor
+	hasValidatorRoot bool
+	// Merkle root for validator public keys
+	validatorRoot H256
+}
```
philipstanislaus, line 40: It seems to me that we are mixing up a MMR leaf and a commitment here. Do you have a link to the data structure of the MMR leafs? I was unable to find it in https://github.com/paritytech/grandpa-bridge-gadget. 

The commitment should contain the MMR root and have the following data structure: 

```rust
/// A commitment signed by GRANDPA validators as part of BEEFY protocol.
///
/// The commitment contains a [payload] extracted from the finalized block at height [block_number].
/// GRANDPA validators collect signatures on commitments and a stream of such signed commitments
/// (see [SignedCommitment]) forms the BEEFY protocol.
#[derive(Debug, PartialEq, Eq, codec::Encode, codec::Decode)]
pub struct Commitment<TBlockNumber, TPayload> {
	/// The payload being signed.
	///
	/// This should be some form of cumulative representation of the chain (think MMR root hash).
	/// For transition blocks it also MUST contain details of the next validator set.
	pub payload: TPayload,

	/// Finalized block number this commitment is for.
	///
	/// GRANDPA validators agree on a block they create a commitment for and start collecting
	/// signatures. This process is called a round.
	/// There might be multiple rounds in progress (depending on the block choice rule), however
	/// since the payload is supposed to be cumulative, it is not required to import all
	/// commitments.
	/// BEEFY light client is expected to import at least one commitment per epoch (the one with
	/// [is_set_transition_block] set), but is free to import as many as it requires.
	pub block_number: TBlockNumber,

	/// BEEFY validator set supposed to sign this commitment.
	///
	/// Validator set is changing once per epoch in the commitment with [is_set_transition_block]
	/// set to `true`. Such "epoch commitments" MUST provide the light client with details of the
	/// new validator set as part of the payload. The protocol itself doesn't enforce how these
	/// details are provided though.
	pub validator_set_id: ValidatorSetId,

	/// Indicator of the last block of the epoch.
	///
	/// The payload will contain some form of the NEW validator set public keys information,
	/// yet the block is signed by the current validator set.
	/// When this commitment is imported, the client MUST increment the `validator_set_id`.
	pub is_set_transition_block: bool,
}
```

(From https://github.com/paritytech/grandpa-bridge-gadget/blob/master/primitives/src/commitment.rs#L26)

philipstanislaus (commented):
Agree with Aidan that we can get started what we have in here and then later change the data structures, the overhead should not be massive. 

@tomusdrw it would be super helpful if you could comment on the proposed data structure for a MMR leaf. I have seen a few ideas floating around, but could not find a doc/reference implementation for the payload of a commitment/MMR leaves (maybe did not dig deep enough).

philipstanislaus:
@tomusdrw just saw you are working on a reference implementation, so no need to comment here. Would be great if you could link to that implementation once merged/PR for it is open. Many thanks!

relayer/tools/gen-mmr.go
vgeddes: I was using this as a rough reference: https://github.com/Snowfork/verification-solidity/blob/master/contracts/MMRVerification.sol#L8.

Issue #162, opened 2020-11-19
Handle Infura connection drops


vgeddes:
Infura is known to drop idle websocket connections, and our relayer does not handle these robustly. This means that after a period of idleness, the header & event subscriptions will be terminated silently.

vgeddes:
Obsolete and already fixed.

PR #163, opened 2020-11-19
Terminate relayer after dropped Infura connection

vgeddes:
This is a stop-gap solution for #162 until we have more robust handling of dropped connections and missed blocks. Currently the relayer is losing its Infura connection after a while.

In our Rococo infrastructure, when the relayer is terminated, systemd will restart it automatically. 

Here's someone else's solution, but I don't think its the right approach: https://github.com/hydroscan/hydroscan-api/blob/54c32f524f36c1797af8cfada6eac3861a9f575a/task/event_subscriber.go#L53

Open to other ideas too.


musnit (approved):

vgeddes:
Also fixed our integration tests which were actually too cumbersome too use in everyday practice. Bootstrapping the services for the integration test now takes a minute or so, instead of 20 minutes.

PR #164, opened 2020-11-20
Inject contract addresses from genesis storage

vgeddes:
Currently, for message routing within the parachain, ethereum contract addresses are hardcoded in the parachain code. When building the parachain, a cargo build script reads the addresses from environment variables and dynamically generates code which is injected into the compilation of the parachain.

This is bad for several reasons:
* Complicates build process for parachain (have to set environment variables before building chain)
* In development environments, the parachain needs to be rebuilt every time the addresses change
* Lots of supporting boilerplate code to support the above.
* Can't query contract addresses from parachain (so no way to debug possibly mismatched addresses, etc)

With this change, all of this is greatly simplified as the addresses can now be injected via the chainspec:

myCustomSpecWithContractAddrs.json:
```json
      "ethApp": {
        "address": "0xfc97a6197dc90bef6bbefd672742ed75e9768553"
      },
      "erc20App": {
        "address": "0xeda338e4dc46038493b885327842fd3e301cab39"
      }
```

```
artemis-node --chain myCustomSpecWithContractAddrs.json
```

Other changes:
* For development, `truffle develop` is deprecated as it doesn't provide enough fine-grained control as `ganache-cli`. By using the latter we can generate deterministic contract addresses, which simplifies development.
* E2E tests are faster and more robust. They are now bootstrapped with a script `test/scripts/start-services.sh` instead of the previous ill-fitting docker-compose method.


philipstanislaus (approved):
Nice, LGTM 🎉 

Putting the addresses into storage makes a lot of sense to me, especially since it allows updates at a later point through extrinsics after a runtime upgrade. 

Long term, we might want to think about a registry that allows trustless addition of further apps on-chain.

parachain/node/Cargo.toml
```diff
@@ -138,6 +138,7 @@ version = "2.0.0-rc6"
 [dependencies]
 jsonrpc-core = "14.0.3"
 structopt = "0.3.8"
+hex-literal = "0.3.1"
```
Rizziepit, line 141: Alphabetical order?

parachain/node/src/chain_spec.rs
```diff
@@ -1,13 +1,18 @@
...
+	SystemConfig, VerifierConfig, VerifierLightclientConfig,
+	ETHConfig, ERC20Config,
+	WASM_BINARY, Signature
 };
 use sp_consensus_aura::sr25519::AuthorityId as AuraId;
 use sp_finality_grandpa::AuthorityId as GrandpaId;
 use sp_runtime::traits::{Verify, IdentifyAccount};
 use sc_service::ChainType;
 
+use hex_literal::hex;
```
Rizziepit, line 14: Heads up for a conflict with https://github.com/Snowfork/polkadot-ethereum/pull/166/, depending on who merges first

parachain/pallets/bridge/src/lib.rs
```diff
@@ -83,12 +82,12 @@ impl<T: Trait> Module<T> {
...
 
-	fn dispatch(app: App, message: Message) -> DispatchResult {
-		match app {
-			App::ETH => T::AppETH::handle(message.payload),
-			App::ERC20 => T::AppERC20::handle(message.payload)
+	fn dispatch(address: &H160, message: &Message) -> DispatchResult {
+		match address {
+			_ if *address == T::AppETH::address() => T::AppETH::handle(message.payload.as_ref()),
+			_ if *address == T::AppERC20::address() => T::AppERC20::handle(message.payload.as_ref()),
+			_ => Err(Error::<T>::AppNotFound.into())
```
Rizziepit, line 89: Is there a reason you're using match instead of an if-else statement?

Rizziepit (commented):
LGTM

parachain/node/src/chain_spec.rs
vgeddes: fixed the merge conflicts

parachain/pallets/bridge/src/lib.rs
vgeddes: Was mostly for aesthetic reasons. I like the match because it lines up the conditionals/statements more neatly than if-else. But I've reverted to if-else now as I admit in this case the code looks a bit convoluted.

PR #166, opened 2020-11-24
Sync Eth headers from genesis in Substrate node

Rizziepit:
This PR builds on the simple Ethereum header forwarding in https://github.com/Snowfork/polkadot-ethereum/pull/159 to add header syncing. This entails:
* Adding a mainnet genesis block in Substrate to sync from
* `Init` channels passed to `chain.Start` so that init params (such as a genesis block) can be exchanged before we start streaming headers/events
* Good-enough-for-now syncing logic in `ethereum.Listener`
* Retrying header writes with backoff
* Polling for a new nonce before writing extrinsics (see comment for explanation)

Below are logs from the relayer as it's syncing the node. Look out for "Witnessed new block header" - the relayer needs to know what the latest block is even though it won't forward it until it's caught up.
Screenshot from 2020-11-23 19-54-27

### DIFF
--- a/relayer/chain/ethereum/listener.go
+++ b/relayer/chain/ethereum/listener.go
@@ -5,6 +5,9 @@ package ethereum
 
 import (
 	"context"
+	"fmt"
+	"math/big"
+	"time"
 
 	geth "github.com/ethereum/go-ethereum"
 	gethCommon "github.com/ethereum/go-ethereum/common"
@@ -34,18 +37,18 @@ func NewListener(conn *Connection, messages chan<- chain.Message, headers chan<-
 	}, nil
 }
 
-func (li *Listener) Start(cxt context.Context, eg *errgroup.Group) error {
+func (li *Listener) Start(cxt context.Context, eg *errgroup.Group, start *HeaderID) error {
 	eg.Go(func() error {
-		return li.pollEventsAndHeaders(cxt)
+		return li.pollEventsAndHeaders(cxt, start)
 	})
 
 	return nil
 }
 
-func (li *Listener) pollEventsAndHeaders(ctx context.Context) error {
+func (li *Listener) pollEventsAndHeaders(ctx context.Context, start *HeaderID) error {
 	events := make(chan gethTypes.Log)
 	var eventsSubscriptionErr <-chan error
-	headers := make(chan *gethTypes.Header)
+	headers := make(chan *gethTypes.Header, 5)
 	var headersSubscriptionErr <-chan error
 
 	if li.messages == nil {
@@ -72,12 +75,16 @@ func (li *Listener) pollEventsAndHeaders(ctx context.Context) error {
 
 	li.log.Info("Polling headers started")
 
+	currentBlock := uint64(start.Number)
+	newestBlock := uint64(0)
 	subscription, err := li.conn.client.SubscribeNewHead(ctx, headers)
 	if err != nil {
 		li.log.WithError(err).Error("Failed to subscribe to new headers")
 		return err
 	}
 	headersSubscriptionErr = subscription.Err()
+
+	rateLimit := time.Duration(1) * time.Second
 
 	for {
 		select {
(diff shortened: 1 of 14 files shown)


relayer/chain/ethereum/listener.go
```diff
@@ -107,23 +110,47 @@ func (li *Listener) pollEventsAndHeaders(ctx context.Context) error {
...
+			blockNumber := gethheader.Number.Uint64()
+			newestBlock = blockNumber
 			li.log.WithFields(logrus.Fields{
-				"blockNumber": gethheader.Number,
-			}).Info("Witnessed block header")
-
-			header, err := MakeHeaderFromEthHeader(gethheader, li.log)
+				"blockNumber": blockNumber,
+			}).Info("Witnessed new block header")
+			if blockNumber <= currentBlock+1 {
```
philipstanislaus, line 118: I guess to enforce ordered submission of blocks we want to have a strict equality check here, right?

```suggestion
			if blockNumber == currentBlock+1 {
```

relayer/chain/ethereum/listener.go
```diff
@@ -107,23 +110,47 @@ func (li *Listener) pollEventsAndHeaders(ctx context.Context) error {
 				li.messages <- *msg
 			}
 		case gethheader := <-headers:
```
philipstanislaus, line 112: I feel this code needs to be made fork aware for subsequent submission of blocks. It should have some sort of queue where we connect the blocks and submit them one by one, reordering any chains of blocks when they are coming in. 

As this code currently stands, it might not pick up forks if the more difficult fork is a bit slower.

relayer/chain/substrate/chain.go
```diff
@@ -93,3 +109,19 @@ func (ch *Chain) Stop() {
...
+	}
+
+	var headerID ethereum.HeaderID
+	_, err = ch.conn.api.RPC.State.GetStorageLatest(storageKey, &headerID)
+	if err != nil {
+		return nil, err
+	}
+
+	headerID.Number = types.NewU64(uint64(headerID.Number) + 1)
+	return &headerID, nil
```
philipstanislaus, line 126: Again, this block should be referenced by hash, not by height, to prevent a situation where we jump to a different fork at that height.

relayer/chain/substrate/writer.go
```diff
@@ -56,15 +60,16 @@ func (wr *Writer) writeLoop(ctx context.Context) error {
 			err := wr.WriteHeader(ctx, &header)
 			if err != nil {
 				wr.log.WithFields(logrus.Fields{
-					"error": err,
+					"blockNumber": header.HeaderData.(ethereum.Header).Number,
```
philipstanislaus, line 63: Let's also add the hash here for better debuggability.

relayer/chain/substrate/writer.go
```diff
@@ -85,16 +90,32 @@ func (wr *Writer) write(c types.Call) error {
...
-	ok, err := wr.conn.api.RPC.State.GetStorageLatest(key, &accountInfo)
-	if err != nil {
-		return err
-	}
-	if !ok {
-		return fmt.Errorf("no account info found for %s", wr.conn.kp.URI)
-	}
+	// Poll account state until we have a new nonce. This matches our extrinsic
+	// sending rate with the node's extrinsic processing rate, and avoids
+	// the transaction being dropped from the node's transaction pool.
```
philipstanislaus, line 95: It limits the intake of the substrate chain though to 1 extrinsic per block. Is that really what we want? There might be severe delays if for some reason we need to catch up after a downtime. 

A less restrictive way here would be to have a sending queue and just back off exponentially if we hit errors.

Any thoughts @musnit ?

relayer/chain/substrate/writer.go
```diff
@@ -141,18 +162,45 @@ func (wr *Writer) WriteMessage(_ context.Context, msg *chain.Message) error {
...
+				"retries":     retries,
+			}).Error("Failure submitting header to substrate")
+		} else {
+			break
+		}
+
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-time.After(getHeaderBackoffDelay(retries)):
```
philipstanislaus, line 187: Nice :+1:

philipstanislaus (commented):
Nice, thanks 🙏 

There is one bigger concern here from my side: Forks, see comments below for details.

Happy to jump on a call to discuss that.

relayer/chain/ethereum/listener.go
musnit: @philipstanislaus the onchain code is fork aware already - though I think you're right in that we want to make the relayer eventually fork aware too. I don't think we should address this in the current PR though.

relayer/chain/substrate/writer.go
musnit: i think 1 extrinsic per block should be fine - remember substrate is 10 - 100x faster than ethereum, especially if we end up submitting transaction proofs in the same extrinsic as new block header. lets not prematurely optimize :), though could create a github issue for it

Rizziepit: @philipstanislaus this is not intended to be a final implementation. It's a workaround for dropped transactions (see https://stackoverflow.com/questions/63475813/transaction-failed-error-1014-priority-is-too-low-515000139-vs-515000139). Only one transaction with `(sender, nonce)` is allowed in the pool at a time. But I don't know much about this stuff in Substrate. Does the nonce here have to match the `accountInfo.Nonce`? If it can be an arbitrary number, I can easily fix this. @vgeddes maybe you know?

Rizziepit: There's also the possibility of customizing the transaction pool logic in our Substrate node. But I want to avoid doing that right now

musnit: oh yeh there could be some interesting designs that come out when considering node transaction pool changes too - but yeh lets not worry about for now

relayer/chain/ethereum/listener.go
Rizziepit: The `<=` is intentional because it means that `newestBlock` and `currentBlock` rewind when we hit a block from a slower fork. If we then subsequently encounter a block at, say, `currentBlock+2` from a faster fork, this case will ignore it, but the `default` case will re-request `currentBlock+1` and `currentBlock+2`. This... works.... sort of. But I'm sure this logic can be thwarted. At the very least, it's really tough to reason about. The benefit is that it allows us to combine logic for sync-forwarding an non-sync-forward (i.e. when the Substrate node is caught up). But I'm going to rewrite all of this later

relayer/chain/substrate/chain.go
Rizziepit: It's not possible to provide the hash corresponding to `headerID.Number + 1` here. The newest header that the Substrate node knows is `headerID`. It doesn't have the hashes for headers that come after.

For the purpose of syncing from an old header, this is ok. As long as the height is at least 35 blocks away from the tip, we assume whatever header is returned by `HeaderByNumber` is final (see https://github.com/Snowfork/polkadot-ethereum/blob/main/parachain/runtime/src/lib.rs#L268). Admittedly, I'm not handling the case where we get < 35 away from the tip here. I'll put that in the sync rewrite

Rizziepit:
To summarize discussion elsewhere: this fork handling is fickle and unintuitive. I'll re-implement it later (tracking [here](https://github.com/Snowfork/polkadot-ethereum/projects/4#card-50008947)). We'll be able to have simpler relayer logic once I implement [this](https://github.com/Snowfork/polkadot-ethereum/projects/4#card-47985524) for the Substrate node.

@philipstanislaus @vgeddes can one of you sign off on this PR so I can get rid of some branches? I'll make small style changes in a follow-up PR

relayer/chain/substrate/writer.go
Rizziepit: Created an issue for this https://github.com/Snowfork/polkadot-ethereum/issues/170

relayer/chain/ethereum/listener.go
philipstanislaus: Okay, understood. Would be great if we could discuss the architecture of this before you rewrite it. Thanks!

relayer/chain/substrate/chain.go
philipstanislaus: Sounds good. And of course we don't know future hashes ;)

relayer/chain/substrate/writer.go
philipstanislaus: Sounds good to me. Nonce needs to be sequential, if we don't customize the pool.

PR #167, opened 2020-11-24
Compute & forward Ethash proof data 

Rizziepit:
This PR finally adds Ethash proof data to the header payload so that Substrate can fully validate and import the header (see Substrate logs at bottom)

There are two main things happening in this PR:
1. I added ethashproof as a git submodule (we've forked it: https://github.com/Snowfork/ethashproof)
2. I added `HeaderChainState` to track... uh.... header state. The biggest piece of work it does is keeping the current + next epoch's cache data in memory for ethashproof

Unfortunately ethashproof uses `fmt.Printf` instead of a logger. So our output is a bit messy and cannot be filtered using log levels:
Screenshot from 2020-11-23 20-58-24

Here are the Substrate logs for successful imports:
Screenshot from 2020-11-23 20-39-49


relayer/chain/ethereum/header_chain_state.go
```diff
@@ -0,0 +1,143 @@
...
+	})
+
+	return &state, nil
+}
+
+// GetEthashProofCache returns the cache used for proof generation. It will return
+// immediately if `number` is in the current or next epoch. Outside that range, it
+// might block for multiple minutes to generate the cache. Calling GetEthashproofCache
+// will also update the current epoch to `number` / 30000.
+func (s *HeaderChainState) GetEthashproofCache(number uint64) (*ethashproof.DatasetMerkleTreeCache, error) {
```
Rizziepit, line 59: This logic should really have a test. @vgeddes @musnit is there a mocking library you folks use?

relayer/go.mod
```diff
@@ -7,7 +7,7 @@ require (
 	github.com/btcsuite/btcd v0.20.1-beta // indirect
 	github.com/centrifuge/go-substrate-rpc-client v2.0.0-rc6+incompatible
 	github.com/deckarep/golang-set v1.7.1 // indirect
-	github.com/ethereum/go-ethereum v1.9.22
+	github.com/ethereum/go-ethereum v1.8.23
```
Rizziepit, line 10: This is the highest version that ethashproof builds with :( I didn't want to spend time upgrading ethashproof but happy to create an issue and do this later if you think it's important

philipstanislaus: Since this should be highly security relevant, I feel we need to upgrade ethashproof before going to production. See https://github.com/ethereum/go-ethereum/releases for details on the security issues fixed in previous releases.

relayer/chain/ethereum/header.go
```diff
@@ -82,12 +89,45 @@ func MakeHeaderFromEthHeader(gethheader *etypes.Header, log *logrus.Entry) (*cha
...
+
+		proofData[i] = DoubleNodeWithMerkleProof{
+			DagNodes: [2]types.H512{
+				types.NewH512(node1Bytes),
+				types.NewH512(node2Bytes),
+			},
+			Proof: proofH128,
+		}
+	}
+
```
philipstanislaus, line 127: Can we put this functionality into a helper function and add a test for it?

relayer/chain/ethereum/header_chain_state.go
```diff
@@ -0,0 +1,143 @@
...
+)
+
+type EthashproofCacheState struct {
+	sync.Mutex
+	currentCache *ethashproof.DatasetMerkleTreeCache
+	nextCache    *ethashproof.DatasetMerkleTreeCache
+}
+
+// HeaderChainState is a helper for tracking where we are
+// in the Ethereum chain. It warms up caches as we go.
```
philipstanislaus, line 21: As mentioned in #166, how does this logic deal with different forks?

relayer/chain/ethereum/header_chain_state.go
philipstanislaus: Agree, tests would be great for this file. 

I prefer https://github.com/stretchr/testify over https://github.com/golang/mock for the better output, but both are fine in my view.

relayer/chain/ethereum/header_chain_state.go
philipstanislaus: I guess the ethash proof logic is fork independent, so that should be fine actually.

philipstanislaus: Would it make sense to separate current/newest block (since that's fork dependent) from the ethashproofCache?

relayer/chain/ethereum/header_chain_state.go
```diff
@@ -0,0 +1,143 @@
...
+
+func (s *HeaderChainState) prepareNextEthashproofCache() error {
+	cacheState := s.ethashproofCacheState
+	cacheState.Mutex.Lock()
+	defer cacheState.Mutex.Unlock()
+
+	// prepareNextEthashproofCache should only ever be called after
+	// nextCache has been set to nil
+	if cacheState.nextCache != nil {
+		return fmt.Errorf("prepareNextEthashproofCache encountered non-nil nextCache")
```
philipstanislaus, line 107: Use logrus here and in other places?

philipstanislaus (commented):
Great job on the `EthashproofCacheState`, I like how you implemented that 💪 

Left a few comments below.

relayer/chain/ethereum/header.go
Rizziepit: For sure

relayer/chain/ethereum/header_chain_state.go
Rizziepit: I think it makes sense to separate tehm. My intuition was to keep this state together since current epoch is related to current forks. But it's not actually relevant to the implementation

relayer/go.mod
musnit: Hm I don't think ethash has changed at all since then but it would still be a good idea to upgrade yeh. we can add a github issue for this to address later.

Rizziepit: https://github.com/Snowfork/polkadot-ethereum/issues/169

Rizziepit:
The caching logic and proof construction now have decent tests. @philipstanislaus I think I've addressed all your feedback. The only thing I skipped is logging via logrus in `HeaderCacheState`. It felt better to bubble up errors and log them in the listener

philipstanislaus:
Cool, will review tomorrow. Thanks!

vgeddes:
Is there reason we need to make `ethashproof` a submodule? Can't we just list it as a dependency in `go.mod`, like `github.com/Snowfork/ethashproof`?

relayer/chain/ethereum/listener.go
```diff
@@ -157,7 +172,7 @@ func makeFilterQuery(contracts []Contract) geth.FilterQuery {
 
 	for _, contract := range contracts {
 		addresses = append(addresses, contract.Address)
-		signature := contract.ABI.Events["AppTransfer"].ID.Hex()
+		signature := contract.ABI.Events["AppTransfer"].Id().Hex()
```
vgeddes, line 175: Does the `Id()` method actually exist? I don't see one defined in https://github.com/ethereum/go-ethereum/blob/master/accounts/abi/event.go.

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "relayer/ethashproof"]
+	path = relayer/ethashproof
+	url = git@github.com:Snowfork/ethashproof.git
```
vgeddes, line 3: Is there reason we need to make ethashproof a submodule? Can't we just list it as a dependency in go.mod, like `github.com/Snowfork/ethashproof`? Then everywhere we import `github.com/tranvictor/ethashproof` we would import `github.com/Snowfork/ethashproof` instead?

musnit: ye agree seems better

Rizziepit: No reason we _have_ to make it a submodule. I got the idea from https://github.com/KyberNetwork/bridge_eos_smart_contracts. I think that the monorepo-like situation might benefit us. Two main reasons for this:
* We might want to make changes to ethashproof down the line (definitely for https://github.com/Snowfork/polkadot-ethereum/issues/169), and those changes will have to be in lockstep with the relayer.
* I'm using bits of ethashproof that I would classify as "internals", not the public API. I might reuse bits for transaction proofs too. So managing ethashproof as an internal module is fitting in that sense.

I haven't used submodules before. So I could very well be missing an important consideration. I'm kinda curious to see how this unfolds but will defer to you folks

Rizziepit: Clarification: `github.com/tranvictor/ethashproof` is the module name declared in ethashproof's go.mod file. So I'm assuming I'd need to update ethashproof's go.mod if I want to import it using `github.com/Snowfork/ethashproof`. That's independent of whether ethashproof is a submodule. I wanted to avoid making changes to ethashproof for this PR but I'll update in another PR

PR #172, opened 2020-11-26
Misc improvements to relayer <-> Substrate

Rizziepit:
This diff addresses feedback on #135 #159 and #166, and fixes up some annoying relayer issues. This includes:
* removing panics caused by `unwrap` in Substrate
* improved errors / logging in relayer. e.g. including block hash in logs
* `onDone` function in `Listener` and `Writer` to clean up channels and avoid deadlock when you try to kill the relayer
* incrementing the nonce used in Substrate extrinsics in the relayer
* fixing relayer unit tests that were broken (but undetected in Github test runs?)

### DIFF
--- a/relayer/chain/ethereum/writer.go
+++ b/relayer/chain/ethereum/writer.go
@@ -62,11 +62,20 @@ func (wr *Writer) Start(ctx context.Context, eg *errgroup.Group) error {
 	return nil
 }
 
+func (wr *Writer) onDone(ctx context.Context) error {
+	wr.log.Info("Shutting down writer...")
+	// Avoid deadlock if a listener is still trying to send to a channel
+	for m := range wr.messages {
+		wr.log.WithFields(logrus.Fields{"message": m}).Debug("Discarded")
+	}
+	return ctx.Err()
+}
+
 func (wr *Writer) writeLoop(ctx context.Context) error {
 	for {
 		select {
 		case <-ctx.Done():
-			return ctx.Err()
+			return wr.onDone(ctx)
 		case msg := <-wr.messages:
 			err := wr.Write(ctx, &msg)
 			if err != nil {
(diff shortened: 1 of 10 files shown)


vgeddes:
> * fixing relayer unit tests that were broken (but undetected in Github test runs?)

Nice catch 👍 

Currently our Github Actions workflows only run the parachain tests, but of course we can change that.

relayer/chain/ethereum/writer.go
```diff
@@ -62,11 +62,20 @@ func (wr *Writer) Start(ctx context.Context, eg *errgroup.Group) error {
 	return nil
 }
 
+func (wr *Writer) onDone(ctx context.Context) error {
+	wr.log.Info("Shutting down writer...")
+	// Avoid deadlock if a listener is still trying to send to a channel
+	for m := range wr.messages {
+		wr.log.WithFields(logrus.Fields{"message": m}).Debug("Discarded")
```
vgeddes, line 69: logrus has a shorter api when only passing a single field:
```suggestion
		wr.log.WithField("message", m).Debug("Discarded")
```

relayer/chain/substrate/writer.go
```diff
@@ -19,11 +19,12 @@ import (
...
-	messages <-chan chain.Message
-	headers  <-chan chain.Header
-	log      *logrus.Entry
-	nonce    uint32
+	conn         *Connection
+	messages     <-chan chain.Message
+	headers      <-chan chain.Header
+	log          *logrus.Entry
+	nonce        uint32
+	nonceCounter uint32
```
vgeddes, line 27: `nonceCounter` looks to be a bit redundant. Can't we just increment `nonce` after every sending every transaction?

vgeddes (approved):
LGTM, though added some minor comments

PR #174, opened 2020-12-01
Use custom parachain spec for integration tests & forward Eth headers

Rizziepit:
This PR changes the integration test environment such that the parachain uses a json chain spec (instead of `--dev`). More importantly, it allows us to make arbitrary changes to that chain spec before launching the parachain. The end result of this is that I can insert the latest Ganache block into the parachain's genesis config, thus allowing the relayer to correctly forward Ethereum block headers to the parachain.

The alternative to this approach is to hard code a chain spec for integration tests. I prefer my current approach because it's super flexible.


test/scripts/start-services.sh
```diff
@@ -9,7 +9,7 @@ start_ganache()
     echo "Starting Ganache"
     yarn run ganache-cli \
         --port=8545 \
-        --blockTime=0 \
+        --blockTime=15 \
```
vgeddes, line 12: Is this strictly necessary? Its very convenient for the E2E tests to have blocks mined instantly after each transaction. Having a 15second delay means we'll need to update our tests for this change in behavior. And the tests will run slower too as they will need to wait up to 30s after submitting each transaction

test/scripts/start-services.sh
```diff
@@ -41,11 +41,31 @@ start_parachain()
     pushd ../parachain
 
     cargo build
-    target/debug/artemis-node --tmp --dev >$logfile 2>&1 &
+
+    echo "Generating Parachain spec"
+    target/debug/artemis-node build-spec --dev --disable-default-bootnode > $configdir/spec.json
+
+    echo "Inserting Ganache chain info into genesis spec"
+    ethereum_initial_header=`curl http://localhost:8545 \
```
vgeddes, line 49: Rather use `$( ... )` instead of backticks for command substitution. The latter has better behavior and is POSIX-compatible: https://stackoverflow.com/questions/9449778/what-is-the-benefit-of-using-instead-of-backticks-in-shell-scripts

test/scripts/start-services.sh
```diff
@@ -41,11 +41,31 @@ start_parachain()
...
+        | node ../test/scripts/helpers/transformEthHeader.js`
+    node ../test/scripts/helpers/overrideParachainSpec.js $configdir/spec.json \
+        genesis.runtime.verifierLightclient.initialDifficulty 0x0 \
+        genesis.runtime.verifierLightclient.initialHeader "$ethereum_initial_header"
+
+    target/debug/artemis-node -lruntime=debug \
+        --alice \
+        --tmp \
+        --rpc-port 11133 \
+        --ws-port 11144 \
```
vgeddes, line 62: The Mocha tests also need to be updated for this change of port: https://github.com/Snowfork/polkadot-ethereum/blob/main/test/test/test.js#L30

test/scripts/start-services.sh
```diff
@@ -41,11 +41,31 @@ start_parachain()
...
+        -X POST \
+        -H "Content-Type: application/json" \
+        -d '{"jsonrpc":"2.0","method":"eth_getBlockByNumber","params": ["latest", false],"id":1}' \
+        | node ../test/scripts/helpers/transformEthHeader.js`
+    node ../test/scripts/helpers/overrideParachainSpec.js $configdir/spec.json \
+        genesis.runtime.verifierLightclient.initialDifficulty 0x0 \
+        genesis.runtime.verifierLightclient.initialHeader "$ethereum_initial_header"
+
+    target/debug/artemis-node -lruntime=debug \
+        --alice \
```
vgeddes, line 59: Adding `--alice` will probably break things as it will run the chain in PoA mode. From my vague recollection PoA mode requires at least two PoA validators to be running in the network. Can we remove this flag?

vgeddes (commented):
Looks good! I agree with your approach as it definitely adds the flexibility we'll be needing as the E2E tests become more comprehensive. 🚀

test/scripts/start-services.sh
Rizziepit: The relayer relies on a stream of incoming blocks to figure out what the latest block is. Otherwise it gets stuck in a loop retrieving a next header that doesn't exist yet. I'm going to overhaul this implementation, but in the mean time, would it be fast enough to set block time to 1 second?

test/scripts/start-services.sh
Rizziepit: Without this flag, the Substrate node doesn't produce blocks / process transactions in its transaction pool. I initially didn't have `--alice` and then realized the `--dev` flag isn't equivalent to using a chain spec with type "Development". Lots of digging later, and `--alice` is the easiest way to replicate `--dev` behaviour. Have a look at https://github.com/paritytech/substrate/blob/master/client/cli/src/commands/run_cmd.rs. I think `--dev` also runs in PoA mode with "forced authoring" of blocks

test/scripts/start-services.sh
vgeddes: ah I see. Makes sense. I think its OK then and I'll update the tests to wait 20-30s after each transaction.

test/scripts/start-services.sh
vgeddes: Ahh Good to know 👍

Rizziepit: You're right. Tests are broken... back to the drawing board

Rizziepit: It was missing the `--force-authoring` flag. Now it's fairly equivalent to `--dev`. Tests pass

vgeddes (approved):
LGTM! lets merge it in. I'll follow up and merge in !176 which updates the timeouts in the E2E tests.

PR #175, opened 2020-12-02
Create & forward receipt proofs from Relayer -> Substrate

Rizziepit:
This PR creates receipt proofs in the relayer and then forwards them to our Substrate bridge pallet. This entails:
* Fetching all receipts for blocks with events we're subscribed to
* Constructing and caching the Merkle Patricia trie for receipts
* Using the trie above to generate proofs
* Adding those proofs to the `VerificationInput` struct in `Message`

Some notable things that are *not included* here:
* Verifying the merkle proof in Substrate
* Restructuring of verification modules / responsibilities in Substrate
* Bulk sending of events


Rizziepit:
@vgeddes @musnit right now, we allow a single `VerificationInput` for a message. That implies a single verifier module will do replay protection + inclusion proof verification. Will there ever be a case where an app/message doesn't need replay protection? Does it make sense to put replay protection in the lightclient module?

parachain/primitives/core/src/types.rs
```diff
@@ -36,6 +37,22 @@ pub enum VerificationInput {
...
+	/// TODO: use specialized structs, not byte arrays
+	LightReceipt {
+		// The receipt in the block
+		receipt: Vec<u8>,
+		// Merkle proof of inclusion in the block's receipts
+		proof: Vec<u8>,
+	},
+	/// Light-client-based scheme checks transaction inclusion proof
+	/// /// TODO: use specialized structs, not byte arrays
+	LightTransaction {
```
Rizziepit, line 50: I'm not sure if there's a use case for _only_ transaction inclusion proof, i.e. proving that a transaction was processed, but not proving its outcome

musnit: yeh. there may be for some obscure uses but we don't need to implement this for now

vgeddes:
> @vgeddes @musnit right now, we allow a single `VerificationInput` for a message. That implies a single verifier module will do replay protection + inclusion proof verification. Will there ever be a case where an app/message doesn't need replay protection? Does it make sense to put replay protection in the lightclient module?

Right now I can't envision a case where replay protection wouldn't be needed.

Regarding replay protection, yeah thanks for bringing that up. It may make sense to split the current verifier module into 2 pallets, each of which has its area of responsibility:

Verifier module:
* Handles replay protection
* Delegates to lightclient module for message verification:

Lightclient module:
* Does lightclient stuff :)

musnit:
> > @vgeddes @musnit right now, we allow a single `VerificationInput` for a message. That implies a single verifier module will do replay protection + inclusion proof verification. Will there ever be a case where an app/message doesn't need replay protection? Does it make sense to put replay protection in the lightclient module?
> 
> Right now I can't envision a case where replay protection wouldn't be needed.
> 
> Regarding replay protection, yeah thanks for bringing that up. It may make sense to split the current verifier module into 2 pallets, each of which has its area of responsibility:
> 
> Verifier module:
> 
> * Handles replay protection
> * Delegates to lightclient module for message verification:
> 
> Lightclient module:
> 
> * Does lightclient stuff :)

lets discuss this on a call plz

vgeddes (approved):
LGTM!

relayer/chain/ethereum/header_cache_state.go
```diff
@@ -4,13 +4,100 @@
...
+	hash := block.Hash().Hex()
+	_, exists := bc.blocks[hash]
+	if exists {
+		return
+	}
+
+	number := block.Number().Uint64()
+	hashesAtNumber, numberExists := bc.hashesByNumber[number]
+	// Remove oldest blocks if we've reached capacity
+	if !numberExists && len(bc.hashesByNumber) == bc.capacity {
```
philipstanislaus, line 66: Why only prune if there is no other block at that height?

Can we only skip pruning if the block to be inserted is at the lowest block number?

relayer/chain/ethereum/header_cache_state.go
```diff
@@ -4,13 +4,100 @@
...
+		var minNumber uint64 = math.MaxUint64
+		for number := range bc.hashesByNumber {
+			if number < minNumber {
+				minNumber = number
+			}
+		}
```
philipstanislaus, line 72: Alternatively, just store the minimum cached block number in `BlockCache`?

relayer/chain/ethereum/header_cache_state.go
```diff
@@ -49,30 +138,39 @@ type HeaderCacheState struct {
 func NewHeaderCacheState(
 	eg *errgroup.Group,
 	initBlockHeight uint64,
-	cl EthashproofCacheLoader,
+	bl BlockLoader,
+	ecl EthashproofCacheLoader,
 ) (*HeaderCacheState, error) {
-	cacheState := EthashproofCacheState{
+	blockCache := NewBlockCache(5)
```
philipstanislaus, line 144: Is `5` a good choice here? I feel it's a bit low for a real chain with latency and forks.

relayer/chain/ethereum/header_cache_state_test.go
```diff
@@ -64,6 +151,74 @@ func TestHeaderCacheState(t *testing.T) {
...
+	_, _ = hcs.GetReceiptTrie(ctx, data[4].Hash)
+	_, _ = hcs.GetReceiptTrie(ctx, data[5].Hash)
+	blockLoader.AssertNumberOfCalls(t, "GetBlock", 6)
+	blockLoader.AssertNumberOfCalls(t, "GetAllReceipts", 6)
+
+	// Should have been deleted in cache because capacity = 5 was reached above
+	_, _ = hcs.GetReceiptTrie(ctx, data[0].Hash)
+	blockLoader.AssertNumberOfCalls(t, "GetBlock", 7)
+	blockLoader.AssertNumberOfCalls(t, "GetAllReceipts", 7)
+}
```
philipstanislaus, line 199: Very nice tests 👍

relayer/chain/ethereum/listener.go
```diff
@@ -157,21 +152,45 @@ func (li *Listener) pollEventsAndHeaders(ctx context.Context, initBlockHeight ui
...
+		}).WithError(err).Error("Failed to get receipt trie for event")
+		return
+	}
+
+	msg, err := MakeMessageFromEvent(event, receiptTrie, li.log)
+	if err != nil {
+		li.log.WithFields(logrus.Fields{
+			"address":     event.Address.Hex(),
+			"txHash":      event.TxHash.Hex(),
+			"blockNumber": event.BlockNumber,
```
philipstanislaus, line 171: Can we add `blockHash` here as well?

philipstanislaus (commented):
Nice, left a few minor comments 👍

relayer/chain/ethereum/header_cache_state.go
Rizziepit: > Why only prune if there is no other block at that height?

If there's no other block at that height, we need to create a new height slot. So, before doing that, we remove the oldest height slot here. `BlockCache.capacity` refers to the max number of heights we'll store, not the max number of blocks (which is a bit confusing, but simplifies the logic).

relayer/chain/ethereum/header_cache_state.go
Rizziepit: That's def better :D

relayer/chain/ethereum/header_cache_state.go
Rizziepit: It might very well need to be adjusted. Fwiw, this cache is not used by the header forwarding / syncing logic. I might use it for that, but with significant changes. My understanding is that, if we want to track forks, we will need to keep the latest 35 heights in a cache. Does that make sense to you @philipstanislaus?

relayer/chain/ethereum/header_cache_state.go
Rizziepit: Does my answer resolve your questions?

PR #176, opened 2020-12-02
Update timeouts for E2E tests

vgeddes:
This makes the tests compatible with necessary changes required for header forwarding.


test/scripts/start-services.sh
```diff
@@ -75,9 +98,17 @@ deploy_contracts
...
-# TODO: Exit when any child process dies
-#  https://stackoverflow.com/questions/37496896/exit-a-bash-script-when-one-of-the-subprocesses-exits
-
+echo "Process Tree:"
 pstree $$
 
+for i in {0..4}; do
+    echo "Waiting for relayer to generate DAG cache..."
+    sleep 30
+done
```
Rizziepit, line 107: I think this will be a bit better:
```
until $(grep "Polling headers starting..." $(pwd)/relay.log > /dev/null); do
    echo "Waiting for relayer to generate DAG cache...."
    sleep 20
done
```

Rizziepit (approved):
LGTM. Check out my one suggestion before merging though

test/scripts/start-services.sh
vgeddes: awesome, thanks Riz, that looks much better. Will make the change

PR #182, opened 2020-12-16
Verify Ethereum receipt proofs in Substrate

Rizziepit:
This PR implements Ethereum receipt proof verification in Substrate. This consists of two parts:

**New Ethereum primitives & functions:**
* `Header::check_receipt_proof` - checks if the receipt is included in the header (but not whether the header is valid).
* New types `Receipt`, `FullNode` and `ShortNode`. `FullNode` and `ShortNode` represent Eth's Merkle Patricia Trie nodes (a proof is simply a traversal of these nodes)
* RLP decoding for above + `Bloom` type

**Lightclient verification & changes to verification API:**
* `Verifier::verify` now returns `Result<VerificationOutput, DispatchError>`. `VerificationOutput` allows the app handlers to confirm that the verification matches their payload. The alternative is to make the verifier do this, but then the verifier would have to parse and interpret the payload for each event / app type.
* `pallet_verifier_lightclient::Module::verify` verifies `VerificationInput::ReceiptProof` now. In addition to calling `Header::check_receipt_proof`, it checks that the header is part of the finalized chain.


parachain/pallets/erc20-app/src/lib.rs
```diff
@@ -121,9 +121,10 @@ impl<T: Trait> Module<T> {
 }
 
 impl<T: Trait> Application for Module<T> {
-	fn handle(payload: &[u8]) -> DispatchResult {
+	fn handle(payload: &[u8], _verification_output: &VerificationOutput) -> DispatchResult {
 		let payload_decoded = InPayload::decode(payload)
 			.map_err(|_| Error::<T>::InvalidPayload)?;
+		// TODO: check that payload event exists in VerificationOutput.Receipt
```
musnit, line 127: can we add github issue with security flag for this, rather than TODO, as its pretty important not to miss.

parachain/pallets/eth-app/src/lib.rs
```diff
@@ -108,10 +108,11 @@ impl<T: Trait> Module<T> {
 }
 
 impl<T: Trait> Application for Module<T> {
-	fn handle(payload: &[u8]) -> DispatchResult {
+	fn handle(payload: &[u8], _verification_output: &VerificationOutput) -> DispatchResult {
 		let payload_decoded = InPayload::decode(payload)
 			.map_err(|_| Error::<T>::InvalidPayload)?;
-
+		// TODO: check that payload event exists in VerificationOutput.Receipt
```
musnit, line 114: can we add github issue with security flag for this, rather than TODO, as its pretty important not to miss.

parachain/pallets/erc20-app/src/lib.rs
Rizziepit: Sure. I do consider this work part of https://github.com/Snowfork/polkadot-ethereum/projects/4#card-47985774 though. So it will get done in the next few days

Rizziepit: This is taken care of in https://github.com/Snowfork/polkadot-ethereum/pull/184

parachain/pallets/eth-app/src/lib.rs
Rizziepit: This is taken care of in https://github.com/Snowfork/polkadot-ethereum/pull/184

parachain/primitives/ethereum/src/lib.rs
```diff
@@ -99,6 +102,32 @@ impl Header {
...
+		match self.apply_merkle_proof(proof) {
+			Some((root, data)) if root == self.receipts_root => rlp::decode(&data).ok(),
+			Some((_, _)) => None,
+			None => None,
+		}
+	}
+
+	pub fn apply_merkle_proof(&self, proof: &[Vec<u8>]) -> Option<(H256, Vec<u8>)> {
+		let mut iter = proof.iter().rev();
+		let bytes = match iter.next() {
```
vgeddes, line 115: Can we rename `bytes` to something else as there is another local variable `bytes` in the  closure passed to `iter.fold`. Is a bit confusing.

parachain/primitives/ethereum/src/lib.rs
```diff
@@ -99,6 +102,32 @@ impl Header {
...
+			Some(b) => b,
+			None => return None,
+		};
+		let item_to_prove: mpt::ShortNode = rlp::decode(bytes).ok()?;
+	
+		let final_hash: Option<[u8; 32]> = iter.fold(Some(keccak_256(bytes)), |maybe_hash, bytes| {
+			let expected_hash = maybe_hash?;
+			let node: mpt::FullNode = rlp::decode(bytes).ok()?;
+			let found_hash = node.children.iter().find(|&&hash| Some(expected_hash.into()) == hash);
```
vgeddes, line 124: if you use `into_iter()` here then there is a chance you can simplify `|&&hash|` to `|&hash|` or  `|hash|`

parachain/primitives/ethereum/src/lib.rs
```diff
@@ -99,6 +102,32 @@ impl Header {
...
+	pub fn check_receipt_proof(&self, proof: &[Vec<u8>]) -> Option<Receipt> {
+		match self.apply_merkle_proof(proof) {
+			Some((root, data)) if root == self.receipts_root => rlp::decode(&data).ok(),
+			Some((_, _)) => None,
+			None => None,
+		}
+	}
+
+	pub fn apply_merkle_proof(&self, proof: &[Vec<u8>]) -> Option<(H256, Vec<u8>)> {
+		let mut iter = proof.iter().rev();
```
vgeddes, line 114: If you're consuming `proof` and don't intend to use it again, consider using `into_iter()` ([IntoIter](https://doc.rust-lang.org/std/iter/trait.IntoIterator.html#tymethod.into_iter)), then you can work with values directly instead of references to values.

vgeddes (approved):
LGTM! Added some minor comments.

parachain/pallets/verifier-lightclient/src/tests.rs
```diff
@@ -287,3 +288,135 @@ fn it_validates_proof_of_work() {
...
+		initial_difficulty: 0.into(),
+	}).execute_with(|| {
+		let output = Verifier::verify(
+			Default::default(),
+			Default::default(),
+			&message_with_receipt_proof(finalized_header_hash, receipt_proof),
+		);
+		assert_ok!(&output);
+		assert!(match output.unwrap() {
+			VerificationOutput::Receipt(receipt) => true,
```
philipstanislaus, line 310: Can we compare the actual receipt here?

philipstanislaus (approved):
Nice, especially the tests 👍

parachain/primitives/ethereum/src/lib.rs
```diff
@@ -99,6 +102,32 @@ impl Header {
...
+		let item_to_prove: mpt::ShortNode = rlp::decode(bytes).ok()?;
+	
+		let final_hash: Option<[u8; 32]> = iter.fold(Some(keccak_256(bytes)), |maybe_hash, bytes| {
+			let expected_hash = maybe_hash?;
+			let node: mpt::FullNode = rlp::decode(bytes).ok()?;
+			let found_hash = node.children.iter().find(|&&hash| Some(expected_hash.into()) == hash);
+			found_hash.map(|_| keccak_256(bytes))
+		});
+		
+		final_hash.map(|hash| (hash.into(), item_to_prove.value))
```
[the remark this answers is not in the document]
Rizziepit: > Can you point me to the reference implementation for this function?

The proof is generated by Geth. The closest I have to a reference is Geth's verification: https://github.com/ethereum/go-ethereum/blob/master/trie/proof.go#L106. That verification actually navigates the trie from root -> item using the item key (which is a function of its transaction index). I tried to implement something similar but ran into issues deriving the item key from the transaction index. I didn't want to replicate all the Ethereum trie key logic in Substrate: https://github.com/ethereum/go-ethereum/blob/master/trie/encoding.go

To avoid having to derive a key, I'm going in the reverse direction: hashing from item -> root, checking at each depth that the hash I computed at `depth + 1` exists at `depth`. This does mean that I'm not technically verifying the transaction / receipt index. I'm not sure if that can be exploited. We're not using the `tx_index` right now (I'll update this PR to remove the field entirely)

> Here each full node includes all the child hashes, is that correct?

I'm not sure what "all the child hashes" refers to. To clarify, it contains [h<sub>1</sub>, h<sub>2</sub>, h<sub>3</sub>...] where h<sub>n</sub> is the hash of an entire subtree (a subtree could consist of a single leaf node though, as is the case for the receipt node).

parachain/primitives/ethereum/src/mpt.rs
```diff
@@ -0,0 +1,92 @@
...
+impl rlp::Decodable for FullNode {
+    fn decode(rlp: &rlp::Rlp) -> Result<Self, rlp::DecoderError> {
+        let maybe_children: Result<Vec<Option<H256>>, rlp::DecoderError> = rlp.iter()
+            .map(|item| {
+                let v: Vec<u8> = item.as_val()?;
+                match v.len() {
+                    0 => Ok(None),
+                    32 => {
+                        let mut bytes = [0u8; 32];
+                        bytes[..].copy_from_slice(&v[..]);
```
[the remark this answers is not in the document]
Rizziepit: Do you prefer `v.as_ref()` over `&v`? I've been favouring the latter

parachain/primitives/ethereum/src/mpt.rs
```diff
@@ -0,0 +1,92 @@
...
+/// Intermediate trie node with children (refers to node with same name in Geth).
+/// This struct only handles the proof representation, i.e. a child is either empty
+/// or a 32-byte hash of its subtree.
+pub struct FullNode {
+    pub children: Vec<Option<H256>>
+}
+
+impl rlp::Decodable for FullNode {
+    fn decode(rlp: &rlp::Rlp) -> Result<Self, rlp::DecoderError> {
+        let maybe_children: Result<Vec<Option<H256>>, rlp::DecoderError> = rlp.iter()
```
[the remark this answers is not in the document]
Rizziepit: It needs a turbofish to work (like `collect::<Result<_, rlp::DecoderError>>()?`) but it's still an improvement

parachain/primitives/ethereum/src/lib.rs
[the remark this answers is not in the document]
Rizziepit: I'm going to merge to get my branch situation under control. But I'll keep an eye out for comments here and put improvements in another PR

philipstanislaus (commented):
@Rizziepit just realized that my last comment was still `Pending` here on GitHub. Let's move the discussion to #202

PR #184, opened 2020-12-17
Check verification when decoding app payload

Rizziepit:


Rizziepit:
@musnit @vgeddes interested in your thoughts on this approach to checking the verification output in apps. These are the alternatives I considered (this PR implements a variant of 2):
1. Introduce `Application::check_verification` and call it in the bridge before calling `Application::handle`
  Could work, but both functions need to decode the payload. So there's some duplicated work.
2. Add verification checks to `Application::handle`
  Less boilerplate, but also nothing that forces you to add verification checks.

parachain/pallets/erc20-app/src/payload.rs
```diff
@@ -28,10 +30,28 @@ pub struct InPayload<AccountId: codec::Decode> {
 
 impl<AccountId: codec::Decode> InPayload<AccountId> {
 
-	/// Decode payload into an Ethereum log event and then extract ABI-encoded data from log event
-	pub fn decode(payload: &[u8]) -> Result<Self, DecodeError> {
+	/// Decode payload into an Ethereum log event, check that it matches the verified receipt,
+	/// and then extract ABI-encoded data from log event
+	pub fn decode_verified(payload: &[u8], verification_output: &VerificationOutput) -> Result<Self, DecodeError> {
```
musnit, line 35: this function looks the same across both apps - could it be in a shared library?

parachain/pallets/erc20-app/src/lib.rs
```diff
@@ -121,10 +121,9 @@ impl<T: Trait> Module<T> {
 }
 
 impl<T: Trait> Application for Module<T> {
-	fn handle(payload: &[u8], _verification_output: &VerificationOutput) -> DispatchResult {
-		let payload_decoded = InPayload::decode(payload)
+	fn handle(payload: &[u8], verification_output: &VerificationOutput) -> DispatchResult {
+		let payload_decoded = InPayload::decode_verified(payload, verification_output)
 			.map_err(|_| Error::<T>::InvalidPayload)?;
```
musnit, line 126: not too familiar with Rust, but looks like even if there is an error here, it will still continue to process the payload, rather than fail? is that correct?

musnit:
@Rizziepit busy writing a doc to get into a lot of stuff related to this, but i think your approach is fine for now and we can refactor if needed in future milestone.

parachain/pallets/erc20-app/src/lib.rs
vgeddes: Nope, the `?` operator basically means that the function will exit early with the `InvalidPayload` error as the result: https://doc.rust-lang.org/edition-guide/rust-2018/error-handling-and-panics/the-question-mark-operator-for-easier-error-handling.html

vgeddes:
Yeah I much prefer option 2. 👍

parachain/pallets/erc20-app/src/payload.rs
vgeddes: Well I'm thinking 2 things:

1. Since Riz factored out most of the log decoding into the `TryFrom<Log>`  impl, we should consider moving the body of the reduced `decode_verified` into `Application::handle`. I'm thinking that's a more appropriate location since:
   * `Application::handle` doesn't do much else except call `decode_verified` and pass on the results.
   * This `payload.rs` module was mostly intended for decoding/unmarshalling, and "business rules" such as verification are maybe better suited for `Application::handle`.
2. I generally don't worry about factoring out duplicated code unless there are more than 2 duplications. As I've learnt from experience that sometimes those factorings, while nice and elegant, are often premature or wasted work in the end. That said, when our code is more stable and mature, we should definitely revisit and see how we can make it easier for users to develop new Substrate apps. - Just my 2cents.

parachain/pallets/erc20-app/src/payload.rs
```diff
@@ -28,10 +30,28 @@ pub struct InPayload<AccountId: codec::Decode> {
...
 		// Decode ethereum Log event from RLP-encoded data
 		let log: Log = rlp::decode(payload)?;
+
+		if let VerificationOutput::Receipt(receipt) = verification_output {
+			if !receipt.contains_log(&log) {
+				return Err(DecodeError::InvalidVerification);
```
vgeddes, line 41: Strictly speaking a failed verification wouldn't be a decoding error. Maybe `InvalidVerification` would be better defined as a pallet error (which ties in somewhat with my previous point about moving this code to `Application::handle`).

vgeddes (commented):
Looks good! I generally like the changes!

parachain/pallets/erc20-app/src/payload.rs
Rizziepit: Agreed. I put this in `InPayload` to reduce the likelihood of someone inadvertently dropping the verification check when they change the code. But hiding it in `InPayload` isn't a pretty poor strategy

vgeddes (approved):
Approved! :+1:

vgeddes:
@Rizziepit actually having the application do the final verification may conflict with our designs around relayer incentivization. Added some comments here: https://docs.google.com/document/d/10kEMqRy6v4nANNqXmqG7S8Px49CWvcEaX4EppSq9_-U/edit?disco=AAAALPtfo4E.

Rizziepit:
> @Rizziepit actually having the application do the final verification may conflict with our designs around relayer incentivization. Added some comments here: https://docs.google.com/document/d/10kEMqRy6v4nANNqXmqG7S8Px49CWvcEaX4EppSq9_-U/edit?disco=AAAALPtfo4E.

I responded in the doc. Let me know what you think

parachain/pallets/eth-app/src/lib.rs
```diff
@@ -116,12 +120,20 @@ impl<T: Trait> Module<T> {
...
-		// TODO: check that payload event exists in VerificationOutput.Receipt
-	
-		Self::handle_event(payload_decoded)
+
+		if let VerificationOutput::Receipt(receipt) = verification_output {
+			if !receipt.contains_log(&log) {
+				return Err(Error::<T>::InvalidVerification.into());
+			}
+		} else {
+			return Err(Error::<T>::InvalidVerification.into());
```
musnit, line 133: is this still needed?

parachain/pallets/erc20-app/src/lib.rs
```diff
@@ -125,12 +129,20 @@ impl<T: Trait> Module<T> {
...
-		// TODO: check that payload event exists in VerificationOutput.Receipt
+		
+		if let VerificationOutput::Receipt(receipt) = verification_output {
+			if !receipt.contains_log(&log) {
+				return Err(Error::<T>::InvalidVerification.into());
+			}
+		} else {
+			return Err(Error::<T>::InvalidVerification.into());
+		}
```
musnit, line 143: is this still needed?

Rizziepit:
@musnit @vgeddes this has been updated now. Instead of putting `receipt.contains_log(&log)` in the bridge, I put it in the LC verifier. This simplifies the design - essentially it's the original design again :)

Reasoning: I don't feel like there's a strong case for having `VerificationOutput` anymore since we don't want the `Application` to do further verification. And if the bridge can assume all payloads are logs / events, then so can the verifier.

vgeddes (approved):
LGTM!

parachain/pallets/erc20-app/src/lib.rs
```diff
@@ -125,10 +127,12 @@ impl<T: Trait> Module<T> {
 }
 
 impl<T: Trait> Application for Module<T> {
-	fn handle(payload: &[u8], _verification_output: &VerificationOutput) -> DispatchResult {
-		let payload_decoded = InPayload::decode(payload)
+	fn handle(payload: &[u8]) -> DispatchResult {
```
musnit, line 130: stupid question likely: it's not possible to call this handle() function directly from an extrinsic, bypassing the bridge - the only way to call it is via the bridge submit method, right? @vgeddes @Rizziepit

Rizziepit: Correct

PR #192, opened 2020-12-30
Fork-aware Eth header sync in relayer

Rizziepit:
This PR is a follow-up on #166 which added header forwarding from Eth -> Substrate. In this PR I'm upgrading it to track all forks (for unfinalized headers) and their forwarding state. This allows me to recursively forward ancestors of a header until we get to the finalized ancestors.

Note that recursing over all 35 unfinalized ancestors will probably only occur once. After that, we usually won't need to forward any ancestors at all (i.e. recursion depth = 1).

### DIFF
--- a/relayer/chain/ethereum/syncer/cache.go
+++ b/relayer/chain/ethereum/syncer/cache.go
@@ -1,0 +1,89 @@
+// Copyright 2020 Snowfork
+// SPDX-License-Identifier: LGPL-3.0-only
+
+package syncer
+
+import (
+	gethCommon "github.com/ethereum/go-ethereum/common"
+	gethTypes "github.com/ethereum/go-ethereum/core/types"
+)
+
+type HeaderCacheItem struct {
+	Header    *gethTypes.Header
+	Forwarded bool
+}
+
+type HeaderCache struct {
+	headers           map[string]*HeaderCacheItem
+	hashesByHeight    map[uint64][]string
+	maxHeight         uint64
+	minHeight         uint64
+	numHeightsToTrack uint64
+}
+
+func NewHeaderCache(numHeightsToTrack uint64) *HeaderCache {
+	return &HeaderCache{
+		headers:           make(map[string]*HeaderCacheItem, numHeightsToTrack),
+		hashesByHeight:    make(map[uint64][]string, numHeightsToTrack),
+		maxHeight:         0,
+		minHeight:         ^uint64(0),
+		numHeightsToTrack: numHeightsToTrack,
+	}
+}
+
+func (hc *HeaderCache) Insert(header *gethTypes.Header) bool {
+	hash := header.Hash().Hex()
+	_, exists := hc.headers[hash]
+	if exists {
+		return true
+	}
+
+	// Don't track headers older than numHeightsToTrack. This means
+	// the range [minHeight, maxHeight] is always moving forward
+	height := header.Number.Uint64()
+	if hc.maxHeight >= hc.numHeightsToTrack && height <= hc.maxHeight-hc.numHeightsToTrack {
+		return false
+	}
+
+	hc.headers[hash] = &HeaderCacheItem{
+		Header:    header,
+		Forwarded: false,
+	}
+
+	hashesAtHeight, heightExists := hc.hashesByHeight[height]
+	if heightExists {
+		hc.hashesByHeight[height] = append(hashesAtHeight, hash)
+	} else {
+		hc.hashesByHeight[height] = []string{hash}
+		if height < hc.minHeight {
+			hc.minHeight = height
+		} else if height > hc.maxHeight {
+			hc.maxHeight = height
+		}
+	}
+
+	if hc.maxHeight >= hc.numHeightsToTrack {
+		hc.pruneUpTo(hc.maxHeight - hc.numHeightsToTrack + 1)
+	}
+	return true
+}
+
+func (hc *HeaderCache) pruneUpTo(minHeightToKeep uint64) {
+	for hc.minHeight < minHeightToKeep {
+		hashesToRemove := hc.hashesByHeight[hc.minHeight]
+		delete(hc.hashesByHeight, hc.minHeight)
+		hc.minHeight++
+		for _, hashToRemove := range hashesToRemove {
+			delete(hc.headers, hashToRemove)
+		}
+	}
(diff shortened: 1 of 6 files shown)


relayer/chain/ethereum/syncer/cache.go
```diff
@@ -0,0 +1,96 @@
...
+	Forwarded bool
+}
+
+// This is used to store the latest headers as they are published. Up to
+// `numHeightsToTrack` heights are stored. Once this number is reached, an old
+// height is pruned each time a new height is added. The current stored height
+// range is given by [minHeight, maxHeight].
+type HeaderCache struct {
+	headers           map[string]*HeaderCacheItem
+	hashesByHeight    map[uint64][]string
```
philipstanislaus, line 22: To save the indirection of hashes, you could just store the pointers here too, i. e. 

```suggestion
	hashesByHeight    map[uint64][]*HeaderChacheItem
```

relayer/chain/ethereum/syncer/syncer.go
```diff
@@ -0,0 +1,178 @@
...
+			lbi.fetchFinalizedDone = true
+			lbi.Unlock()
+
+			s.log.WithField("blockNumber", syncedUpUntil).Debug("Done retrieving finalized headers")
+
+			break
+		}
+		lbi.Unlock()
+
+		header, err := s.loader.HeaderByNumber(ctx, new(big.Int).SetUint64(syncedUpUntil+1))
```
philipstanislaus, line 86: Fetching by height here only gives you a chain if the node we are connected to has the same definition of finality. If they are more conservative (they require more confimrations), the node could give us headers from another fork at some point.

I think we should check whether the headers connect in this group, and reset the loop if we find out that the fork changed. Alternatively, we could just go backwards from finalized, and query headers by hash.

What do you think?

relayer/chain/ethereum/syncer/syncer.go
```diff
@@ -0,0 +1,178 @@
...
+
+func (s *Syncer) forwardAncestry(ctx context.Context, hash gethCommon.Hash, oldestHeight uint64) error {
+	item, exists := s.headerCache.Get(hash)
+	if !exists {
+		header, err := s.loader.HeaderByHash(ctx, hash)
+		if err != nil {
+			return err
+		}
+
+		// If a header is too old, it cannot be inserted. We can assume it's already been forwarded
```
philipstanislaus, line 157: As described in my comment regarding the cache, this implies that if we are too far behind, older headers will never be forwarded. 

Could we add a test for this case with a low `numHeightsToTrack` on the cache?

relayer/chain/ethereum/syncer/syncer.go
```diff
@@ -0,0 +1,178 @@
...
+
+	return nil
+}
+
+func (s *Syncer) fetchFinalizedHeaders(ctx context.Context, initBlockHeight uint64, lbi *latestBlockInfo) error {
+	syncedUpUntil := initBlockHeight
+
+	for {
+		lbi.Lock()
+		latestFinalizedHeight := lbi.height - s.descendantsUntilFinal
```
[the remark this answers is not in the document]
Rizziepit: This is intended. This is fast enough for mainnet Eth (and testnets going by https://ethereum.stackexchange.com/questions/27048/comparison-of-the-different-testnets). We can speed up old header retrieval by fetching e.g. 100 in parallel. However, if 100x is not enough, then we have a more fundamental syncing problem, i.e. individual header forwarding to sync won't work

relayer/chain/ethereum/syncer/syncer.go
Rizziepit: > Fetching by height here only gives you a chain if the node we are connected to has the same definition of finality.

This is a fair point.

> I think we should check whether the headers connect in this group, and reset the loop if we find out that the fork changed. Alternatively, we could just go backwards from finalized, and query headers by hash.

Sounds sensible to put a sanity check that headers connect here and rewinding if they don't - but how far does one rewind? If the `descendantsUntilFinal` value of 35 isn't sufficient to prevent this from happening, what value is? The alternative (fetching backwards) doesn't scale because it forces serial fetching. I'd like to give us the option of concurrent fetching + batching in future. Thoughts?

relayer/chain/ethereum/syncer/syncer.go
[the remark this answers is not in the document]
philipstanislaus: Okay, makes sense 👍

relayer/chain/ethereum/syncer/syncer.go
philipstanislaus: Good points. 

I guess it's fine to not have any checks for now, since a failure would just mean that the light clients will stop proceeding, which should be monitored anyways. A simple restart of the relayer should suffice to fix the issue. 

It would be a bit prettier if we actually check whether the headers connect, and just fail the relayer if they don't. 
I'd be fine though to just add a TODO explaining those potential issues for now.

relayer/chain/ethereum/syncer/cache.go
```diff
@@ -0,0 +1,96 @@
...
+	maxHeight         uint64
+	minHeight         uint64
+	numHeightsToTrack uint64
```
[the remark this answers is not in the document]
Rizziepit: > Doesn't that imply that if we are more than numHeightsToTrack behind the current tip of the blockchain, the cache will fail?

Yes and no. `numHeightsToTrack = descendantsUntilFinal + 1` in `NewSyncer`. So, provided that this cache is only used to retrieve headers after we've synced old blocks, it shouldn't matter that it doesn't contain older heights. After syncing old blocks, we will be `descendantsUntilFinal` behind the current tip

> There will be new blocks coming in at the top, but we try to sync old blocks at the beginning. It might make more sense to instead use a LRU cache, where we just store pointers in an array, pushing recently used ones to the top?

I'm not sure I understand this. We don't store old blocks in this cache. It's exclusively used for new headers as they come in. And it is a type of LRU cache in that we keep the latest heights and discard old heights.

relayer/chain/ethereum/syncer/syncer.go
Rizziepit: Failing loudly in the relayer is a good idea. I'll also add a comment

PR #195, opened 2021-01-04
Fixed DAG roots for ethashproof

Rizziepit:
This PR follows up on https://github.com/Snowfork/ethashproof/pull/2 which fixes a DAG generation bug by upgrading to go-ethereum v1.9.25. DAG generation broke at epoch 385. This PR contains the recomputed DAG roots for epoch 385 onwards so that we can validate Eth headers once again.

Note: I'm also upgrading the relayer to go-ethereum v1.9.25 here since the relayer and ethashproof must use the same version.


vgeddes (approved):
Yay! Thanks for the upgrade to geth 1.9!

philipstanislaus (commented):
👍

PR #132, opened 2020-10-20
Rococo Integration

vgeddes:
I've used [substrate-parachain-template](https://github.com/substrate-developer-hub/substrate-parachain-template) as a reference for this integration work.

Leaving this in draft mode until upstream work on Rococo, Cumulus, and parachains V1 has stabilized a bit more.

NOTE: This has been superseded by #152 which supports both Rococo and XCMP

PR #200, opened 2021-01-07
Support for Parachains V1 + XCMP

vgeddes:
This is a follow-up to my previous draft Rococo PRs (#132 and #152) , and represents the final code required to integrate with the Rococo V1 network that is set to go live soon.

Parachains V1 has bleeding-edge dependencies, so we now depend on the master branches of Polkadot, Substrate, and Cumulus. As a side-effect such we've had to update or change quite a few things.

Changes:
* Upgrade to Substrate/Cumulus/Polkadot master
* Support for updated XCM apis in Polkadot/Cumulus master
* Renamed `Trait` to `Config` across all pallets (Upstream change in Substrate)
* Support for MultiAddress (https://github.com/paritytech/substrate/pull/7380) - In future this would allow us to map an ethereum address to an `AccountId` which is useful for a bunch of things.


parachain/node/src/command.rs
```diff
@@ -1,69 +1,404 @@
...
+		)
 	}
 
 	fn author() -> String {
 		env!("CARGO_PKG_AUTHORS").into()
 	}
 
 	fn support_url() -> String {
-		"http://www.snowfork.com/".into()
+		"http://www.snowfork.com".into()
```
musnit, line 56: snowbridge.snowfork.com

musnit (approved):

PR #201, opened 2021-01-08
rename snowdot to wrapped dot

musnit:


musnit:
@philipstanislaus thoughts on naming for this asset? i think maybe wrapped dot is better.

vgeddes:
What about `sDOT`. Is short, sweet and disambiguates from other wrapped DOT implementations.

musnit:
possibly, but i'm not sure we want disambiguation - we likely want ours to be the default/defacto one given its the most decentralized/trustless.

vgeddes:
Sure, even so, "sDOT" (or even eDOT) reads better from a marketing perspective than "WrappedDOT". The latter term is quite technical/generic and means we need to explain what "wrapped" means to laypersons.

musnit:
Wrapped is the most common terminology in the space. But yeh eDOT could
work too.

vgeddes:
> Wrapped is the most common terminology in the space. But yeh eDOT could work too.

Well, it depends on our audience for these docs I guess. Technical docs vs layperson docs. Wrapped would only be understood understood by a subset of crypto developers. Maybe we could use both "wrapped DOT" and "eDOT" in these docs, depending on the context

musnit:
sounds like you're not at all familiar with the crypto crowd :P. wrapped is
common throughout ethereum defi

philipstanislaus:
While wrapped is definitely the right technical term, I also have slight preference for a very short and sweet name. `sDOT` and `eDOT` feel shorter and easier to communicate about than `WrappedDOT` in my view.

musnit:
> While wrapped is definitely the right technical term, I also have slight preference for a very short and sweet name. `sDOT` and `eDOT` feel shorter and easier to communicate about than `WrappedDOT` in my view.

why not wDOT?

vgeddes:
I still prefer `sDOT` over `WDOT` given that we are using `sETH` for wrapped Ethereum. It would be kinda weird to have `WDOT` (on ethereum) and `sETH` (on polkadot). If we want to be consistent across the board, and reduce confusion, we should exclusively use `W` or exclusively use `s`.

philipstanislaus:
Yeah. And having branding in there `s` for Snowfork/Snowbridge does not hurt, does it?

philipstanislaus:
The `s` prefix might not be such a good idea as it's used by Synthetix: https://synthetix.io/

philipstanislaus:
We leave it at SnowETH/SnowDOT, right? 

Can we close this PR?

PR #203, opened 2021-01-12
Send bulk events from Eth -> Sub on header finalization

Rizziepit:
This PR makes the necessary changes to Substrate and the relayer to forward all messages in a block once the block is finalized.

Key parts of this PR:
* Substrate bridge module gets a `submit_bulk` dispatchable that accepts multiple message for multiple app IDs
* Substrate verifier module gets a `verify_bulk` function for the sake of efficiency, e.g. we only have to query storage once per header to check finality
* Relayer message channel changes from type `chain.Message` to `[]chain.Message`. Additionally, the messages are forwarded in chunks (Substrate will reject an extrinsic that's too large)


parachain/primitives/core/src/types.rs
```diff
@@ -24,6 +24,9 @@ pub struct Message {
 	pub verification: VerificationInput,
 }
 
+#[derive(Clone, Encode, Decode, Default, PartialEq, RuntimeDebug)]
+pub struct Messages(pub AppId, pub Vec<Message>);
```
vgeddes, line 28: Would prefer the name `MessageBatch` (even over other suggestions like `BulkMessages`, etc), as `Messages` could be confused for plain list of messages.

parachain/pallets/bridge/src/lib.rs
```diff
@@ -75,18 +77,51 @@ decl_module! {
...
+					let handler = handlers[i];
+					messages.1.iter()
+						.map(|msg| handler(&msg.payload))
+						.collect::<Vec<DispatchResult>>()
+				})
+				.flatten()
+				.filter_map(|r| r.err())
+				.collect();
+
+			if errors.is_empty() { Ok(()) } else { Err(Error::<T>::SkippedFailedMessages.into()) }
```
vgeddes, line 107: I don't think there's much value in returning a `SkippedFailedMessages` error. Its not really actionable, especially since we can't specify the specific underlying app error(s) that occurred.

Also, users who really want to know whether their message submission failed could always query storage to see whether the apps failed to emit an event.

I would consider just returning `Ok(())`, irrespective of whether we got errors from individual apps.

vgeddes (commented):
Looks very good! 🚀 

Though I still need to do a more thorough review tomorrow.

Also am wondering if we should make `submit_bulk` the only way to send messages. I'll sleep on it :-)

parachain/pallets/bridge/src/lib.rs
Rizziepit: > Also, users who really want to know whether their batch message submission failed could always query storage to see whether the apps emitted an event or not

Makes sense to me. I'll change

relayer/chain/ethereum/message.go
```diff
@@ -150,12 +150,49 @@ func MakeMessageFromEvent(event *etypes.Log, receiptsTrie *etrie.Trie, log *logr
...
 
-	return &msg, nil
+	return func() ([]chain.Message, bool) {
+		chunk := make([]chain.Message, 0)
+		count := 0
+
+		for i < len(addresses) {
+			if count == chunkSize {
+				return chunk, true
+			}
```
philipstanislaus, line 175: If `chunkSize == 0` (which might not make much sense, but is a valid input), this leads to an infinite loop in the caller.

philipstanislaus (commented):
Nice, thanks @Rizziepit 🎉 

There is one bigger concern if I'm not mistaken: The current design does not maintain order of messages across and also not within blocks, it only maintains order within applications. This means that messages from one app could be executed before messages from earlier blocks in another app are executed. If those apps interact, this could lead to indeterministic and unexpected behaviour.

If we wanted to maintain order of messages across, but also within blocks, the `MessageBatch` data structure does not give us many efficiency gains. You could only batch messages that are in the same block and all talk to the same target app. 

What is the reason to batch messages by app?

musnit:
> Nice, thanks @Rizziepit 🎉
> 
> There is one bigger concern if I'm not mistaken: The current design does not maintain order of messages across and also not within blocks, it only maintains order within applications. This means that messages from one app could be executed before messages from earlier blocks in another app are executed. If those apps interact, this could lead to indeterministic and unexpected behaviour.
> 
> If we wanted to maintain order of messages across, but also within blocks, the `MessageBatch` data structure does not give us many efficiency gains. You could only batch messages that are in the same block and all talk to the same target app.
> 
> What is the reason to batch messages by app?

I'm not sure the benefit of batching messages by app either, but ordering and inclusion of messages is guaranteed only by going through specific channels, not by the bridge as a whole. If apps or users expect guarantees around ordering and inclusion, they need to send their messages through a channel that guarantees this, so I don't see this as a problem at the bridge submission stage in this PR.

Follow up PR will add channels which can deal with that.

Related question though (@Rizziepit @vgeddes @philipstanislaus ): A channel can basically be implemented as an app too, ie, it's just an app that adds some guarantees on ordering/replay protection, and then a second time routes its payload to a second app.

eg: bridge -> channelapp -> otherapp

There are a some considerations for implementing channels in the design: 

1 - Are they the same thing as apps or not? Essentially, do we REQUIRE all apps to send/receive their messages via a channel, or can apps opt to send their messages directly, as in the current implementation?

2 - Do we implement them in their own pallets, or here in the bridge pallet?

I'd say it's probably simplest to implement our channels within the bridge pallet and require all apps to route their messages through a channel such that our chain and relayer only watches channel smart contract events and no direct app events. Then we disconnect the ETHApp and ERC20App from the bridge once channels are implemented. Thoughts?

vgeddes:
> I'm not sure the benefit of batching messages by app either, but ordering and inclusion of messages is guaranteed only by going through specific channels, not by the bridge as a whole. If apps or users expect guarantees around ordering and inclusion, they need to send their messages through a channel that guarantees this, so I don't see this as a problem at the bridge submission stage in this PR.
> 
> Follow up PR will add channels which can deal with that.

I remember now in the arch docs you envisioned channels doing things like ordering. Thinking about it some more though, dynamic reordering of messages on-chain sounds quite expensive. How do you envision it to work? 🤔 

I'm thinking the relayer is in a very good place to perform cheap message ordering, That ordering can then be verified by the on-chain Receive channel. So they can work together to ensure that ordering takes place.

> Related question though (@Rizziepit @vgeddes @philipstanislaus ): A channel can basically be implemented as an app too, ie, it's just an app that adds some guarantees on ordering/replay protection, and then a second time routes its payload to a second app.

I don't have any immediate answers or thoughts yet. Now that Rococo/parachain stuff is mostly out the way, my plan is to focus on these implementation details.

Rizziepit:
> There is one bigger concern if I'm not mistaken: The current design does not maintain order of messages across and also not within blocks, it only maintains order within applications. This means that messages from one app could be executed before messages from earlier blocks in another app are executed. If those apps interact, this could lead to indeterministic and unexpected behaviour.
> 
> If we wanted to maintain order of messages across, but also within blocks, the `MessageBatch` data structure does not give us many efficiency gains. You could only batch messages that are in the same block and all talk to the same target app.
> 
> What is the reason to batch messages by app?

Batching by app made processing a bit easier in the parachain, but it's not necessary. Your concern makes sense @philipstanislaus. I didn't take ordering guarantees into account when implementing this.

**My proposal for this PR:** make the relayer batch by block and maintain the event ordering within the block - parachain doesn't enforce this though, nor are there any guarantees of ordering across blocks
**Follow-up PR:** When I implement the in-channel in Substrate, the parachain starts enforcing ordering by using the nonce @musnit added in https://github.com/Snowfork/polkadot-ethereum/pull/208. This would enforce ordering in and across blocks. 

> I remember now in the arch docs you envisioned channels doing things like ordering. Thinking about it some more though, dynamic reordering of messages on-chain sounds quite expensive. How do you envision it to work?

 I share @vgeddes' concern. I want the ordering to be a channel concern, but implementation will be quite complicated unless we have a strict ordering where nonces in block<sub>x</sub> < nonces in block<sub>x+1</sub>. @musnit is it possible to implement the nonce that way?

musnit:
> Batching by app made processing a bit easier in the parachain, but it's not necessary. Your concern makes sense @philipstanislaus. I didn't take ordering guarantees into account when implementing this.
> 
> **My proposal for this PR:** make the relayer batch by block and maintain the event ordering within the block - parachain doesn't enforce this though, nor are there any guarantees of ordering across blocks
sounds fine, though the relayer will be enforcing this so may as well make the default behaviour of the relayer follow the enforced ordering. I think it should actually naturally, because the order in which events are stored and emitted with new blocks/subscriptions should match the order of nonces in #208, though ideally the relayer does keep track of what it's submitting, whether those submissions succeed or not, and what nonce it should be sending based on that.

> **Follow-up PR:** When I implement the in-channel in Substrate, the parachain starts enforcing ordering by using the nonce @musnit added in #208. This would enforce ordering in and across blocks.
yep sounds good

> > I remember now in the arch docs you envisioned channels doing things like ordering. Thinking about it some more though, dynamic reordering of messages on-chain sounds quite expensive. How do you envision it to work?
> 
> I share @vgeddes' concern. I want the ordering to be a channel concern, but implementation will be quite complicated unless we have a strict ordering where nonces in blockx < nonces in blockx+1. @musnit is it possible to implement the nonce that way?
there won't be any reordering happening on chain, nor anything dynamic. the relayer MUST submit events in order based on the ordering from the channel, and the channel will enforce on the ethereum side, when events are created, that nonces in blockx < nonces in blockx+1. The implementation in #208 already should do this.

musnit:
Also batching by app may still be valuable once channels are fully implemented, as our different channels will essentially be our apps, so we'll effectively be batching by channel.

parachain/config.json
```diff
@@ -22,7 +22,16 @@
 			"wsPort": 9988,
 			"port": 31200,
 			"balance": "1000000000000000000000",
-			"flags": ["--rpc-cors=all", "--enable-offchain-indexing", "--discover-local", "--", "--execution=wasm"]
+			"flags": [
+				"-lbasic_authorship=trace,import_header=trace,runtime=debug",
+				"--execution=native",
```
Rizziepit, line 27: @vgeddes I'm adding `--execution=native`. Without it, the PoW merkle proofs fail and Eth header imports won't work. I think it relates to the hashing functions but haven't gotten to the bottom of it yet

Rizziepit:
Updated it so that we're not batching by block or app ID. Technically, the relayer batches by block, but it's not explicit in the data structure. I think this is best - ordering enforcement will be done in another PR

parachain/config.json
vgeddes: ah that's concerning, but sure, fine for now :+1:

musnit: Um what does that do? Does it mess with the WASM code/stop compilation to wasm? May be issues with connecting to the relay chain if so

vgeddes: The parachain node on its host can execute either using its native or WASM runtime (both are builtin to the node executable). Note this is all after the WASM has been compiled and registered on the relay chain. Even if running natively, the node can only execute blocks that were validated on the relay chain using the registered WASM validation function.

musnit:
So for a relay chain node running our wasm validation function (which
includes pow verification), will it run successfully even with this flag?

Rizziepit:
> So for a relay chain node running our wasm validation function (which includes pow verification), will it run successfully even with this flag?

Doubtful. That might explain this issue I'm seeing: https://snowfork.slack.com/archives/G01C5QC3D4M/p1611125308034600?thread_ts=1611124625.034400&cid=G01C5QC3D4M

Rizziepit:
In addition to fixing ethash in WASM, I added a few things to this PR:
* The [`wasm-bindgen-test`](https://rustwasm.github.io/wasm-bindgen/wasm-bindgen-test/index.html) crate lets us run unit tests in WASM using `--target wasm32-unknown-unknown`. Note that you have to be in the crate dir. This works because of configuration at parachain/primitives/ethereum/.cargo/config:
```
cd parachain/primitives/ethereum
cargo test --target wasm32-unknown-unknown
```
* Substrate logging to target `submit_messages` to verify that message batches are received and processed (see screenshot)
Screenshot from 2021-01-25 23-20-33

cc @vgeddes

PR #205, opened 2021-01-12
Use offchain storage for message commitments

vgeddes:
Since the commitment hash is on-chain, in the block header, it is not necessary to store the committed messages on-chain too. We can securely store them offchain in Substrate's [Off-chain Indexing DB](https://substrate.dev/docs/en/knowledgebase/learn-substrate/off-chain-features#off-chain-indexing). This reduces storage costs, especially seeing as the ABI-encoded message vector is very bloated.

The Indexing DB is accessible via JSON-RPC. One catch is that we will need to configure nodes to maintain the DB using the `--enable-offchain-indexing` flag.


musnit:
can we add the --enable-offchain-indexing flag to this PR too plz :)?

parachain/pallets/commitments/src/lib.rs
```diff
@@ -105,11 +105,11 @@ impl<T: Config> Module<T> {
 		let commitment = Self::encode_commitment(&messages);
 		let commitment_hash: H256 = keccak_256(&commitment).into();
 
-		<Self as Store>::Commitment::set(commitment);
-
 		let digest_item = CustomDigestItem::Commitment(commitment_hash.clone()).into();
 		<frame_system::Module<T>>::deposit_log(digest_item);
 
+		sp_io::offchain_index::set(T::INDEXING_KEY, &commitment);
```
musnit, line 111: is this setting a new key for each commitment, or overwriting the same indexing key? we should be indexing by the commitment_hash so that we can query with the commitment_hash as input, right?

musnit:
also, we'll need an api/rpc endpoint to query for commitment by hash:
 - do we get rpc/apis for free, like with state queries? or do we need to add it?
 - are commitments being stored in the offchaindb in a way that allows for querying by hash?

vgeddes:
> also, we'll need an api/rpc endpoint to query for commitment by hash:
> 
> * do we get rpc/apis for free, like with state queries? or do we need to add it?

The DB can be queried using the json-rpc `offchain_localStorageGet` method, assuming persisten offchain indexing is enabled.

Example:

```bash
curl -H "Content-Type: application/json" -d '{"id":1, "jsonrpc":"2.0", "method": "offchain_localStorageGet", "params":["PERSISTENT", "0xdeabdbeef"]}' https://parachain-rpc.snowbridge.network
```


> * are commitments being stored in the offchaindb in a way that allows for querying by hash?

Good catch! I was just updating the data stored under a single key `commitments`, but I think your approach is better as we can have a persistent historical record of committed messages.

parachain/pallets/commitments/src/lib.rs
vgeddes: Good catch, that is the better approach. Will update

musnit:
looks good now - if test pass and the curl for commitment by hash works, ship it!

philipstanislaus (approved):
Looks good to me. 

Generic question, maybe I'm misreading the docs: Offchain storage is not populated among the blockchain network, which means that new nodes might not have the messages, right? 

Does that imply that a rotation of validators could lead to loss of commitments?

musnit:
yeh i think so. data availability is a problem for most bridge/layer 2
solutions in general. in our case, we don't really care about commitments
once they have been sent over to ethereum, which should happen within 2min,
so as long as there is not some major rotation happening within a really
short time period, with nodes going offline, we should be fine?

philipstanislaus:
I'm actually not sure. Even if offchain storage is not populated, the messages themselves are as the node is syncing the chain. Not sure how that works with light sync. Created #209 to check that.

vgeddes:
According to https://substrate.dev/docs/en/knowledgebase/learn-substrate/off-chain-features#off-chain-indexing the offchain DB can be easily recreated:

> Unlike OCWs, which are not executed during initial blockchain synchronization, off-chain indexing is populating the storage every time a block is processed, so the data is always consistent and will be exactly the same for every node with indexing enabled.

PR #208, opened 2021-01-13
Ethereum out-channel

musnit:
This PR adds a solidity implementation of a channel based on https://snowbridge-docs.snowfork.com/concepts/basic-ethereum-to-polkadot-message-channel

Some notes:
 - Incentivized and unincentivized are the same for now. On the receiving side, the former will use ordered replay protection and the latter will use unordered. Later on, incentives will be added to the former.
 - Just ETHApp implemented for now
 - Still WIP on erc20app + relayer + parachain changes, will look into them next


musnit:
@Rizziepit @vgeddes question: what would be best to use for the target application id (https://github.com/Snowfork/polkadot-ethereum/pull/208/files#diff-303a8400e2b020c3f34f316d13e19ec7b0648d4cf79601a653209891f455047eR15) such that it can be routed cleanly by our bridge pallet to the right destination pallet? Is 'eth-app' fine?

ethereum/contracts/OutChannel.sol
```diff
@@ -0,0 +1,14 @@
+ // SPDX-License-Identifier: MIT
+pragma solidity >=0.6.2;
+
+// OutChannel contains methods that all outgoing channels must implement
+abstract contract OutChannel {
+
+    event NewMessage(uint256 nonce, address senderAddress, string targetApplicationId, bytes payload);
```
philipstanislaus, line 7: `nonce` in here implies that we use nonces as replay protection on every `OutChannel`. Can we fing a more generic term?  Maybe `replayProtection`?

ethereum/contracts/ETHApp.sol
```diff
@@ -36,6 +49,16 @@ contract ETHApp is Application {
         totalETH = totalETH.add(msg.value);
 
         emit AppTransfer(msg.sender, _recipient, msg.value);
```
philipstanislaus, line 51: Do we still want to emit that?

ethereum/test/test_eth_app.js
```diff
@@ -41,11 +48,14 @@ contract("EthApp", function (accounts) {
       const weiAmount = web3.utils.toWei("0.25", "ether");
 
       // Deposit Ethereum to the contract and get the logs of the transaction
-      const { logs } = await this.ethApp.sendETH(
+      const tx = await this.ethApp.sendETH(
         recipient,
+        true,
```
philipstanislaus, line 53: Shall we add a test case for non incentivized as well?

ethereum/test/test_eth_app.js
```diff
@@ -81,6 +120,7 @@ contract("EthApp", function (accounts) {
         // Send to a substrate recipient to load contract with unlockable ETH
         await this.ethApp.sendETH(
           substrateRecipient,
+          true,
```
philipstanislaus, line 123: Likewise?

philipstanislaus (commented):
Nice!

ethereum/contracts/OutChannel.sol
musnit: nonce is the generic term :P - https://en.wikipedia.org/wiki/Cryptographic_nonce

ethereum/contracts/ETHApp.sol
musnit: could be useful for things like block explorers and whatever else, though can remove later on if we need to hyper optimize for gas

ethereum/contracts/BasicOutChannel.sol
```diff
@@ -0,0 +1,23 @@
...
+    constructor() public {
+        currentNonce = 0;
+    }
+
+    /**
+     * @dev Submits a message to the channel
+     */
+    function submit(string memory targetApplicationId, bytes memory payload) public override {
+        emit NewMessage(currentNonce, msg.sender, targetApplicationId, payload);
+        currentNonce++;
```
Rizziepit, line 20: Just checking my understanding... I'm assuming `payload` cannot be replayed here because of the way events are implemented in the EVM?

Rizziepit:
> @Rizziepit @vgeddes question: what would be best to use for the target application id (https://github.com/Snowfork/polkadot-ethereum/pull/208/files#diff-303a8400e2b020c3f34f316d13e19ec7b0648d4cf79601a653209891f455047eR15) such that it can be routed cleanly by our bridge pallet to the right destination pallet? Is 'eth-app' fine?

I don't see any issues with `eth-app`

ethereum/contracts/ETHApp.sol
vgeddes: yeah is useful for informational purposes. The substrate side also emits an informational `Minted` event. 

I would rename it to `Locked` though.

ethereum/contracts/BasicOutChannel.sol
```diff
@@ -0,0 +1,23 @@
...
+    uint256 public currentNonce;
+
+    constructor() public {
+        currentNonce = 0;
+    }
+
+    /**
+     * @dev Submits a message to the channel
+     */
+    function submit(string memory targetApplicationId, bytes memory payload) public override {
```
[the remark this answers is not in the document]
vgeddes: Also conceptually, at level of a channel, `send` is more commonly used than `submit`.

```suggestion
    function send(string memory target, bytes memory payload) public override {
```

vgeddes (commented):
Looks good! Added a few comments mostly about naming.

Further on that topic I'd like us to consider renaming `OutChannel` to `SendChannel`. The latter is much more common in the APIs that I've seen and worked with over the years, and so developers would prefer it.

ethereum/contracts/BasicOutChannel.sol
musnit: well, the intention here is that this NewMessage will only be accepted by a corresponding InChannel on the parachain side, not by any other other pallet/channel, and that this corresponding InChannel will use the nonce for replay protection. The payload itself/EVM event implementation details are not really part of the replay protection consideration?

ethereum/contracts/BasicOutChannel.sol
```diff
@@ -0,0 +1,23 @@
+ // SPDX-License-Identifier: MIT
+pragma solidity >=0.6.2;
+
+import "./OutChannel.sol";
+
+// BasicOutChannel is a basic channel that just outputs messages with a nonce.
+contract BasicOutChannel is OutChannel  {
+
+    uint256 public currentNonce;
```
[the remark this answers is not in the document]
musnit: not sure what you mean by nonce being inherently current? nonces in general do not need to be current/have any ordering/time based attribute. the ones in this contract do specifically have ordering due to them being increasing integers, and so i think currentNonce (or maybe nextNonce, rather?) does make sense?

ethereum/contracts/BasicOutChannel.sol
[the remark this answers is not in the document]
musnit: totally disagree, i'm a fan of long variable names as they increase readability :D 

send sounds better though yeh, agree.

ethereum/scripts/dumpRelayerConfig.js
```diff
@@ -4,18 +4,18 @@ const fs = require('fs');
 const os = require('os');
 
 const Bridge = artifacts.require("Bridge")
-const ETHApp = artifacts.require("ETHApp")
-const ERC20App = artifacts.require("ERC20App")
```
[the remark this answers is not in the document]
musnit: yeh, need to implement for erc20 too

but anyway, i don't think breaking the e2e tests is a problem. they're kinda useless now that they're testing the wrong thing anyway, so breaking and then fixing them in follow up PRs should be fine?

relayer/README.md
```diff
@@ -77,13 +77,13 @@ endpoint = "ws://localhost:9545/"
 address = "0x17f7C1e314180D8b8588CA50cF09A0e0847c77F6"
 abi = "/tmp/Bridge.json"
 
-[ethereum.apps.eth]
```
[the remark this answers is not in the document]
musnit: was hoping to actually get the substrate changes in this same PR such that we could merge the whole PR in together and keep the main branch fully functional, rather than merging in one piece at a time. anyway either way is fine, i don't really think it makes a major difference.

ethereum/test/test_eth_app.js
musnit: refactored tests, have separate tests for the apps and for channels, will push soon.

ethereum/test/test_eth_app.js
musnit: refactored tests, have separate tests for the apps and for channels, will push soon.

ethereum/contracts/ETHApp.sol
```diff
@@ -1,32 +1,45 @@
 // SPDX-License-Identifier: MIT
 pragma solidity >=0.6.2;
+pragma experimental ABIEncoderV2;
```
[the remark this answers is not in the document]
musnit: @vgeddes actually this is a mission because our dependencies (openzeppelin) dont support 0.8.0 on master yet and truffle doesn't easily allow mixing versions.

ethereum/contracts/BasicOutChannel.sol
Rizziepit: I'll continue this convo elsewhere. I think my lack of Solidity knowledge is the problem

PR #218, opened 2021-01-19
Change relayer to listen for message commitments

vgeddes:
Plus improvements to the commitments pallet.

Depends on this GSRPC PR: https://github.com/Snowfork/go-substrate-rpc-client/pull/1


parachain/pallets/commitments/src/tests.rs
```diff
@@ -64,3 +64,13 @@ fn test_add_message() {
 
 	});
 }
+
+
+#[test]
+fn test_make_offline_key() {
```
musnit, line 70: offline_key or offchain_key?

relayer/chain/substrate/writer.go
```diff
@@ -110,7 +110,7 @@ func (wr *Writer) write(_ context.Context, c types.Call) error {
 
 	ext := types.NewExtrinsic(c)
 
-	era := types.ExtrinsicEra{IsMortalEra: false}
+	era := types.ExtrinsicEra{IsImmortalEra: true}
```
musnit, line 113: why immortal?

parachain/pallets/commitments/src/lib.rs
```diff
@@ -107,15 +110,13 @@ impl<T: Config> Module<T> {
 		let messages: Vec<Message> = <Self as Store>::MessageQueue::take();
```
musnit, line 110: Not too related to this PR, though could be included here - can we skip creating a commitment if the MessageQueue is empty (eg: in the scenario where a few 100 blocks pass and there is no activity)?

parachain/pallets/commitments/src/tests.rs
vgeddes: Actually I removed the test as I was only using it to print the key

relayer/chain/substrate/writer.go
vgeddes: It was always immortal. I just changed the initialization to be more explicit about the fact that it is immortal.

I tried to make it mortal but it seems fairly complicated to do due to lack of docs. Anyway we do have a ticket for it.

parachain/pallets/commitments/src/lib.rs
vgeddes: Fixed :+1:

relayer/chain/substrate/listener.go
```diff
@@ -103,37 +94,39 @@ func (li *Listener) pollBlocks(ctx context.Context) error {
 				continue
 			}
 
-			// Get hash for latest block, sleep and retry if not ready
-			hash, err := li.conn.api.RPC.Chain.GetBlockHash(currentBlock)
+			digestItem, err := digest.GetAuxiliaryDigestItem(finalizedHeader.Digest)
```
musnit, line 97: do we need to query for a digestitem on each block, ie, proactively, or can we just subscribe to some event like NewDigestItemCreated() and then query it reactively?

musnit: maybe should add an issue for this?

relayer/chain/substrate/listener.go
```diff
@@ -103,37 +94,39 @@ func (li *Listener) pollBlocks(ctx context.Context) error {
...
+				if data != nil {
+					li.log.WithFields(logrus.Fields{
+						"block":               finalizedHeader.Number,
+						"commitmentSizeBytes": len(*data),
+					}).Debug("Retrieved commitment from offchain storage")
+				} else {
+					li.log.WithError(err).Error("Commitment not found in offchain storage")
+				}
 
-			li.handleEvents(currentBlock, events)
```
musnit, line 136: will need li.handleCommitment(...), though can do in follow up PR. see https://github.com/Snowfork/polkadot-ethereum/pull/219/files#diff-42f6d75aa7d06e027e7684545996c8f782c82c6fc53f288c6c5852c6e26bfb71R36 for the expected interface to submit it to solidity

relayer/chain/substrate/listener.go
vgeddes: > do we need to query for a digestitem on each block, ie, proactively, or can we just subscribe to some event like NewDigestItemCreated() and then query it reactively? i guess it doesn't matter too much either way.

Using events as a signal wouldn't add any value and in fact would complicate things. The relayer already fetches each block header, so its much simpler and cheaper to just pro-actively look for the commitment hash in each header. 

Additionally if at all possible I'd like the relayer to avoid having to query any events, as it means we need to maintain all the event types in the relayer so that gsrpc can decode them. That caused great pain in the past for me 🙂 

As an aside, even the Polkadot-JS UI shows all digest items for a block header, which does help with debugging.

> in a future PR the relayer should query the eth contract for the latest parachain block that has been processed and then resume the substrate listener from that block so that it can catch up even if it crashes and falls behind.

Added https://github.com/Snowfork/polkadot-ethereum/issues/224

musnit: yeh, thinking about how this integrates with the mmr stuff too - i'll draw up a flow chart now quickly and share for review.

parachain/pallets/erc20-app/src/lib.rs
```diff
@@ -88,19 +86,15 @@ decl_module! {
...
-			let message = OutPayload {
+			let message = OutboundPayload {
 				token_addr: token_addr,
 				sender_addr: who.clone(),
 				recipient_addr: recipient,
 				amount: amount
 			};
-			T::Commitments::add(Self::address(), message.encode());
-
+			T::MessageCommitment::add(Self::address(), 0, &message.encode());
```
philipstanislaus, line 95: How do we deal with nonces here? This `0` is static, but should be incrementing, right?

parachain/pallets/eth-app/src/lib.rs
```diff
@@ -90,18 +87,14 @@ decl_module! {
...
 
-			let message = OutPayload {
+			let message = OutboundPayload {
 				sender_addr: who.clone(),
 				recipient_addr: recipient,
 				amount: amount
 			};
-			T::Commitments::add(Self::address(), message.encode());
-
+			T::MessageCommitment::add(Self::address(), 0, &message.encode());
```
philipstanislaus, line 95: As above, should this not be incrementing?

philipstanislaus (commented):
Nice, happy to see we finally use the commitments 🎉 

Left a few comments below. One generic thought: how does the trust model change if the message sender can pick a nonce, as opposed to the bridge assigning message nonces? Did we discuss those trade-offs somewhere?

parachain/pallets/erc20-app/src/lib.rs
vgeddes: That was just a placeholder for the nonce handling I've implemented in this follow-up PR: https://github.com/Snowfork/polkadot-ethereum/pull/223/files#diff-a3f0c878bcb36ffd0b0fabe45b220da7797857e800a53208b5fce46161dd50a4R41

The outbound channel in the above PR is responsible for incrementing the nonce.

parachain/pallets/eth-app/src/lib.rs
vgeddes: That was just a placeholder for the nonce handling I've implemented in this follow-up PR: https://github.com/Snowfork/polkadot-ethereum/pull/223/files#diff-a3f0c878bcb36ffd0b0fabe45b220da7797857e800a53208b5fce46161dd50a4R41

parachain/pallets/erc20-app/src/lib.rs
philipstanislaus: Cool, thanks for clarifying 👍

parachain/pallets/eth-app/src/lib.rs
philipstanislaus: 👍

PR #219, opened 2021-01-19
Incentivized receive channel

musnit:
This PR is the result of initial architecture and pseudocode design for our ethereum Incentivized Receive Channel that supports dispatch to arbitrary contracts and controls for gas constraints.

Next step will be to flesh out the code into a working MVP with better abstractions.

@denalimarsh will take over implementation of working verifyCommitment and processCommitmentContents logic that can be integrated with the relayer for an end to end test.

The plan is:
 - [ ] _(in this PR)_ Complete verifyCommitment, however keep the MMRLeaf verification and ParachainBlockHeader verification commented out, such that only commitment<>commitmentContents verification is done properly
 - [ ]  _(in this PR)_ Complete processCommitmentContents, implementing the replay protection and arbitrary contract call dispatch of payload, however leave out the gas controls/constraints.
 -  _(soon after this PR)_ Integrate relayer with verifyCommitment endpoint to get working e2e test of Polkadot -> Ethereum flow with channels. This may require some slight modifications to eth-app and erc2-app to ensure their OutPayload is correctly ABI-encoded bytes that fit into the [call() opcode](https://docs.soliditylang.org/en/v0.5.3/assembly.html) nicely. cc @vgeddes @Rizziepit 
 - _(later PR)_ add test cases for actual verification implementation, covering various attack vectors and edge cases
 - _(later PR)_ start to integrate with MMRLeaf/Beefy/parachain block header verification for full trustlessness and get all test cases passing
  - _(later PR)_ add test cases for gas constraints/attack vectors and edge cases on gas attacks or dispatch attacks
 - _(later PR)_ Complete gas constraints for full deliverability guarantees and get test cases passing
 - _(later PR)_ Complete incentives for full delivery guarantees


ethereum/contracts/IncentivizedReceiveChannel.sol
```diff
@@ -0,0 +1,139 @@
...
+        for (uint256 i = 0; i < commitmentContents.messages.length; i++) {
+            // Check message nonce is correct and increment nonce for replay protection
+            Message memory message = commitmentContents.messages[i];
+            require(message.nonce == lastProcessedNonce + 1, "invalid nonce");
+
+            lastProcessedNonce = lastProcessedNonce + 1;
+
+            // Deliver the message to the destination
+            // Delivery will have fixed maximum gas allowed for the destination app.
+            bytes memory callInput = message.payload;
```
vgeddes, line 109: Following on from our convo this morning - I suspect you are wanting to implement a generic RPC protocol by including the method ID in the message payload, in essence allowing a substrate application pallet to call any method on an Ethereum contract.

I think some kind of generic RPC like this can be useful but there are a few considerations we need to take into account. 

Keen to hop onto a call later when you're ready to discuss this.

denalimarsh: I'd like to join the call as well, the considerations are probably relevant to ongoing work on https://github.com/Snowfork/polkadot-ethereum/pull/228

denalimarsh: Can reference https://github.com/Snowfork/polkadot-ethereum/pull/228 for the proper payload encoding. The payload byte contents are:
- the function signature. This is the first 8 bytes of the result of `keccak256(functionName(argVariableType1,argVariableType2,...))`
- each variable padded to 32 bytes

An example ETHApp Unlock function payload is `0xed581fd300000000000000000000000022448f19a3cb4ee570c2ec6d3323761d3399bbbd0000000000000000000000000000000000000000000000000000000000000002`
- Function signature: `0xed581fd3` derived from `keccak256(Unlock(address, uint256))`
- Recipient address: `00000000000000000000000022448f19a3cb4ee570c2ec6d3323761d3399bbbd`
- Amount: `0000000000000000000000000000000000000000000000000000000000000002`

denalimarsh:
PR https://github.com/Snowfork/polkadot-ethereum/pull/228 implements `verifyCommitment` and `processCommitmentContents`, merging it into this PR will complete this PRs TODOs.

denalimarsh:
Previously Ethereum applications needed to implement a 'handle' function that parsed message data and delivered it to the correct internal function. This means that smart contracts already deployed to Ethereum can't receive messages from the Bridge (unless someone deployed an application-specific parser for each contract). Instead of instantiating a contract and calling a specific method, Ethereum enables transactions containing data to be sent directly to any address. If the receiving address is a contract it uses a function signature (described [here](https://github.com/Snowfork/polkadot-ethereum/pull/219#discussion_r563605329)) prepended to the data to decide how to process the data. By adding the function signature to the message payload we can deliver data to the correct function of any arbitrary Ethereum application.

ethereum/contracts/IncentivizedReceiveChannel.sol
vgeddes: As I've discussed with Denali on Slack, the ability to call any method on a target contract does change the design a bit, and that leads to inconsistencies in several areas:

* On the Substrate side, apps still need to implement a `handle` method. Theoretically yes we can update the design to support calling any dispatchable, but practically that's not going to be implemented any time soon. 
* One suggested use-case for the proposed feature was to have the bridge support Ethereum contracts which are already deployed and live, and of course don't implement `handle` methods. I can see where the feature helps with that. However we need to consider the opposite direction where these old previously-deployed contracts need to submit mesages to the Send Channel contract. That's currently not possible without deploying an intermediary adaptor contract. But now we have this inconsistency where the adaptor isn't needed in the inbound direction, and only in the outbound direction. This adaptor contract needs to somehow fit in with the design and I don't think we have thought that out enough.

Then apart from the above design inconsistencies, there are other concerns:
* Substrate applications now need to know what method to call on their peer contract, so this adds complexity and configuration to the Substrate side where previously none existed.
* Previously, end-users who want to develop Ethereum apps on our bridge would have only needed to write a contract that implements an `IApplication` interface with a `handle` function. This makes things simpler for users as they know what they need to implement. They just need to fill in an implementation for `handle` and its done.

Given all the above, I suggest we limit the scope of this PR (incentivized channel) and defer this message dispatch feature until later on when we've reached some kind of consensus on the above.

vgeddes: > i think these inconsistencies are totally fine, don't see any major issue with them.

Its hard to reply to that. Design can be subjective.

Shouldn't this PR be focused on receive channel stuff? The proposed message dispatch feature is really orthogonal to that I think.

> Not really - our pallets already have configuration/complexity requirements, i don't think this adds much more?

It still adds up though.

> Huh? No, now they are not constrained to any interface, they can develop ethereum apps with solidity as normal - seems much simpler?

Nope, I still think it makes it harder for users for the reasons I mentioned in my original point.

ethereum/contracts/IncentivizedReceiveChannel.sol
```diff
@@ -0,0 +1,173 @@
...
+                    callInputDataPointer, // Store output over input (saves space)
+                    callInputLength // Outputs can be as long as input length
+                )
+            }
+
+            emit MessageDelivered(message.nonce, result);
+        }
+    }
+
+    function validateCommitment(
```
vgeddes, line 141: The implementation here doesn't match how commitment hashes are actually produced on the Substrate side:

```
hash = keccak256(abiEncode(messages))
```

I implemented it like that for several reasons, one of them being that it should be theoretically simpler and cheaper to reproduce the hash on the Ethereum side.

Since `messages` can be made available in the calldata for the  `newParachainCommitment()` (if we specify that storage location), we can just extract the bytes from the calldata and hash it.

musnit: yep, @denalimarsh mentioned some issues he encountered with the hashing, so still need to resolve/fix this

musnit: agree accessing and hashing the exact bytes via calldata slots should be more efficient though yep abi.encode(messages) will be ok as an alternative if there are issues with that

ethereum/contracts/IncentivizedReceiveChannel.sol
```diff
@@ -0,0 +1,173 @@
...
+            bytes memory callInput = message.payload;
+
+            bool result;
+            uint256 callInputLength = callInput.length;
+            assembly {
+                let callInputDataPointer := add(callInput, 32) // Move 32 ahead of callInput length slot to get to actual data slot (TODO: verify this works in all cases)
+
+                // Dispatch call to the receiver - it is expected to be fire and forget. If the call reverts, runs out of gas, error,
+                // etc, its the fault of the sender
+                result := call(
```
musnit, line 126: can probably replace with contract.call in solidity instead of assembly

philipstanislaus:
Bit late to the party. Just a generic comment on architecture: 

I would describe 2 "pure" bridge modes:

1. Pure push mode (could also call it RPC mode): sending blockchain defines destination, contract/method, arguments. Works without any changes to the receiving contract by just doing regular contract calls. Sender pays for the execution. Use case: send data to another blockchain, transfer tokens.
2. Pure pull mode (could also call it publish/subscribe mode): receiving blockchain defines events to listen for/subscribe, defines handler as entrypoint to handle those events. Works without any changes to publisher/emitter contract by just subscribing to events. Receiver pays for execution. Use case: react to things happen on another blockchain, e. g. oracle functionality for trades on uniswap to have up to date pricing information.

Both have their valid use cases, and we might want to implement both on our bridge at a later point.

What we currently have/had is a mix where sender and receiver contracts need both to adhere to an interface for sending/dispatching and receiving. 

I feel the bridge would be more usable by becoming more pure push or pure pull for now. Since most of our application is implementing a push mode, I am all for going further into that direction.

@musnit shall we add those push/pull modes to the documentation? They might have some users understand better how the bridge works.

ethereum/contracts/IncentivizedReceiveChannel.sol
```diff
@@ -0,0 +1,173 @@
...
+contract IncentivizedReceiveChannel {
+    using Decoder for bytes;
+
+    uint256 public lastProcessedNonce;
+
+    uint256 public MAX_PAYLOAD_BYTE_SIZE = 1000;
+    uint256 public MAX_PAYLOAD_GAS_COST = 500000;
+    uint256 public EXTERNAL_CALL_COST = 21000;
+    uint256 public MAX_GAS_PER_MESSAGE =
+        EXTERNAL_CALL_COST + MAX_PAYLOAD_BYTE_SIZE + EXTERNAL_CALL_COST;
```
philipstanislaus, line 16: Shall we not multiply `MAX_PAYLOAD_BYTE_SIZE` by a `GAS_COST_PER_PAYLOAD_BYTE` here?

ethereum/contracts/IncentivizedReceiveChannel.sol
```diff
@@ -0,0 +1,173 @@
...
+    constructor() public {
+        lastProcessedNonce = 0;
+    }
+
+    event MessageDelivered(uint256 _nonce, bool _result);
+
+    function newParachainCommitment(
+        Commitment memory commitment,
+        CommitmentContents memory commitmentContents,
+        uint256 parachainBlockNumber,
```
philipstanislaus, line 42: This is not used – do we really need it?

ethereum/contracts/IncentivizedReceiveChannel.sol
```diff
@@ -0,0 +1,173 @@
...
+            validateCommitment(commitment, commitmentContents),
+            "invalid commitment"
+        );
+
+        // Require there is enough gas to play all messages
+        require(
+            gasleft() >=
+                commitmentContents.messages.length * MAX_GAS_PER_MESSAGE,
+            "insufficient gas for delivery of all messages"
+        );
```
philipstanislaus, line 90: We also need to account for gas that's used by the rest of the current function. Probably need to benchmark to find out how much we need to add exactly.

ethereum/contracts/IncentivizedReceiveChannel.sol
```diff
@@ -0,0 +1,173 @@
...
+        return true;
+    }
+
+    function processCommitmentContents(
+        CommitmentContents memory commitmentContents
+    ) internal {
+        for (uint256 i = 0; i < commitmentContents.messages.length; i++) {
+            // Check message nonce is correct and increment nonce for replay protection
+            Message memory message = commitmentContents.messages[i];
+            require(message.nonce == lastProcessedNonce + 1, "invalid nonce");
```
philipstanislaus, line 109: Why do we need message nonces? Since we process all messages from one commitments, a nonce per commitment would be sufficient.

ethereum/contracts/IncentivizedReceiveChannel.sol
```diff
@@ -0,0 +1,173 @@
...
+                // etc, its the fault of the sender
+                result := call(
+                    allowedGas, // Allowed gas
+                    targetApplicationAddress, // To addr
+                    0, // No ether value being sent
+                    callInputDataPointer, // Inputs are stored at callInputDataPointer
+                    callInputLength, // Input length
+                    callInputDataPointer, // Store output over input (saves space)
+                    callInputLength // Outputs can be as long as input length
+                )
```
philipstanislaus, line 134: Can we add a task to think about security vulnerabilities of this mechanism? Re-entrancy into the receive channel does not seem to be possible (a message would need to know the commitment hash for re-entrancy), but we should look into other potential attacks.

ethereum/contracts/IncentivizedReceiveChannel.sol
```diff
@@ -0,0 +1,173 @@
...
+                bytes32 messageHash =
+                    hashMessage(commitmentContents.messages[i]);
+                commitmentHash = keccak256(
+                    abi.encodePacked(commitmentHash, messageHash)
+                );
+            }
+        }
+        return
+            keccak256(abi.encodePacked(commitmentHash)) ==
+            keccak256(abi.encodePacked(commitment.commitmentHash));
```
philipstanislaus, line 159: Why do we need to hash again here? We could just do the following, right?

```suggestion
            commitmentHash == commitment.commitmentHash;
```

philipstanislaus (approved):
LGTM

musnit:
I like those modes/abstractions in theory, but I'm not sure how they hold up in practice...thinking...:
- Eth -> Polkadot:
  - Pure Push: Polkadot side still needs some kind of channel/logic for handling replay protection, ordering. I suppose if pure push was implemented with ETH waiting for a second feedback/confirmation event for delivery confirmation then these could perhaps be avoided, but this seems more complex too?
  - Pure Pull: I think this could work, though not sure how the receiver would pay for execution? Eg: in the case of PolkaETH, do they need to send a transaction to polkadot first, to notify their intention to receive, and then send their transaction to Ethereum to lockup PolkaETH after? Could be a worse UX
- Polkadot -> Eth:
 - Pure Push: We still basically need our ReceiveChannel as above to handle MMR stuff/replay protection?
 - Pure Pull: This could work, though potentially same UX issues as above?

There may be another way to think about/abstract things/resolve things that preserves your abstractions though - can perhaps discuss on call.

ethereum/contracts/IncentivizedReceiveChannel.sol
musnit: ye looks like a bug, will fix, though in general fee/incentive/gas constraints are planned to do properly in follow up milestone/PR

ethereum/contracts/IncentivizedReceiveChannel.sol
musnit: yeh, will need to do proper gas-related review and changes in next pr/milestone

ethereum/contracts/IncentivizedReceiveChannel.sol
musnit: i feel like i put this here for a reason, though can't remember now lol, will think about it and see

ethereum/contracts/IncentivizedReceiveChannel.sol
musnit: yeh, will create an issue for this potential optimization

ethereum/contracts/IncentivizedReceiveChannel.sol
musnit: sounds good - will be simpler too if we replace this with a solidity call rather than a assembly call

ethereum/contracts/IncentivizedReceiveChannel.sol
musnit: ye @denalimarsh is working on fixing this.

philipstanislaus:
> * Eth -> Polkadot:
>   * Pure Push: Polkadot side still needs some kind of channel/logic for handling replay protection, ordering. I suppose if pure push was implemented with ETH waiting for a second feedback/confirmation event for delivery confirmation then these could perhaps be avoided, but this seems more complex too?

Absolutely, by "pure" above, I just imply that the receiving contract/function does not need to know about the bridge. Of course verification, reply protection, ordering and routing are necessary before the receiving contract/function is called.

>   * Pure Pull: I think this could work, though not sure how the receiver would pay for execution? Eg: in the case of PolkaETH, do they need to send a transaction to polkadot first, to notify their intention to receive, and then send their transaction to Ethereum to lockup PolkaETH after? Could be a worse UX

Yes, that flow would be different: The subscriber would deposit funds to the bridge on Polkadot that will cover future cost of received messages. These fees could be withdrawn at the receiving side from users that receive funds.

> * Polkadot -> Eth:
>   * Pure Push: We still basically need our ReceiveChannel as above to handle MMR stuff/replay protection?

Pure push is in my view what we are implementing with this PR. There of course must still be a recipient contract (our ReceiveChannel) on Ethereum that verifies the commitment and messages. Replay protection is handled by a nonce on the commitment. 

>   * Pure Pull: This could work, though potentially same UX issues as above?

This is almost impossible to do with our current architecture, since the light client uses messages in commitments, which require an on-chain call to be included. For this to work, we would need to be able to prove any event on Polkadot efficiently.

> There may be another way to think about/abstract things/resolve things that preserves your abstractions though - can perhaps discuss on call.

We are getting quite off-topic here – let's discuss those points next time we speak

philipstanislaus (approved):

musnit:
ah right, yeh i think we agree. we are going for pure push. polkadot -> eth has this implemented already. eth -> polkadot will likely evolve into this in future when we want to support dynamic dispatch to other parachains/pallets.

ethereum/contracts/IncentivizedReceiveChannel.sol
denalimarsh: Addressed in https://github.com/Snowfork/polkadot-ethereum/pull/219/commits/35478272b36387ce85ca211d150cf839f39abb57

ethereum/contracts/IncentivizedReceiveChannel.sol
denalimarsh: Addressed in https://github.com/Snowfork/polkadot-ethereum/pull/219/commits/c1c84636367643bd86fe7319869912a7dc2cdb2d

ethereum/contracts/IncentivizedReceiveChannel.sol
denalimarsh: Updated the `validateCommitment` implementation in https://github.com/Snowfork/polkadot-ethereum/pull/219/commits/35478272b36387ce85ca211d150cf839f39abb57

musnit:
e2e tests all pass, should be good to merge this after e2e-test-fixes is merged

ethereum/contracts/IncentivizedReceiveChannel.sol
musnit: https://github.com/Snowfork/polkadot-ethereum/issues/251

ethereum/contracts/IncentivizedReceiveChannel.sol
musnit: noted in https://github.com/Snowfork/polkadot-ethereum/pull/250

ethereum/contracts/IncentivizedReceiveChannel.sol
musnit: added in https://github.com/Snowfork/polkadot-ethereum/pull/250

PR #221, opened 2021-01-21
WIP prototype for channels in Substrate

vgeddes:
WIP prototype for channels in Substrate

Added some half-baked prototype code to show how we can model channels. Doesn't compile and still needs more things implemented. But it shows the overall idea.

Inspired by the message lane concept in parity-bridges-common (See https://github.com/paritytech/parity-bridges-common/blob/master/modules/message-lane/src/inbound_lane.rs for example).

Channels are represented as plain structs with attached methods (ie no pallets). Using some fancy storage logic, persistent data (nonces, etc) is stored in the bridge module.

For simplicitly of implementation:
* All code is in the bridge module instead of being spread out over multiple pallets
* Fixed number of channels:
  * 1 pair of outbound channels (basic and incentivized),
  * 1 pair of inbound channels (basic and incentivized).
* Each channel pair shares the same persistent storage area.

Have renamed SendChannel/ReceiveChannel concepts to OutboundChannel/InboundChannel.

Of course open to all ideas and discussion points on the above.

### DIFF
--- a/parachain/pallets/bridge/src/lib.rs
+++ b/parachain/pallets/bridge/src/lib.rs
@@ -31,17 +31,18 @@ use sp_runtime::RuntimeDebug;
 use codec::{Encode, Decode}
 
 use artemis_core::{AppId, Application, Message, Verifier};
+use primitives::{self, ChannelId, InboundChannel, OutboundChannel};
 
-use primitives::{self, ChannelId, MessageNonce};
+use inbound_channel::make_inbound_channel;
+use outbound_channel::make_outbound_channel;
 
 pub mod primitives;
 pub mod inbound_channel;
+pub mod outbound_channel;
 
 pub trait ApplicationRegistry {
-
 	fn iter() -> Iterator<Application>;
 }
-
 pub trait Config: system::Config {
 	type Event: From<Event> + Into<<Self as system::Config>::Event>;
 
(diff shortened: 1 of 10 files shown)


musnit:
> Have renamed SendChannel/ReceiveChannel concepts to OutboundChannel/InboundChannel.
lol we just renamed them from out/in to send/receive based on last ethereum PR

vgeddes:
> > Have renamed SendChannel/ReceiveChannel concepts to OutboundChannel/InboundChannel.
> > lol we just renamed them from out/in to send/receive based on last ethereum PR

Sorry!  😃

parachain/pallets/bridge/src/lib.rs
```diff
@@ -24,32 +26,75 @@ use frame_system::{self as system, ensure_signed};
...
 	}
 }
 
-decl_event!(
+decl_event! {
     /// Events for the Bridge module.
 	pub enum Event {
+
+		/// Message accepted for delivery to Ethereum
+		MessageAccepted(ChannelId, Option<MessageNonce>)
```
vgeddes, line 95: Parity-bridges-common also emits a `MessageDelivered` if the Ethereum side actually got the message. Pretty neat!

parachain/pallets/bridge/src/lib.rs
```diff
@@ -65,28 +110,32 @@ decl_module! {
...
-		pub fn submit(origin, app_id: AppId, message: Message) -> DispatchResult {
+		pub fn submit(origin, channel_id: ChannelId, message: Message) -> DispatchResult {
 			let who = ensure_signed(origin)?;
 
-			// TODO: move replay protection here
+			// Construct channel object from storage
+			let channel = Self::get_inbound_channel(channel_id);
+
+			// Design choices:
+			//   Do verification and then submit to channel?
```
musnit, line 121: there needs to be some verification inside the channel, for example, the receiving channel needs to verify the smart contract address of the sending channel to make sure it comes from a trusted source. i think maybe all the verification could go in the channel even?

parachain/pallets/eth-app/src/lib.rs
```diff
@@ -85,7 +87,7 @@ decl_module! {
 		// Users should burn their holdings to release funds on the Ethereum side
 		// TODO: Calculate weights
 		#[weight = 0]
-		pub fn burn(origin, recipient: H160, amount: U256) -> DispatchResult {
+		pub fn burn(origin, channel_id: ChannelId, recipient: H160, amount: U256, ) -> DispatchResult {
```
vgeddes, line 90: actually users will Specify a more higher-level `DeliveryOptions` parameter instead of channel_id

musnit: lets just keep it simple and get it working and we can improve things after :)

parachain/pallets/bridge/src/lib.rs
musnit: you mean on the ethereum smart contract? or where? link?

vgeddes: Here we go: https://github.com/paritytech/parity-bridges-common/blob/master/modules/message-lane/src/lib.rs#L212

musnit: oh, yeh that's with a delivery confirmation flow too - our design doesn't rely on this or want/need this

vgeddes: In theory could be implemented like this: We just need the ReceiveChannel on the Ethereum side ask its SendChannel sibling to send some kind of "message delivered" control message across the bridge.

Of course this is a nice-to-have and can wait till after we've delivered.

musnit: yes but it's not useful/valuable at all in our current design, so not even a nice-to-have.

for alternative designs/setups, yes it could be added as such

parachain/pallets/eth-app/src/lib.rs
vgeddes: fine with me :+1:

parachain/pallets/bridge/src/lib.rs
vgeddes: Have removed `MessageAccepted`

PR #223, opened 2021-01-21
Channels

vgeddes:
Added an implementation for the channel concept we've introduced into the design.

Channels are represented as plain structs with attached methods (i.e. no pallets). They are able to persist data in the bridge pallet's storage. The design was loosely inspired by the [message lane](https://github.com/paritytech/parity-bridges-common/blob/master/modules/message-lane/src/inbound_lane.rs) concept in parity-bridges-common.

For simplicity of implementation:
- All code is in the bridge module instead of being spread out over multiple pallets
- Fixed number of channels:
  - 1 pair of outbound channels (basic and incentivized),
  - 1 pair of inbound channels (basic and incentivized).

Related changes:
* Introduced a type `ChannelId` for identification of whether a channel is basic or incentivized.
* Added an application registry which simplified message dispatch for inbound channels.
* Commitments pallet now has a per-channel message queue.


parachain/runtime/src/lib.rs
```diff
@@ -333,11 +335,19 @@ impl xcm_handler::Config for Runtime {
 
 // Our pallets
 
+parameter_types! {
+	pub Apps = {
```
vgeddes, line 339: This is where we initialize the application registry.

parachain/pallets/bridge/src/lib.rs
```diff
@@ -18,38 +18,47 @@
...
 
 	/// The verifier module responsible for verifying submitted messages.
 	type Verifier: Verifier<<Self as system::Config>::AccountId>;
 
-	/// ETH Application
-	type AppETH: Application;
-
-	/// ERC20 Application
-	type AppERC20: Application;
+	type Apps: Get<BTreeMap<H160, Box<dyn Application>>>;
```
vgeddes, line 46: channels can lookup apps with `T::Apps::get(ID)`

parachain/pallets/erc20-app/src/lib.rs
```diff
@@ -131,7 +132,16 @@ impl<T: Config> Application for Module<T> {
...
+#[derive(Copy, Clone)]
+struct Proxy<T: Config>(PhantomData<T>);
+
+impl<T: Config> Application for Proxy<T> {
+	fn handle(&self, payload: &[u8]) -> DispatchResult {
+		<Module<T>>::handle(payload)
 	}
 }
+
+pub fn make_proxy<T: Config>() -> Box<dyn Application> {
```
vgeddes, line 145: We need to create a proxy object as pallets can't be instantiated and passed around like normal objects.

parachain/runtime/src/lib.rs
```diff
@@ -333,11 +335,19 @@ impl xcm_handler::Config for Runtime {
...
+		let mut registry = make_registry();
+		registry.insert(H160::repeat_byte(1), eth_app::make_proxy<Runtime>())
```
vgeddes, line 341: Still need to settle on a suitable type for application IDs (Still using `H160` here), so channels can dispatch to them.

Apps still need to know the contract address of their peers on the Ethereum side though. That code hasn't been removed.

parachain/pallets/commitments/src/lib.rs
```diff
@@ -133,14 +137,13 @@ impl<T: Config> Module<T> {
 
 		ethabi::encode(&vec![Token::FixedArray(messages)])
 	}
-
 }
 
 impl<T: Config> MessageCommitment for Module<T> {
 
 	// Add a message for eventual inclusion in a commitment
 	// TODO: Number of messages per commitment should be bounded
```
musnit, line 145: can we move this todo to a github issue flagged with Security?

parachain/pallets/commitments/src/lib.rs
```diff
@@ -98,21 +92,31 @@ decl_module! {
...
+		let messages: Vec<Message> = <Self as Store>::MessageQueues::take(channel_id);
 		if messages.len() == 0 {
 			return 0
 		}
 
 		let commitment = Self::encode_commitment(&messages);
 		let commitment_hash = <T as Config>::Hashing::hash(&commitment);
 
-		let digest_item = AuxiliaryDigestItem::CommitmentHash(commitment_hash.clone()).into();
+		let digest_item = AuxiliaryDigestItem::Commitment(ChannelId::Basic, commitment_hash.clone()).into();
```
musnit, line 119: not sure what this is diong but looks like its using a hardcoded ChannelId::Basic rather than the actual channel_id?

parachain/pallets/bridge/src/channel/inbound.rs
```diff
@@ -0,0 +1,116 @@
...
+			storage: Storage::new(ChannelId::Basic)
+		}
+	}
+}
+
+impl<T: Config> InboundChannel<T::AccountId> for BasicInboundChannel<T> {
+	fn submit(&mut self, relayer: &T::AccountId, message: &Message) -> DispatchResult {
+		self.storage.try_mutate(|data| {
+			// Example: Increment nonce
+			data.nonce += 1;
```
musnit, line 42: need to check message.nonce?

parachain/pallets/bridge/src/channel/inbound.rs
```diff
@@ -0,0 +1,116 @@
...
+
```
musnit, line 43: need to check message.sender, still need to decide where best to do this:
 - either here in the channel (in which case, the App will need to register authorizedSenders, or have a method like getAuthorizedSenders, or somethin)
 - in the app (in which case we forward msg.sender into the app and let it check there)
 
I don't think it matters which for now, but longer term when we think about how we'll be supporting cross-parachain dispatch, this distinction may matter.

parachain/pallets/commitments/src/lib.rs
vgeddes: Oopsie!, will fix :)

parachain/pallets/commitments/src/lib.rs
vgeddes: Sure, added #226.

parachain/pallets/bridge/src/channel/inbound.rs
vgeddes: Added a comment to these `submit` methods indicating that they are still WIP. The goal of this PR is the overall design and API for channels in substrate. So the implementations of these methods will come eventually.

parachain/pallets/bridge/src/lib.rs
```diff
@@ -65,28 +75,20 @@ decl_module! {
...
-			let who = ensure_signed(origin)?;
-
-			// TODO: move replay protection here
+		pub fn submit(origin, channel_id: ChannelId, message: Message) -> DispatchResult {
+			let relayer = ensure_signed(origin)?;
 
-			T::Verifier::verify(who, app_id, &message)?;
-			Self::dispatch(app_id.into(), &message)
+			let mut channel = make_inbound_channel::<T>(channel_id);
+			channel.submit(&relayer, &message)
```
Rizziepit, line 83: I'm assuming dropping the call to `T::Verifier::verify` is temporary for the sake of dev. Just a heads up that I'm busy with e2e debugging of message proofs. So it would be good to keep verification around. Consider adding a config param like https://github.com/Snowfork/polkadot-ethereum/blob/aa43d98b0665245b8a7327d17e5b316d0c8928fc/parachain/pallets/verifier-lightclient/src/mock.rs#L114 to disable verification in the pallet in tests

parachain/pallets/bridge/src/primitives.rs
```diff
@@ -0,0 +1,27 @@
...
+#[derive(Encode, Decode, Copy, Clone, PartialEq, Eq, Default, RuntimeDebug)]
+pub struct OutboundChannelData {
+	pub nonce: u64
+}
+
+pub trait InboundChannel<AccountId>
+where
+	AccountId: Parameter + Member + Ord + MaybeSerializeDeserialize + Debug
+{
+	fn submit(&mut self, relayer: &AccountId, message: &Message) -> DispatchResult;
```
Rizziepit, line 22: I landed #203 which adds `submit_bulk` to the bridge and the verifier. I think it's fine to call `channel.submit` one by one after verification though... unless it's a real perf bottleneck. Am I correct it saying there will be 10 writes to storage for a batch of 10 messages? Regardless, premature optimization and all that jazz

Rizziepit (commented):
I like this architecture. I was going to implement traits directly for the bridge module, but this is better given that we have multiple channel types

Not approving since this is a draft but I don't have any objections (aside from verification call being missing)

parachain/pallets/bridge/src/lib.rs
vgeddes: Thanks for the heads up. Though for unit tests I'm just disabling the verifier using this technique

bridge/src/mock.rs
```rust
impl Config for Test {
	type Event = TestEvent;
	type Verifier = ();
	type AppETH = ();
	type AppERC20 = ();
	type MessageCommitment = ();
}
```

primitives/cores/sr/lib:
```rust
impl<AccountId> Verifier<AccountId> for () {
	fn verify(sender: AccountId, app_id: AppId, message: &Message) -> DispatchResult {
		Ok(())
	}
	fn verify_bulk(sender: AccountId, messages: &[(AppId, Message)]) -> DispatchResult {
		Ok(())
	}
}
```

parachain/pallets/bridge/src/primitives.rs
vgeddes: Am gonna leave `submit_bulk` alone for now. Can always move it into a channel in the next PR.

parachain/pallets/bridge/src/channel/inbound.rs
```diff
@@ -0,0 +1,117 @@
...
+			storage: Storage::new(ChannelId::Basic)
+		}
+	}
+}
+
+impl<T: Config> InboundChannel<T::AccountId> for BasicInboundChannel<T> {
+	// This implementation is a WIP!
+	fn submit(&mut self, relayer: &T::AccountId, app_id: AppId, message: &Message) -> DispatchResult {
+		T::Verifier::verify(relayer.clone(), app_id, &message)?;
+		Module::<T>::dispatch(app_id.into(), message)
```
musnit, line 43: Nonce check for replay protection should be here right?

musnit: Similar for incentivized channel

vgeddes: yeah, but wasn't planning to do it in this PR though. The focus for this PR is the channel API.

vgeddes: Also the nonce check can't be implemented until the other side includes a nonce.

musnit: The pr for the other side has this already but sure, merging and doing that in future PR sounds fine

parachain/pallets/bridge/src/channel/inbound.rs
musnit: Don't forget about this plz - add to issue/todo for the next PR if not planning to address in this PR

vgeddes: Added https://github.com/Snowfork/polkadot-ethereum/issues/231

parachain/primitives/core/src/lib.rs
```diff
@@ -31,9 +31,27 @@ pub trait Verifier<AccountId> {
...
+/// Outbound submission for applications
+pub trait SubmitOutbound {
+	fn submit(channel_id: ChannelId, payload: &[u8]) -> DispatchResult;
+}
+
+impl SubmitOutbound for () {
+	fn submit(channel_id: ChannelId, payload: &[u8]) -> DispatchResult {
+		Ok(())
+	}
+}
```
philipstanislaus, line 52: Could you add a comment here why we have a default implementation that does not err?

parachain/primitives/core/src/lib.rs
```diff
@@ -42,7 +60,21 @@ pub trait Application {
...
+
+/// Add a message to a commitment
 pub trait MessageCommitment {
+	fn add(channel_id: ChannelId, address: H160, nonce: u64, payload: &[u8]);
+}
 
-	fn add(address: H160, nonce: u64, payload: &[u8]);
+impl MessageCommitment for () {
+	fn add(channel_id: ChannelId, address: H160, nonce: u64, payload: &[u8]) { }
```
philipstanislaus, line 79: Again, please add a comment why this is needed

parachain/primitives/core/src/lib.rs
```diff
@@ -42,7 +60,21 @@ pub trait Application {
...
 
+impl Application for () {
+	fn handle(payload: &[u8]) -> DispatchResult {
+		Ok(())
+	}
+
+	fn address() -> H160 {
+		H160::zero()
+	}
+}
```
philipstanislaus, line 71: And likewise, please add a comment

philipstanislaus (commented):
Nice, left just 3 additional comments

PR #225, opened 2021-01-22
Relayer flows

musnit:
This is an initial draft version of the relayer flow for the Polkadot -> Ethereum direction.

I thought about it mostly in the context of imagining a single relayer, but ideally long term it needs to work in the context of multiple relayers run by different parties running at the same time and competing to submit ethereum transactions.

I'm not too concerned with optimising for profit, but I am concerned with ensuring the flows handle cases in which an alternative relayer gets ahead. Ideally any of these should be handled gracefully, possibly retrying and then moving on without crashing/getting stuck at a dead end.

I'll work on adding/fixing any broken flows in the context of this competitive environment next, but help would be appreciated and so when reviewing/auditing, please try analyze it in the context of this competitive environment. 

This is an important piece, and catching problems here early will make the relayer work a lot easier, so will each of the reviewers please make an effort to review properly.

@Rizziepit I'm thinking you'd likely be best to implement all this as you're likely to be free/done with your stream soon and also have the most familiarity having implemented similar stuff for the Ethereum -> Polkadot direction, but lets see.


docs/markdown/using-snowchain/validating-and-relaying/relayer-flow.md
```diff
@@ -0,0 +1,127 @@
...
+```
+### Ethereum Block Processing Queue
+```mermaid!
+graph TD;
+  Eb1[New Block In Queue];
+  Eb1-- contains updated latestRelayedEpoch -->Eb2
+  Eb2[Save as latestRelayedEpoch];
+  Eb1-- no update -->Eb3
+  Eb3[Ignore];
+```
```
philipstanislaus, line 39: This should not just track epochs, but also actual MMR roots, probably by block height.

An epoch is likely taking between 6 and 24 hours (has this been decided yet?). For the bridge to be fast, we should target to commit every single (or at least skip only a few) MMR roots.

docs/markdown/using-snowchain/validating-and-relaying/relayer-flow.md
```diff
@@ -0,0 +1,127 @@
...
+  M1[New MMR Root In Queue];
+  M1-->M1a;
+  M1a[Is epoch == latestRelayedEpoch + 1?];
+  M1a-- No -->M2;
+  M1a-- Yes -->M3;
+  M2[Ignore];
+  M3[Is last block of epoch?];
+  M3-- Yes -->M4;
+  M4[Add new MMR Root to MMR Root Relay queue];
+  M3-- No -->M2;
```
philipstanislaus, line 52: Again, let's use block height here, and process every MMR root for a higher block.

docs/markdown/using-snowchain/validating-and-relaying/relayer-flow.md
```diff
@@ -0,0 +1,127 @@
...
+```
+### MMR Relay Queue
+```mermaid!
+graph TD;
+  Mr00[New MMR to Relay];
```
philipstanislaus, line 57:
```suggestion
  Mr00[New MMR Root to Relay];
```

docs/markdown/using-snowchain/validating-and-relaying/relayer-flow.md
```diff
@@ -0,0 +1,127 @@
...
+  M1a[Is epoch == latestRelayedEpoch + 1?];
+  M1a-- No -->M2;
+  M1a-- Yes -->M3;
+  M2[Ignore];
+  M3[Is last block of epoch?];
+  M3-- Yes -->M4;
+  M4[Add new MMR Root to MMR Root Relay queue];
+  M3-- No -->M2;
+```
+### MMR Relay Queue
```
philipstanislaus, line 54:
```suggestion
### MMR Relay Queue

Note that this queue needs to support relaying multiple MMR Roots in parallel. Since there is a waiting period of `N` Ethereum blocks, and there `~2.8`  Polkadot blocks per Ethereum block, there might be `N * 2.8` MMR Roots being relayed at any time. (For a realistic `N=45`, that would be `126` parallel MMR Root submissions.
```

docs/markdown/using-snowchain/validating-and-relaying/relayer-flow.md
```diff
@@ -0,0 +1,127 @@
...
+nav_order: 2
+---
+
+# Relayer Flow
+
+## Polkadot to Ethereum
+The relayer can be started in 2 different modes for relaying from Polkadot to Ethereum.
+
+## MMR-Epochs-Only mode
+This mode runs the relayer in a mode where it ensures that the Ethereum contract receives at least 1 MMR proof per epoch. It only interacts with the Relay Chain Light Client, not the Parachain Light Client.
```
philipstanislaus, line 14:
```suggestion
## MMR-Only mode
This mode runs the relayer in a mode where it ensures that the Ethereum contract receives MMR roots, where at least one MMR proof per epoch is guaranteed. It only interacts with the Relay Chain Light Client, not the Parachain Light Client.
```

docs/markdown/using-snowchain/validating-and-relaying/relayer-flow.md
```diff
@@ -0,0 +1,127 @@
...
+  Mr1[Watch and wait for x Ethereum blocks until block wait period is over];
+  Mr1-->Mr10;
+  Mr10[Query blockhash for random seed];
+  Mr10-->Mr11;
+  Mr11[Generate random number, selected Merkle Proofs and transaction input data];
+  Mr11-->Mr2;
+  Mr2[Submit final verification transaction to relay chain light client];
+  Mr2--Transaction Successful -->Mr3;
+  Mr3[Done, Ethereum Block Processing Queue should incidentally receive this new block and update latestRelayedEpoch];
+```
```
philipstanislaus, line 74: I thought a bit more about the mechanism for preventing multiple competing relayers starting the process for the same block height.

It should be enough to just prevent completion for any relay chain block height that already has a completed MMR root. That just requires one additional storage entry (maybe `latestCompletedHeight`). 

With that approach, multiple completions will never occur, just multiple initiations, which are not that expensive.

docs/markdown/using-snowchain/validating-and-relaying/relayer-flow.md
```diff
@@ -0,0 +1,127 @@
...
+layout: default
+title: Relayer Flow
+parent: Validating and Relaying
+nav_order: 2
+---
+
+# Relayer Flow
+
+## Polkadot to Ethereum
+The relayer can be started in 2 different modes for relaying from Polkadot to Ethereum.
```
philipstanislaus, line 11:
```suggestion
The relayer can be started in 2 different modes for relaying from Polkadot to Ethereum. Both can be run in parallel.
```

docs/markdown/using-snowchain/validating-and-relaying/relayer-flow.md
```diff
@@ -0,0 +1,127 @@
...
+  M5[Get MMR Root -> MMRLeaf proof for relevant parachain block];
+  M6[Get MMRLeaf parachain_heads -> our parachain head merkle proof];
+  M7[Query Parachain for all messages in commitment];
+  M8[Generate Transaction input data];
+  M9[Submit Transaction to Incentivized Channel];
+  M5-->M6
+  M6-->M7
+  M7-->M8
+  M8-->M9
+  M9--Transaction Successful -->M10;
```
philipstanislaus, line 125:
```suggestion
  M9--Transaction Successful -->M10;
  M9--Transaction Unsuccessful, MMR Root not matched => Retry -->M5;
```

docs/markdown/using-snowchain/validating-and-relaying/relayer-flow.md
```diff
@@ -0,0 +1,127 @@
...
+  M10[Done];
+```
```
philipstanislaus, line 127: Here multiple relayers are less problematic, since we use nonces anyways. The only cost incurring a second relayer submitting the transaction is for calldata, then the contract should revert.

philipstanislaus (commented):
Nice, left a few thoughts below. Thanks!

docs/markdown/using-snowchain/validating-and-relaying/relayer-flow.md
Rizziepit, line 14: When you say 1 per epoch @musnit, do you mean Polkadot epoch or Ethereum epoch? The diagram below also shows that you're watching Eth blocks, not Polkadot blocks. I prob missing something. How do Ethereum blocks and MMR roots relate?

docs/markdown/using-snowchain/validating-and-relaying/relayer-flow.md
musnit: @philipstanislaus This is for MMR-Epochs-Only mode. This is intended to ensure only that the contracts receive at least 1 proof per epoch. Incentivized-Channel Mode should handle sending every single new commitment, so give us fast finality.

docs/markdown/using-snowchain/validating-and-relaying/relayer-flow.md
musnit: same comment as above - this mode is only to ensure the bridge doesn't skip any epochs

docs/markdown/using-snowchain/validating-and-relaying/relayer-flow.md
musnit: yep this works unless multiple relayers submit their completion transaction in the same ethereum block, which is pretty much guaranteed in a hypercompetitive environment?

docs/markdown/using-snowchain/validating-and-relaying/relayer-flow.md
musnit: hmm if this is unsuccessful something unexpected/weird likely happened, so retry is probably not the best flow. probably need something like:
 - check why unsuccessful, if because someone else did it already, then great, done. if not, error/alarm/alert for manual intervention

docs/markdown/using-snowchain/validating-and-relaying/relayer-flow.md
musnit: yeh not too major

docs/markdown/using-snowchain/validating-and-relaying/relayer-flow.md
musnit: @Rizziepit Polkadot epoch. Epochs mentioned anywhere in this context refer to polkadot epochs, as eth epochs are not relevant.

MMR R
And correct, we do watch eth blocks too, basically:
 - MMR Roots are basically like new Polkadot blocks
 - MMR Roots are relayed to Ethereum and ethereum stores the latest MMR Root that has been relayed after the 2-step interactive relay process completes.
 - The relayer needs to keep track of the most recent Polkadot Epoch that Ethereum is aware of. This is latestRelayedEpoch.

We watch Polkadot to detect new MMR Roots that have been **generated**, so that we can relay them to Ethereum.
We watch Ethereum to detect new MMR Roots that have been **relayed** (eg: by someone else, or by the relayer in a different mode) so that if they contain a new epoch, we can update our awareness of Ethereum to this new epoch.

Can discuss in more detail on next call if needed.

docs/markdown/using-snowchain/validating-and-relaying/relayer-flow.md
musnit: I've updated the New Incentivized Channel Commitment Processing Queue and the preview link to include this - please review.

docs/markdown/using-snowchain/validating-and-relaying/relayer-flow.md
musnit: The MMR-Epochs-Only is basically intended only to ensure that the light client contract never skips an epoch, nothing else. In a normal scenario, where there is actual message flow through the bridge and at least one message per epoch, the Incentivized-Relay mode should pickup that message and relay the respective required MMR Root for it, and so the process running in MMR-Epochs-Only node effectively does nothing/never relays. It's only important in cases where there are no messages at all for more than 1 epoch.

I've also updated the New Incentivized Channel Commitment Processing Queue and the preview link to include some changes that reflect how the Incentivized-Relay mode should pickup and relay required MMR Roots - please review.

Rizziepit: > We watch Ethereum to detect new MMR Roots that have been relayed (eg: by someone else, or by the relayer in a different mode) so that if they contain a new epoch, we can update our awareness of Ethereum to this new epoch.

In the Eth -> Polkadot direction, we don't continuously keep track of the latest Eth block that Polkadot is aware of. We only retrieve this info when the relayer starts, and then forward each new block as we become aware of it.

So my question is, how does continously keeping track of the latest Polkadot epoch that Ethereum is aware of change this logic? At best we can save some network bandwidth by skipping an MMR root send if we know it's already out of date... but it's prob simpler to send it regardless, as soon as we become aware of it, and hope for the best. Why won't this strategy work?

musnit: I think because in MMR-Epochs-Only mode, the expected default is that another relayer running in Incentivized-Relay mode will be moving the light client forward new blocks and so the MMR-Epochs-Only mode won't really ever need to forward blocks in many circumstances. It could do this anyway without keeping track, as you say, but consider the following:
 - Need to handle failure properly, detecting why and then skipping over onto the next one as in normal circumstances failure due to someone else relaying first is likely
 - The relay is a 2-transaction process, not 1-transaction, so again may require watching Ethereum to handle failure anyway, depending on when the failure occurs, though I haven't thought about this too much
 - It's not just about saving network bandwidth, its also about saving ethereum gas costs, which will be significant because the cost for the 2-transaction process is high even in the case of failure/reversion.

Having said that, your idea for simple/greedy strategy sounds fine to implement for MVP presuming it can handle failure and continue correctly, and then can optimize for saving gas by watching after.

docs/markdown/using-snowchain/validating-and-relaying/relayer-flow.md
philipstanislaus: With `N = 45`, this means that there will only be one MMR every 12 minutes 45 seconds. Is that good fast enough for now?

docs/markdown/using-snowchain/validating-and-relaying/relayer-flow.md
philipstanislaus: Not sure that is correct, the second transaction from the same block would still be blocked? The only exception would be a re-entrancy problem, but there is no call to external contracts.

docs/markdown/using-snowchain/validating-and-relaying/relayer-flow.md
philipstanislaus: Yes, good point 👍

philipstanislaus (approved):
🎉

docs/markdown/using-snowchain/validating-and-relaying/relayer-flow.md
musnit: yeh the second transaction in the block will fail and revert, but the relayer then still ends up having to pay gas costs for the calldata, which is likely to be quite expensive still, so ideally still to be avoided in future optimization

docs/markdown/using-snowchain/validating-and-relaying/relayer-flow.md
musnit: yeh that is a bit slow, though probably still acceptable. should review  https://snowbridge-docs.snowfork.com/concepts/polkadot-light-client-verifier/interactive-protocol again and confirm if we really need a high N

PR #228, opened 2021-01-23
Test inline assembly call

denalimarsh:


ethereum/contracts/Decoder.sol
```diff
@@ -84,4 +84,18 @@ library Decoder {
         }
         return temp;
     }
+
+    // Slice 32 bytes and cast to an address
+    function sliceUint(bytes memory _bytes, uint256 _start)
```
musnit, line 89: does this work properly? its no longer used, can remove it if it doesn't

denalimarsh: It works, but it's no longer used.

philipstanislaus (commented):
LGTM

PR #229, opened 2021-01-26
Relayer command to get Eth genesis block for parachain spec

Rizziepit:
The relayer currently forwards headers one by one, from the newest finalized known to the parachain to the latest Ethereum block. This can take days. Now that we're using `polkadot-launch`, the relayer also seems to overwhelm the parachain easily when syncing thousands of old blocks, i.e. transactions stop being processed when the transaction pool is full (??). For now, it's best to avoid long syncs by setting a new Eth header in the parachain's genesis spec.

PS. I haven't figured out the exact mechanism of this "overwhelment" yet. Are new transactions from the relayer ignored when the transaction pool is full or are old transactions dropped? why doesn't the client library return an error?

### DIFF
--- a/relayer/cmd/getblock.go
+++ b/relayer/cmd/getblock.go
@@ -1,0 +1,198 @@
+// Copyright 2020 Snowfork
+// SPDX-License-Identifier: LGPL-3.0-only
+
+package cmd
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"math/big"
+
+	"github.com/ethereum/go-ethereum/ethclient"
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+
+	gethTypes "github.com/ethereum/go-ethereum/core/types"
+	"github.com/snowfork/polkadot-ethereum/relayer/chain/ethereum"
+	"github.com/snowfork/polkadot-ethereum/relayer/core"
+)
+
+type Format string
+
+const (
+	RustFmt Format = "rust"
+	JSONFmt Format = "json"
+)
+
+func getBlockCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "getblock",
+		Short:   "Retrieve the latest finalized block",
+		Args:    cobra.ExactArgs(0),
+		Example: "artemis-relay getblock",
+		RunE:    GetBlockFn,
+	}
+	cmd.Flags().StringP(
+		"format",
+		"f",
+		"rust",
+		"The output format. Options are 'rust' and 'json'. They correspond to the Substrate genesis config formats.",
+	)
+	return cmd
+}
+
+func GetBlockFn(cmd *cobra.Command, _ []string) error {
+	config, err := core.LoadConfig()
+	if err != nil {
+		return err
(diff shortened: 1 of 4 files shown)


relayer/cmd/getblock.go
```diff
@@ -0,0 +1,198 @@
...
+
+const (
+	RustFmt Format = "rust"
+	JSONFmt Format = "json"
+)
+
+func getBlockCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "getblock",
+		Short:   "Retrieve the latest finalized block",
```
Rizziepit, line 31: Would it be useful to extend this command for Polkadot blocks as well @vgeddes?

vgeddes: yeah definitely, it would be useful :+1:

Though currently our ethereum contracts aren't storing the latest finalized block, so we can query for that info yet

vgeddes:
> PS. I haven't figured out the exact mechanism of this "overwhelment" yet. Are new transactions from the relayer ignored when the transaction pool is full or are old transactions dropped? why doesn't the client library return an error?

Yeah I don't the exact mechanism either. Though tangentially, it doesn't help that the relayer is sending immortal transactions which are never dropped even if they aren't executed within a certain amount of time.

Our substrate client, the "writer" thingy, doesn't watch the submitted transactions to see what eventually happens to them. We'd have to use the `SubmitAndWatchExtrinsic` method to do that: https://pkg.go.dev/github.com/centrifuge/go-substrate-rpc-client@v2.0.0-alpha.1+incompatible/rpc/author#Author.SubmitAndWatchExtrinsic. Though I'm not sure how the relayer would be able to action errors besides backing off and waiting.

I guess we could send also the headers in batches, either by modifying our `import_header` API, or using the `Utility.batch` extrinsic (from the Utilities pallet), which allows us to call `import_header` several times in one transaction.

vgeddes (approved):

musnit:
i think explicitly not supporting long-range sync in an automated/reliable way is fine and removes most of our concerns. 

short-range sync, with a new block being set on genesis and then keeping the relayer running on an ongoing basis without too much downtime should be fine.

Rizziepit:
> Yeah I don't the exact mechanism either. Though tangentially, it doesn't help that the relayer is sending immortal transactions which are never dropped even if they aren't executed within a certain amount of time.
> 
> Our substrate client, the "writer" thingy, doesn't watch the submitted transactions to see what eventually happens to them. We'd have to use the `SubmitAndWatchExtrinsic` method to do that: https://pkg.go.dev/github.com/centrifuge/go-substrate-rpc-client@v2.0.0-alpha.1+incompatible/rpc/author#Author.SubmitAndWatchExtrinsic. Though I'm not sure how the relayer would be able to action errors besides backing off and waiting.
> 
> I guess we could improve the situation by sending the headers in batches, either by modifying our `import_header` API, or using the `Utility.batch` extrinsic (from the Utilities pallet), which allows us to call `import_header` several times in one transaction.

This is useful info. I'll prob be putting up more relayer reliability PRs soon

PR #230, opened 2021-01-27
Update to latest rococo-v1 dependencies

vgeddes:


Rizziepit:
Heads up that something about this change is preventing transactions from being included in blocks. I see these errors that seem related:
```bash
2021-01-28 15:49:48  failed to assemble or send inherent data err=CanceledBackedCandidates(Canceled)
2021-01-28 15:49:48  could not get provisioner inherent data; injecting default data err=ClosedChannelAwaitingInherentData
```
I've also seen this, I think when I'm running without the `--execution=native` flag:
```bash
2021-01-28 16:01:19  failed to get collation from collator protocol subsystem err=OneshotRecv(Canceled)
```
I'll try and solve this later if you don't beat me to it @vgeddes

vgeddes:
Ah good catch. I see now that users on the Rococo channel on Element were reporting all sorts of bugs about this version. The maintainers have fixed some things and urged us to update.

Running `cargo update` should update our Polkadot, Substrate, and Cumulus dependencies.

vgeddes:
All sorted out in https://github.com/Snowfork/polkadot-ethereum/pull/234. I've tested to ensure that transactions are being executed.

PR #235, opened 2021-01-29
Always restart relayer

Rizziepit:
The relayer can exit due to an error with exit code 0 right now. So we can't rely on the exit code for restarts. But I'm not sure we want to rely on the exit code regardless - the relayer should always be up, right?


vgeddes (approved):
Thanks Riz! I actually forget to make this exact change because as you say its hard to make the relayer exit with a non-zero error code.

PR #237, opened 2021-02-01
Get channels working end-to-end

vgeddes:
Lots of changes everywhere to get channels operational end-to-end. Particularly in ensuring that all components adhere to the same wire formats.

Substrate:
* Bridge module verifies that a submitted message came from a known Ethereum outbound channel.
* Fleshed out implementation of inbound channels (i.e. nonce-checking, etc)
* Apps now encode the outbound payload as an ABI-encoded function call.
* Inbound payload for apps is now SCALE-encoded
* Verifier `verify` method now returns the Ethereum log included in the message.
* The `VerificationInput` is no longer an enum, and is renamed to `Proof`, as we aren't going to support multiple verification backends.
* Temporarily removed the bulk submission feature, as its interface complicated integration with the other bridge changes (soz @Rizziepit, we can definitely add this back later, I just ran out of time)
* Lots of added tests!

Relay:
* Used `abigen` (a tool from go-ethereum) to generate high-level clients for the Ethereum `OutboundChannel` and `InboundChannel` contracts
* Updated various type definitions and wire-format protocols as needed by Substrate and Ethereuem components
* Updated configuration for channels

Ethereum:
* Removed obsolete Bridge, Decoder, and Verifier contracts (@denalimarsh I just removed these as they weren't being used but if there was particular code which was still needed for something we can certainly add them back) 
* Added missing abstract contract for Outbound channel (or maybe it was the Inbound channel)
* Nonce variables are now part of the abstract channel contracts (as it is shared by descendant contracts)
* Updated nonce verification (as nonce's start at zero on the substrate side).
* SCALE-encode outgoing payloads.
* Added `ScaleCodec` contract (which includes some things from the now-obsolete `Decoder` contract)
* Lots of changes to ETHApp and ERC20App for uniformity with the Substrate side (naming, channels, etc)
* @denalimarsh I still need to update to fully update the unit tests. Though the `test_eth_app` tests work fully.
* Commented out the commitment verification code for now. Can add back later.
* Upgrade to solc 0.7.6 as it promises better support for `calldata` parameters (useful for optimized hashing of commitments)
* Upgrade `@openzeppelin/contracts` for compatibility with solc 0.7.6

Outcomes:
* Messages from Ethereum to Substrate and vice-versa work 100% according to my manual testing


ethereum/contracts/InboundChannel.sol
```diff
@@ -0,0 +1,18 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.7.6;
+pragma experimental ABIEncoderV2;
+
+abstract contract InboundChannel {
+
+    uint64 public nonce;
+
+    struct Message {
+        address target;
```
musnit, line 10:
```suggestion
        address destination;
```

ethereum/contracts/IncentivizedInboundChannel.sol
```diff
@@ -0,0 +1,59 @@
...
+        Message[] memory commitment
+    ) internal {
+        for (uint256 i = 0; i < commitment.length; i++) {
+            // Check message nonce is correct and increment nonce for replay protection
+            Message memory message = commitment[i];
+            require(message.nonce == nonce, "invalid nonce");
+
+            nonce = nonce + 1;
+
+            (bool success,) = message.target.call{value: 0, gas: MAX_GAS_PER_MESSAGE}(message.payload);
```
musnit, line 53:
```suggestion
            (bool success,) = message.destination.call{value: 0, gas: MAX_GAS_PER_MESSAGE}(message.payload);
```

ethereum/scripts/dumpRelayerConfig.js
```diff
@@ -3,38 +3,44 @@ const TOML = require('@iarna/toml');
...
             "descendants-until-final": 35,
-            bridge: {
-                address: bridge.address,
-                abi: bridgeAbiFile,
-            },
-            apps:{
-                eth:{
-                    address: ethApp.address,
-                    abi: ethAbiFile,
+            channels: {
```
musnit, line 34: need to dump apps too so that apps on substrate can verify the sending/source contract address

ethereum/scripts/dumpTestConfig.js
```diff
@@ -1,68 +1,60 @@
...
+            channels: {
+                basic: {
+                    inbound: channels.basic.inbound.address,
+                    outbound: channels.basic.outbound.address,
+                },
+                incentivized: {
+                    inbound: channels.incentivized.inbound.address,
+                    outbound: channels.incentivized.outbound.address,
                 },
-                erc20: {
```
musnit, line 37: need to dump apps too

parachain/node/src/chain_spec.rs
```diff
@@ -111,6 +111,16 @@ fn testnet_genesis(
 			// Configure endowed accounts with initial balance of 1 << 60.
 			balances: endowed_accounts.iter().cloned().map(|k|(k, 1 << 60)).collect(),
 		}),
+		bridge: Some(BridgeConfig {
+			source_channels: SourceChannelConfig {
+				basic: SourceChannel {
+					address: hex!["2ffa5ecdbe006d30397c7636d3e015eee251369f"].into(),
```
musnit, line 117: need to make these all dynamic

parachain/pallets/bridge/src/lib.rs
```diff
@@ -85,54 +104,26 @@ decl_module! {
...
 }
 
 impl<T: Config> Module<T> {
-	fn dispatch(address: H160, message: &Message) -> DispatchResult {
-		if address == T::AppETH::address() {
-			T::AppETH::handle(message.payload.as_ref())
-		} else if address == T::AppERC20::address() {
-			T::AppERC20::handle(message.payload.as_ref())
+	fn dispatch(source: H160, payload: &[u8]) -> DispatchResult {
+		if source == T::AppETH::address() {
```
musnit, line 123: i see the source->destination mapping from eth to substrate is now hardcoded/defined here, rather than being based on something like a pallet_id - is this correct? if so, sounds good for now and we can evolve it later.

ethereum/scripts/dumpTestConfig.js
vgeddes: Isn't necessary tho - `dumpTestConfig.js` actually only generates configuration for the relayer, and the relayer only needs to know the addresses for channels (by design its unaware of application contracts).

Since we use a ganache instance that generates deterministic addresses based upon a static seed, the E2E mocha tests and parachain genesis config have hardcoded addresses for the ETH and ERC20 apps. Is a very useful simplification, though the catch is if we change the order in which contracts are deployed then we would need to update the addresses everywhere.

parachain/pallets/commitments/src/lib.rs
```diff
@@ -119,7 +119,7 @@ impl<T: Config> Module<T> {
 		let digest_item = AuxiliaryDigestItem::Commitment(channel_id, commitment_hash.clone()).into();
 		<frame_system::Module<T>>::deposit_log(digest_item);
 
-		offchain_index::set(&Self::offchain_key(channel_id, commitment_hash), &commitment);
+		offchain_index::set(&Self::offchain_key(channel_id, commitment_hash), &messages.encode());
```
musnit, line 122: can we use messages.encode() a few lines above, instead of encode_commitment too? or what's the difference?

ethereum/scripts/dumpRelayerConfig.js
vgeddes: ah this script is kinda unmaintained. I should delete it. The E2E tests now use dumpTestConfig.js.

ethereum/contracts/BasicInboundChannel.sol
```diff
@@ -0,0 +1,59 @@
...
+        Message[] memory commitment
+    ) internal {
+        for (uint256 i = 0; i < commitment.length; i++) {
+            // Check message nonce is correct and increment nonce for replay protection
+            Message memory message = commitment[i];
+            require(message.nonce == nonce, "invalid nonce");
+
+            nonce = nonce + 1;
+
+            (bool success,) = message.target.call{value: 0, gas: MAX_GAS_PER_MESSAGE}(message.payload);
```
musnit, line 53:
```suggestion
            (bool success,) = message.destination.call{value: 0, gas: MAX_GAS_PER_MESSAGE}(message.payload);
```

relayer/chain/ethereum/writer.go
```diff
@@ -87,54 +89,46 @@ func (wr *Writer) writeLoop(ctx context.Context) error {
...
 
-	tx := types.NewTransaction(nonce, wr.bridgeContract.Address, value, gasLimit, gasPrice, txData)
-	signedTx, err := types.SignTx(tx, types.HomesteadSigner{}, wr.conn.kp.PrivateKey())
-	if err != nil {
-		return err
+	var messages []inbound.InboundChannelMessage
+	for _, m := range msg.Commitment {
+		messages = append(messages,
+			inbound.InboundChannelMessage{
+				Target:  m.Target,
```
musnit, line 117:
```suggestion
				Destination:  m.Destination,
```

relayer/substrate/commitment.go
```diff
@@ -0,0 +1,7 @@
+package substrate
+
+type CommitmentMessage struct {
+	Target  [20]byte
```
musnit, line 4:
```suggestion
	Destination  [20]byte
```

parachain/pallets/commitments/src/lib.rs
vgeddes: Well originally I was going to persist `commitment` which is ABI-encoded. However due to certain unresolved issues in the relayer, I need to pass the commitment in SCALE-encoded form (the output of `messages.encode()`), have the relayer decode that into normal structs, and then pass those structs directly to the auto-generated client for the `InboundChannel` contract. That client then ABI-encodes the structs before submitting the TX.

A bit redundant and inefficient, yes.

parachain/pallets/bridge/src/lib.rs
vgeddes: Yeah we definitely need to resolve that by encoding the `(PalletId, CallIndex)` in the solidity contract. Its gonna make this much cleaner.

I have a plan for doing it and should be straightforward, after seeing how they did it in parity-bridges-common.

With this change, the ETH pallet will have this dispatchable: 
```
fn mint(origin: Origin, recipient, amount) {
  sender_contract = ensure_signed!(origin);
  ensure!(sender_contract == AccountIdForETHApp)
  ...
}
```

where `AccountIdForETHApp` is a special AccountId that doesn't have a private key and is derived from the H160 address of the ETHApp contract.

parachain/pallets/bridge/src/lib.rs
```diff
@@ -85,54 +104,26 @@ decl_module! {
...
-				.filter_map(|r| r.err())
-				.collect();
-
-			debug::trace!(
-				target: "submit_messages",
-				"Messages were dispatched",
-			);
+			let log = T::Verifier::verify(&message)?;
+			let envelope = Envelope::try_from(log).map_err(|_| Error::<T>::InvalidEnvelope)?;
+			let channel_id = SourceChannels::get(envelope.channel)
```
musnit, line 112: channel_id is the contract address of the corresponding outbound channel right?

musnit (commented):
First pass review done. Overall, things look good, just a few minor comments/suggestions.

One big q:
 - How does the "target"/destination get set on the substrate side when sending out a message? I don't see the code for that anywhere? Is something missing?

Also note:
 - I have not reviewed any test code
 - I have not reviewed any light client verifier code (@Rizziepit plz do)
 - I have reviewed any relayer code properly

ethereum/scripts/dumpTestConfig.js
musnit: ah, yeh i thought the parachain genesis config used the dump.

on ropsten ofc the genesis will need to be changed to use whatever ropsten addresses we get. do we have a dump for this yet or do we just have to manually change the genesis config on ropsten?

parachain/pallets/commitments/src/lib.rs
musnit: ah yeh - do we have a github issue to fix this yet? if not plz create

parachain/pallets/bridge/src/lib.rs
musnit: cool leave it for a follow up PR though :P

parachain/node/src/chain_spec.rs
musnit: actually its probably fine to leave hardcoded indefinitely:
 - works fine as is for local/testing
 - for ropsten+production, we should only need to do genesis once and so making it dynamic doesn't save too much time as hardcoded modification should be fine. maybe we can create a chain_spec_local.rs and chain_spec_ropsten.rs or something for then?

parachain/pallets/verifier-lightclient/src/lib.rs
```diff
@@ -412,20 +408,20 @@ impl<T: Config> Module<T> {
...
+		let header = Headers::<T>::get(proof.block_hash)
 			.ok_or(Error::<T>::MissingHeader)?
 			.header;
-	
-		let receipt = header.check_receipt_proof(&proof)
+
+		let receipt = header.check_receipt_proof(&proof.merkle_proof.1)
 			.ok_or(Error::<T>::InvalidProof)?;
-		
+
```
Rizziepit, line 418: Do you have something in your vscode config that removes white space in blank lines?

parachain/pallets/verifier-lightclient/src/lib.rs
```diff
@@ -469,64 +465,18 @@ fn ancestry<T: Config>(mut hash: H256) -> impl Iterator<Item = (H256, EthereumHe
...
-					return Err(Error::<T>::InvalidProof.into());
-				}
-		
-				Ok(())
-			},
-			_ => Err(Error::<T>::UnsupportedVerificationScheme.into())
-		}
-	}
-
-	fn verify_bulk(_: T::AccountId, messages: &[(AppId, Message)]) -> DispatchResult {
```
Rizziepit, line 491: Will you add this back in in future? I do think it's worth it, along with `submit_bulk`, mostly for catching up the node after downtime. In another PR though

parachain/pallets/verifier-lightclient/src/lib.rs
```diff
@@ -469,64 +465,18 @@ fn ancestry<T: Config>(mut hash: H256) -> impl Iterator<Item = (H256, EthereumHe
...
 }
 
-impl<T: Config> Verifier<T::AccountId> for Module<T> {
+impl<T: Config> Verifier for Module<T> {
 
-	fn verify(_: T::AccountId, _: AppId, message: &Message) -> DispatchResult {
-		match message.verification {
-			VerificationInput::ReceiptProof { ref block_hash, ref tx_index, ref proof } => {
-				let receipt = Self::verify_receipt_inclusion(block_hash, &proof.1)?;
+	fn verify(message: &Message) -> Result<Log, DispatchError> {
```
Rizziepit, line 470: Nice to see this simplified :)

relayer/chain/ethereum/listener.go
```diff
@@ -105,52 +108,87 @@ func (li *Listener) pollEventsAndHeaders(
...
+		more := iter.Next()
+		if !more {
+			err = iter.Error()
+			if err != nil {
+				return nil, err
+			}
+			break
+		}
+
+		fmt.Printf("Event %+v", iter.Event)
```
Rizziepit, line 158: nit: remove stray print

relayer/chain/ethereum/message.go
```diff
@@ -6,156 +6,52 @@ package ethereum
...
 	}
 
-	receiptKey, err := rlp.EncodeToBytes(event.TxIndex)
+	receiptKey, err := rlp.EncodeToBytes(event.Raw.TxIndex)
 	if err != nil {
 		return nil, err
 	}
 
-	proof := NewProof()
+	proof := substrate.NewMerkleProof()
```
Rizziepit, line 32: I'd personally drop the "merkle" because we have the header proofs around as well - also technically merkle proofs - but formatted and validated using very different code. The "merkle" commonality doesn't translate to any shared code in substrate or the relayer

relayer/chain/substrate/writer.go
```diff
@@ -86,12 +86,25 @@ func (wr *Writer) writeLoop(ctx context.Context) error {
 		case <-ctx.Done():
 			return wr.onDone(ctx)
 		case msgs := <-wr.messages:
-			err := wr.WriteMessages(ctx, msgs)
+
+			fmt.Println("Got messages!")
```
Rizziepit, line 90: nit: stray print

relayer/chain/substrate/writer.go
```diff
@@ -86,12 +86,25 @@ func (wr *Writer) writeLoop(ctx context.Context) error {
...
+
+			var concreteMsgs []*chain.EthereumOutboundMessage
+			for _, msg := range msgs {
+
+				fmt.Println("MESSAGE ", msg)
```
Rizziepit, line 95: nit: stray print

relayer/chain/substrate/writer.go
```diff
@@ -149,38 +161,25 @@ func (wr *Writer) write(_ context.Context, c types.Call) error {
...
-	if err != nil {
-		return err
-	}
+// WriteMessages submits a "Bridge.submit_bulk" call
+func (wr *Writer) WriteMessages(ctx context.Context, msgs []*chain.EthereumOutboundMessage) error {
 
-	wr.log.WithFields(logrus.Fields{
-		"appid": hex.EncodeToString(msg.AppID[:]),
-	}).Info("Submitted message to Substrate")
+	fmt.Println("Write messages!")
```
Rizziepit, line 167: nit: stray print

relayer/chain/substrate/writer.go
```diff
@@ -149,38 +161,25 @@ func (wr *Writer) write(_ context.Context, c types.Call) error {
...
-	c, err := types.NewCall(&wr.conn.metadata, "Bridge.submit", msg.AppID, msg.Payload)
-	if err != nil {
-		return err
-	}
-
-	err = wr.write(ctx, c)
-	if err != nil {
-		return err
-	}
+// WriteMessages submits a "Bridge.submit_bulk" call
```
Rizziepit, line 164: Comment is outdated, but doesn't matter if you're planning on adding `submit_bulk` back

relayer/chain/ethereum/message.go
Rizziepit: Ditto on the substrate side

Rizziepit (commented):
I reviewed the relayer + substrate and left a few minor comments. All good from my side if our tests are passing

parachain/pallets/verifier-lightclient/src/lib.rs
vgeddes: I'll definitely add it back. I originally removed it as the interface some of the implementation would have needed to be updated for the changes in the bridge design, and well I just ran out of time to do that before the audit started.

I'm thinking that I'll change `verify_bulk` to return an iterator that produces values of the form `Result<Log, DispatchError>`. Then the bridge can sequentially iterate over the values and process them or act on the first error found.

parachain/pallets/verifier-lightclient/src/lib.rs
vgeddes: yup! these settings automagically clean up white space on save:

```json
{
  "files.trimFinalNewlines": true,
  "files.trimTrailingWhitespace": true,
  "files.insertFinalNewline": true
}
```

parachain/pallets/verifier-lightclient/src/lib.rs
vgeddes: yeah I think the auditors would have appreciated the simpler API.

relayer/chain/ethereum/message.go
vgeddes: ah good to know. Since its generic proof data, how about this renaming?:

```diff
pub struct Message {
	pub data: Vec<u8>,
	pub proof: Proof,
}

pub struct Proof {
	pub block_hash: H256,
	pub tx_index: u32,
-	pub merkle_proof: (Vec<Vec<u8>>, Vec<Vec<u8>>),
+	pub data: (Vec<Vec<u8>>, Vec<Vec<u8>>),
}

File: ethereum/contracts/BasicOutboundChannel.sol [hunk 21]
```diff
@@ -0,0 +1,24 @@
...
+
+    /**
+     * @dev Sends a message across the channel
+     */
+    function submit(bytes memory payload)
+        public
+        override
+    {
+        emit Message(msg.sender, nonce, payload);
+        nonce = nonce + 1;
```
philipstanislaus, line 22: Minor point, but for consistency: Seems that the nonce here tracks the last consumed one, while in the inbound channel the nonce tracks the next free one. I think next free one is the idiomatic way, right?

philipstanislaus (commented):
1. Is SCALE encoding on Ethereum necessary? In terms of gas optimization, it might more sense to just rely on abi encoding here, right?

vgeddes:
> 1. Is SCALE encoding on Ethereum necessary? In terms of gas optimization, it might more sense to just rely on abi encoding here, right?

@philipstanislaus yeah I'd say its necessary, as it has the potential to greatly simplify message dispatch to applications on the Substrate side. Right now the bridge module on Substrate has [hardwired links](https://github.com/Snowfork/polkadot-ethereum/blob/vincent/ethereum-receive-channel/parachain/pallets/bridge/src/lib.rs#L122) to the various application pallets, and that really hurts the bridge design in a number of ways.

The ultimate goal is to SCALE-encode a [Call](https://substrate.dev/rustdocs/v2.0.1/frame_support/dispatch/trait.Dispatchable.html) object which can be dispatched as-is on the Substrate side. This means Solidity contracts can issue messages which can result in calls to arbitrary dispatchables on any pallet (Obviously locked down though with verification of origins). @musnit and I discovered that `parity-bridges-common` devs were following the same approach for message dispatch (ref https://github.com/paritytech/parity-bridges-common/blob/master/modules/call-dispatch/src/lib.rs#L317)

Apparently, encoding a Call object is fairly straightforward, (after seeing how its done in GSRPC):
```
call = ScaleCodec.Encode([PalletIndex, CallableIndex, Arg1, Arg2, ..., ArgN])
```

Then regarding gas costs, turns out that its not much expensive than ABI-encoding, since in many cases the Solidity `abi.encodePacked` generates data that has the same wire-format as SCALE. Specifically, this is true for the `bytes32`,  and `address` types. So in our current contracts we only need special handling for `uint256` which needs to be encoded in little-endian form:

```solidity
function encodePayload(OutboundPayload memory payload) private pure returns (bytes memory) {
    return abi.encodePacked(
        payload.sender,
        payload.recipient,
        payload.amount.toBytes32LE()
    );
}
```

vgeddes:
@musnit @denalimarsh I've just updated the Solidity unit tests so we're getting coverage of all new channel stuff and their coupling with the apps:

```
  Contract: ETHApp
    deposits
      ✓ should lock funds (91ms)
      ✓ should send payload to the basic outbound channel (48ms)
      ✓ should send payload to the incentivized outbound channel (46ms)
    withdrawals
      ✓ should unlock via the basic inbound channel (139ms)
      ✓ should unlock via the incentivized inbound channel (137ms)

  Contract: ERC20App
    deposits
      ✓ should lock funds (166ms)
      ✓ should send payload to the basic outbound channel (99ms)
      ✓ should send payload to the incentivized outbound channel (102ms)
    withdrawals
      ✓ should unlock via the basic inbound channel (157ms)
      ✓ should unlock via the incentivized inbound channel (171ms)
```

relayer/chain/ethereum/message.go
vgeddes: Fixed. Renamed as described above.

parachain/pallets/verifier-lightclient/src/lib.rs
vgeddes: Added https://github.com/Snowfork/polkadot-ethereum/issues/240 for this.

parachain/pallets/bridge/src/lib.rs
vgeddes: Is a bit more complicated. the `SourceChannels` storage item is a mapping between a ChannelId and a corresponding address for a ethereum outbound channel.

```
pub enum ChannelId {
	Basic,
	Incentivized
}
```

parachain/pallets/bridge/src/lib.rs
vgeddes: Added #239 to track this.

parachain/node/src/chain_spec.rs
vgeddes: @musnit the hard-coded chainspec can easily be overriden for production without having to edit and recompile that rust file:

```bash
# Dump generated spec:
target/release/artemis build-spec > config.json

# Edit spec and add custom config:
vim config.json

# Launch chain with custom config
target/release/artemis --chain config.json
```

In future though, as you say, we will want to add hardcoded specs for rococo/ropsten though. In which case, launching a chain with such a spec would be as simple as:

```bash
target/release/artemis --chain rococo-ropsten
```

@Rizziepit Yes! Though since our chain is trust-minimized, `sudo` is a no-go. So in practice we'd need to use democratic measures to vote on such changes. We'd need to incorporate the [Democracy](https://substrate.dev/docs/en/knowledgebase/runtime/frame#democracy) and [Elections](https://substrate.dev/docs/en/knowledgebase/runtime/frame#elections) for that though.

ethereum/scripts/dumpTestConfig.js
vgeddes: I've explained the process here a bit more in https://github.com/Snowfork/polkadot-ethereum/pull/237#discussion_r568577491

vgeddes:
> First pass review done. Overall, things look good, just a few minor comments/suggestions.
> 
> One big q:
> 
> * How does the "target"/destination get set on the substrate side when sending out a message? I don't see the code for that anywhere? Is something missing?
> 

Good catch, I actually missed that and was sending a zero address `H160::zero()` as the destination. That was some stub code from the previous PR #223, and was causing message delivery to fail. 

I've fixed that now. Look at https://github.com/Snowfork/polkadot-ethereum/blob/vincent/ethereum-receive-channel/parachain/pallets/eth-app/src/lib.rs#L95 to see how the substrate apps specify the target contract.

> Also note:
> 
> * I have not reviewed any test code

Unit tests for the Substrate and Ethereum sides are now in very good shape. See below example output from Ethereum tests:

```
$ truffle test test/test_eth_app.js test/test_erc20_app.js
  Contract: ETHApp
    deposits
      ✓ should lock funds (91ms)
      ✓ should send payload to the basic outbound channel (48ms)
      ✓ should send payload to the incentivized outbound channel (46ms)
    withdrawals
      ✓ should unlock via the basic inbound channel (139ms)
      ✓ should unlock via the incentivized inbound channel (137ms)

  Contract: ERC20App
    deposits
      ✓ should lock funds (166ms)
      ✓ should send payload to the basic outbound channel (99ms)
      ✓ should send payload to the incentivized outbound channel (102ms)
    withdrawals
      ✓ should unlock via the basic inbound channel (157ms)
      ✓ should unlock via the incentivized inbound channel (171ms)
```

The E2E mocha tests haven't been updated though. They will probably fail. But I've manually tested message sending in both directions with success.

> * I have not reviewed any light client verifier code (@Rizziepit plz do)
> * I have reviewed any relayer code properly

musnit:
> > First pass review done. Overall, things look good, just a few minor comments/suggestions.
> > One big q:
> > 
> > * How does the "target"/destination get set on the substrate side when sending out a message? I don't see the code for that anywhere? Is something missing?
> 
> Good catch, I actually missed that and was sending a zero address `H160::zero()` as the destination. That was some stub code from the previous PR #223, and was causing message delivery to fail.
> 
> I've fixed that now. Look at https://github.com/Snowfork/polkadot-ethereum/blob/vincent/ethereum-receive-channel/parachain/pallets/eth-app/src/lib.rs#L95 to see how the substrate apps specify the target contract.

looks good!

musnit (approved):
looks good to me - once other reviewer's feedback is addressed, i think we can merge

ethereum/contracts/InboundChannel.sol
vgeddes: I'll probably leave this low-hanging fruit for another PR. If we go ahead with this, for consistency we'll need to rename the corresponding `source` to `origin`, and then a search/replace gets a bit more involved.

ethereum/contracts/BasicOutboundChannel.sol
vgeddes: Good point! If that's the idiom then I'm onboard. Though I've just got everything working end-to-end, so am rather keen to address this in a future PR. Added https://github.com/Snowfork/polkadot-ethereum/issues/241 to track the issue.

parachain/pallets/commitments/src/lib.rs
vgeddes: Sure, added https://github.com/Snowfork/polkadot-ethereum/issues/242

PR #253, opened 2021-02-03
Handle edge case for Eth transaction proof validation in Substrate

Rizziepit:
My original implementation of trie node decoding made the incorrect assumption that, in a proof, all intermediate nodes are "full nodes", i.e. they contain a list of child hashes. However, it's possible for an intermediate node to be a "short node", i.e. it has a single hash value.

Note that this is not a complete implementation of the Ethereum trie node encoding, but it is sufficient for processing a receipt proof (at least, I didn't encounter a failure across 200 000 events / 5000 blocks). We could try to re-use a complete Eth trie implementation instead, like https://github.com/provable-things/ptokens-core/blob/master/src/chains/eth/trie.rs, but it's very heavy and doesn't clearly facilitate the proof validation use case. More investigation needed


musnit:
ah nice catch, thanks!

vgeddes (approved):
LGMT!

PR #255, opened 2021-02-04
Relayer `getmessages` command

Rizziepit:
I added the `getmessages` command to make it easy for myself to get samples for testing, e.g. https://github.com/Snowfork/polkadot-ethereum/pull/253/ uses test data I retrieved using this command (before I updated it to only load channel events). It will probably be useful again in future.


relayer/cmd/getblock.go
```diff
@@ -72,11 +85,16 @@ func getEthBlock(config *ethereum.Config) (*gethTypes.Header, error) {
...
 	}).Info("Connected to chain")
 
-	latestHeader, err := client.HeaderByNumber(ctx, nil)
-	header, err := client.HeaderByNumber(
-		ctx,
-		new(big.Int).Sub(latestHeader.Number, big.NewInt(int64(config.DescendantsUntilFinal))),
-	)
+	var header *gethTypes.Header
+	if blockHash == nil {
+		header, err = client.HeaderByNumber(ctx, nil)
```
musnit, line 90:
```suggestion
```

musnit (approved):
just the 1 comment

relayer/README.md
```diff
@@ -50,7 +50,7 @@ Example Configuration:
 ```toml
 [ethereum]
 endpoint = "ws://localhost:8545/"
-descendants-until-final = 0
+descendants-until-final = 35
```
vgeddes, line 53: Just an FYI, a value of `35` caused the relayer to break during E2E tests (segfaults, etc).

relayer/cmd/getmessages.go
```diff
@@ -0,0 +1,219 @@
...
+	"github.com/snowfork/polkadot-ethereum/relayer/chain/ethereum"
+	"github.com/snowfork/polkadot-ethereum/relayer/contracts/outbound"
+	"github.com/snowfork/polkadot-ethereum/relayer/core"
+	"github.com/snowfork/polkadot-ethereum/relayer/crypto/secp256k1"
+	"github.com/snowfork/polkadot-ethereum/relayer/substrate"
+)
+
+func getMessagesCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "getmessages",
```
vgeddes, line 33: Nit: can we rename the command to `fetch-messages`? I prefer hyphens in between the words of command names.

vgeddes (approved):

relayer/README.md
vgeddes: Added https://github.com/Snowfork/polkadot-ethereum/issues/256 to track

Rizziepit: I'm guessing I introduced an underflow error at https://github.com/Snowfork/polkadot-ethereum/blob/bd53ee173bd04b55ccdf179464e3e0d56b1d3e82/relayer/chain/ethereum/listener.go#L114 when I switched to bulk message sends. I'll fix, although it shouldn't be a problem if the e2e config keeps using 0 (which is what it should be using?)

I'm using 0 in e2e config so that blocks are immediately finalized. So messages are immediately forwarded. It makes e2e test runs *much* faster. But I'm weary of putting 0 in a readme config in case it makes its way into production - 0 is only ok when forks aren't possible.

relayer/cmd/getblock.go
Rizziepit: I'm using header.Number right below to calculate the finalized block number. Admittedly I prob shouldn't be re-using the variable

musnit: oh right my bad - but ye maybe use different names :)

PR #257, opened 2021-02-05
Eth difficulty check fix

Rizziepit:
As per title. I didn't notice this error in the original implementation in https://github.com/near/rainbow-bridge :(

PS. I mined the test block using `ethash.Seal` in geth fwiw


musnit (approved):
nice!

Issue #241, opened 2021-02-02
Channels should track nonce's more idiomatically


vgeddes:
Minor point, but for consistency: Seems that the nonce here tracks the last consumed one, while in the inbound channel the nonce tracks the next free one. I think next free one is the idiomatic way, right?

_Originally posted by @philipstanislaus in https://github.com/Snowfork/polkadot-ethereum/pull/237#discussion_r568411993_

PR #258, opened 2021-02-05
Improved message dispatch to pallets

### LINKED ISSUE #239: Solidity applications should encode Substrate Call objects
> 1. Is SCALE encoding on Ethereum necessary? In terms of gas optimization, it might more sense to just rely on abi encoding here, right?

@philipstanislaus yeah I'd say its necessary, as it has the potential to greatly simplify message dispatch to applications on the Substrate side. Right now the bridge module on Substrate has [hardwired links](https://github.com/Snowfork/polkadot-ethereum/blob/vincent/ethereum-receive-channel/parachain/pallets/bridge/src/lib.rs#L122) to the various application pallets, and that really hurts the bridge design in a number of ways.

The ultimate goal is to SCALE-encode a [Call](https://substrate.dev/rustdocs/v2.0.1/frame_support/dispatch/trait.Dispatchable.html) object which can be dispatched as-is on the Substrate side. This means Solidity contracts can issue messages which can result in calls to arbitrary dispatchables on any pallet (Obviously locked down though with verification of origins). @musnit and I discovered that `parity-bridges-common` devs were following the same approach for message dispatch (ref https://github.com/paritytech/parity-bridges-common/blob/master/modules/call-dispatch/src/lib.rs#L317)

Apparently, encoding a Call object is fairly straightforward, (after seeing how its done in GSRPC):
```javascript
call = ScaleCodec.Encode([PalletIndex, CallableIndex, Arg1, Arg2, ..., ArgN])
```

Then regarding gas costs, turns out that its not much expensive than ABI-encoding, since in many cases the Solidity `abi.encodePacked` generates data that has the same wire-format as SCALE. Specifically, this is true for the `bytes32`,  and `address` types. So in our current contracts we only need special handling for `uint256` which needs to be encoded in little-endian form:

```solidity
function encodePayload(OutboundPayload memory payload) private pure returns (bytes memory) {
    return abi.encodePacked(
        payload.sender,
        payload.recipient,
        payload.amount.toBytes32LE()
    );
}
```

Also `abi.encodePacked` isn't free either. It's a language primitive that gets compiled down to optimized assembly, which has  a gas cost regardless.

_Originally posted by @vgeddes in https://github.com/Snowfork/polkadot-ethereum/pull/237#issuecomment-771488489_

vgeddes:
Solidity apps can now call any dispatchable function on any substrate pallet, allowing for greater flexibility in messaging.

Changes:
* Added `artemis-dispatch` pallet, inspired by the [call-dispatch](https://github.com/paritytech/parity-bridges-common/blob/master/modules/call-dispatch/src/lib.rs) module in in parity-bridges-common
* Added a custom Origin for calls dispatched on behalf of a remote contract. Used by apps for authorization.
* Removed obsolete `Application` trait.
* Updated Solidity ETH and ERC20 contracts to encode Call objects
* Updated Substrate ETH and ERC20 pallets now have `mint` dispatchables.

Other changes which somehow crept in:
* Fixed nonce handling (#241)

Closes: #239 #241

The E2E tests pass successfully with these changes


ethereum/contracts/ETHApp.sol
```diff
@@ -80,17 +74,11 @@ contract ETHApp {
...
         pure
         returns (bytes memory)
     {
-        return
-            abi.encodePacked(
-                payload.sender,
-                payload.recipient,
-                payload.amount.toBytes32LE()
-            );
+        return abi.encodePacked(CALL_INDEX, _sender, _recipient, _amount.toBytes32LE());
```
musnit, line 82: do you have a guess for how much this adds to the cost of the transaction in USD?

vgeddes: You mean the new SCALE-encoding vs older ABI-encoding?

Well I ran a benchmark and the SCALE-encoding in this PR adds 6.64% to the gas used (76240 vs 81307) by the `lock()` transaction. At current gas prices (average confirmation times) that adds an extra $1.77 to the cost of the transaction, coming up to a total of $26.64879 according to https://ethgasstation.info/calculatorTxV.php.

vgeddes: Though gas prices are unusually high today, roughly 50% more than the average over the past 4 months.

musnit: you mean 6% just for adding the  CALL_INDEX parameter? or 6% in total compared to abi encoding?

vgeddes: the latter, 6% in total compared to abi encoding.

vgeddes: I've further optimized the SCALE-encoding of `uint256`, so its now about 21.5% faster than the previous algorithm. So total  overhead should now only be about 5% instead of 6%.

vgeddes:
@philipstanislaus 

I've improved nonces. Here's how it works now in pseudocode:

Outbound channel:
```
submit(message):
  nonce = nonce + 1
  send_message(message, nonce)
```

Inbound channel:
```
receive(message):
   if message.nonce != nonce + 1:
      raise Error
   nonce = nonce + 1
```

So consumed nonce's now start at 1 instead of 0, and the outbound channel tracks the last consumed nonce, and the inbound channel tracks the nonce for the last received message. Hope this addresses your concerns?

musnit (approved):
looks good overall - lets merge to unblock other work and can address comments from other reviewers in follow up PR

Issue #256, opened 2021-02-04
Config descendants-until-final=35 breaks the relayer during E2E tests


vgeddes:
Just an FYI, a value of `35` caused the relayer to break during E2E tests (segfaults, etc).

_Originally posted by @vgeddes in https://github.com/Snowfork/polkadot-ethereum/pull/255#discussion_r570394247_

vgeddes:
Basically the relayer was trying to process block -35 😜

Rizziepit:
Just so it's documented here: we shouldn't use 35 in e2e tests because it will make the tests very slow by making us wait for 35 blocks before forwarding a message from Eth -> Sub. A number like 0 or 1 is much preferred, and it's fine in an environment where there aren't forks (i.e. ganache). e2e tests run against Ropsten should use > 0. 35 will add 9 - 18 minutes to a test on Ropsten

Regardless, this should be fixed :)

PR #259, opened 2021-02-05
Minor fixes for relayer

Rizziepit:
This should solve https://github.com/Snowfork/polkadot-ethereum/issues/256

@vgeddes you mentioned that you saw segfaults relating to #256. I didn't see those, but I saw error logs relating to underflow:
```bash
time="2021-02-05T14:08:34-08:00" level=error msg="Failure fetching event logs" chain=Ethereum error="Number can only safely store up to 53 bits"
```
You might have seen a segfault right at the end if the relayer crashed, but usually the root cause isn't the segfault - the segfault is a symptom of deallocation that happens in the wrong order during shutdown. The segfault is tricky to fix but harmless. If you experience an issue, also have a look at the logs prior to shutdown

### DIFF
--- a/relayer/chain/ethereum/listener.go
+++ b/relayer/chain/ethereum/listener.go
@@ -111,6 +111,11 @@ func (li *Listener) pollEventsAndHeaders(
 				li.log.Info("Not polling events since channel is nil")
 			}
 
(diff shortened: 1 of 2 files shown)


vgeddes (approved):
LGTM!

PR #263, opened 2021-02-09
Get XCMP working for rococo-v1

vgeddes:


parachain/pallets/transfer/src/lib.rs
```diff
@@ -0,0 +1,185 @@
+#![cfg_attr(not(feature = "std"), no_std)]
```
philipstanislaus, line 1: What's the reasoning behind adding another transfer pallet?

vgeddes: ah this pallet was originally called `TokenDealer` and has been in the codebase for quite a while. I just renamed it to `Transfer`, which I think better reflects its purpose - allowing users to initiate the transfer of assets to other chains.

The original name was quite a bad choice and was copied over from some very early XCMP demo code provided by the substrate devs.

PR #266, opened 2021-02-09
Update master with direct bytes encoding and latest inbound channel changes

denalimarsh:
Incorporates the following PRs into the main branch:
- https://github.com/Snowfork/polkadot-ethereum/pull/238
- https://github.com/Snowfork/polkadot-ethereum/pull/250
- https://github.com/Snowfork/polkadot-ethereum/pull/265


ethereum/contracts/ERC20App.sol
```diff
@@ -94,17 +96,19 @@ contract ERC20App {
...
+        bytes32 _recipient,
+        uint256 _amount
+    ) private pure returns (bytes memory) {
         return
             abi.encodePacked(
                 MINT_CALL,
                 _token,
                 _sender,
-                byte(0x00), // Encode recipient as MultiAddress::Id
+                bytes1(0x00), // Encode recipient as MultiAddress::Id
```
musnit, line 111: does this do anything different?

musnit: looks like it would make it a byte array of length 1, rather than a single byte?

denalimarsh: I copied in the changes from https://github.com/Snowfork/polkadot-ethereum/commit/9c55d78da5727e87c322f49a42661e368087adb9 which has it as `bytes(0x00`), so not sure how it ended up as `bytes1`. Reverted to `bytes` in https://github.com/Snowfork/polkadot-ethereum/pull/266/commits/3c8835e6cc412c8c31fdcc7b62ca716c5a14d72d.

vgeddes: Can we leave it as `byte`? I made it `byte` since according to the solidity docs, `byte` is an alias for `bytes1` (There isn't a single byte type in Solidity, as a fixed-length byte array of length 1 is equal to a single byte).

test/src/subclient/index.js
```diff
@@ -23,11 +23,6 @@ class SubClient {
             "Incentivized": null
           }
         },
-        "MessageNonce": "u64",
```
vgeddes, line 26: Looks like you need to pull in from `main` :)

ethereum/contracts/ERC20App.sol
```diff
@@ -69,7 +70,8 @@ contract ERC20App {
 
         bytes memory call = encodeCall(_token, msg.sender, _recipient, _amount);
 
-        OutboundChannel channel = OutboundChannel(channels[_channelId].outbound);
+        OutboundChannel channel =
```
vgeddes, line 73: Small q - Is it recommended for Solidity code to not go over 80 columns? As I see both you and @musnit seem to be following that.

ethereum/test/test_eth_app.js
```diff
@@ -147,23 +150,24 @@ contract("ETHApp", function (accounts) {
       const recipient = "0xcCb3C82493AC988CEBE552779E7195A3a9DC651f";
 
       // expected amount to unlock
-      const amount = BigNumber(web3.utils.toWei("1", "ether"));
```
vgeddes, line 150: I think it makes the code easier to maintain and extend if we stick to a single number type, such as `BigNumber`.

I distrust the ad-hoc and unspecified conversions between `String` , `BN`, `BigNumber` and `Number` that occur behind the scenes. It means I never really know the exact type of some variable that holds a number.

ethereum/test/test_eth_app.js
```diff
@@ -147,23 +150,24 @@ contract("ETHApp", function (accounts) {
...
+      const amount = web3.utils.toWei("1", "ether");
 
       const beforeBalance = BigNumber(await this.app.balance());
       const beforeRecipientBalance = BigNumber(await web3.eth.getBalance(recipient));
 
       // Commitment payload generated using:
       //   cd parachain/pallets/eth-app
       //   cargo test test_outbound_payload_encode -- --nocapture
-      const commitment = [
+      const messages = [
```
vgeddes, line 161: Strictly speaking, `messages` is the commitment, so I would wouldnt rename this. The `commitment` variable you added is better referred to as the `commitmentHash`

ethereum/contracts/ETHApp.sol
```diff
@@ -74,16 +75,17 @@ contract ETHApp {
...
+        address _sender,
+        bytes32 _recipient,
+        uint256 _amount
+    ) private pure returns (bytes memory) {
         return
             abi.encodePacked(
                 MINT_CALL,
                 _sender,
-                byte(0x00), // Encode recipient as MultiAddress::Id
+                bytes(0x00), // Encode recipient as MultiAddress::Id
```
vgeddes, line 88: Can we leave it as `byte`? I made it byte since according to the solidity docs, `byte` is an alias for `bytes1` (There isn't a single byte type in Solidity, as a fixed-length byte array of length 1 is equal to a single byte).

philipstanislaus: Yes, `byte` is more efficient than `bytes` (bytes is a dynamic array, which encoded needs a length prefix). See https://docs.soliditylang.org/en/v0.5.3/types.html#fixed-size-byte-arrays

ethereum/contracts/IncentivizedInboundChannel.sol
```diff
@@ -54,19 +58,25 @@ contract IncentivizedInboundChannel is InboundChannel {
...
+        // TODO: Test/review further.
+        uint256 maxMessageProcessingCost =
+            _messages.length * MAX_GAS_PER_MESSAGE;
+        require(
+            gasleft() >=
+                maxMessageProcessingCost +
+                    POST_PROCESSING_LEFTOVER_GAS_REQUIRED,
+            "insufficient gas for delivery of all messages"
+        );
+
```
philipstanislaus, line 79: Do we even need those gas checks? When we execute the calls, we limit the gas, which might be enough?

ethereum/contracts/IncentivizedInboundChannel.sol
```diff
@@ -93,16 +103,6 @@ contract IncentivizedInboundChannel is InboundChannel {
...
-            bytes memory messageBytes =
-                abi.encodePacked(
-                    _messages[i].target,
-                    _messages[i].nonce,
-                    _messages[i].payload
-                );
-            messagesBytes = abi.encodePacked(messagesBytes, messageBytes);
-        }
-        return keccak256(messagesBytes) == _commitment;
+        return keccak256(abi.encode(_messages)) == _commitment;
```
philipstanislaus, line 106: Nice!

philipstanislaus (approved):
Great, left some minor comments. 

Nice test by the way 💪

philipstanislaus:
Also, could you close the other PRs mentioned above? Thanks!

ethereum/contracts/BasicInboundChannel.sol
```diff
@@ -17,10 +17,12 @@ contract BasicInboundChannel is InboundChannel {
     }
 
     // TODO: Submit should take in all inputs required for verification,
-    // including eg: _commitment, _parachainBlockNumber, _parachainMerkleProof, parachainHeadsMMRProof
-    function submit(Message[] memory _messages) public override {
-        //TODO: Verify messages
-        //verifyMessages(_messages, _commitment, ...);
+    // including eg: _parachainBlockNumber, _parachainMerkleProof, parachainHeadsMMRProof
+    function submit(Message[] memory _messages, bytes32 _commitment)
```
musnit, line 21: maybe we can use calldata instead of memory for these fields to avoid them being copied over for each internal function call? (i'm not sure if we can easily access the calldata variables from internal calls - need to look into it)

^ let's leave this for a follow up PR/future optimization though if you don't have a quick and easy answer+fix - https://github.com/Snowfork/polkadot-ethereum/issues/252

ethereum/contracts/ERC20App.sol
musnit: i think @denalimarsh just copied my code and my setup is doing this

ethereum/contracts/IncentivizedInboundChannel.sol
musnit: yeh i think you're right

ethereum/test/test_eth_app.js
musnit: i think its cleaner to leave it as messages - i know the data is all being generated by the commitment pallet, but the commitment generated by that pallet is just the hash, not the messages?

parachain/pallets/erc20-app/src/payload.rs
```diff
@@ -5,6 +5,15 @@ use codec::Encode;
 use ethabi::{self, Token};
 use artemis_ethereum::{H160, U256};
 
+// Message from Ethereum (SCALE-encoded)
```
musnit, line 8: should this be in this PR?

ethereum/contracts/IncentivizedInboundChannel.sol
denalimarsh: Should we remove the gas checks this PR or make an issue for a follow-up PR?

ethereum/contracts/ERC20App.sol
denalimarsh: Looks like my Solidity linter is automatically changing `byte` to `bytes1` on save, which is probably why this was originally changed to `bytes1`

denalimarsh: Changed it back to `byte` in https://github.com/Snowfork/polkadot-ethereum/pull/266/commits/7ef0b553f98b69069fb5b4f00f21f9fe24b147c7 :)

test/src/subclient/index.js
denalimarsh: Ah this must have gotten cut out during the merge. Added it back in https://github.com/Snowfork/polkadot-ethereum/pull/266/commits/58c393183e6cbe7fc00e483d4bd5a007e27598b0

ethereum/contracts/ETHApp.sol
denalimarsh: Changed it back to `byte` in https://github.com/Snowfork/polkadot-ethereum/pull/266/commits/7ef0b553f98b69069fb5b4f00f21f9fe24b147c7

ethereum/test/test_eth_app.js
denalimarsh: I agree. I added `BigNumber` back in here in https://github.com/Snowfork/polkadot-ethereum/pull/266/commits/47fbb4dd81b9dd8a432ad9e043ec5e6634d897a5.

parachain/pallets/erc20-app/src/payload.rs
denalimarsh: No; generally speaking, this PR shouldn't touch anything outside the `ethereum` folder. This must have been changed in master at some point and this legacy code block got orphaned in this branch. I removed it in https://github.com/Snowfork/polkadot-ethereum/pull/266/commits/a23900ee7e7063222e39bb4ae4ef65ef579cee6b so that the pallet matches what's in master.

ethereum/contracts/IncentivizedInboundChannel.sol
denalimarsh: I created issue https://github.com/Snowfork/polkadot-ethereum/issues/268 for investigating these potentially redundant gas checks. I can address it in this PR if it's blocking.

ethereum/contracts/BasicInboundChannel.sol
denalimarsh: Looks like copying dynamically encoded structs from calldata to memory is supported as of [v0.7.6](https://github.com/ethereum/solidity/releases/tag/v0.7.6). I transitioned the inbound channels from calldata to memory in https://github.com/Snowfork/polkadot-ethereum/pull/266/commits/20a8833a91a4bfddd6037033ebf2944bdc9191f1.

denalimarsh:
> Also, could you close the other PRs mentioned above? Thanks!

Closed all the legacy PRs.

denalimarsh:
Pushed revisions and added some issues for ongoing tasks. Should be ready for another quick round of reviews before merge @vgeddes @musnit

ethereum/contracts/BasicInboundChannel.sol
musnit: Yeh copying works but what I meant was to check if it is possible and cheaper to avoid copying by accessing the calldata directly within our internal functions instead of sending them as memory parameters.

ethereum/test/test_eth_app.js
vgeddes: The pallet generates two things, a commitment (list of messages), and a hash of the former. That's the terminology used across the system (parachain, relayer, solidity) since I landed the Big Channels PR. The relayer passes a `commitment` and `commitmentHash` to the contract. With that in mind, I think its important to be consistent about naming across our system, as it leaves less room for confusion.

ethereum/contracts/BasicInboundChannel.sol
vgeddes: @musnit I think Denali did implement that (the `message` parameter is now passed via calldata in all internal functions)

vgeddes:
@denalimarsh We still need to update relayer to pass the `commitmentHash` to the inbound channel. Let me make that change and push it to this branch. Can then proceed with a merge (after addressing any remaining/new comments)

ethereum/contracts/BasicInboundChannel.sol
vgeddes: ah wait, I think I see what you mean - the `_message` parameter already lives in `msg.data` in ABI-encoded form. So we should just extract the bytes from there:

So something like this in pseudocode:
```
commitmentBytes= msg.data[START...END]
commitmentHash = keccak256(commitmentBytes) 
```

From https://ethereum.stackexchange.com/questions/14037/what-is-msg-data

vgeddes: actually even better, can extract the data using the `CALLDATACOPY` instruction (described in https://blog.openzeppelin.com/ethereum-in-depth-part-2-6339cf6bddb9)

denalimarsh: @vgeddes he's referring to copying each message into memory inside the `processMessages` function and then operating upon it instead of just reading directly from the calldata. Just pushed the related updates in https://github.com/Snowfork/polkadot-ethereum/pull/266/commits/0ce021b7cb499d5c43e67f3130feb8a9334dabc3. The strategy does save us a bit of gas:
- Before: `Submitted 4 messages. Gas used: 154248`
- After: `Submitted 4 messages. Gas used: 153816`

denalimarsh: Looks like it saves exactly 108 gas per message 👍

denalimarsh:
> @denalimarsh We still need to update relayer to pass the `commitmentHash` to the inbound channel. Let me make that change and push it to this branch. Can then proceed with a merge (after addressing any remaining/new comments)

Sounds good. It looks the contract binding in `relayer/contracts/inbound/contract.go` will need to be updated as part of that process. Its `submit` method is outdated, needs to be passing both messages and the commitment.

ethereum/contracts/BasicInboundChannel.sol
vgeddes: ahh OK! yeah that makes sense.

ethereum/test/test_eth_app.js
denalimarsh: I think `messages`/`commitment` or `messages`/`commitmentHash` makes more sense but I agree we should stick to the existing terminology.

musnit: @vgeddes hmm, yeh the existing terminology is basically wrong though then. the message list is a list of messages, it's not a commitment :S

musnit: only the hash is actually a commitment

musnit: so idk what's more confusing, inconsistent but correct naming, or consistent but incorrect naming :D? i'd say the former is better at least for solidity developers to understand the solidity code. i'd go with the former, ie messages/commitment

ethereum/contracts/BasicInboundChannel.sol
musnit: looks good, thanks!

ethereum/test/test_eth_app.js
vgeddes: Well, ok, it seems there is lack of consensus on a what the term "commitment" should refer to - the actual value being committed, or some proxy or identifier (the hash) for that thing. IMO both are equally correct in the end so no one's really wrong here 🙂.

Y'all can leave it as messages/commitment, if you prefer though. 👍

musnit: haha well commitment refers to something that someone has commited to, ie, their is an identifiable individual that has made the commitment. for the hash, we have that, as the hash is actually signed, onchain, by each polkadot validator in the beefy process. for the messages, we don't - nobody really commits to them/signs them directly and they're offchain.

vgeddes: I mean it really depends on what we want "commitment" to mean, so I think we are just disputing semantics here.

In my mind, the parachain pledges that a set of messages are fit and proper to be relayed to Ethereum. Those messages are being committed to, ergo they are the commitment. The commitment happens to be uniquely identified by a hash of the messages being committed. In the same way that you can commit code with git and refer to that commit/commitment using a "commit hash" (which is a valid term in git).

I do think `commitment/commitmentHash` is preferable over `message/commitment` as it is more descriptive. It indicates that there is a relationship between the two things.

denalimarsh: For this PR let's just leave it as it is. I created issue https://github.com/Snowfork/polkadot-ethereum/issues/273 for any further discussions about messages/commitment vs. commitment/commitment hash naming conventions.

PR #267, opened 2021-02-10
Relayer tracks extrinsic status

Rizziepit:
This PR improves the situation where extrinsics are occasionally dropped in Substrate. It's not a complete solution, but it does make the relayer behave in a way that's easier to manage. The two key improvements:
1. The relayer implicitly rate limits by restricting the "in flight" extrinsics to `MaxWatchedExtrinsics`. This should prevent dropped transactions (not invalid ones though).
2. It retries all extrinsics, not just headers, after a grace period. If a retry fails, it crashes. This is a bit harsh but it's unlikely that a retry will succeed without giving the Substrate node substantial time to resolve whatever is going wrong, e.g. [a transaction ban only resolves after 30 minutes](https://stackoverflow.com/questions/57270007/in-substrate-what-does-code-1012-transaction-is-temporarily-banned-mean).

I recommend we try to restart the relayer in 10 minute intervals. The problem usually resolves itself after the 1st or 2nd restart. I know this is fine for the Eth -> Sub direction - it replays headers + events. But is this fine in the opposite direction?

### DIFF
--- a/relayer/chain/substrate/outgoing_extrinsics.go
+++ b/relayer/chain/substrate/outgoing_extrinsics.go
@@ -1,0 +1,101 @@
+// Copyright 2020 Snowfork
+// SPDX-License-Identifier: LGPL-3.0-only
+
+package substrate
+
+import (
+	"context"
+	"fmt"
+	"sync"
+	"time"
+
+	"github.com/sirupsen/logrus"
+	"github.com/snowfork/go-substrate-rpc-client/v2/types"
+	"golang.org/x/sync/errgroup"
+)
+
+const MaxWatchedExtrinsics = 100
+
+type extrinsicPool struct {
+	sync.Mutex
+	conn    *Connection
+	eg      *errgroup.Group
+	log     *logrus.Entry
+	watched uint
+}
+
+func newExtrinsicPool(eg *errgroup.Group, conn *Connection, log *logrus.Entry) *extrinsicPool {
+	ep := extrinsicPool{
+		conn:    conn,
+		eg:      eg,
+		log:     log,
+		watched: 0,
+	}
+	return &ep
+}
+
+func (ep *extrinsicPool) WaitForSubmitAndWatch(ctx context.Context, ext *types.Extrinsic) {
+	defer ep.Unlock()
+
+	for {
+		ep.Lock()
+		if ep.hasCapacity() {
+			ep.eg.Go(func() error {
+				return ep.submitAndWatchLoop(ctx, ext)
+			})
+
+			ep.watched++
+			return
+		}
+		ep.Unlock()
+
+		time.Sleep(100 * time.Millisecond)
+	}
+}
+
+func (ep *extrinsicPool) hasCapacity() bool {
+	return ep.watched < MaxWatchedExtrinsics
+}
+
+func (ep *extrinsicPool) submitAndWatchLoop(ctx context.Context, ext *types.Extrinsic) error {
+	sub, err := ep.conn.api.RPC.Author.SubmitAndWatchExtrinsic(*ext)
+	if err != nil {
+		return err
+	}
+
+	for {
+		select {
+		case <-ctx.Done():
+			return fmt.Errorf("Context was canceled. Stopping extrinsic monitoring")
+		case status := <-sub.Chan():
+			ep.logStatusUpdate(status)
+
+			// Indicates that the extrinsic wasn't processed. We need to retry
+			if status.IsDropped || status.IsInvalid {
+				sub.Unsubscribe()
+				newSub, err := ep.conn.api.RPC.Author.SubmitAndWatchExtrinsic(*ext)
+				if err != nil {
+					return err
+				}
+				sub = newSub
+			} else if !status.IsReady && !status.IsFuture && !status.IsBroadcast {
+				// We assume all other status codes indicate that the extrinsic was processed
+				// and account nonce was incremented.
+				// See details at:
+				// https://github.com/paritytech/substrate/blob/29aca981db5e8bf8b5538e6c7920ded917013ef3/primitives/transaction-pool/src/pool.rs#L56-L127
+				sub.Unsubscribe()
+				ep.Lock()
+				defer ep.Unlock()
+				ep.watched--
+				return nil
+			}
+		case err := <-sub.Err():
+			return err
+		}
+	}
+}
+
+func (ep *extrinsicPool) logStatusUpdate(status types.ExtrinsicStatus) {
+	statusBytes, _ := status.MarshalJSON()
+	ep.log.WithField("status", string(statusBytes)).Debug("Received extrinsic status update")
+}
--- a/relayer/chain/substrate/outgoing_extrinsics.go
+++ b/relayer/chain/substrate/outgoing_extrinsics.go
@@ -14,7 +14,7 @@ import (
 	"golang.org/x/sync/errgroup"
 )
 
-const MaxWatchedExtrinsics = 500
+const MaxWatchedExtrinsics = 200
 
 type extrinsicPool struct {
 	sync.Mutex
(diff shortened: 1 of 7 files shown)


relayer/chain/substrate/writer.go
```diff
@@ -116,7 +119,7 @@ func (wr *Writer) writeLoop(ctx context.Context) error {
 }
 
 // Write submits a transaction to the chain
-func (wr *Writer) write(_ context.Context, c types.Call) error {
+func (wr *Writer) write(ctx context.Context, c types.Call) error {
 	ext := types.NewExtrinsic(c)
 
 	era := types.ExtrinsicEra{IsImmortalEra: true}
```
vgeddes, line 125: Would you design still be compatible if we change the extrinsic [mortality](https://wiki.polkadot.network/docs/en/build-protocol-info#transaction-mortality) send mortal extrinsics instead of immortal ones?

Currently, we're sending immortal extrinsics which from my understanding will never be dropped from the txpool unless it runs out of resources or suffers some other abnormal error.

We do need to change over to mortal transactions at some point, and ideally sometime before we deliver the project. I heard immortal extrinsics can have security complications.

philipstanislaus: Immortal extrinsics are an issue indeed: Since nonces are tracked per account, and accounts can be reaped (nonce will be reset), any immortal transaction can be replayed.

philipstanislaus: Hence, anyone can mess up our bridge by replaying immortal extrinsics after a reap happens (even if accidentally).

relayer/core/relay.go
```diff
@@ -174,7 +174,7 @@ func (re *Relay) Start() {
 		// Goroutines are either shutting down or deadlocked.
 		// Give them a second...
```
philipstanislaus, line 175:
```suggestion
		// Give them a few seconds...
```

relayer/chain/substrate/outgoing_extrinsics.go
```diff
@@ -0,0 +1,149 @@
...
+				return ep.submitAndWatchLoop(ctx, nonce, ext)
+			})
+
+			ep.watched++
+			return
+		}
+		ep.Unlock()
+
+		time.Sleep(100 * time.Millisecond)
+	}
```
philipstanislaus, line 54: Instead of a loop with a sleep, the more idiomatic way would be to use a buffered channel of size `MaxWatchedExtrinsics`. No need to change that, just noticed it ;)

relayer/chain/substrate/outgoing_extrinsics.go
```diff
@@ -0,0 +1,149 @@
...
+	ep.Lock()
+	defer ep.Unlock()
+	if nonce <= ep.maxNonce {
+		// No delay because we don't need to retry
+		return 0
+	}
+
+	// Stagger retries in the case that we have a series of failed extrinsics
+	noncesToCatchUp := nonce - ep.maxNonce
+	return time.Duration(noncesToCatchUp) * time.Second * 5
```
philipstanislaus, line 140: Is a 5 second delay not a bit long? What extrinsic load to we expect in production? Imagine a scenario where the relayer crashes and is offline for 8 hours. Would this allow the relayer to ever catch up?

relayer/chain/substrate/outgoing_extrinsics.go
```diff
@@ -0,0 +1,149 @@
...
+					"nonce":  nonce,
+					"status": statusStr,
+				}).Debug("Re-submitting failed extrinsic")
+				newSub, err := ep.conn.api.RPC.Author.SubmitAndWatchExtrinsic(*ext)
+				if err != nil {
+					return err
+				}
+				sub = newSub
+
+			} else if !status.IsReady && !status.IsFuture && !status.IsBroadcast {
```
philipstanislaus, line 109: Would it make sense to check for `Finalized` or `FinalityTimeout` instead?

philipstanislaus (approved):
Nice, left just a few tiny comments 👍

relayer/chain/substrate/writer.go
vgeddes: ah, good to know the exact reason why its harmful. I've updated our ticket https://github.com/Snowfork/polkadot-ethereum/issues/76 for after looking into how we can fix the problem.

musnit:
> I recommend we try to restart the relayer in 10 minute intervals. The problem usually resolves itself after the 1st or 2nd restart. I know this is fine for the Eth -> Sub direction - it replays headers + events. But is this fine in the opposite direction?

I think its a good idea for the opposite direction to be able to handle repeated restarts, so let's make sure it's fine :)

relayer/chain/substrate/outgoing_extrinsics.go
Rizziepit: | Would this allow the relayer to ever catch up

This retry logic shouldn't affect the relayer's ability to catch up... Honestly, I am not sure how this would play out in production though. In dev, this retry logic is almost never used because `Dropped` extrinsics are non-existent, and `Invalid` extrinsics are rare and I still don't know their root cause. This retry logic comes from the intuition that, if the problem is caused by a full txpool, it will take at least one block production cycle to consume transactions in the pool and resolve the issue. Block time in Polkadot is 6 seconds I believe? The staggered nature of re-sends though... that's risky. The idea is to go easy for a while if the txpool is full, but other relayers won't play by the same rules. I'm going to update to remove the staggering, and refine by strategy for `Dropped` vs `Invalid` a bit

relayer/chain/substrate/writer.go
Rizziepit: > Would your design still be compatible if we change the extrinsic mortality send mortal extrinsics instead of immortal ones?

Yeah, I believe so. Regardless, now's a good time to make the switch and test. I'll update this PR

relayer/chain/substrate/outgoing_extrinsics.go
Rizziepit: Maybe? My aim was for `extrinsicPool` to guarantee that Substrate processes all nonces without gaps. So `InBlock` is good enough. But it might make sense for the pool to not only guarantee that the nonce is provided, but additionally guarantee that the extrinsic is included in a block. Which do you think makes more sense?

philipstanislaus: Sorry, have missed your ping here. I believe the relayer should fail load and noisy if an extrinsic fails, but for the pool it should be fine to just guarantee nonce processing 👍

philipstanislaus (commented):
@Rizziepit I just had another look at these changes, and introducing mortal extrinsics might cause an issue with the current submit and watch logic: 

I just quickly scanned the substrate code, but it seems that expired transactions would have `TransactionStatus::Invalid`. If that's indeed the logic there, then our relayer would be stuck – it would wait forever.

Do you have a bit capacity verifying whether expired mortal extrinsics indeed have `TransactionStatus::Invalid`?

Rizziepit:
> @Rizziepit I just had another look at these changes, and introducing mortal extrinsics might cause an issue with the current submit and watch logic:
> 
> I just quickly scanned the substrate code, but it seems that expired transactions would have `TransactionStatus::Invalid`. If that's indeed the logic there, then our relayer would be stuck – it would wait forever.
> 
> Do you have a bit capacity verifying whether expired mortal extrinsics indeed have `TransactionStatus::Invalid`?
> Many thanks!

@philipstanislaus you're correct that expired transactions would have `TransactionStatus::Invalid`. What happens in Substrate is that invalid transactions are temporarily banned. So when we re-submit the transaction, we get an immediate "temporarily banned" error response, and the relayer crashes. So we're not actually stuck.

Rizziepit:
I should mention that this crashing behaviour is intentional on my part. I considered re-signing expired extrinsics with a fresh era and re-submitting. But I ended up erring on the side of "fail loudly as soon as things go wrong". Lots of room for improvement here though. Let me know if you have ideas

philipstanislaus:
Awesome, thanks for looking into this again. 🙌 

I'm fully on board with failing loudly instead of retrying – we can always refine that once we are running with actual traffic.

PR #271, opened 2021-02-12
Relayer Rewards + Wrapped DOT

vgeddes:
Relayer accounts now get paid instantaneously per message submitted. These rewards are sourced from the `DotApp` pallet, which controls DOT locked up by users.

When users lock up DOT, SnowDOT gets minted on the Ethereum side.

Currently relayers are rewarded 0 DOT, since we still need a few more pieces that are out of scope for this PR.

SnowDOT is implemented as an [ERC777](https://eips.ethereum.org/EIPS/eip-777) token, which is fully-backwards compatible with ERC20. ERC777 has numerous benefits, including being able to transfer funds to and notify a contract within a single transaction. Since ERC777 is backwards compatible, it is supported by Uniswap V2.

The only potential caveat with ERC777 is an increased risk of reentrancy attacks. For example, in 2019, a ERC777 pool/pair on Uniswap V1 was exploited using a reentrancy attack. See details [here](https://blog.openzeppelin.com/exploiting-uniswap-from-reentrancy-to-actual-profit/). However strictly speaking this is not actually an ERC777 bug, and more about Uniswap V1 not guarding against reentracy exploits. Uniswap V2 is [not vulnerable](https://uniswap.org/blog/uniswap-v2/#technical-improvements) to this exploit.

While DOT has 10 decimal places, WrappedDOT has 18 decimal places to match Ethereum and most ERC20 tokens. This simplifies integration with other dapps, frontends, and scripts. There is also increasing [consensus](https://github.com/ethereum/EIPs/issues/724) within the Ethereum community that customizable decimal places in ERC20 contracts was in hindsight a bad design decision.

Supporting 18 decimal places means SnowDOT holders may eventually accumulate tiny amounts of unredeemable "dust" in their accounts. I don't believe this should be of much concern though.

Another example of an ERC777 token in production is the wrapped [pBTC](https://etherscan.io/token/0x5228a22e72ccc52d415ecfd199f99d0665e7733b) token (created by [ptokens](https://ptokens.io/)), which also supports 18 decimals (unlike WBTC), and is actively traded on Uniswap V2.

The DOTApp contract owns the SnowDOT token (WrappedToken.sol). All burning and minting of SnowDOT has to be done via   DOTApp which is connected to the bridge.

DOTApp does not emit any Burned or Minted events as they would be redundant and cost extra gas. The SnowDOT contract emits Burned and Minted events which are part of the ERC777 interface.


parachain/runtime/src/lib.rs
```diff
@@ -398,6 +405,18 @@ impl assets::Config for Runtime {
 	type Event = Event;
 }
 
+parameter_types! {
+	pub const DotModuleId: ModuleId = ModuleId(*b"s/dotapp");
```
vgeddes, line 409: The `AccountId` that controls the locked DOT is derived from this module id.

parachain/primitives/core/src/rewards.rs
```diff
@@ -0,0 +1,31 @@
...
+use sp_std::marker::PhantomData;
+
+
+pub trait RewardRelayer<AccountId, Balance> {
+	fn pay_relayer(source: &AccountId, relayer: &AccountId, reward: Balance);
+}
+
+pub struct InstantRewards<T, C>(PhantomData<(T, C)>);
+
+impl<T, C> RewardRelayer<T::AccountId, C::Balance> for InstantRewards<T, C>
```
vgeddes, line 13: This is a thing that actually pays relayers. It pays them instantaneously per message.

parachain/primitives/core/src/rewards.rs
```diff
@@ -0,0 +1,31 @@
...
+where
+	T: Config,
+	C: Currency<T::AccountId>
+{
+	fn pay_relayer(source: &T::AccountId, relayer: &T::AccountId, reward: C::Balance) {
+		if reward.is_zero() {
+			return;
+		}
+
+		let _ = C::transfer(
```
philipstanislaus, line 23: Note that transfer will be logged in Polkadot explorers, so it will allow users to identify the relayer easily, and also allow relayers to see where the rewards came from. Fees in substrate are normally not transferred, but rather withdrawn and then deposited, which means that fees do not show up as transfers in explorers.

I have a slight opinion that rewards/fees should not show up as transfers, because such a transfer was not the user intention.

parachain/runtime/src/lib.rs
```diff
@@ -359,11 +359,18 @@ impl dispatch::Config for Runtime {
 	type CallFilter = CallFilter;
 }
 
+parameter_types! {
+	pub RewardsAccount: AccountId = DotModuleId::get().into_account();
```
philipstanislaus, line 363: Do we really want to use the same account for rewards as for locking of DOT?

parachain/pallets/dot-app/src/lib.rs
```diff
@@ -0,0 +1,116 @@
...
+
+			T::Currency::transfer(&who, &Self::account_id(), amount, AllowDeath)?;
+
+			let message = OutboundPayload {
+				sender: who.clone(),
+				recipient: recipient.clone(),
+				amount: amount.saturated_into::<u128>(),
+			};
+
+			T::SubmitOutbound::submit(channel_id, Address::get(), &message.encode().unwrap())?;
```
philipstanislaus, line 91: Instead of panicing here, can we return the error?

philipstanislaus: Also, please encode and return errors first before doing the transfer and the submit. Substrate does not roll back the application state, so an error in message encoding would lead to a transfer without a submit.

philipstanislaus (approved):
Nice, left a few minor comments. 

🎉

parachain/primitives/core/src/rewards.rs
vgeddes: ah good catch! Will add a fix for it.

parachain/pallets/dot-app/src/lib.rs
vgeddes: good catch! Not sure how I overlooked that. Will add a fix.

ethereum/contracts/DOTApp.sol
```diff
@@ -0,0 +1,91 @@
...
+    function encodeCall(address _sender, bytes32 _recipient, uint128 _amount)
+        private
+        pure
+        returns (bytes memory)
+    {
+        return
+            abi.encodePacked(
+                UNLOCK_CALL,
+                _sender,
+                byte(0x00), // Encode recipient as MultiAddress::Id
```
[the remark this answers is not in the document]
vgeddes: Nope, that comment is purely informational. Basically telling readers that we're encoding the recipient address using [MultiAddress::ID](https://crates.parity.io/sp_runtime/enum.MultiAddress.html) (which is now the preferred way of referring to on-chain account ids).

ethereum/contracts/WrappedToken.sol
```diff
@@ -0,0 +1,36 @@
...
+import "@openzeppelin/contracts/access/Ownable.sol";
+
+contract WrappedToken is ERC777, Ownable {
+
+    constructor(
+        string memory _name,
+        string memory _symbol,
+        address[] memory _defaultOperators
+    )
+        ERC777(_name, _symbol, _defaultOperators)
```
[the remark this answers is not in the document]
vgeddes: It isn't necessary as the constructor for `Ownable` takes zero parameters, so its just called implicitly I think.

https://docs.openzeppelin.com/contracts/3.x/api/access#Ownable-constructor--

parachain/runtime/src/lib.rs
vgeddes: I thought about this too. I agree its a bit weird. My thinking was that relayers are paid instantaneously on receipt of each incoming message. So the fee will be deducted from the user's account and find its way into the relayers account within a single transaction. That means having an intermediary rewards account to hold collected fees would be kinda redundant.

But I think I will adding an intermediary rewards account is still a good thing especially as we may want to allocate fees differently (ie. relayers only get 80% of the fee charged to a user).

Will make the change.

ethereum/contracts/DOTApp.sol
```diff
@@ -0,0 +1,91 @@
...
+    uint256 constant internal GRANULARITY = 10 ** (18 - POLKADOT_DECIMALS);
+
+    struct Channel {
+        address inbound;
+        address outbound;
+    }
+
+    constructor(Channel memory _basic, Channel memory _incentivized) {
+        address[] memory defaultOperators;
+        token = new WrappedToken("Wrapped DOT", "WDOT", defaultOperators);
```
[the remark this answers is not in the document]
vgeddes: I don't think we ever came to a consensus on `SnowDOT` vs `WDOT` as the discussion got side-tracked on a tangent about what to call SnowETH (sETH vs SnowETH).

I'm fine with either really, I think both work. The `WBTC` folks may get grumpy if we invade their namespace 😄 

@musnit what's your call on this?

parachain/pallets/dot-app/src/lib.rs
vgeddes: Fixed now. I've wrapped both the `unlock` and `lock` calls with the [transactional](https://substrate.dev/rustdocs/v3.0.0/frame_support/attr.transactional.html) macro.

I've also added a test in [6d1513](https://github.com/Snowfork/polkadot-ethereum/commit/6d15131e28bf266e32a13701c7523554e6b85dec) to verify that if `T::SubmitOutbound::submit()` fails, then any previous storage changes (such as the asset transfer) would be rolled back.

parachain/pallets/dot-app/src/lib.rs
```diff
@@ -0,0 +1,118 @@
...
+		type Error = Error<T>;
+
+		fn deposit_event() = default;
+
+		#[weight = 0]
+		#[transactional]
+		pub fn lock(origin, channel_id: ChannelId, recipient: H160, amount: BalanceOf<T>) -> DispatchResult {
+			let who = ensure_signed(origin)?;
+
+			T::Currency::transfer(&who, &Self::account_id(), amount, AllowDeath)?;
```
musnit, line 85: do we want to AllowDeath ? I think probably it's fine, but just checking?

parachain/pallets/bridge/src/channel/inbound.rs
```diff
@@ -42,6 +43,8 @@ impl<T: Config> InboundChannel<T::AccountId> for BasicInboundChannel<T> {
 			Ok(())
 		})?;
 
+		T::RewardRelayer::pay_relayer(&T::RewardsAccount::get(), relayer, 10.into());
```
musnit, line 46: this shouldn't be here for basic inbound?

musnit: can we put the amount here into a config somewhere?

ideally something that can eventually be modified by governance too, though can defer that if its not quick and easy

ethereum/migrations/2_next.js
```diff
@@ -63,5 +64,30 @@ module.exports = function(deployer, network, accounts) {
...
+    // Deploy ERC1820 Registry for our E2E stack.
+    if (network == 'e2e_test')  {
+
+      require('@openzeppelin/test-helpers/configure')({ web3 });
+      const { singletons } = require('@openzeppelin/test-helpers');
+
+      await singletons.ERC1820Registry(accounts[0]);
+    }
+
+    // only deploy this contract to non-development networks
```
musnit, line 77: why?

ethereum/migrations/2_next.js
```diff
@@ -63,5 +64,30 @@ module.exports = function(deployer, network, accounts) {
     );
 
     await deployer.deploy(TestToken, 100000000, "Test Token", "TEST");
+
+    // Deploy ERC1820 Registry for our E2E stack.
```
musnit, line 68: whats this for? seems unused?

musnit (commented):
DOT stuff looks good on both sides, besides for above comments.

Incentive/Fee stuff looks still changing/to review

Can we also add e2e integration tests for this in a follow up PR?

parachain/pallets/dot-app/src/lib.rs
philipstanislaus: AllowDeath means that the account will be deleted, including the nonce state. That might allow a replay of transactions that are not yet expired (especially immortal transactions are a problem).

I feel that allowing death here could be problematic, but would need to do a bit more research.

musnit: Well, even if we don't allow death for our transfer, the account could get itself deleted through some other mechanism right, and so we're still vulnerable to hypothetical attack?

Though I think replay, even if possible, would only fuck over the reborn account rather than the DOTApp/Bridge in any way? My substrate knowledge in this area is limited though so will leave it to u and vincent to resolve

parachain/runtime/src/lib.rs
vgeddes: Actually I'm going to leave this one for the next related piece of work that needs to happen - Governance. We'll be reworking rewards/fee allocation in that task anyway.

parachain/pallets/bridge/src/channel/inbound.rs
vgeddes: Yeah this will be all sorted out in the next PRs related to Governance & the incentivized bridge. Making fees configurable is a bit out-of-scope for this current PR, and I'm actually just gonna set the reward to `0` for now.

ethereum/migrations/2_next.js
vgeddes: The ERC1820 registry contract is a requirement for ERC777 tokens. Its fundamental to how ERC777 can implement token transfers & notification within a single transaction.

It seems unused, but actually the ERC777 contract references the deployed contract using a hardcoded address:
https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC777/ERC777.sol#L31

ethereum/migrations/2_next.js
vgeddes: Because these migration scripts interfere with unit tests (which use the development network). 

Truffle assumes that the unit tests need the contracts deployed, so it runs this migration script before running tests. But actually most of the unit tests need to redeploy the contracts before each individual test in a `beforeEach` function.

So there's this duplicated deployment which has the potential to introduce deployment errors, because the migration script creates state that is incompatible with what the unit tests are doing (or vice versa).

This has worked up until now without errors. But now we've introduced DOTApp and its indirect dependency on an ERC1820 registry: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC777/ERC777.sol#L31.

The only way I've got it to work is to deploy the ERC1820 registry in the unit tests.

ethereum/migrations/2_next.js
vgeddes: Also to add, The registry is already deployed to mainnet and ropsten. So that's why we only need to deploy it for E2E tests.

parachain/primitives/core/src/rewards.rs
vgeddes: done.

ethereum/contracts/DOTApp.sol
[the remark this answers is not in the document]
vgeddes: Renamed to "Snowfork DOT" / "SnowDOT" in https://github.com/Snowfork/polkadot-ethereum/pull/271/commits/d4718cf703bb31ea361784c39b20340a223b7f16

ethereum/contracts/DOTApp.sol
```diff
@@ -0,0 +1,91 @@
...
+        - Wrapped DOT has 256 bits of precision and 18 decimal places.
+    */
+
+    function wrap(uint128 _value) pure internal returns (uint256) {
+        return uint256(_value) * GRANULARITY;
+    }
+
+    function unwrap(uint256 _value) pure internal returns (uint128) {
+        return uint128(_value / GRANULARITY);
+    }
```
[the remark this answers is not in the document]
vgeddes: Potentially, though we may be able to get away with it since  `GRANULARITY` is a known constant.

I'm re-evaluating the security of these unwrap/wrap methods to see if there are any potential attack vectors.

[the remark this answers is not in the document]
vgeddes: We should be fine. Did some more investigation documented [here](https://github.com/Snowfork/polkadot-ethereum/pull/271/commits/775c20188a377ec67bf245e9eee774d390891e26#diff-2ba5dfc6637a7877a4585d96d345e1b4de0dd5b2ad99acca5443ea42e918ae20R91) in the latest upload, showing why we don't need SafeMath.

vgeddes:
> Can we also add e2e integration tests for this in a follow up PR?

@musnit E2E Tests added in latest upload.

Issue #268, opened 2021-02-10
Potentially redundant gas checking in InboundChannel


denalimarsh:
Test if gas checking in `InboundChannel.sol`, `IncentivizedInboundChannel.sol` is redunant and if so remove it.

Do we even need those gas checks? When we execute the calls, we limit the gas, which might be enough?

_Originally posted by @philipstanislaus in https://github.com/Snowfork/polkadot-ethereum/pull/266#discussion_r573777242_

vgeddes:
Agree, I bumped into issues around this while testing the new DOT app (which seems to be gas-heavy until I optimize it further), and had to comment out the gas checks in the solidity code.

I also changed the relayer to use a static gas limit when making calls, instead of using `estimateGas` as that caused ganache to crash.

Anyway will submit a PR for this.

vgeddes:
The gas checks in Solidity also have the potential to break the bridge by refusing messages, even though they were accepted for delivery by the Substrate side. The refusals would lead to mismatched nonces and effectively prevent further messages being sent over the bridge.

musnit:
@vgeddes No, our gas checking logic nd constants should be high enough to ensure that all messages in a commitment will be accepted. It should not refuse messages. Do you have an example of your concern?

vgeddes:
We have a linear issue for this: https://linear.app/snowfork/issue/SNO-89/channel-flow-control-limits

PR #278, opened 2021-02-16
Fixes for commitment generation and forwarding

vgeddes:
Changes:
* Parachain
  * Fixed outdated format of ABI-encoded payload (necessary for correct commitment generation)
* Relayer
  * Updated contract bindings
  * Added documentation about how to update bindings
  * Forward commitment hash to Inbound channel
  * Use a static gas limit of `500000` instead of calling `eth_estimateGas`
* Solidity
  * Remove redundant gas checks

E2E tests pass successfully.

Closes: #268


relayer/chain/ethereum/writer.go
```diff
@@ -105,9 +105,10 @@ func (wr *Writer) Write(ctx context.Context, msg *chain.SubstrateOutboundMessage
 	}
 
 	options := bind.TransactOpts{
-		From:    wr.conn.kp.CommonAddress(),
-		Signer:  wr.signerFn,
-		Context: ctx,
+		From:     wr.conn.kp.CommonAddress(),
+		Signer:   wr.signerFn,
+		Context:  ctx,
+		GasLimit: 500000,
```
musnit, line 111: @vgeddes can we put this into an env/config file rather?

relayer/contracts/generate.go
```diff
@@ -0,0 +1,4 @@
+//go:generate bash -c "jq .abi ../../ethereum/build/contracts/InboundChannel.json | abigen --abi - --type contract --pkg inbound --out inbound/contract.go"
+//go:generate bash -c "jq .abi ../../ethereum/build/contracts/OutboundChannel.json | abigen --abi - --type contract --pkg outbound --out outbound/contract.go"
+
+package contracts
```
philipstanislaus, line 4: Nice!

ethereum/contracts/BasicInboundChannel.sol
```diff
@@ -80,11 +62,8 @@ contract BasicInboundChannel is InboundChannel {
             nonce = nonce + 1;
 
             // Deliver the message to the target
-            // Delivery will have fixed maximum gas allowed for the target app
             (bool success, ) =
-                _messages[i].target.call{value: 0, gas: MAX_GAS_PER_MESSAGE}(
-                    _messages[i].payload
-                );
```
philipstanislaus, line 87: Maybe I'm missing something, but don't we still need this gas check? I think only the ones above with the `require` statements were unnecessary, right?

ethereum/contracts/IncentivizedInboundChannel.sol
```diff
@@ -90,9 +64,7 @@ contract IncentivizedInboundChannel is InboundChannel {
             // Deliver the message to the target
             // Delivery will have fixed maximum gas allowed for the target app
             (bool success, ) =
-                _messages[i].target.call{value: 0, gas: MAX_GAS_PER_MESSAGE}(
-                    _messages[i].payload
-                );
```
philipstanislaus, line 95: Likewise, I think only the checks above with the `require` statements were unnecessary, right?

ethereum/contracts/BasicInboundChannel.sol
philipstanislaus: Nevermind, has been added here again: https://github.com/Snowfork/polkadot-ethereum/pull/279

ethereum/contracts/IncentivizedInboundChannel.sol
philipstanislaus: Nevermind, has been added here again: https://github.com/Snowfork/polkadot-ethereum/pull/279

PR #279, opened 2021-02-16
Add gas-checking back to the Solidity contracts

vgeddes:
Passed the E2E tests.


musnit (approved):
looks fine. we'll do another review/audit and change this further in our final solidity milestone too, so if this is working for you with the relayer changes/latest code, i think we're good :)

denalimarsh (commented):
LGTM

philipstanislaus (commented):
LGTM

PR #282, opened 2021-02-19
Incentivized channel

rubydog:


ethereum/contracts/IncentivizedOutboundChannel.sol
```diff
@@ -7,10 +7,21 @@ import "./OutboundChannel.sol";
 // IncentivizedOutboundChannel is a channel that sends ordered messages with an increasing nonce. It will have incentivization too.
 contract IncentivizedOutboundChannel is OutboundChannel {
 
+    uint256 gasFee;
+    address feeController;
```
philipstanislaus, line 11: Where is this set?

ethereum/contracts/IncentivizedOutboundChannel.sol
```diff
@@ -19,6 +30,16 @@ contract IncentivizedOutboundChannel is OutboundChannel {
         override
     {
         nonce = nonce + 1;
-        emit Message(msg.sender, nonce, payload);
+        emit Message(msg.sender, nonce, payload, msg.value);
```
philipstanislaus, line 33: Don't we need a check here whether the fee is enough?

ethereum/contracts/IncentivizedOutboundChannel.sol
```diff
@@ -7,10 +7,21 @@ import "./OutboundChannel.sol";
 // IncentivizedOutboundChannel is a channel that sends ordered messages with an increasing nonce. It will have incentivization too.
 contract IncentivizedOutboundChannel is OutboundChannel {
 
+    uint256 gasFee;
+    address feeController;
+
     constructor() {
         nonce = 0;
+        gasFee = 0;
```
denalimarsh, line 15: I think it might make sense here to pass `gasFee` as an argument to the constructor.

ethereum/contracts/IncentivizedOutboundChannel.sol
```diff
@@ -19,6 +30,16 @@ contract IncentivizedOutboundChannel is OutboundChannel {
...
-        emit Message(msg.sender, nonce, payload);
+        emit Message(msg.sender, nonce, payload, msg.value);
+    }
+
+    modifier onlyFeeController {
+        require(msg.sender == feeController);
+        _;
+    }
+
+    function setGasFee(uint256 _fee) public onlyFeeController {
```
denalimarsh, line 41: Unless `feeController` is set in a contract that extends `IncentivizedOutboundChannel` the modifier will reject all attempts to update the gas fee.

ethereum/contracts/IncentivizedOutboundChannel.sol
musnit, line 33:
```suggestion
        emit Message(msg.sender, nonce, payload, gasFee);
```

rubydog:
This is new PR https://github.com/Snowfork/polkadot-ethereum/pull/294

PR #284, opened 2021-02-20
Basic Channel pallet

alvatar:


parachain/pallets/bridge/src/channel/outbound.rs
```diff
@@ -39,9 +39,9 @@ impl<T: Config> OutboundChannel for BasicOutboundChannel<T> {
 	// This implementation is a WIP!
 	fn submit(&self, target: H160, payload: &[u8]) -> DispatchResult {
 		self.storage.try_mutate(|data| {
-			data.nonce += 1;
-			T::MessageCommitment::add(self.id, target, data.nonce, payload)?;
-			<Module<T>>::deposit_event(Event::MessageAccepted(self.id, data.nonce));
+			// data.nonce += 1;
+			// T::MessageCommitment::add(self.id, target, data.nonce, payload)?;
+			// <Module<T>>::deposit_event(Event::MessageAccepted(self.id, data.nonce));
```
philipstanislaus, line 44: Are those commented out intentionally?

parachain/pallets/bridge/src/channel/outbound.rs
```diff
@@ -65,9 +65,9 @@ impl<T: Config> OutboundChannel for IncentivizedOutboundChannel<T> {
...
```
philipstanislaus, line 70: Are those commented out intentionally?

philipstanislaus (commented):
Great, left a few comments 👍

parachain/pallets/basic-channel/src/lib.rs
```diff
@@ -0,0 +1,115 @@
...
+		#[weight = 0]
+		pub fn submit(origin, message: Message) -> DispatchResult {
+			let relayer = ensure_signed(origin)?;
+			// submit message to verifier for verification
+			let log = T::Verifier::verify(&message)?;
+
+			// Decode log into an Envelope
+			let envelope = Envelope::try_from(log).map_err(|_| Error::<T>::InvalidEnvelope)?;
+
+			let channel = BasicInboundChannel::<T>::new(envelope.source);
```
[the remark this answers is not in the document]
musnit: source will be the end user that sent the message into the channel i think, not the source contract?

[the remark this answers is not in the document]
alvatar: I think envelope.source comes from `msg.sender` on the Ethereum side. Prior to this change, there was a single inbound channel (well, one for basic and one for incentivized). This change creates a channel per user. So if we need to have multiple channels per user, we need a different approach.

parachain/pallets/basic-channel/src/tests.rs
```diff
@@ -0,0 +1,23 @@
...
+#[test]
+fn correct_error_for_none_value() {
+	new_test_ext().execute_with(|| {
+		// Ensure the expected error is thrown when no value is present.
+		assert_noop!(
+			TemplateModule::cause_error(Origin::signed(1)),
+			Error::<Test>::NoneValue
+		);
+	});
+}
```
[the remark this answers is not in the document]
alvatar: Definitely, I need to fix all tests before calling this PR final. I just needed to have some pointers before continuing with it.

parachain/pallets/bridge/src/channel/outbound.rs
alvatar: This code is pending the decision whether to leave the bridge or not and which functionality to keep here exactly, since some changes are necessary for this code to work again. It's just WIP, in other words.

parachain/pallets/bridge/src/channel/outbound.rs
alvatar: Ditto

parachain/pallets/basic-channel/src/lib.rs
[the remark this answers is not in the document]
vgeddes: `envelope.source` is currently the contract that sent the message. The end user account is encoded within `envelope.payload` for our current set of apps (eth, erc20, dot)

parachain/pallets/basic-channel/src/lib.rs
```diff
@@ -0,0 +1,115 @@
...
+	type MessageCommitment: MessageCommitment;
+
+	/// Verifier module for message verification.
+	type MessageDispatch: MessageDispatch<MessageId>;
+}
+
+decl_storage! {
+	trait Store for Module<T: Config> as BasicChannelModule {
+		/// Storage for inbound channels.
+		pub InboundChannels: map hasher(identity) EthAddress => InboundChannelData;
```
vgeddes, line 59: By convention we just use `H160` to represent ethereum addresses, without using an alias like `EthAddress`.

parachain/pallets/basic-channel/src/lib.rs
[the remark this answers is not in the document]
blablacio: Hmmm, I'm currently looking into this because tests are failing when I use `tx.origin` instead of `msg.sender` in the outbound channel. Is this the reason?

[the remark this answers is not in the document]
musnit: hmm, well for incentivizedchannel, we need the source address of the contract that send the message. for basic, we need that too, but in addition we need the tx.origin of the user. so we'll need an extra field in the basic for this.

parachain/pallets/basic-channel/src/lib.rs
```diff
@@ -0,0 +1,115 @@
...
+	type Verifier: Verifier;
+
+	/// Used by outbound channels to persist messages for outbound delivery.
+	type MessageCommitment: MessageCommitment;
+
+	/// Verifier module for message verification.
+	type MessageDispatch: MessageDispatch<MessageId>;
+}
+
+decl_storage! {
```
vgeddes, line 56: Now that we've split out the basic channel from the bridge, some of the abstractions and associated boilerplate in the old design no longer make sense.

Particularly, we can just store the account nonces directly, without putting them into those `InboundChannelData` or `OutboundChannelData` structs.

For example
```rust
decl_storage! {
	trait Store for Module<T: Config> as BasicChannelModule {
		/// Nonces for inbound channels.
		pub InboundNonces: map hasher(identity) H160 => u64;
		/// Storage for outbound channels.
		pub OutboundNonces: map hasher(identity) T::AccountId => u64;
	}
}
```

By implication we can mostly just remove all the code under `channels/`, except for the `InboundChannel::submit` and `OutboundChannel::submit` methods. We should move those methods to a impl block like this:

```
impl<T: Config> Module<T> {
  fn submit_inbound(...) { ... }
  fn submit_outbound(...) { ... }
}
```

parachain/pallets/basic-channel/src/lib.rs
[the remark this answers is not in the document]
vgeddes: > `envelope.source` is currently the contract that sent the message. The end user account is encoded within `envelope.payload` for our current set of apps (eth, erc20, dot)

So of course we'll need to add another field like `origin` (or some other name) to the envelope to track the end user who sent the message.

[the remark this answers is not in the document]
alvatar: Sounds good.
We need the mapping to be tx.origin of the user -> nonce then. 
The remaining question then is if this is enough granularity or we want to map (user, channel or contract) -> nonce
If that's the case, we'd need both the tx.origin _and_ the envelope.source.
What do you think?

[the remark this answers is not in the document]
musnit: i think we need both

tx.origin -> nonce mapping for replay protection checks
envelope.source for application contract address, so that the application receiving the payload can verify it came from the right contract app

parachain/pallets/basic-channel/src/lib.rs
```diff
@@ -0,0 +1,115 @@
...
+
+		#[weight = 0]
+		pub fn submit(origin, message: Message) -> DispatchResult {
+			let relayer = ensure_signed(origin)?;
+			// submit message to verifier for verification
+			let log = T::Verifier::verify(&message)?;
+
+			// Decode log into an Envelope
+			let envelope = Envelope::try_from(log).map_err(|_| Error::<T>::InvalidEnvelope)?;
+
```
vgeddes, line 101: We still need to port over the check from https://github.com/Snowfork/polkadot-ethereum/blob/main/parachain/pallets/bridge/src/lib.rs#L121.

The basic inbound channel needs to verify that the message was forwarded from a known outbound channel on the Ethereum side. Otherwise anyone can send messages and corrupt the bridge.

The address of the outbound channel contract is recorded in `envelope.channel`.

I suggest adding the following storage item:
```
  SourceChannel get(fn source_channel) config(): H160;
```

vgeddes (commented):
Is a good start, thanks Alvaro! 🚀 

Added a few comments inline.

parachain/pallets/basic-channel/src/lib.rs
alvatar: Yeah totally, I was thinking of it too.
The only thing that would prevent it is if we need something more complex for a (user, channel) -> nonce mapping, but I was thinking on using double maps for that (need to look further)

parachain/pallets/basic-channel/src/lib.rs
alvatar: Great, thanks :+1:

.github/workflows/parachain.yml
```diff
@@ -77,11 +77,17 @@ jobs:
...
-            --manifest-path parachain/Cargo.toml
-            --verbose --workspace
-            --exclude artemis,artemis-runtime
-            --exclude-files 'src/command.rs,*/mock.rs,*/test.rs'
-            --coveralls ${{ secrets.COVERALLS_REPO_TOKEN }}
-          toolchain: nightly-2021-01-10
+          command: test
+          toolchain: stable
+          args: --manifest-path parachain/Cargo.toml --workspace --exclude artemis
+      # - name: cargo test
```
alvatar, line 83: Tarpaulin is producing a strange error during `cargo test`, it looks like a bug. I disabled it temporarily because it seems obscure, but I can try and dig deeper into this.

alvatar:
Merged `into feature/basic-channel` and working on integration there with @blablacio

PR #287, opened 2021-02-21
Add ropsten-compatible deployment configuration

vgeddes:
I set `DescendantsUntilFinalized` to 3 until we have a solution for https://github.com/Snowfork/polkadot-ethereum/issues/286.

### DIFF
--- a/parachain/runtime/src/lib.rs
+++ b/parachain/runtime/src/lib.rs
@@ -368,7 +368,7 @@ impl bridge::Config for Runtime {
 
 #[cfg(not(feature = "test-e2e"))]
 parameter_types! {
-	pub const DescendantsUntilFinalized: u8 = 35;
+	pub const DescendantsUntilFinalized: u8 = 3;
 	pub const VerifyPoW: bool = true;
 }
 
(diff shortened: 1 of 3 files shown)


parachain/runtime/src/lib.rs
```diff
@@ -368,7 +368,7 @@ impl bridge::Config for Runtime {
 
 #[cfg(not(feature = "test-e2e"))]
 parameter_types! {
-	pub const DescendantsUntilFinalized: u8 = 35;
+	pub const DescendantsUntilFinalized: u8 = 3;
```
Rizziepit, line 371: This is quite dangerous for mainnet I think. But as a temporary thing, that's fine

vgeddes: yeah is a total temporary thing, as ropsten is our de-facto "live" environment until we actually deploy a production stack connected to mainnet

philipstanislaus: Can we add a ticket for that though?

philipstanislaus: Done actually: https://github.com/Snowfork/polkadot-ethereum/issues/291

PR #288, opened 2021-02-22
Add a type definitions package

vgeddes:
The goal is to have a reusable NPM package containing all the type definitions for our parachain.

For this purpose, I've previously created https://github.com/Snowfork/snowbridge-types, which I've now merged into the monorepo.

Before we can merge this work, we need to submit a PR to `@polkadot/apps` so that the webapp will use our types package automatically when connecting to our chain.


parachain/runtime/src/lib.rs
```diff
@@ -106,10 +106,10 @@ pub mod opaque {
 
 /// This runtime version.
 pub const VERSION: RuntimeVersion = RuntimeVersion {
-	spec_name: create_runtime_str!("artemis-node"),
-	impl_name: create_runtime_str!("artemis-node"),
+	spec_name: create_runtime_str!("snowbridge"),
```
musnit, line 109: will need to rename stuff in the cumulus node to be called snowbridge too

vgeddes: You mean the name of the binary? If so, sure yeah that will happen in time.

For more context, The polkadot webapp specifically uses `spec_name` to apply our custom types when someone browses our chain: See https://github.com/polkadot-js/apps/blob/master/packages/apps-config/src/api/spec/index.ts#L59 for example.

PR #289, opened 2021-02-22
Add types definition package

vgeddes:


musnit:
can use it from test/../subclient.ts too plz

musnit:
substrate currently has no way to automatically generate this right? has to be manually created?

vgeddes:
> can use it from test/../subclient.ts too plz

This is just the first of several PRs, as this change has to be made in several phases:
* First add the types package to the monorepo
* Ask upstream to use our types package in the polkadot webapp
* Once they OK it, then follow up and make the necessary changes to get everything else to use our types package.

vgeddes:
> substrate currently has no way to automatically generate this right? has to be manually created?

Correct. However there are long-term plans to update the SCALE codec to include type metadata. Thats a long way off though.

PR #294, opened 2021-02-23
Refactor IncentivizedOutboundChannel to accept and pay fees 

rubydog:


ethereum/contracts/IncentivizedOutboundChannel.sol
```diff
@@ -19,6 +31,28 @@ contract IncentivizedOutboundChannel is OutboundChannel {
...
+    */
+    function setGasFee(uint256 _fee) public onlyFeeController {
+        require(_fee > 0, "fee must be positive");
+        gasFee = _fee;
+    }
+
+    /**
+    * @dev Change feeController address
+    */
+    function setFeeController(address _feeControllerAddress) public onlyFeeController {
```
blablacio, line 55: I think this contract should be `Ownable` and the modifier should be set to `onlyOwner`.

ethereum/contracts/IncentivizedOutboundChannel.sol
```diff
@@ -31,11 +31,13 @@ contract IncentivizedOutboundChannel is OutboundChannel {
         override
     {
         nonce = nonce + 1;
-        emit Message(msg.sender, nonce, payload, msg.value);
+
+
+        emit Message(msg.sender, nonce, payload, gasFee);
```
blablacio, line 36: Shouldn't `msg.value` be preserved here?

ethereum/contracts/IncentivizedOutboundChannel.sol
rubydog: You mean to setFeeController?

rubydog:
Automated tests are work in progress.

ethereum/contracts/IncentivizedOutboundChannel.sol
blablacio: No I mean that the contract should be `Ownable` and that the modifier should be changed from `onlyFeeController` to `onlyOwner`. Not sure what the implementation details are here, but I think that makes sense.

rubydog: feeController is different from the owner of the contract. Only feeController is allowed to set gasFee and change itself.

ethereum/contracts/IncentivizedOutboundChannel.sol
```diff
@@ -7,10 +7,22 @@ import "./OutboundChannel.sol";
 // IncentivizedOutboundChannel is a channel that sends ordered messages with an increasing nonce. It will have incentivization too.
 contract IncentivizedOutboundChannel is OutboundChannel {
 
-    constructor() {
+    uint256 public gasFee;
```
denalimarsh, line 10:
```suggestion
    uint256 public fee;
```
This terminology may be a bit misleading. The fee here isn't gas paid to Ethereum network miners for quicker relay, but rather is paid to Relayers. I think `fee` or `relayFee` may be clearer.

denalimarsh (commented):
Looks like this replaces https://github.com/Snowfork/polkadot-ethereum/pull/282, if so could you close that PR with a comment that says this is the new PR? Thanks!

ethereum/contracts/IncentivizedOutboundChannel.sol
denalimarsh: I agree with @blablacio here. The `owner` is just an address that has elevated permissions. In this case, the `owner` would have permission to update fees. It's better to reuse existing secure contracts (such as OpenZeppelin's `Ownable.sol`) when possible to reduce potential attack vectors on new code.

musnit: disagree - I don't think the terminology owner is good or reflects the meaning of this role. using existing libs for roles/authorization though, yeh that's good.

ethereum/contracts/IncentivizedOutboundChannel.sol
musnit: this has nothing to with msg.value. msg.value is for eth. fees are not paid in eth, they're paid in SnowDOT. we should possibly even reject transactions that have any msg.value.

ethereum/contracts/IncentivizedOutboundChannel.sol
musnit: yeh relayFee is good

ethereum/contracts/IncentivizedOutboundChannel.sol
denalimarsh: Ah okay, that perspective makes sense. Worth looking into [AccessControl.sol](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/access/AccessControl.sol) here then.

ethereum/contracts/IncentivizedOutboundChannel.sol
denalimarsh: Ahh gotcha, my mental model for the Outbound channel was off.

philipstanislaus (approved):
LGTM

musnit:
@rubydog looks good overall now - next step is to implement the actual fee payment/integrate with SnowDOT and DOTApp

ethereum/contracts/IncentivizedOutboundChannel.sol
rubydog: [ AccessControl.sol ](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/access/AccessControl.sol) also has `DEFAULT_ADMIN_ROLE` which has all the permissions (like a super user/role). It by default set to `0x00`, can this be a problem? Should we set owner of the contract as DEFAULT_ADMIN?

rubydog:
@musnit @denalimarsh I don't think we can use [AccessControl.sol](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/access/AccessControl.sol). It only lets an account with admin role to grant any other role to an account. In our case we want to let **only** feeController change itself, not an admin.

ethereum/contracts/IncentivizedOutboundChannel.sol
vgeddes: That sounds fine. Either way, it may not matter much soon. One of my upcoming tasks is to introduce collective governance mechanisms. This work will likely end up refactoring any code related to on-chain configuration and permissions. so this `setFeeController` may not even be necessary right now.

ethereum/contracts/IncentivizedOutboundChannel.sol
```diff
@@ -3,14 +3,33 @@ pragma solidity >=0.7.6;
...
 
 // IncentivizedOutboundChannel is a channel that sends ordered messages with an increasing nonce. It will have incentivization too.
 contract IncentivizedOutboundChannel is OutboundChannel {
 
-    constructor() {
+    uint256 public relayFee;
+    address public feeController;
+    DOTApp private _DOTApp;
+
+    constructor(uint256 _relayFee, address _feeControllerAddress) {
```
vgeddes, line 15: I'd prefer if we have zero constructor parameters for now, mostly due to the current structure of our unit tests.

I suggest defaulting `_relayFee` to 0, and `_feeControllerAddress` to `msg.sender`.

ethereum/contracts/IncentivizedOutboundChannel.sol
```diff
@@ -3,14 +3,33 @@ pragma solidity >=0.7.6;
...
 import "./OutboundChannel.sol";
+import "./DOTApp.sol";
 
 // IncentivizedOutboundChannel is a channel that sends ordered messages with an increasing nonce. It will have incentivization too.
 contract IncentivizedOutboundChannel is OutboundChannel {
 
-    constructor() {
+    uint256 public relayFee;
+    address public feeController;
+    DOTApp private _DOTApp;
```
vgeddes, line 13: The underscore in `_DOTApp` isn't necessary here, as we've chosen to only use them in function parameter names.

So just `dotApp` instead.

vgeddes (commented):
Looks great! Thanks Bhushan! Added some comments...

ethereum/test/test_incentivized_outbound_channel.js
```diff
@@ -13,18 +14,25 @@ require("chai")
...
-      this.channel = await IncentivizedOutboundChannel.new();
+      this.channel = await IncentivizedOutboundChannel.new(relayFee, feeController);
     });
   });
 
   describe("send", function () {
     beforeEach(async function () {
-      this.channel = await IncentivizedOutboundChannel.new();
+      this.channel = await IncentivizedOutboundChannel.new(relayFee, feeController);
+      const mock = await MockContract.new()
```
vgeddes, line 33: You shouldn't need to use mock for DOTApp. You can make use of the `deployAppContractWithChannels` helper by copying over these lines from` test_dot_app.js`:
```
      this.erc1820 = await singletons.ERC1820Registry(owner);
      [this.channels, this.app] = await deployAppContractWithChannels(DOTApp);
      this.token = await Token.at(await this.app.token());
```

But for that to work, you either need remove the constructor parameters from IncentivizedOutboundChannel, or update the `deployAppContractWithChannels` helper to pass the constructor parameters when creating the channel.
https://github.com/Snowfork/polkadot-ethereum/pull/294#discussion_r584705075

Its probably easiest to remove the constructor parameters.

ethereum/contracts/IncentivizedOutboundChannel.sol
vgeddes: Ah ok I see you've updated the unit tests at https://github.com/Snowfork/polkadot-ethereum/pull/294/files#diff-5a92558b750f5ec20439a92fe71c62cbbf40115e3b721e8d8ce106d9d3a336baR167, so feel free to ignore this.

ethereum/test/test_incentivized_outbound_channel.js
vgeddes: Ah I see you updated deployAppContractWithChannels already. sorry, I missed that. Did deployAppContractWithChannels  still not work for you after that?

ethereum/test/test_incentivized_outbound_channel.js
```diff
@@ -13,18 +14,25 @@ require("chai")
...
+      await mock.givenAnyReturnBool(true)
+      this.channel.setDOTApp(mock.address, { from: owner })
```
vgeddes, line 35: this expression is a contract call so it needs to be `await`'ed.

ethereum/contracts/IncentivizedOutboundChannel.sol
```diff
@@ -3,22 +3,64 @@ pragma solidity >=0.7.6;
...
 
 import "./OutboundChannel.sol";
+import "./DOTApp.sol";
 
 // IncentivizedOutboundChannel is a channel that sends ordered messages with an increasing nonce. It will have incentivization too.
 contract IncentivizedOutboundChannel is OutboundChannel {
 
+    uint256 public relayFee;
+    address public feeController;
+    DOTApp private dotApp;
```
vgeddes, line 13: To make your branch compatible with latest main:
```suggestion
    BaseDOTApp private dotApp;
```

ethereum/contracts/IncentivizedOutboundChannel.sol
```diff
@@ -3,22 +3,64 @@ pragma solidity >=0.7.6;
 pragma experimental ABIEncoderV2;
 
 import "./OutboundChannel.sol";
+import "./DOTApp.sol";
```
vgeddes, line 6:
```suggestion
import "./BaseDOTApp.sol";
```

vgeddes:
Closing this for now. I'm gonna incorporate the various learnings and discoveries from this work into a new PR. thanks @rubydog!

PR #295, opened 2021-02-23
Beefy relay compatibility

musnit:
This PR adds support for running a relay chain with the beefy pallet support. All e2e tests still pass.

Related:
 - Using polkadot-launch from: https://github.com/paritytech/polkadot-launch/pull/62
 - Using polkadot from: https://github.com/paritytech/polkadot/pull/2101

We may not want to merge this to main yet, as eventually the above 2 PRs will be part of polkadot-launch/polkadot master, but I do think the team and staging and all future PRs should base themselves off this and work to maintain compatibility with this.

@denalimarsh @philipstanislaus This PR also includes a node script (subscribeBeefyJustifications), which, if you run it after everything is setup, will connect to the appropriate RPC endpoints on the relay chain and parachain and subscribe to all the data we'll need to be relayed into the Solidity light client. This script should help make it easy to integrate all the endpoints and relaying of all data types into our Golang relayer. It includes:
 1 - Getting initial list of authorities (note: currently a small bug, so hardcoded)
 2 - Getting each new beefy commitment produced and the collection of signatures on that commitment (which we need for the Relay Chain Light Client)
 3 - Getting the latest new MMR block that the beefy commitment commits to, which we need to process the commitment for validator set updates/epoch changes/new parachain headers and the proof for that MMR
 4 (WIP/TODO) - Getting the parachain heads commitment in that MMR, a merkle proof of our parachain's new head within parachain heads, and the new head data itself

Note: I had a minor issue building the suggested branch of Polkadot with an error related to `polkadot-node-primitives`. If you encounter this, you can fix it by deleting the redundant `node-primitives` line from `node/network/statement-distribution/Cargo.toml`

### DIFF
--- a/parachain/config.json
+++ b/parachain/config.json
@@ -13,6 +13,10 @@
 				"wsPort": 9955,
 				"port": 30555
 			}
+		],
+		"flags": [
+			"--enable-offchain-indexing=true",
+			"--offchain-worker=Always"
 		]
 	},
 	"parachains": [
--- a/parachain/config.json
+++ b/parachain/config.json
@@ -6,17 +6,21 @@
 			{
 				"name": "alice",
 				"wsPort": 9944,
-				"port": 30444
+				"port": 30444,
+				"flags": [
+					"--enable-offchain-indexing=true",
+					"--offchain-worker=Always"
+				]
 			},
 			{
 				"name": "bob",
 				"wsPort": 9955,
-				"port": 30555
+				"port": 30555,
+				"flags": [
+					"--enable-offchain-indexing=true",
+					"--offchain-worker=Always"
+				]
 			}
-		],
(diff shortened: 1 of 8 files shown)


parachain/config.json
```diff
@@ -32,6 +32,7 @@
 		}
 	],
 	"simpleParachains": [],
+	"hrmpChannels": [],
```
musnit, line 35: @vgeddes unrelated q: could something related to this config be what caused weirdness when you implemented XCMP?

alvatar: FWIW this is one of the things I've tried during my journey setting up the local chain

musnit:
included in https://github.com/Snowfork/polkadot-ethereum/pull/343

PR #296, opened 2021-02-24
Fix relayer tests

blablacio:
Just a few fixes for the relayer tests.

One test is skipped due to dependency on external data (a specific file under `~/.ethash`).


vgeddes (approved):
LGTM! Thanks!

PR #298, opened 2021-02-24
Update README for E2E tests

vgeddes:
A new step is now required before setting up the E2E test stack:

```bash
cd ../types  # change into the toplevel types directory
yarn install
yarn build


# continue with rest of setup...
```

This is so we can link the `@snowfork/snowbridge-types` package correctly into the dependencies for the e2e tests.


denalimarsh (approved):
LGTM

PR #301, opened 2021-02-28
Start using canonical Eth difficulty calculation

Rizziepit:
We were using a simplified difficulty check where a child block's difficulty must fall within a +/- 1% range from the parent's difficulty. This was a super convenient simplification but it's actually incorrect, specifically when we get close to the "ice age". It also has security implications that are best avoided. So this PR replaces it with the canonical Ethereum difficulty calculation.

Some implementation notes:
* This only supports **blocks from the Byzantium fork onwards**. Earlier blocks will always return `InvalidHeader`.
* Test cases come from https://github.com/ethereum/tests/.
* I added `DifficultyConfig` and `BombDelay` so that it's trivial to add another bomb delay fork block. Behold, another EIP is coming for exactly this: https://eips.ethereum.org/EIPS/eip-3238


Rizziepit:
I moved the Eth header types from ethereum/lib.rs -> ethereum/header.rs. Only new code there is the `Header::has_ommers` function

parachain/primitives/ethereum/src/difficulty.rs
```diff
@@ -0,0 +1,253 @@
...
+	config: &DifficultyConfig,
+	time: u64,
+	parent: &Header,
+) -> Result<U256, &'static str> {
+	let bomb_delay = config.bomb_delay(parent.number + 1)
+		.ok_or("Cannot calculate difficulty for block number prior to Byzantium")?;
+
+	let block_time = time.checked_sub(parent.timestamp).ok_or("Invalid block time")?;
+	let block_time_div_9 = i32::try_from(block_time / 9).or(Err("Invalid block time"))?;
+	let sigma2 = match parent.has_ommers() {
```
vgeddes, line 76: For the benefit of readers, can we explicitly add a type here: like `sigma: i32` I'm assuming its `i32`.

parachain/primitives/ethereum/src/difficulty.rs
```diff
@@ -0,0 +1,253 @@
...
+	let mut difficulty_without_exp = parent.difficulty;
+	if sigma2 < 0 {
+		difficulty_without_exp -= (parent.difficulty >> DIFFICULTY_BOUND_DIVISOR) * (sigma2 * -1) as u32;
+	} else {
+		difficulty_without_exp += (parent.difficulty >> DIFFICULTY_BOUND_DIVISOR) * sigma2 as u32;
+	}
+
+	difficulty_without_exp = difficulty_without_exp.max(MINIMUM_DIFFICULTY.into());
+
+	let fake_block_number: u32 = u32::try_from(parent.number)
```
vgeddes, line 90: If `Header::number` is `u64`, why would we want to reduce precision to `u32`?

parachain/primitives/ethereum/src/difficulty.rs
```diff
@@ -0,0 +1,253 @@
...
+pub fn calc_difficulty(
+	config: &DifficultyConfig,
+	time: u64,
+	parent: &Header,
+) -> Result<U256, &'static str> {
+	let bomb_delay = config.bomb_delay(parent.number + 1)
+		.ok_or("Cannot calculate difficulty for block number prior to Byzantium")?;
+
+	let block_time = time.checked_sub(parent.timestamp).ok_or("Invalid block time")?;
+	let block_time_div_9 = i32::try_from(block_time / 9).or(Err("Invalid block time"))?;
```
vgeddes, line 75: If `Header::timestamp` is `u64`, why would we want to reduce precision to `i32` after the division? Wouldn't `i64` be fine?

parachain/primitives/ethereum/src/difficulty.rs
```diff
@@ -0,0 +1,253 @@
...
+	let sigma2 = match parent.has_ommers() {
+		true => 2 - block_time_div_9,
+		false => 1 - block_time_div_9,
+	}.max(-99);
+
+	let mut difficulty_without_exp = parent.difficulty;
+	if sigma2 < 0 {
+		difficulty_without_exp -= (parent.difficulty >> DIFFICULTY_BOUND_DIVISOR) * (sigma2 * -1) as u32;
```
vgeddes, line 83: would prefer `sigma2.checked_abs()`, as its more descriptive, and of course is safer.

parachain/primitives/ethereum/src/difficulty.rs
Rizziepit: I feel like there was a reason but I can't remember what it was :D I'll keep it u64 and see if I run into an issue again

parachain/primitives/ethereum/src/difficulty.rs
Rizziepit: Block time should be fairly small, e.g. 10 - 20 seconds. If it doesn't fit into `i32`, something has probably gone wrong. But, thinking about it some more, I don't think this function is the place to enforce block time. At least it doesn't appear in Geth's implementation

parachain/primitives/ethereum/src/difficulty.rs
Rizziepit: I'm going to go with `abs()` for brevity. `sigma2`'s range is strictly [-99, 2]

PR #304, opened 2021-03-02
Merkle Tree commitments (on new basic channel implementation)

alvatar:
This is intended to be merged on the basic-channel-pallet branch, but done in a separate PR for now


parachain/Cargo.toml
```diff
@@ -67,7 +68,7 @@ polkadot-parachain = { git = "https://github.com/paritytech/polkadot.git", branc
 
 artemis-core = { path = "primitives/core" }
 artemis-runtime = { path = "runtime" }
-tarpaulin = "0.1.0"
+#tarpaulin = "0.1.0"
```
alvatar, line 71: Tarpaulin is giving me trouble, both on my machine while compiling, and on Github's CI

parachain/pallets/commitments/src/lib.rs
```diff
@@ -98,69 +106,145 @@ decl_module! {
...
 
 	// TODO: return proper weight
 	fn commit() -> Weight {
-		let mut weight: Weight = 0;
+		Self::commit_for_basic_channel() + Self::commit_for_incentivized_channel()
+	}
 
-		for channel_id in ChannelId::into_enum_iter() {
-			weight += Self::commit_for_channel(channel_id);
+	fn commit_for_basic_channel() -> Weight {
```
vgeddes, line 124: I think we're at the point where this commitments pallet has outlived its usefulness, as it was originally designed to generate one kind of commitment. Now its generating a basic and an incentivized commitment, which breaks abstraction boundaries.

Lets move the commitment code to the respective outbound channels, and then delete this pallet.

parachain/pallets/commitments/src/lib.rs
```diff
@@ -1,42 +1,41 @@
...
+mod merkle_tree;
 #[cfg(test)]
 mod mock;
-
 #[cfg(test)]
 mod tests;
 
 /// Auxiliary [`DigestItem`] to include in header digest.
 #[derive(Encode, Decode, Copy, Clone, PartialEq, RuntimeDebug)]
 pub enum AuxiliaryDigestItem {
```
vgeddes, line 28: Lets move `AuxiliaryDigestItem` to `primitives/core` so it can be reused by the both outbound channels which need to generate their own commitments.

parachain/pallets/commitments/src/lib.rs
alvatar: Good point, I like the idea.

parachain/primitives/core/src/assets.rs
```diff
@@ -1,20 +1,19 @@
 use frame_support::dispatch::DispatchResult;
 
-use codec::{Encode, Decode};
```
vgeddes, line 3: Lets please leave major linting/whitespace changes for another PR.

parachain/pallets/erc20-app/.rustfmt.toml
```diff
@@ -0,0 +1 @@
+hard_tabs = true
```
vgeddes, line 1: We don't currently use `rustfmt` as its incompatible with Substrate, so I'd rather we not check these in.

parachain/pallets/commitments/src/merkle_tree.rs
```diff
@@ -0,0 +1,62 @@
+use codec::Encode;
+use sp_core::{Hasher, H256};
+use sp_std::prelude::*;
+use sp_trie::TrieConfiguration;
+
+use hash256_std_hasher::Hash256StdHasher;
```
vgeddes, line 6: hmm, you should be using this hasher if possible, its already provided to the commitments pallet: https://github.com/Snowfork/polkadot-ethereum/blob/main/parachain/pallets/commitments/src/lib.rs#L55

Its a keccak256 hasher so is compatible.

alvatar: I'll try to make it work with that one

parachain/pallets/erc20-app/.rustfmt.toml
alvatar: Sure, I'll get rid of this. But I think it's compatible as long as the tabs are used, it just reformats code outside of macros.

parachain/primitives/core/src/assets.rs
alvatar: Sure, I will just remove the use of rustfmt as mentioned above.

parachain/pallets/commitments/src/merkle_tree.rs
```diff
@@ -0,0 +1,62 @@
...
+impl MerkleTree {
+	pub fn new<T: Encode>(items: impl Iterator<Item = T>) -> Self {
+		let ordered_items = items
+			.map(|x| Encode::encode(&x))
+			.enumerate()
+			.map(|(i, v)| (Layout::encode_index(i as u32), v))
+			.collect::<Vec<(Vec<u8>, Vec<u8>)>>();
+
+		let mut db = sp_trie::MemoryDB::<KeccakHasher>::default();
+		let mut cb = trie_db::TrieBuilder::new(&mut db);
```
vgeddes, line 40: Looks like this would build a trie as opposed to a binary merkle tree. A bit skeptical that this is would work out well for us. Is it even a merkle trie?

I'd really prefer a binary merkle tree unless there's a good reason for going with a trie.

We don't expect huge numbers of people to use the basic channel, so the advantages of a trie would likely not matter much in the end.

vgeddes: but keen to hear your thoughts though.

alvatar: Should be a Merkle trie, yes. I'm assuming that the trie_visit does the heavy lifting. See here, where I adapted the code form: https://github.com/paritytech/grandpa-bridge-gadget/blob/master/beefy-cli/src/cli/merkle_tree.rs#L118

But this is not a binary Merkle tree as you pointed out. I have no strong reasons to use the trie over the binary tree, so I can implement the binary tree.

alvatar: @musnit pointed out this implementation, so perhaps he has a reason for this to be a trie, though.

vgeddes: Ah ok, is fine then. If it’s being used in that beefy codebase, then it makes sense to reuse that.

vgeddes: I do have another other concern. We're building the tree all in one go, every K blocks, using messages collected from the queue. I see the following issues arising:

- Generating the tree could easily consume all the available block [weight](https://substrate.dev/docs/en/knowledgebase/learn-substrate/weight), which means no one else can include transactions in that block.
- Since the max block weight is currently 0.5 seconds, that puts a hard cap on the time we can take to generate the tree, and the number of messages we can include in a commitment.
- If generating the tree takes more than 0.5 seconds then the parachain kinda gets stuck.
- Since the tree is generated using an `on_initialize` hook, users who want to send messages over the bridge aren't paying their fair share for the time it takes to generate the tree. This makes our design susceptible to DOS attacks. Strictly speaking this was already an issue in the old commitment design - except its much more exploitable now since users don't have to pay fees to add messages to the basic commitment.

Am thinking that building the merkle tree incrementally over K blocks has better characteristics. That way, user's pay their fair share, and we aren't limited by the maximum block weight.

musnit: @vgeddes makes sense but incremental would be a bunch more complex to implement, so sounds like premature optimization? 

I suppose this also implies we need to bottleneck the acceptance of transactions to our parachain, both for basic and incentivized send, to ensure we dont get blocked unable to make commitments, both for the basic and incentivized channel?

vgeddes: It doesn't feel like premature optimization to me. Am worried this is a design decision which could come back to bite us. Especially the way users aren't paying for commitment generation, even though its derived from user-initiated activity.

I realise building an incremental merkle tree would take a fair amount of engineering effort. What about going back to a flat list commitment for Snowbridge V1? It would simplify things on the Ethereum side too.

Then the commitment for the basic channel could look like this:
```
Keccack256(
  [
    (account, message)
    (account, message)
    (account, message)
    (account, message)
  ]
)
```
And of course that can still be built up incrementally every block.

> I suppose this also implies we need to bottleneck the acceptance of transactions to our parachain, both for basic and incentivized send, to ensure we dont get blocked unable to make commitments, both for the basic and incentivized channel?

Yeah, that would definitely be necessary for non-incremental commitment generation. For incremental generation, our bottleneck would likely only be the gas cost of dispatching that commitment to Ethereum.

musnit: @vgeddes a flat commitment still has this same issue - if too many messages come in, then even hashing them could be overly gas intensive? or no?

musnit: also, if we do go with flat commitment, then account is not really that important to have in it.

anyway, please estimate some numbers on actual number of messages that would result in commitment generation issues. if its v high, then we can defer. if its low and likely to be hit, we'll go with flat and defer.

vgeddes: Ah I see we do need a commitment per user for the sake of the relayer. Ignore what I said about the flat list commitment.

vgeddes: > a flat commitment still has this same issue - if too many messages come in, then even hashing them could be overly gas intensive? or no?

@musnit A flat commitment can still be generated incrementally (as I said in that comment when I introduced the idea)

> anyway, please estimate some numbers on actual number of messages that would result in commitment generation issues. if its v high, then we can defer. if its low and likely to be hit, we'll go with flat and defer.

Isn't that simple as we'd need to do benchmarks & weights to figure this out - In fact it would be very similar to the work @Rizziepit is doing now.

musnit: @vgeddes ok. given in practice i expect the basic channel to have much less use, i really doubt this will be an issue in the short term. though lets definitely create a github security issue to deal with it in future and deal with it before full production release, but leave as is for this PR.

alvatar: To confirm, are we keeping the Merkle Trie implementation as it is now, bundling all messages?
With the current implementation, I think the performance issue could arise with a "many users, few messages per user" use case, as opposed to "few users with many messages per user", since the more you tend towards the latter, the closer it will resemble the flat commitments setting. Do you have a specific use case in mind, or is it designed with the intention to support all use cases?

There is another thing to consider too: proof generation. Using this code, and if I'm understanding its mechanics well, we'd need to generate a proof for each leaf and bundle it in the messages blob. At least this was my intention.
But I think that in this scenario it would make sense to _not_ store the proofs, and rather let the node generate them on-demand. If I'm understanding substrate correctly, that would be performed by the off-chain worker and as a consequence run in native code and also not be subject to the weight limitations. Is this the case @vgeddes ?

musnit: Yeh no proof-related stuff should be computed or stored onchain. Offchain proof generation and proof storage is fine. Vincent's concern is with the onchain merkle root generation, as you describe. It should support all use cases, but let's defer any optimizations/changes until later. Please create a github issue for this todo at a later point

alvatar: This would require adding a [custom RPC](https://substrate.dev/recipes/custom-rpc.html), but seems easy enough.

musnit: Or we can out the logic in the relayer, rather than a custom RPC. Either is fine - whichever you think is easiest

alvatar:
Outdated. It's part of the basic-channel branch now.

PR #308, opened 2021-03-05
Refactor channel architecture

vgeddes:
Changes:
* Substrate
  * Bridge module has been refactored into the following pallets
    * BasicInboundChannel
    * BasicOutboundChannel
    * IncentivizedInboundChannel
    * IncentivizedOutboundChannel
* Relayer
  * Updated contract bindings generator in `relayer/contracts/generate.go`.
  * In the Ethereum to Substrate direction, the relayer submits messages to either of these pallet dispatchables:
    * `BasicInboundChannel.submit`
    * `IncentivizedInboundChannel.submit`
* Ethereum
  * Outbound channels now accept an `origin` address for submission of messages. The basic and incentivized channel can use this origin address for different purposes (charging fees, etc).

All the unit tests and E2E tests pass!


relayer/chain/ethereum/listener.go
```diff
@@ -117,27 +125,31 @@ func (li *Listener) pollEventsAndHeaders(
...
+			var events []*etypes.Log
+
+			filterOptions := bind.FilterOpts{Start: finalizedBlockNumber, End: &finalizedBlockNumber, Context: ctx}
 
-			for _, channelContract := range li.contracts {
-				channelEvents, err := li.queryEvents(ctx, channelContract, finalizedBlockNumber, &finalizedBlockNumber)
-				if err != nil {
-					li.log.WithError(err).Error("Failure fetching event logs")
-				}
+			basicEvents, err := li.queryBasicEvents(li.basicOutboundChannel, &filterOptions)
```
alvatar, line 132: Just want to point out that we are reading events from the two channels serially. Is there a potential scenario in which reading basic channel events could hinder the performance of the incentivized one?
Probably not a problem, or at least not yet, but it caught my attention.

relayer/chain/ethereum/listener.go
```diff
@@ -151,34 +163,54 @@ func (li *Listener) queryEvents(ctx context.Context, contract *outbound.Contract
...
+	if err != nil {
+		return nil, err
 	}
 
+	for {
+		more := iter.Next()
+		if !more {
+			err = iter.Error()
+			if err != nil {
+				return nil, err
```
alvatar, line 184: What's the reason to not return the already-processed events in case of error?

relayer/cmd/fetch_messages.go
```diff
@@ -104,75 +120,99 @@ func getEthContractEventsAndTrie(
...
-			if uint64(iter.Event.Raw.TxIndex) != index {
-				continue
+	for {
+		more := iter.Next()
+		if !more {
+			err = iter.Error()
+			if err != nil {
+				return nil, err
 			}
-			events = append(events, iter.Event)
```
alvatar, line 161: Was this a bug? (appending an event when there were none even if there was no error). As above, I ask to see if I'm getting the semantics of the function right.
BTW I think the code would be clearer if directly returns `return events, nil` instead of breaking.

alvatar (approved):

parachain/pallets/basic-channel/src/inbound/mod.rs
```diff
@@ -0,0 +1,93 @@
...
+	/// Verifier module for message verification.
+	type Verifier: Verifier;
+
+	/// Verifier module for message verification.
+	type MessageDispatch: MessageDispatch<MessageId>;
+}
+
+decl_storage! {
+	trait Store for Module<T: Config> as BasicInboundModule {
+		pub SourceChannel get(fn source_channel) config(): H160;
```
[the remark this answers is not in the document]
vgeddes: When the `config()` thing is added, the custom getter has an alternative use, and that is to provide an initial value for the storage item in the genesis configuration:

https://github.com/Snowfork/polkadot-ethereum/blob/vincent/channel-contracts-v2/parachain/src/chain_spec.rs#L116

parachain/pallets/basic-channel/src/inbound/test.rs
```diff
@@ -0,0 +1,234 @@
...
+		};
+		assert_ok!(BasicInboundChannel::submit(origin.clone(), message.clone()));
+		let nonce: u64 = Nonce::get();
+		assert_eq!(nonce, 1);
+
+		// Submit the same again
+		assert_noop!(
+			BasicInboundChannel::submit(origin.clone(), message.clone()),
+			Error::<Test>::InvalidNonce
+		);
```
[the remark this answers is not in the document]
vgeddes: That's what [assert_noop](https://crates.parity.io/frame_support/macro.assert_noop.html) does 🙂 . It checks that runtime storage hasn't been modified.

parachain/pallets/erc20-app/src/tests.rs
```diff
@@ -57,3 +57,25 @@ fn burn_should_emit_bridge_event() {
...
+fn should_not_burn_on_commitment_failure() {
+	new_tester().execute_with(|| {
+		let token_id = H160::repeat_byte(1);
+		let sender: AccountId = Keyring::Bob.into();
+		let recipient = H160::repeat_byte(9);
+
+		Assets::deposit(AssetId::Token(token_id), &sender, 500.into()).unwrap();
+
+		assert_noop!(
+			ERC20App::burn(
```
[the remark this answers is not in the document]
vgeddes: assert_noop does a check to ensure that storage hasn't been mutated.

relayer/cmd/fetch_messages.go
vgeddes: I can't see a bug here. We were never calling `append` within the `if !more` block. The patch diff is just very messy which can be misleading.

Agree on the `return events, nil` instead of breaking, however my VS code linter complains about it, so leaving it in for now.

relayer/chain/ethereum/listener.go
vgeddes: yeah is potentially an issue. Though at some point we will probably modify the relayer so that it works with only one channel, either the basic one, or the incentivized one.

relayer/chain/ethereum/listener.go
vgeddes: When relaying messages, we're following an all-or-nothing approach. The relayer either successfully relays all the messages, or it fails-fast and restarts.

This simplifies error-handling. We can't easily just forward a subset of the messages/events because the message nonce's need to match up with what the Substrate inbound channel expects.

relayer/cmd/fetch_messages.go
alvatar: The diff was misleading: I read it as this line `events = append(events, iter.Event)` right after 
```go
			err = iter.Error()
			if err != nil {
				return nil, err
			}
                        // here
```
which wasn't the case looking at line numbers.

parachain/pallets/basic-channel/src/inbound/mod.rs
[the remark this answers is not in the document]
alvatar: I see, thank you

parachain/pallets/basic-channel/src/inbound/test.rs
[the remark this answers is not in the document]
alvatar: ah cool!

ethereum/contracts/DOTApp.sol
```diff
@@ -46,21 +63,28 @@ abstract contract BaseDOTApp {
...
-        channel.submit(call);
+        channel.submit(msg.sender, call);
     }
 
     function mint(bytes32 _sender, address _recipient, uint128 _amount) public {
         // TODO: Ensure message sender is a known inbound channel
         token.mint(_recipient, wrap(_amount), abi.encodePacked(_sender));
     }
 
+    function burnFee(address _account, uint256 _amount) external returns (uint128) {
```
vgeddes, line 81: Note that this and other DOT-related changes is being updated/modified/removed in this follow-up PR: #309

ethereum/contracts/BasicOutboundChannel.sol
```diff
@@ -6,14 +6,19 @@ import "./OutboundChannel.sol";
...
+        address source,
+        uint64 nonce,
+        bytes payload
+    );
 
     /**
      * @dev Sends a message across the channel
      */
-    function submit(bytes memory payload) public override {
+    function submit(address, bytes calldata payload) external override {
```
musnit, line 21: havent seen this syntax, with unnamed parameter. what's it do? is the parameter optional?

ethereum/contracts/WrappedToken.sol
```diff
@@ -23,8 +23,8 @@ contract WrappedToken is ERC777, Ownable {
         _mint(recipient, amount, data, "");
     }
 
-    // Don't allow users to directly burn their SnowDOT via the IERC777 burn API, as it won't have
-    // the desired effect.
+    // Don't allow users to directly burn their SnowDOT via the IERC777 burn API, as it won't redeem
```
musnit, line 26: comment is DOT specific, but contract isn't. may wanna reword?

relayer/chain/ethereum/listener.go
musnit: yep @blablacio is working on a more configurable relayer atm already

ethereum/contracts/BasicOutboundChannel.sol
vgeddes: The parameter is still mandatory, and needs to be passed to submit. However in this specific implementation of submit, the parameter doesn't need a name because its not being used. Its the only way to workaround the "unused parameter" warnings emitted by the compiler

PR #309, opened 2021-03-05
DOT App V2

vgeddes:
The major change here is move the conversions between native and wrapped dot to the Substrate DOT pallet.

This has the following benefits:
* Simplifies some logic around extracting fees on the Ethereum side
* Considerably reduces gas use on the Ethereum side (was previously using around 2000 gas per conversion)
* Rust has safer/better primitives for numerical conversions

Only downside is that on the Ethereum side, if users burn tiny amounts of wrapped DOT that is outside the precision of native DOT, then that burnt amount won't be redeemable on the Substrate side. I don't think this a huge issue though. We're talking about amounts like 0.00000000001 DOT or less.

All the tests pass.


parachain/pallets/dot-app/src/primitives.rs
```diff
@@ -0,0 +1,87 @@
...
+		// largest possible value
+		let max_possible_amount = Balance::MAX;
+		let min_possible_decimals = 0;
+		assert_ne!(
+			wrap::<Test>(max_possible_amount, min_possible_decimals),
+			None
+		);
+
+		// smallest possible value
+		let min_possible_amount = Balance::MAX;
```
vgeddes, line 60: should be Balance::MIN

parachain/pallets/dot-app/src/primitives.rs
```diff
@@ -0,0 +1,87 @@
...
+		Some(value) => value,
+		None => return None,
+	};
+
+	let value_u256 = match value.checked_into::<u128>() {
+		Some(value) => U256::from(value),
+		None => return None,
+	};
+
+	match value_u256.checked_mul(granularity) {
```
vgeddes, line 34: match is redundant

parachain/pallets/dot-app/src/primitives.rs
```diff
@@ -0,0 +1,87 @@
...
+		Some(value) => value,
+		None => return None,
+	};
+
+	let unwrapped = match value.checked_div(granularity) {
+		Some(value) => value,
+		None => return None,
+	};
+
+	match unwrapped.low_u128().checked_into() {
```
vgeddes, line 17: match is redundant

blablacio (approved):
LGTM, at least the Solidity part.

parachain/pallets/dot-app/src/lib.rs
```diff
@@ -77,17 +83,32 @@ decl_module! {
...
 		#[weight = 0]
 		#[transactional]
 		pub fn lock(origin, channel_id: ChannelId, recipient: H160, amount: BalanceOf<T>) -> DispatchResult {
 			let who = ensure_signed(origin)?;
 
 			T::Currency::transfer(&who, &Self::account_id(), amount, AllowDeath)?;
 
+			let amount_wrapped = match wrap::<T>(amount, T::Decimals::get()) {
+				Some(value) => value,
+				None => panic!("Runtime is misconfigured"),
```
philipstanislaus, line 105: We should not panic here – quoting from the Substrate docs: 

> Runtime code should explicitly and gracefully handle all error cases, which is to say that runtime code must be "non-throwing", or must never "panic" to use Rust terminology. 

Can we instead return a runtime error, or panic during node initialization?

parachain/pallets/dot-app/src/lib.rs
```diff
@@ -97,15 +118,20 @@ decl_module! {
...
-		pub fn unlock(origin, sender: H160, recipient: <T::Lookup as StaticLookup>::Source, amount: BalanceOf<T>) -> DispatchResult {
+		pub fn unlock(origin, sender: H160, recipient: <T::Lookup as StaticLookup>::Source, amount: U256) -> DispatchResult {
 			let who = T::CallOrigin::ensure_origin(origin)?;
 			if who != Address::get() {
 				return Err(DispatchError::BadOrigin.into());
 			}
 
+			let amount_unwrapped = match unwrap::<T>(amount, T::Decimals::get()) {
+				Some(value) => value,
+				None => panic!("Runtime is misconfigured"),
```
philipstanislaus, line 129: Likewise

philipstanislaus (commented):
Cool, just left one comment regarding panicking. 👍

parachain/pallets/dot-app/src/lib.rs
vgeddes: Hey, yeah you're right - Will add a fix.

vgeddes: Fixed in https://github.com/Snowfork/polkadot-ethereum/pull/330

Issue #286, opened 2021-02-21
Use separate runtimes for our deployment stages


vgeddes:
Specifically because the verifier DescendantsUntilFinalized config setting needs to be different for local vs Ropsten vs Mainnet.

We need to set it to 3 for ropsten, 35 for ethereum, and 1 for E2E tests.

Rizziepit:
Some context here: the hardcoding comes from parity-bridges-common's architecture, and I do think it's the best option. The runtime performance, and thus weights + security, is affected by this value (`DescendantsUntilFinalized` directly determines how many parent headers we need to read from storage). So our weight annotation for `import_header` will look something like:
```
#[weight = <T as Trait>::WeightInfo::import_header(DescendantsUntilFinalized)]
fn import_header(
```
The weight calculation should add 0 overhead. So loading a value from storage is not an option. We also have to make sure our benchmarks account for each value of `DescendantsUntilFinalized` that we might want to use. So it's appropriate to hard code a few supported values.

If we want to get rid of feature flags and needing to re-build for e2e tests, a better option is to construct multiple runtimes like parity-bridges-common

Rizziepit:
There's probably some middle ground here between building multiple runtimes and completely dynamic values via chain spec. I'll leave that thinking to someone else :)

vgeddes:
Ahh I understand better now. Yeah then given how the value affects weight, multiple runtimes will likely be the way to go.

I'll figure out how we can structure our code to support multiple runtimes.

Rizziepit:
Heads up that https://github.com/Snowfork/polkadot-ethereum/pull/301/ introduces `DifficultyConfig` parameter which is also network-specific

PR #310, opened 2021-03-05
Multiple runtime support

vgeddes:
Added the following runtimes:
* runtime/snowbridge: Production runtime. Enable with `--features with-snowbridge-runtime`
* runtime/rococo: Staging runtime. Enable with `--features with-rococo-runtime`
* runtime/local: Runtime for local development. Enable with `--features with-local-runtime`

Closes: #286

@Rizziepit this means the `test-e2e` feature flag is no longer necessary. I've removed it and updated individual runtimes accordingly.

So for local testing we should now compile with:
```bash
cargo build --release --no-default-features --features with-local-runtime
```


parachain/Cargo.toml
```diff
@@ -75,12 +77,16 @@ frame-benchmarking-cli = { git = "https://github.com/paritytech/substrate.git",
...
 runtime-benchmarks = [
-    "artemis-runtime/runtime-benchmarks",
+    "snowbridge-runtime/runtime-benchmarks",
+    "rococo-runtime/runtime-benchmarks",
+    "local-runtime/runtime-benchmarks",
     "polkadot-service/runtime-benchmarks",
 ]
-test-e2e = ["artemis-runtime/test-e2e"]
+with-snowbridge-runtime = ["snowbridge-runtime"]
+with-karura-runtime = ["rococo-runtime"]
```
vgeddes, line 88: should be with-rococo-runtime

parachain/runtime/rococo/Cargo.toml
```diff
@@ -0,0 +1,122 @@
...
+    "sp-version/std",
+	"cumulus-primitives-core/std",
+	"cumulus-pallet-parachain-system/std",
+	"cumulus-pallet-xcm-handler/std",
+	"parachain-info/std",
+	"xcm/std",
+	"xcm-executor/std",
+	"xcm-builder/std",
+	"polkadot-parachain/std",
+    "basic-channel/std",
```
vgeddes, line 99: fix whitespace

parachain/runtime/snowbridge/Cargo.toml
```diff
@@ -0,0 +1,122 @@
...
+    "sp-api/std",
+    "sp-block-builder/std",
+    "sp-core/std",
+    "sp-inherents/std",
+    "sp-offchain/std",
+    "sp-runtime/std",
+    "sp-session/std",
+    "sp-std/std",
+    "sp-transaction-pool/std",
+    "sp-version/std",
```
vgeddes, line 90: fix whitespace

Rizziepit (approved):
This looks great to me. I just introduced a complication in https://github.com/Snowfork/polkadot-ethereum/pull/307/commits/3c03d94304ec3a4b0d707eed5b61399695881575 though. It adds a `weights` submodule to the runtime and configures pallets like this:
```
type WeightInfo = weights::verifier_lightclient_weights::WeightInfo<Runtime>;
```

Options we have:
* Add a blanket implementation of `WeightInfo` for each pallet that can be used by staging + local runtimes, i.e. `type WeightInfo = ()` (Substrate pallets already have this). Only production runtime gets a generated weights submodule.
  * Pro: easy to configure runtimes; con: extra boilerplate in each pallet, blanket implementation is insecure
* All runtimes use the same `weights` module
  * Pro: less code; con: obscures the fact that the weights are runtime-specific, and thus incorrect for 2 of the runtimes

Rizziepit:
I went in the direction of option 1 with this blanket implementation of `WeightInfo`: https://github.com/Snowfork/polkadot-ethereum/pull/307/commits/0aa225ad681142563ec5a4c8e43d99c72ed377bc. Also check out https://github.com/Snowfork/polkadot-ethereum/pull/311/

vgeddes:
Thanks Riz. That sounds good, though can we have the staging runtime have its own `weights` module too? I'd like to to treat staging as just another production network, with similar transaction fee behavior.

Rizziepit:
> Thanks Riz. That sounds good, though can we have the staging runtime have its own specific `weights` module too? I'd like to to treat staging as just another production network, with similar transaction fee behavior.

Sounds good. It should be trivial to extend https://github.com/Snowfork/polkadot-ethereum/pull/311/ to generate weights for other runtimes

parachain/README.md
```diff
@@ -34,7 +35,7 @@ Once the development environment is set up, build the parachain. This command wi
 [native](https://substrate.dev/docs/en/knowledgebase/advanced/executor#native-execution) code:
 
 ```bash
-cargo build --release
+--no-default-features --features with-local-runtime
```
philipstanislaus, line 38:
```suggestion
cargo build --release --no-default-features --features with-local-runtime
```

philipstanislaus (commented):
LGTM, though I realize that there is a lot of duplicate code now (I'd say around 1.5k LOC). 

Don't need to improve this now, but maybe there are better ways to organize those runtimes longer term.

PR #314, opened 2021-03-12
Interactive BEEFY commitment relay protocol

denalimarsh:
An in-memory database is used for managing the two-step BEEFY commitment relay protocol. State transitions:
1. BEEFY commitment is witnessed by the Substrate listener and passed to the Ethereum writer.
2. The Ethereum writer sends an InitialVerificationTransaction to the PolkadotRelayChainBridge. An Item is created in database containing the BEEFY commitment's information, status `InitialVerificationTxSent`, and the recently sent tx's hash.
3. Ethereum listener witnesses an InitialVerificationSuccessful event on the PolkadotRelayChainBridge. Fetches relevant item from database by the event's tx hash, then updates the item with status `InitialVerificationTxConfirmed` and `CloseOnBlock`.
4. Every block, the Ethereum listener checks each commitment with status `InitialVerificationTxConfirmed` and if its `CloseOnBlock` is <= the current Ethereum block then the database item is updated with status `ReadyToComplete` and a `RandomSeed` pulled from the intended completion block's hash.
5. Ethereum writer fetches the commitment from the database and uses the `RandomSeed` to select some validator addresses/proofs then sends a CompleteVerificationTransaction to the PolkadotRelayChainBridge. The item is updated in database with status `CompleteVerificationTxSent`.

TODO:
- Gracefully sync the in-memory database from any previous block.
- Re-introduce PolkadotRelayChainBridge validation checks.
- ~~Resolve edge case account nonce conflict that occurs sometimes when steps 2 and 5 overlap pending state.~~ fixed in https://github.com/Snowfork/polkadot-ethereum/pull/314/commits/ddf9ae44cf627f88ff7189bce0332388f7ea779d

This PR is currently R4R but is *not* intended to be merged to master at this time.


relayer/core/relay.go
```diff
@@ -23,8 +24,9 @@ import (
 )
 
 type Relay struct {
-	ethChain chain.Chain
-	subChain chain.Chain
+	subChain    chain.Chain
+	ethSubChain chain.Chain
+	paraChain   chain.Chain
```
musnit, line 29: this, and all the beefy stuff is actually the relayChain, not the paraChain - may want to rename/keep in mind when you refactor the copied files in this PR over to the main substrate folder

subChain is the parachain.

musnit: ideally i think we want a parachain and relaychain config, both of which use shared substrate libraries/files for things in common, and then seperate files of their own for their respective commitments/messages

relayer/chain/parachain/listener.go
```diff
@@ -0,0 +1,272 @@
...
+			err := types.DecodeFromHexString(msg.(string), signedCommitment)
+			if err != nil {
+				li.log.WithError(err).Error("Faild to decode beefy commitment messages")
+			}
+
+			received++
+			li.log.Info("Msg number:", received)
+
+			// Construct BEEFY merkle tree
+			beefyValidatorAddresses := []string{
```
musnit, line 87: are you having in issues querying the authorities from the chain via beefy.authorities? lmk if stuck

ethereum/contracts/LightClientBridge.sol
```diff
@@ -0,0 +1,272 @@
...
+     * @param validatorSignatureCommitment the signature of an arbitrary validator
+     * @param validatorPublicKeyMerkleProof Proof required for validation of the Merkle tree
+     */
+    function newSignatureCommitment(
+        bytes32 payload,
+        uint256 validatorClaimsBitfield,
+        bytes memory validatorSignatureCommitment,
+        address validatorPublicKey,
+        bytes32[] calldata validatorPublicKeyMerkleProof
+    ) public payable {
```
vgeddes, line 87: Shouldn't this be `external` instead of public? I vaguely recall `calldata` doesn't work with `public`.

ethereum/contracts/LightClientBridge.sol
```diff
@@ -0,0 +1,272 @@
...
+     * @param randomPublicKeyMerkleProofs an array of merkle proofs from the chosen validators
+     */
+    function completeSignatureCommitment(
+        uint256 id,
+        bytes32 payload,
+        bytes[] memory randomSignatureCommitments,
+        uint8[] memory randomSignatureBitfieldPositions,
+        address[] memory randomValidatorAddresses,
+        bytes32[][] memory randomPublicKeyMerkleProofs
+    ) public {
```
vgeddes, line 143: This could potentially be quite a lot more efficient if we use `calldata` instead of memory. Especially because of the variable-length arrays we're passing in.

vgeddes (commented):
Looks good so far! Though I do need to take a deeper look after the relayer work progresses further.

ethereum/contracts/LightClientBridge.sol
denalimarsh: I grabbed the contracts from https://github.com/Snowfork/verification-solidity/pull/12 which is currently a WIP, just using them as a placeholder to write the core Relayer functionality against. Once the PR is merged we'll move the finalized contracts into this repo.

ethereum/contracts/LightClientBridge.sol
denalimarsh: See https://github.com/Snowfork/polkadot-ethereum/pull/314#discussion_r598653157

ethereum/contracts/LightClientBridge.sol
```diff
@@ -0,0 +1,272 @@
+// SPDX-License-Identifier: Apache-2.0
```
musnit, line 1: can we rename this file to PolkadotRelayChainBridge to be more clear?

denalimarsh: Integrated latest contracts from https://github.com/Snowfork/verification-solidity/pull/12 in https://github.com/Snowfork/polkadot-ethereum/pull/314/commits/25329c8a44d706bcff17036902b25ffca55cd127 and renamed the contract in https://github.com/Snowfork/polkadot-ethereum/pull/314/commits/83e1d9bf40dc417d18a9fa7b9ee67fab01b52513

musnit:
@Rizziepit would appreciate a quick review from you on the relayer code too plz - ping on slack if any questions/clarifications

relayer/core/relay.go
denalimarsh: Refactored in latest

relayer/chain/ethereum/config.go
```diff
@@ -16,3 +18,8 @@ type ChannelConfig struct {
 	Inbound  string `mapstructure:"inbound"`
 	Outbound string `mapstructure:"outbound"`
 }
+
+type ContractsConfig struct {
+	PolkadotRelayChainBridge string `mapstructure:"polkadot_relay_chain_bridge"`
+	ValidatorRegistry        string `mapstructure:"validator_registry"`
```
musnit, line 24: is this really even used?

relayer/chain/ethereum/connection.go
```diff
@@ -57,3 +57,7 @@ func (co *Connection) Close() {
 func (co *Connection) GetClient() *ethclient.Client {
 	return co.client
 }
+
+func (co *Connection) GetKeyPair() *secp256k1.Keypair {
```
musnit, line 61: seems like elsewhere we just use wr.conn.kp directly - can we be consistent, ie, just remove this and use it directly, or use the GetKeyPair() function throughout the writer?

relayer/chain/ethereum/listener.go
```diff
@@ -117,23 +140,64 @@ func (li *Listener) pollEventsAndHeaders(
...
 				}
 
-				events = append(events, channelEvents...)
+				channelEvents = append(channelEvents, contractEvents...)
+			}
+			li.forwardChannelEvents(ctx, hcs, channelEvents)
+
+			// Query PolkadotRelayChainBridge contract's InitialVerificationSuccessful events
+			blockNumber := gethheader.Number.Uint64()
+			var events []*polkadotrelaychainbridge.ContractInitialVerificationSuccessful
```
musnit, line 156: events -> relayChainBridgeEvents

relayer/chain/ethereum/listener.go
```diff
@@ -117,23 +140,64 @@ func (li *Listener) pollEventsAndHeaders(
...
+
+			contractEvents, err := li.queryLightClientEvents(ctx, blockNumber, &blockNumber)
```
musnit, line 158: for both queryLightClientEvents and processLightClientEvents we should think about how they will work with multiple relayers running. for now, at a glance i think it looks fine in the listener, but want to think through further and also haven't looked at writer yet

relayer/chain/ethereum/listener.go
```diff
@@ -16,28 +18,37 @@ import (
...
+	polkadotRelayChainBridge *polkadotrelaychainbridge.Contract
+	validatorRegistry        *validatorregistry.Contract
+	messages                 chan<- []chain.Message
+	beefyMessages            chan<- store.DatabaseCmd
+	headers                  chan<- chain.Header
+	log                      *logrus.Entry
 }
 
-func NewListener(config *Config, conn *Connection, messages chan<- []chain.Message, headers chan<- chain.Header, contracts []*outbound.Contract, log *logrus.Entry) (*Listener, error) {
+func NewListener(config *Config, conn *Connection, db *store.Database, messages chan<- []chain.Message, beefyMessages chan<- store.DatabaseCmd, headers chan<- chain.Header, contracts []*outbound.Contract, log *logrus.Entry) (*Listener, error) {
```
musnit, line 42: need to think about the cold-start flow, ie, if relayer crashes in middle of something and is restarted, can it recover the state of its DB by reading from both chains? can think about/do in follow up PR/task though

relayer/chain/ethereum/listener.go
```diff
@@ -117,23 +140,64 @@ func (li *Listener) pollEventsAndHeaders(
...
+			if len(items) > 0 {
+				li.log.Info(fmt.Sprintf("Found %d InitialVerificationTxConfirmed items", len(items)))
+			}
+			for _, item := range items {
+				if item.CompleteOnBlock <= blockNumber {
+					li.log.Info("4: Updating item status from 'InitialVerificationTxConfirmed' to 'ReadyToComplete'")
+
+					// Fetch intended completion block's hash
+					blockHash := gethheader.Hash()
+					if item.CompleteOnBlock > blockNumber {
```
musnit, line 180: seems weird, first `if item.CompleteOnBlock <= blockNumber` then `if item.CompleteOnBlock > blockNumber` - does this block ever run?

relayer/chain/ethereum/listener.go
```diff
@@ -117,23 +140,64 @@ func (li *Listener) pollEventsAndHeaders(
...
+			}
+			li.processLightClientEvents(ctx, events)
+
+			// Mark items ReadyToComplete if the current block number has passed their CompleteOnBlock number
+			items := li.db.GetItemsByStatus(store.InitialVerificationTxConfirmed)
+			if len(items) > 0 {
+				li.log.Info(fmt.Sprintf("Found %d InitialVerificationTxConfirmed items", len(items)))
+			}
+			for _, item := range items {
+				if item.CompleteOnBlock <= blockNumber {
```
musnit, line 175: what if a block with CompleteOnBlock is received, but ends up being a short-term fork after 1 or 2 blocks and is lost?

maybe we should wait a few more blocks, like 6 blocks to ensure we use a deeper confirmed block?

relayer/chain/ethereum/listener.go
musnit: maybe we should filter out and ignore all light client events that don't correspond to our own node?

relayer/chain/ethereum/writer.go
```diff
@@ -134,3 +197,118 @@ func (wr *Writer) Write(ctx context.Context, msg *chain.SubstrateOutboundMessage
...
+}
+
+// WriteCompleteSignatureCommitment sends a CompleteSignatureCommitment tx to the PolkadotRelayChainBridge contract
+func (wr *Writer) WriteCompleteSignatureCommitment(ctx context.Context, item *store.BeefyItem) error {
+	beefy, err := item.ToBeefy()
+	if err != nil {
+		wr.log.WithError(err).Error("Error converting database item to Beefy")
+	}
+
+	// Select validator index based on random seed
```
musnit, line 269: seed should come from blockhash of item's completion block, though can worry about getting these parts right in follow up PR once the plumbing/goroutine code is cleaned up and merged

relayer/chain/substrate/chain.go
```diff
@@ -30,21 +32,27 @@ func NewChain(config *Config) (*Chain, error) {
...
+		conn:      NewConnection(config.Parachain.Endpoint, kpParachain.AsKeyringPair(), log),
+		relayconn: NewConnection(config.Relaychain.Endpoint, kpRelaychain.AsKeyringPair(), log),
+		listener:  nil,
+		writer:    nil,
+		log:       log,
 	}, nil
 }
 
-func (ch *Chain) SetReceiver(ethMessages <-chan []chain.Message, ethHeaders <-chan chain.Header) error {
+func (ch *Chain) SetReceiver(ethMessages <-chan []chain.Message, ethHeaders <-chan chain.Header, _ chan<- store.DatabaseCmd) error {
```
musnit, line 55: not sure having the general SetReceiver interface contain the beefy channel makes sense, especially given here its not needed/used. may be worth thinking about an alternative design/interface

relayer/chain/substrate/chain.go
```diff
@@ -53,11 +61,13 @@ func (ch *Chain) SetReceiver(ethMessages <-chan []chain.Message, ethHeaders <-ch
 	return nil
 }
 
-func (ch *Chain) SetSender(subMessages chan<- []chain.Message, _ chan<- chain.Header) error {
+func (ch *Chain) SetSender(subMessages chan<- []chain.Message, _ chan<- chain.Header, beefyMessages chan<- store.DatabaseCmd) error {
 	listener := NewListener(
 		ch.config,
 		ch.conn,
```
musnit, line 67: can rename to paraconn too

relayer/chain/substrate/listener.go
```diff
@@ -5,30 +5,38 @@ package substrate
...
+	config        *Config
+	conn          *Connection
+	relayconn     *Connection
+	messages      chan<- []chain.Message
+	beefyMessages chan<- store.DatabaseCmd
+	log           *logrus.Entry
 }
 
-func NewListener(config *Config, conn *Connection, messages chan<- []chain.Message, log *logrus.Entry) *Listener {
+func NewListener(config *Config, conn *Connection, relayconn *Connection, messages chan<- []chain.Message,
```
musnit, line 31: conn -> parachainConn

relayer/chain/substrate/listener.go
```diff
@@ -149,6 +161,122 @@ func (li *Listener) pollBlocks(ctx context.Context) error {
...
+			}
+
+			li.log.Info("1: Writing BEEFY item to database with status 'WitnessedCommitment'")
+			cmd := store.NewDatabaseCmd(&item, store.Create, nil)
+			li.beefyMessages <- cmd
+		}
+	}
+}
+
+type Authorities = [][33]uint8
```
musnit, line 227: move to another file plz

relayer/core/relay.go
```diff
@@ -52,16 +54,34 @@ func NewRelay() (*Relay, error) {
 		return nil, err
 	}
 
-	ethChain, err := ethereum.NewChain(&config.Eth)
+	// TODO: integrate with config
```
musnit, line 57: do in this PR plz as it is related to general plumbing/structure of relayer

relayer/core/relay.go
```diff
@@ -186,6 +217,8 @@ func (re *Relay) Start() {
 		log.WithError(ctx.Err()).Error("Goroutines appear deadlocked. Killing process")
 		re.ethChain.Stop()
 		re.subChain.Stop()
+		// re.database.Stop() // TODO: graceful shutdown
```
musnit, line 220: do in this PR plz as it is related to general plumbing/structure of relayer

relayer/core/relay.go
```diff
@@ -218,11 +251,20 @@ func LoadConfig() (*Config, error) {
 	}
 	config.Eth.PrivateKey = strings.TrimPrefix(value, "0x")
 
+	// TODO: auto populate contract addresses
```
musnit, line 254: do in this PR plz as it is related to general plumbing/structure of relayer

relayer/core/relay.go
```diff
@@ -218,11 +251,20 @@ func LoadConfig() (*Config, error) {
...
+	config.Eth.Contracts.ValidatorRegistry = "0xB1185EDE04202fE62D38F5db72F71e38Ff3E8305"
+	// TODO: query from 'BLOCK_WAIT_PERIOD' on RelayBridgeLightClient contract
+	config.Eth.BeefyBlockDelay = 5
+
 	value, ok = os.LookupEnv("ARTEMIS_SUBSTRATE_KEY")
 	if !ok {
 		return nil, fmt.Errorf("environment variable not set: ARTEMIS_SUBSTRATE_KEY")
 	}
-	config.Sub.PrivateKey = value
+	config.Sub.Parachain.PrivateKey = value
```
musnit, line 264: do proper config in this PR plz as it is related to general plumbing/structure of relayer

relayer/store/beefy.go
```diff
@@ -0,0 +1,141 @@
...
+type CompleteSignatureCommitmentMessage struct {
+	ID                               *big.Int
+	Payload                          [32]byte
+	RandomSignatureCommitments       [][]byte
+	RandomSignatureBitfieldPositions []*big.Int
+	RandomValidatorAddresses         []common.Address
+	RandomPublicKeyMerkleProofs      [][][32]byte
+}
+
+type Beefy struct {
```
musnit, line 32: I think this is a BeefyWitness right?

relayer/store/beefy.go
```diff
@@ -0,0 +1,141 @@
...
+		ValidatorSignatureCommitment:  sig0Commitment,
+		ValidatorPublicKey:            b.ValidatorAddresses[valAddrIndex],
+		ValidatorPosition:             big.NewInt(int64(valAddrIndex)),
+		ValidatorPublicKeyMerkleProof: sig0ProofContents,
+	}
+
+	return msg, nil
+}
+
+func (b *Beefy) GenerateMmrProofOffchain(valAddrIndex int) ([][32]byte, error) {
```
musnit, line 70: GenerateMerkleProofOffchain not mmr

relayer/store/beefy.go
```diff
@@ -0,0 +1,141 @@
...
+		beefyTreeData[i] = buf
+	}
+
+	// Create the tree
+	beefyMerkleTree, err := merkletree.New(beefyTreeData)
+	if err != nil {
+		return [][32]byte{}, err
+	}
+
+	// Generate MMR proof for validator at index valAddrIndex
```
musnit, line 88: mmr -> merkle

relayer/store/store.go
```diff
@@ -0,0 +1,190 @@
...
+
+const (
+	CommitmentWitnessed            Status = iota // 0
+	InitialVerificationTxSent      Status = iota // 1
+	InitialVerificationTxConfirmed Status = iota // 2
+	ReadyToComplete                Status = iota // 3
+	CompleteVerificationTxSent     Status = iota // 4
+)
+
+type BeefyItem struct {
```
musnit, line 26: BeefyItem -> BeefyWitnessRelay ? or any better name suggestions?

relayer/store/store.go
```diff
@@ -0,0 +1,190 @@
...
+	var beefyItem BeefyItem
+	if !db.HasTable(&beefyItem) {
+		db.CreateTable(&beefyItem)
+		db.Model(&beefyItem)
+	}
+}
+
+func (d *Database) onDone(ctx context.Context) error {
+	d.log.Info("Shutting down database...")
+	// close(d.messages) TODO: properly close channels
```
musnit, line 134: do in this pr plz

relayer/chain/ethereum/writer.go
```diff
@@ -80,7 +115,35 @@ func (wr *Writer) writeLoop(ctx context.Context) error {
...
+	}
+}
+
+func (wr *Writer) writeBeefyLoop(ctx context.Context) error {
+	ticker := time.NewTicker(5 * time.Second)
+	for {
+		select {
+		case <-ctx.Done():
+			return wr.onDone(ctx)
+		case <-ticker.C:
```
musnit, line 133: still haven't got to the substrate reader code yet, but it looks like you're doing something like running a 5 second ticker to check for new CommitmentWitnessed events from the db, rather than reacting instantly to them as they come in?

can you explain the actual flow for the substrate reader -> db -> ethereum writer piece and your thinking behind it?

relayer/chain/ethereum/writer.go
```diff
@@ -80,7 +115,35 @@ func (wr *Writer) writeLoop(ctx context.Context) error {
...
+				if err != nil {
+					wr.log.WithError(err).Error("Error submitting message to ethereum")
+				}
+			}
+		}
+	}
+}
+
+func (wr *Writer) writeBeefyLoop(ctx context.Context) error {
+	ticker := time.NewTicker(5 * time.Second)
```
philipstanislaus, line 128: Can we put the seconds into a config variable?

relayer/chain/ethereum/writer.go
```diff
@@ -134,3 +197,118 @@ func (wr *Writer) Write(ctx context.Context, msg *chain.SubstrateOutboundMessage
...
+		msg.ValidatorClaimsBitfield, msg.ValidatorSignatureCommitment,
+		msg.ValidatorPosition, msg.ValidatorPublicKey, msg.ValidatorPublicKeyMerkleProof)
+	if err != nil {
+		wr.log.WithError(err).Error("Failed to submit transaction")
+		return err
+	}
+
+	wr.log.WithFields(logrus.Fields{
+		"txHash":      tx.Hash().Hex(),
+		"blockNumber": blockNumber,
```
philipstanislaus, line 247: This blockNumber will not be from the block which includes the tx, which might be misleading. What is the intention here?

relayer/chain/ethereum/writer_test.go
```diff
@@ -0,0 +1,206 @@
...
+		time.Sleep(5 * time.Second)
+	}
+
+	// Fetch the tx receipt
+	receipt, err := econn.GetClient().TransactionReceipt(ctx, txHash)
+	if err != nil {
+		log.Fatal(err)
+	}
+	assert.Equal(t, receipt.TxHash, txHash)
+	assert.Equal(t, receipt.Status, uint64(1)) // Successful tx status
```
philipstanislaus, line 142: Nice, let's add another test case for completion, but fine to put that into the next PR.

philipstanislaus (commented):
LGTM, thanks! 

I feel we should add a few e2e tests to this whole logic – but fine to put them into a follow-up PR.

relayer/chain/substrate/chain.go
```diff
@@ -14,14 +14,16 @@ import (
...
-	config   *Config
-	listener *Listener
-	writer   *Writer
-	conn     *Connection
-	log      *logrus.Entry
+	config    *Config
+	listener  *Listener
+	writer    *Writer
+	conn      *Connection
+	relayconn *Connection
```
vgeddes, line 25: In a previous version of this work, I vaguely recall you had the relay chain stuff in a separate chain abstraction called `relayer/chain/parachain`. While not correctly named, it was still the right approach. Was there a reason you moved away from that?

I don't think we should go down this path of making the substrate chain module handle both the parachain and relaychain. This wasn't really how I designed the chain abstractions to work, and adds extra complexity to a module which is already quite complicated and brittle as it is. 

Instead I suggest the following module layout for the chains involved.
* relayer/chain/ethereum
* relayer/chain/parachain
* relayer/chain/relaychain
 
There will some duplicated code, but we can improve that post-launch.

vgeddes (commented):
Overall this looks very good! Thanks Denali, this is quite a chunk of work. 🙂🚀

I do have one concern - that is the use of the `chain/substrate` module to manage both the parachain and relaychain communications. See my inline comments for more detail.

ethereum/contracts/PolkadotRelayChainBridge.sol
```diff
@@ -0,0 +1,296 @@
...
+        /**
+         * @dev Generate an array of numbers
+         */
+        // uint8[] memory randomNumbers = getRandomNumbers(data, requiredNumberOfSamples);
+
+        /**
+         *  @dev For each randomSignature, do:
+         */
+        for (uint256 i = 0; i < requiredNumberOfSamples; i++) {
+            // @note Require random numbers generated onchain match random numbers
```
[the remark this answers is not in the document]
denalimarsh: The plan is to address the contract changes in a follow-on PR that integrates the latest Solidity changes from https://github.com/Snowfork/verification-solidity/pull/16. The contracts in this PR were pulled from prior versions and are essentially mocks to test the Relayer against.

ethereum/contracts/PolkadotRelayChainBridge.sol
```diff
@@ -0,0 +1,296 @@
...
+     * @dev Note that `blockhash(blockNum)` will only work for the 256 most recent blocks
+     * @dev Each generated number must be less than MAXIMUM_NUM_SIGNERS
+     * @param data a storage reference to the validationData struct
+     * @return onChainRandNums an array storing the random numbers generated inside this function
+     */
+    function getRandomNumbers(
+        ValidationData storage data,
+        uint8 requiredNumberOfSamples
+    ) private view returns (uint8[] memory onChainRandNums) {
+        // // @note Get payload.blocknumber, add BLOCK_WAIT_PERIOD
```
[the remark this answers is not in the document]
denalimarsh: See https://github.com/Snowfork/polkadot-ethereum/pull/314#discussion_r606707082

relayer/chain/substrate/chain.go
denalimarsh: Originally the third chain abstraction was primarily used as a dev environment of sorts as I wanted to get everything working before prematurely refactoring the existing code. It hosted all the new ethereum functionality as well, which needed to be refactored into the ethereum reader/writer. As part of that refactor I also moved the substrate-compatible portions into the substrate chain abstraction.

I agree with your suggestion, each chain should have its own independent abstraction. I'll make the suggested changes. It's worth noting that the `relayer/chain/relaychain` abstraction wouldn't need a writer, as we only read information from the relay chain via subscriptions and queries.

ethereum/contracts/PolkadotRelayChainBridge.sol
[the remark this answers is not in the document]
musnit: yeh i think we can do a cleanup PR after - so long as code from this PR actually works properly and is fit for staging, some mess that we cleanup straight after is fine

relayer/chain/substrate/chain.go
musnit: yeh i was imagining the substrate abstraction to be more like a library that is used by separate relaychain/parachain abstractions rather than a module that handles both together

relayer/chain/ethereum/config.go
denalimarsh: Not used for now. The production version could/should check that the validator address passes `ValidatorRegistry`'s validation before making the call initial verification call to `PolkadotRelayChainBridge`. But can remove it from the config for now until we bring validation back into the contracts.

relayer/chain/ethereum/connection.go
denalimarsh: This is orphaned code that was previously used in a test file. Will remove.

relayer/chain/ethereum/config.go
musnit: yeh - tbh i dont even know if we need this for production version

relayer/chain/ethereum/listener.go
denalimarsh: Yeah second conditional should be `if item.CompleteOnBlock < blockNumber`.

relayer/chain/substrate/chain.go
denalimarsh: I went with @vgeddes's suggestion in https://github.com/Snowfork/polkadot-ethereum/pull/314/commits/b428332f96d28dd075dd6dee02c99b9f1b554f5d, https://github.com/Snowfork/polkadot-ethereum/pull/314/commits/9251d5b526bf056a4094d3f29f06e9dec2d0bfa6. The relaychain actually ended up being quite small once all the parachain-specific logic was removed. The `Chain` struct for the two chains is similar (only difference is that relaychain doesn't have a writer) so it seems sensible to base both abstractions off a shared Substrate chain struct. I'll come back to @musnit's suggestion once some of the other revisions are addressed.

relayer/chain/ethereum/config.go
denalimarsh: Removed in https://github.com/Snowfork/polkadot-ethereum/pull/314/commits/e380f1392bab16b837d7bb5da653103eede3b5b2

relayer/chain/ethereum/connection.go
denalimarsh: Removed in https://github.com/Snowfork/polkadot-ethereum/pull/314/commits/bc765754f8d0e0df068a2b39d0e6bf1225b6eb18

relayer/chain/ethereum/listener.go
denalimarsh: Updated in https://github.com/Snowfork/polkadot-ethereum/pull/314/commits/bc765754f8d0e0df068a2b39d0e6bf1225b6eb18

relayer/chain/ethereum/listener.go
denalimarsh: Updated in https://github.com/Snowfork/polkadot-ethereum/pull/314/commits/bc765754f8d0e0df068a2b39d0e6bf1225b6eb18

relayer/chain/substrate/chain.go
denalimarsh: Agreed. The `Chain` interface is currently outdated. The Ethereum, Relaychain, and Parachain writers/readers use different selections of the following channels:
- subMessages/ethMessages <-chan []chain.Message
- ethHeaders chan<- chain.Header
- beefyMessages chan<- store.DatabaseCmd

Could we just deprecate the `SetReceiver`/`SetSender` methods on the `Chain` interface? Then each abstraction can pass only the required args it needs. @vgeddes thoughts?

relayer/chain/substrate/listener.go
denalimarsh: Refactoring to separate relaychain/parachain abstractions left just one connection per abstraction, so I left this as `conn` in https://github.com/Snowfork/polkadot-ethereum/commit/b428332f96d28dd075dd6dee02c99b9f1b554f5d.

relayer/chain/substrate/chain.go
denalimarsh: See https://github.com/Snowfork/polkadot-ethereum/pull/314#discussion_r607245280

relayer/store/beefy.go
denalimarsh: Whoops, updated in https://github.com/Snowfork/polkadot-ethereum/pull/314/commits/9251d5b526bf056a4094d3f29f06e9dec2d0bfa6

relayer/store/beefy.go
denalimarsh: Updated in https://github.com/Snowfork/polkadot-ethereum/pull/314/commits/9251d5b526bf056a4094d3f29f06e9dec2d0bfa6

relayer/core/relay.go
denalimarsh: Updated in https://github.com/Snowfork/polkadot-ethereum/pull/314/commits/ec27500ba0c88d0268ac6faaa85f59f2ab3dbffd

relayer/core/relay.go
denalimarsh: Updated in https://github.com/Snowfork/polkadot-ethereum/pull/314/commits/ec27500ba0c88d0268ac6faaa85f59f2ab3dbffd

relayer/chain/ethereum/writer_test.go
```diff
@@ -0,0 +1,206 @@
...
+)
+
+func TestWriter(t *testing.T) {
+	// ------------------------- Set up Database -------------------------
+	configJson := `
+	{"db_config": {
+			"dialect": "sqlite3",
+			"db_path": "./tmp.db"
+		}
+	}`
```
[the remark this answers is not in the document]
denalimarsh: Migrated this over to the existing configuration in https://github.com/Snowfork/polkadot-ethereum/pull/314/commits/ec27500ba0c88d0268ac6faaa85f59f2ab3dbffd

relayer/store/store_test.go
```diff
@@ -0,0 +1,158 @@
...
+	"github.com/snowfork/polkadot-ethereum/relayer/store"
+)
+
+func TestStore(t *testing.T) {
+	configJson := `
+	{"db_config": {
+			"dialect": "sqlite3",
+			"db_path": "./tmp.db"
+		}
+	}`
```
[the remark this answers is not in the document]
denalimarsh: Updated in https://github.com/Snowfork/polkadot-ethereum/pull/314/commits/ec27500ba0c88d0268ac6faaa85f59f2ab3dbffd

relayer/core/relay.go
denalimarsh: https://github.com/Snowfork/polkadot-ethereum/commit/ec27500ba0c88d0268ac6faaa85f59f2ab3dbffd

relayer/core/relay.go
denalimarsh: https://github.com/Snowfork/polkadot-ethereum/pull/314/commits/f03c815ec36c601637c07e7ca1695ce85ab6e85a, https://github.com/Snowfork/polkadot-ethereum/pull/314/commits/4a14bcb4c831ad4ecb1aebad47afc2cce0ec6339

relayer/store/store.go
denalimarsh: Updated relaychain listener to shut down the beefy channel used by the database write loop in https://github.com/Snowfork/polkadot-ethereum/pull/314/commits/f03c815ec36c601637c07e7ca1695ce85ab6e85a.

relayer/chain/ethereum/writer.go
denalimarsh: Good catch. Removed `blockNumber` from the log in https://github.com/Snowfork/polkadot-ethereum/pull/314/commits/b40cf71eb5ce4c11b59406eab49421bb2245d5d3

relayer/store/store.go
denalimarsh: BeefyRelayInfo?

relayer/store/beefy.go
denalimarsh: Yea it's a witnessed Beefy justification that includes the signed commitment + authority addresses. `BeefyWitness` sounds like it refers to the party running the relayer that 'witnessed' the Beefy justification. How about `WitnessedBeefyJustification` or `BeefyJustification`?

relayer/store/store.go
denalimarsh: It's a database item that contains the information required throughout the relay lifecycle:
- Marshaled original justification contents (ValidatorAddresses, SignedCommitment)
- Lifecycle info i.e. Status, Tx hashes, Seed, etc.

relayer/chain/ethereum/listener.go
denalimarsh: Such that we only attempt to send the CompleteVerification tx if our relayer's Ethereum address was the sender of the CreateVerification tx? Makes sense as we're already requiring that the `msg.sender` is the same in the contract's `completeSignatureCommitment` method.

Should be fine for happy path, but later on will need to allow other nodes to submit the CompleteVerification tx after some time if the original CreateVerification tx sender doesn't complete it.

relayer/chain/ethereum/listener.go
denalimarsh: Yea called this one out in the PR description, it's something I've been thinking about as well. It should be possible to sync from an arbitrary previous block number and properly build up DB state. Can't go through the same sequence of tx sending/checks as we do in live-mode though, so will need some sort of db syncing mode.

relayer/chain/ethereum/listener.go
musnit: yep. and yeh though in V1 the expectation is that other validators will do their own initial+complete tx's rather than take over. after stuff is working, we'll implement some kinda lock/stake or similar mutex to avoid redundant txs

relayer/chain/ethereum/listener.go
musnit: yeh - can u add an issue for it?

relayer/store/beefy.go
denalimarsh: https://github.com/Snowfork/polkadot-ethereum/pull/314/commits/05ed130b01538ab8198d39a72e0e594352c4ae24

relayer/store/store.go
denalimarsh: https://github.com/Snowfork/polkadot-ethereum/pull/314/commits/05ed130b01538ab8198d39a72e0e594352c4ae24

relayer/core/relay.go
```diff
@@ -129,12 +150,22 @@ func (re *Relay) Start() {
...
+		}).Error("Failed to start database")
+		return
+	}
+	log.WithField("database", "Beefy").Info("Started database")
+
 	// Short-lived channels that communicate initialization parameters
 	// between the two chains. The chains close them after startup.
-	ethInit := make(chan chain.Init, 1)
 	subInit := make(chan chain.Init, 1)
+	ethSubInit := make(chan chain.Init, 1)
```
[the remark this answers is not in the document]
denalimarsh: This one is actually legacy code, but updated it to be a non-buffered channel in https://github.com/Snowfork/polkadot-ethereum/pull/314/commits/4e218a923d74e9008188dea6914b4b2ad393c1f6

relayer/core/relay.go
```diff
@@ -52,16 +54,34 @@ func NewRelay() (*Relay, error) {
...
+		}
+	}`
+	dbConfig := store.ParseConfigFromJson(configJson)
+
+	db, err := store.PrepareDatabase(dbConfig)
 	if err != nil {
 		return nil, err
 	}
 
+	beefyMessages := make(chan store.DatabaseCmd, 1)
```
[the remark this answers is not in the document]
denalimarsh: Updated in https://github.com/Snowfork/polkadot-ethereum/pull/314/commits/4e218a923d74e9008188dea6914b4b2ad393c1f6

relayer/chain/substrate/chain.go
vgeddes: Great, thanks!

relayer/chain/ethereum/listener.go
denalimarsh: Added a check so we only process events from our node in https://github.com/Snowfork/polkadot-ethereum/pull/314/commits/c66da2228c36b070ca5b2db081936f342c37dc35

relayer/chain/substrate/listener.go
denalimarsh: Updated in https://github.com/Snowfork/polkadot-ethereum/pull/314/commits/179de7d5ab82396d4d1a473b7b2a1ff4f5e9c302

relayer/store/beefy.go
```diff
@@ -0,0 +1,141 @@
...
+	ValidatorPublicKeyMerkleProof [][32]byte
+}
+
+type CompleteSignatureCommitmentMessage struct {
+	ID                               *big.Int
+	Payload                          [32]byte
+	RandomSignatureCommitments       [][]byte
+	RandomSignatureBitfieldPositions []*big.Int
+	RandomValidatorAddresses         []common.Address
+	RandomPublicKeyMerkleProofs      [][][32]byte
```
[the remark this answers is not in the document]
denalimarsh: Updated contracts to latest in https://github.com/Snowfork/polkadot-ethereum/pull/314/commits/79220513a46d60de6598935a6b1bda1ca3215d80

ethereum/contracts/PolkadotRelayChainBridge.sol
```diff
@@ -0,0 +1,296 @@
...
+import "@openzeppelin/contracts/math/SafeMath.sol";
+import "@openzeppelin/contracts/cryptography/ECDSA.sol";
+import "./utils/Bits.sol";
+import "./utils/Bitfield.sol";
+import "./ValidatorRegistry.sol";
+
+/**
+ * @title A entry contract for the Ethereum light client
+ */
+contract PolkadotRelayChainBridge {
```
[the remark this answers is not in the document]
denalimarsh: Updated contracts to latest (includes rename of the contract to `LightClientBridge`) in https://github.com/Snowfork/polkadot-ethereum/pull/314/commits/79220513a46d60de6598935a6b1bda1ca3215d80. We can sort out the best name for the primary relay bridge contract at other time.

ethereum/contracts/PolkadotRelayChainBridge.sol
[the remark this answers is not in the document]
denalimarsh: Updated contracts to latest in https://github.com/Snowfork/polkadot-ethereum/pull/314/commits/79220513a46d60de6598935a6b1bda1ca3215d80 and then commented out contract verification checks in https://github.com/Snowfork/polkadot-ethereum/pull/314/commits/c55ba8d602d6122c4d8f2f0c185d5a6513fce8c7 that would cause the end-to-end process to fail.

Uncommenting the verification checks will cause the end-to-end BEEFY relay to fail at step #3 because the initial verification event won't be emitted. The idea is that once this PR is merged we immediately integrate the verification checks back in one by one.

relayer/chain/ethereum/writer.go
denalimarsh: Removed in https://github.com/Snowfork/polkadot-ethereum/pull/314/commits/ed14d1307e70ccefd10c98516684c226ab613ec2

relayer/chain/ethereum/listener.go
denalimarsh: Went with @philipstanislaus's suggestion in https://github.com/Snowfork/polkadot-ethereum/pull/314/commits/bf5e0950f7bf09a85aafd3c34bd1f613052a5fc6 - was easy to add it in since the `decendantsUntilFinal` variable was already available within scope.

relayer/chain/ethereum/writer.go
denalimarsh: The database is used as the intermediary for every step of the BEEFY justification relay lifecycle (see PR description) for consistency. The database's primary utility is storing the BEEFY justification's `ValidatorAddresses` and `SignedCommitement` so we can properly build the CompleteSignatureCommitment.

We could favor direct communication between the services and have the Relaychain listener forward the witnessed BEEFY justification directly to the Ethereum writer by introducing another channel to both services. To remove the 5-second cron job entirely, the Ethereum listener checks the database on each new block for any completable items, then forwards them directly to the Ethereum writer.

philipstanislaus: That would still involve a database though, right?

denalimarsh: @philipstanislaus yes, a database is still required in order to store the BEEFY justification so we can build the CompleteSignatureCommitment in the final step. Is it possible to use the `currentId` from `InitialVerificationSuccessful` event to query the LightClientBridge's `ValidationData` and use that to build the CompleteSignatureCommitment instead?

For now, I refactored the Ethereum writer according to the approach described above in https://github.com/Snowfork/polkadot-ethereum/pull/314/commits/cb1c68c0d84516ae5c1ee6344cafc44794a4ea35 and removed the writer's 5-second cron job entirely by having the Ethereum listener fetch completable items and forward them to the writer as well.

relayer/chain/ethereum/writer.go
denalimarsh: Removed this cron job entirely in https://github.com/Snowfork/polkadot-ethereum/pull/314/commits/cb1c68c0d84516ae5c1ee6344cafc44794a4ea35

test/scripts/start-services.sh
```diff
@@ -110,21 +110,21 @@ trap 'kill $(jobs -p)' SIGINT SIGTERM EXIT
 start_ganache
 deploy_contracts
 start_parachain
-start_relayer
+# start_relayer
```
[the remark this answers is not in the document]
denalimarsh: Added it back in https://github.com/Snowfork/polkadot-ethereum/pull/314/commits/8924d7734f763b26467884ac91a9dec885400e8e. Can start everything as before with `./scripts/start-services.sh`.

relayer/chain/parachain/listener.go
denalimarsh: Included this as part of issue https://github.com/Snowfork/polkadot-ethereum/issues/341

relayer/chain/ethereum/writer.go
denalimarsh: Included this as part of issue https://github.com/Snowfork/polkadot-ethereum/issues/340

relayer/chain/substrate/chain.go
denalimarsh: Included this as part of issue https://github.com/Snowfork/polkadot-ethereum/issues/340

relayer/chain/ethereum/listener.go
denalimarsh: Included this in issue https://github.com/Snowfork/polkadot-ethereum/issues/340

denalimarsh:
This is ready for another round of reviews. I've created two issues:
1. Relayer architecture and optimizations: https://github.com/Snowfork/polkadot-ethereum/issues/340
2. TODOs for proper verification of BEEFY commitments: https://github.com/Snowfork/polkadot-ethereum/issues/341

I've been having difficulty properly shutting down all goroutines, the final outstanding bug before this can be merged is that ~50% of the time killing then restarting the Relayer leads to the goroutines being deadlocked. Would appreciate it if the original author of the `SIGINT, SIGTERM` termination in `core/relay.go` could advise on the proper solution, thanks!

musnit:
> This is ready for another round of reviews. I've created two issues:
> 
> 1. Relayer architecture and optimizations: #340
> 2. TODOs for proper verification of BEEFY commitments: #341
> 
> I've been having difficulty properly shutting down all goroutines, the final outstanding bug before this can be merged is that ~50% of the time killing then restarting the Relayer leads to the goroutines being deadlocked. Would appreciate it if the original author of the `SIGINT, SIGTERM` termination in `core/relay.go` could advise on the proper solution, thanks!

cc @Rizziepit

musnit:
@denalimarsh can you branch this off master too and merge in master?

Rizziepit (commented):
This is a lot of code :D Looking good. I mainly reviewed the relayer code. Since I don't know much about the beefy protocol, I can't speak to the correctness of state changes, etc. I'm fine with this being merged though. Then we can iterate on the architecture from there. Please have a look at the notes below - they're for follow-up PRs. @vgeddes should take a look too.

About the `Chain` interface: it does need an update, no doubt. I see three issues that will affect the final interface though:
* I'm not sure the `relaychain` should implement the `Chain` interface... it might be better as a sub-component of the parachain. Have a look at the chain/ethereum/syncer/syncer.go module and how it's used by the Eth listener. At a high-level, the beefy listener appears to have a similar role as the Eth header listener/syncer. Do you think that design could work or am I missing something @denalimarsh?
* The `Chain.SetSender` and `Chain.SetReceiver` should accept a single channel arg. You can blame me for the current situation. The channel should combine messages / beefy stuff / headers. The fact that we have multiple channels right now implies that items in one channel are independent of items in another channel, i.e. it doesn't matter which channel's item is written first. But that's incorrect. I'll update it this week.
* The DB writes probably shouldn't use a channel. It seems that the behaviour we want is for `databaseMessages <- cmd` to block until the query has executed. A non-blocking `databaseMessages` channel would be problematic due to a read/write race condition, right? It seems safer and simpler to directly call `database.execute(cmd)` instead of sending to a channel.

With regards to deadlocks... This PR handles the `ctx.Done()` signal and calls `chain.Stop` correctly. Check out https://github.com/Snowfork/polkadot-ethereum/pull/232. It's unfortunately expected that the relayer deadlocks a lot on shutdown - particularly if it's busy syncing old Eth headers. As long as the deadlock detection works and kills the process this is ok. Is this causing other problems?

vgeddes (approved):

musnit:
merged in with https://github.com/Snowfork/polkadot-ethereum/pull/343 - fixes will be made in upcoming PRs including your comments @Rizziepit

PR #317, opened 2021-03-16
Benchmarks for dot-, erc20- and eth-app pallets

Rizziepit:
This adds benchmarks for the 3 app pallets (dot, erc20 and eth). It also generates implementations of the pallets' `WeightInfo` traits, and integrates the generated weights into the snowbridge and rococo runtimes.


parachain/pallets/dot-app/src/benchmarking.rs
```diff
@@ -0,0 +1,97 @@
...
+        
+        let balance = existential_deposit * 10u32.into();
+        // The balance remaining after unlock will be < existential deposit.
+        // This shouldn't trigger account reaping because we're using
+        // the 'KeepAlive' flag. But just in case that ever changes...
+        let amount = existential_deposit * 9u32.into() + 1u32.into();
+        let amount_wrapped = wrap::<T>(amount, T::Decimals::get()).unwrap();
+        
+        T::Currency::make_free_balance_be(&lock_account, balance);
+	}: _(artemis_dispatch::Origin(caller).into(), sender, recipient_lookup, amount_wrapped)
```
Rizziepit, line 85: @vgeddes I'm having a hard time figuring out how to construct the origin for `unlock`. Can you see an obvious fix? This construction has the following error:
```
cannot infer type for type parameter `T` declared on the trait `Into`
```

I'll go dig deeper if you don't see an immediate issue. Ignore wonky formatting :)

vgeddes: Hey I did some research / experiments and came up with https://github.com/Snowfork/polkadot-ethereum/pull/318 🙂

parachain/pallets/dot-app/Cargo.toml
```diff
@@ -24,6 +25,7 @@ sp-runtime = { git = "https://github.com/paritytech/substrate.git", branch = "ro
 ethabi = { git = "https://github.com/Snowfork/ethabi-decode.git", package = "ethabi-decode", branch = "master", default-features = false }
 
 artemis-core = { path = "../../primitives/core", default-features = false }
+artemis-dispatch = { path = "../../pallets/dispatch", optional = true }
```
vgeddes, line 28: Needs a `default-features = false` here I think.

parachain/pallets/dot-app/src/benchmarking.rs
```diff
@@ -0,0 +1,97 @@
...
+        let recipient: T::AccountId = account("recipient", 0, 0);
+        let recipient_lookup: <T::Lookup as StaticLookup>::Source = T::Lookup::unlookup(recipient.clone());
+        let sender = H160::zero();
+        
+        let balance = existential_deposit * 10u32.into();
+        // The balance remaining after unlock will be < existential deposit.
+        // This shouldn't trigger account reaping because we're using
+        // the 'KeepAlive' flag. But just in case that ever changes...
+        let amount = existential_deposit * 9u32.into() + 1u32.into();
+        let amount_wrapped = wrap::<T>(amount, T::Decimals::get()).unwrap();
```
vgeddes, line 82: Thought I'd add some more context around this "wrapping/unwrapping" business. My documentation for the DOT pallet was pretty non-existent, sorry about that.

Basically on Substrate, native DOT is represented using u128, with 10 decimal places. Our Wrapped DOT token living on Ethereum however is represented using u256 and has 18 decimal places. The 18 decimal places is a convenience to match ETH and most ERC20 tokens which also have 18 decimal places. So the task of `wrap()` is convert a native DOT to 18 decimal places with some arithmetic conversions.

parachain/pallets/eth-app/src/lib.rs
```diff
@@ -32,12 +33,25 @@ use artemis_core::{ChannelId, SingleAsset, OutboundRouter};
...
+mod benchmarking;
+
 #[cfg(test)]
 mod mock;
 
 #[cfg(test)]
 mod tests;
 
+/// Weight functions needed for this pallet.
+pub trait WeightInfo {
```
vgeddes, line 45: I think we're supposed to dynamically generate this trait definition and its implementation for `()`. I looked at two other big projects (the Polkadot relay chain, and Acala), and they are doing roughly this for each module:

1. Calling the `benchmark` command with the `--template` parameter to generate `pallets/{module}/src/weights.rs`. The generated code includes the trait definition and its implementation for `()`.
2. Calling `benchmark` again, this time without a `--template` parameter, to generate `runtime/{runtime}/src/weights/{module}.rs`

After poking around, I found these templates used by Acala and Substrate respectively.
* https://github.com/AcalaNetwork/Acala/blob/master/templates/module-weight-template.hbs
* https://github.com/paritytech/substrate/blob/master/.maintain/frame-weight-template.hbs

And an example `weights.rs` generated for an Acala module:
https://github.com/AcalaNetwork/Acala/blob/master/modules/dex/src/weights.rs

vgeddes: Anway, can leave this for a future PR.

vgeddes (approved):

parachain/pallets/eth-app/src/lib.rs
Rizziepit: It looks like they're using the same template as Substrate itself: https://github.com/paritytech/substrate/blob/master/.maintain/frame-weight-template.hbs

I considered doing the same, but decided against it because I don't think the Substrate approach serves us well. As you can see at https://github.com/AcalaNetwork/Acala/blob/master/modules/dex/src/weights.rs, the template generates weights as a submodule of the pallet. This obscures the fact that weights are runtime-specific. I think there might be a temptation for a dev to use the pallet weights as-is instead of regenerating them for each runtime.

I also don't think generating the `WeightInfo` trait is a great idea. The weight(s) associated with an extrinsic are closely tied to the inner workings of the extrinsic. So defining the interface, i.e. `WeightInfo`, that describes these inner workings in a separate module feels wrong. The benchmarks should fulfil the interface but not define the interface themselves.

Pedantic, misguided or reasonable? Interested in your thoughts @vgeddes

vgeddes: Hey, yeah, I think that's reasonable and makes sense - I did find it odd that they were generating weights alongside the source code for the pallet, which is supposed to be runtime-agnostic. And having to run the benchmark command twice is a bit cumbersome too. I think the benchmarking framework is still a WIP while they iterate towards a holistic solution.

philipstanislaus (commented):
LGTM 🎉

PR #324, opened 2021-03-24
Fix for invalid transaction signing

vgeddes:
This should fix the issue where the relayer self-terminates upon receiving an "Invalid Transaction" error from the Substrate transaction pool.

In my investigation, I saw the following error in the parachain logs, which has this [explanation](https://github.com/paritytech/substrate/blob/9b08105b8c7106d723c4f470304ad9e2868569d9/primitives/runtime/src/transaction_validity.rs#L46):
```
Transaction pool error: Invalid transaction validity: InvalidTransaction::BadProof
```

Looking through our transaction signing logic in the relayer, I think I found the root cause: https://github.com/Snowfork/polkadot-ethereum/blob/af643f51cc3403fe170da68d92bebde40e244068/relayer/chain/substrate/writer.go#L147

The `BlockHash` field needs to refer to the block hash used to compute the transaction era. But that is not the case in our code, as we request the latest block hash again. After this signed transaction is submitted to the pool, its signature verification will fail if `BlockHash` doesn't match the hash used to generate the era.

Other changes:
* Used `GetFinalizedHead` instead of `GetBlockLatest` to ensure we're using latest head from the canonical chain.
* Cache `genesisHash` so we don't need to fetch it every time.

### DIFF
--- a/relayer/chain/substrate/writer.go
+++ b/relayer/chain/substrate/writer.go
@@ -17,12 +17,13 @@ import (
 )
 
 type Writer struct {
-	conn     *Connection
-	messages <-chan []chain.Message
-	headers  <-chan chain.Header
-	log      *logrus.Entry
-	nonce    uint32
-	pool     *extrinsicPool
+	conn        *Connection
+	messages    <-chan []chain.Message
+	headers     <-chan chain.Header
+	log         *logrus.Entry
+	nonce       uint32
+	pool        *extrinsicPool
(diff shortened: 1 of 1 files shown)


musnit:
ah nice catch!

Rizziepit (approved):
Very nice! I'm really glad to see this fix

philipstanislaus (commented):
Great catch, thanks 👍

PR #325, opened 2021-03-26
Benchmarks for basic channel

Rizziepit:
This PR add benchmarks for the basic inbound channel's `submit` call. The incentivized channel will have the exact same benchmarks, but I didn't add them here to keep the PR small(ish). More importantly, it adds stuff to our core APIs / app pallets to facilitate benchmarking:
* `MessageDispatch` gets `successful_dispatch_event` function so that I can verify that a dispatch was successful by inspecting the most recent event.
* `Verifier` gets `initialize_storage` which is used by both benchmarks and genesis code to set up the initial header chain.
* `dot_app.config` gets `initial_module_balance` so that it's possible to set a non-zero balance for the module account via the chain spec


parachain/pallets/basic-channel/src/inbound/benchmarking.rs
```diff
@@ -0,0 +1,255 @@
...
+// should be used in the `submit` benchmark.
+//
+// We rely on configuration via chain spec of the app pallets because
+// we don't have access to their storage here.
+benchmarks! {
+	// Benchmark `submit` extrinsic under worst case conditions:
+	// * `submit` dispatches the DotApp::unlock call
+	// * `unlock` call successfully unlocks DOT
+	submit {
+		let h in 0 .. 20;
```
Rizziepit, line 51: The intention is to add a parameter to `Verifier` to control this upper limit in another PR

Rizziepit: Actually, @vgeddes do you have any thoughts on this? I'm going back and forth on the best solution. `Verifier::verify` has a very high upper limit for storage reads, i.e. 5000. It's currently limited by the pruning range (see [`FINALIZED_HEADERS_TO_KEEP`](https://github.com/Snowfork/polkadot-ethereum/blob/7bdca4f73328a2089ccbd9a0183884a31cbf8b83/parachain/pallets/verifier-lightclient/src/lib.rs#L37)). So a message from 5000 headers ago will trigger an iteration over 5000 ancestors to ensure it's in the finalized chain.

If I were to set `submit`'s weight to the current upper limit, that's unreasonably high. Even if I implement a refund mechanism. So I'm considering the following options:
1. Simply reduce `FINALIZED_HEADERS_TO_KEEP` to a value like 20. This should be fine if messages are always forwarded as soon as their header is finalized. But it feels very aggressive.
2. Let pallets pass `max_header_age` to `Verify::verify`. So each pallet declares its own upper limit.
3. Add `FinalizedHeaderExpiresAfter` config parameter to lightclient pallet. It's worse than (1) in my opinion since we're storing a bunch of headers that we don't use

musnit: eth is 6484.00 per day i think, so 5000 is like max 20 hours downtime

musnit: why does Verify::verify iterate over storage reads to find the header? could it just accept the exact block number and just get that one out?

Rizziepit: The 5000 means we need messages to be forwarded to us within 5000 blocks of when a message's block was finalized. The implication for downtime is we can't have more than 20 hours asymmetrical downtime, i.e. we receive headers but we don't receive their messages. I don't think we can recover from that. If we don't receive both headers and messages, that's fine. Nothing will be pruned and we'll continue where we left off when forwarding is restored.

W.r.t. why we iterate: An old block's existence in storage doesn't indicate it's in the canonical chain because we don't prune stale siblings on finalization. Pruning happens only at the tail of 5000 headers. This comes from parity-bridges-common (see https://github.com/paritytech/parity-bridges-common/blob/master/modules/ethereum/src/lib.rs#L957-L963). Instead of iterating, we could set a flag in storage when a header is finalized. At a glance I think that will work. But the fact that parity-bridges-common didn't implement something like that is conspicuous to me... Shall we give it a shot anyway?

musnit: yeh, the flag sounds like a good idea to me - i guess it means slightly higher initial costs and an extra field being stored when finalizing but then o(1) message verification for older messages and could do much later prunes, like 50k blocks?

i suppose if the expectation is that message relay will happen as soon as a block is finalized 99% of the time, then the flag could be more expensive overall longer term due to the overhead? maybe that's why they didn't add it?

@tomusdrw do you have any insight into why this was or wasn't done in parity-bridges-common?

tomusdrw: The linked `ethereum` crate is dealing with PoA chain, which does have a notion of finality, but it's not necessarily predictable. I.e. the fact that there has been many blocks on top of some block doesn't necessarily indicate finality. We don't use a flag there to prevent a situation where we have to mark as finalized a very long fork.
With PoW I think it's definitely a good idea to have a flag here, since the notion of "finality" is quite predictable - i.e. you just wait for a bunch of confirmations, so with a header import you always finalized at most one header.
I don't think the extra storage cost for the flag is worth to be considered TBH.

Just a random thought that I had now: perhaps it's worth to implement the storage as a ring buffer?

Rizziepit: Thanks @tomusdrw. That's helpful context. I'm going to add the flag.

W.r.t. a ring buffer - I think that would simplify the pallet a lot. The behaviour of the ring buffer would be unexpected though, but it probably doesn't matter. By "unexpected" I mean it cannot guarantee storage of exactly `size` block heights - no more, no less. At least, that would be my expectation of a ring buffer in this context.

Rizziepit:
I added the flag and made `verify` O(1) in https://github.com/Snowfork/polkadot-ethereum/pull/328

parachain/pallets/basic-channel/src/inbound/benchmarking.rs
tomusdrw: > By "unexpected" I mean it cannot guarantee storage of exactly size block heights - no more, no less. 

Ah, I was just writing a comment to ask you to elaborate on why it's not possible, but I think I know what you mean - let me get a confirmation though. So while we know that finality is achieved in `confirmations` number of blocks, we can also specify the `proof_window` (number of finalized blocks to keep) and then set the ring buffer `size = confirmations + proof_window`. However there can be many forks in the network, so we potentially need to store multiple headers under each ring buffer slot (i.e. `Vec<(Header, IsFinalized)>`). I think it's still interesting to see how big that `Vec` would be historically in the worst case, my gut says it's like 2-3 entries at most, which is not huge for decode/encode and a lookup. We could also use an enum:
```rust
enum Headers {
  Finalized(Header),
  Unfinalized(Vec<Header>),
}
```
So instead of marking the header with a flag we simply switch from vec-like enum variant to single-value enum varing in the ring buffer slot.

Rizziepit: > I think it's still interesting to see how big that Vec would be historically in the worst case, my gut says it's like 2-3 entries at most, which is not huge for decode/encode and a lookup

Sounds about right. I'm guessing that the ring buffer will save us 1 or 2 storage writes for the average case, and 8 in the worst case.

Just to clarify, when I said "cannot guarantee storage..." I meant that we cannot guarantee it while maintaining header insertion as an O(1) operation. I was imagining the headers are stored individually, and the ring buffer stores only pointers to those headers. So whenever you insert a header, you have to delete a variable number of stale headers using individual storage writes until you hit a deletion limit. However, your proposal of having a ring buffer slot contain the headers themselves, not pointers to the headers, means that insertion is trivially O(1) in terms of storage writes since you delete 1 slot max. It's not O(1) in terms of encoding/decoding but that's a great tradeoff for perf - at least for the `import_header` extrinsic, which is by far the most expensive.

Rizziepit: The ring buffer has some other annoyances, like not being able to look up a header by hash. So clients always need to know both block height + hash. I'll go think about this some more

musnit: i don't think we should implement ring buffer now but yeh seems like it may provide some optimizations longer term, though with extra complexities

parachain/runtime/snowbridge/src/lib.rs
```diff
@@ -670,6 +670,7 @@ impl_runtime_apis! {
 			add_benchmark!(params, batches, pallet_timestamp, Timestamp);
 			add_benchmark!(params, batches, verifier_lightclient, VerifierLightclient);
 			add_benchmark!(params, batches, assets, Assets);
+			add_benchmark!(params, batches, basic_channel_inbound, BasicInboundChannel);
```
vgeddes, line 673: Maybe its possible to initialize the locked DOT balance here instead of the chain spec?

```rust
use frame_support::traits::Currency;

Balances::make_free_balance_be(&DotModuleId::get().into_account(), 1000u32.into());

add_benchmark!(params, batches, frame_system, SystemBench::<Runtime>);
add_benchmark!(params, batches, pallet_balances, Balances);	
...
```

parachain/pallets/dot-app/src/lib.rs
```diff
@@ -71,6 +71,16 @@ decl_storage! {
 		/// Address of the peer application on the Ethereum side.
 		Address get(fn address) config(): H160;
 	}
+
+	add_extra_genesis {
+		config(initial_module_balance): BalanceOf<T>;
```
vgeddes, line 76: See my other comment about whether its possible to initialize the locked balance in `dispatch_benchmarks`.

If that doesn't work, the current approach works I guess. Though can we rename `initial_module_balance` to `balance`? Its simpler.

parachain/pallets/basic-channel/src/inbound/mod.rs
```diff
@@ -13,6 +13,8 @@ use artemis_core::{
 
 use envelope::Envelope;
 
+mod benchmarking;
```
vgeddes, line 16: I think we should only declare the benchmarking module if the `runtime-benchmarks` is enabled.

```
#[cfg(feature = "runtime-benchmarks")]
pub mod benchmarking;
```

parachain/pallets/basic-channel/src/inbound/benchmarking.rs
```diff
@@ -0,0 +1,255 @@
+//! BasicInboundChannel pallet benchmarking
+
+#![cfg(feature = "runtime-benchmarks")]
```
vgeddes, line 3: Can remove this if we only declare the benchmarking module if the feature is enabled.

vgeddes (commented):
Looks good! Added some minor comments

vgeddes:
@Rizziepit hey so I overlooked that we can set the initial locked DOT balance using the genesis configuration for the balances pallet.

Given that the account id for the DOTApp module is "5EYCAe5jHEQsVPTRQqy6NCeG71Hz1EVXikZxTkr67fM8j2Rd", we can append the following item to the `genesis.runtime.palletBalances.balances` array to initialize the locked balance to a large number.

```json
[
  "5EYCAe5jHEQsVPTRQqy6NCeG71Hz1EVXikZxTkr67fM8j2Rd",
  1152921504606846976
]
```

parachain/pallets/basic-channel/src/inbound/benchmarking.rs
Rizziepit: I agree with not declaring the module if it's not used. I'm going to leave it as is in this PR though, and update all our pallets in another PR. We're using this same pattern in all pallets, and I'd rather update them all in one go to keep the style consistent.

parachain/pallets/basic-channel/src/inbound/mod.rs
Rizziepit: > @Rizziepit hey so I overlooked that we can set the initial locked DOT balance using the genesis configuration for the balances pallet.
> 
> Given that the account id for the DOTApp module is `5EYCAe5jHEQsVPTRQqy6NCeG71Hz1EVXikZxTkr67fM8j2Rd`, we can append the following item to the `genesis.runtime.palletBalances.balances` array to initialize the locked balance to a large number.
> 
> ```json
> [
>   "5EYCAe5jHEQsVPTRQqy6NCeG71Hz1EVXikZxTkr67fM8j2Rd",
>   1152921504606846976
> ]
> ```

Nice! I was struggling to figure out what the module account ID is. How did you convert `ModuleId(*b"s/dotapp")` to account ID?

philipstanislaus: I have used this in the past: https://github.com/paritytech/substrate/blob/9b08105b8c7106d723c4f470304ad9e2868569d9/utils/frame/frame-utilities-cli/src/module_id.rs

philipstanislaus (commented):
LGTM 👍

parachain/pallets/basic-channel/src/inbound/mod.rs
vgeddes: Nice, didn't know about that tool. I managed to get the account id by doing the following in a unit test:

```rust
use sp_runtime::traits::AccountIdConversion;

fn print_account() {
    println!("{}", ModuleId(*b"s/dotapp").into_account())
}
```

philipstanislaus: That‘s also super straight forward, thanks for sharing!

PR #326, opened 2021-03-29
Fees for Ethereum → Sustrate messages

vgeddes:
This change implements fees for sending messages to Sustrate as well as a bunch of related things:

* In Substrate, added a Treasury pallet
* The fee attached to an incoming message is split, 80% going to the relayer account, and 20% going to the treasury. Of course this can be changed as necessary.
* On Ethereum, users need to authorize apps to send messages on their behalf. Our current set of bridge apps however are authorized by default for all users. See ChannelAccess.sol for this code which is used by IncentivizedOutboundChannel.
* On Ethereum, initial governance / configuration mechanisms for some contracts.

The changes have all been tested using the updated E2E and unit tests.


ethereum/contracts/IncentivizedOutboundChannel.sol
```diff
@@ -2,25 +2,85 @@
...
+            _authorizeDefaultOperator(operator);
+        }
+
+        // drop admin privileges
+        renounceRole(DEFAULT_ADMIN_ROLE, msg.sender);
+    }
+
+    // Update message submission fee.
+    function setFee(uint256 _amount) external {
+        require(hasRole(CONFIG_UPDATE_ROLE, msg.sender), "Caller is unauthorized");
```
musnit, line 60: can use a modifier for this like onlyConfigUpdater - https://docs.soliditylang.org/en/v0.6.2/contracts.html#function-modifiers

ethereum/migrations/2_next.js
```diff
@@ -29,6 +29,10 @@ const channels = {
 
 module.exports = function(deployer, network, accounts) {
   deployer.then(async () => {
+
+    // Account of governance contract (to be implemented...)
+    const administrator = accounts[0];
```
musnit, line 34: put in config plz

ethereum/contracts/ChannelAccess.sol
```diff
@@ -0,0 +1,51 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.7.6;
+pragma experimental ABIEncoderV2;
+
+abstract contract ChannelAccess {
```
musnit, line 5: Can we get a high level comment describing the problem being solved by this as a guide to readers/auditors?

ethereum/migrations/2_next.js
```diff
@@ -89,7 +94,17 @@ module.exports = function(deployer, network, accounts) {
...
-    }
 
+      // Do post-construction initialization.
+      await channels.incentivized.outbound.instance.initialize(
+        administrator,
+        dotApp.address,
+        [dotApp.address, ethApp.address, erc20App.address]
+      );
+      await channels.incentivized.outbound.instance.setFee(
+        "1000000000000000000", // 1 SnowDOT
```
musnit, line 105: put in config plz

parachain/pallets/incentivized-channel/src/inbound/mod.rs
```diff
@@ -106,3 +113,37 @@ decl_module! {
 		}
 	}
 }
+
+impl<T: Config> Module<T> {
+
+	fn handle_fee(amount: BalanceOf<T>, relayer: &T::AccountId) {
```
musnit, line 119: @philipstanislaus plz review this function thoroughly too

musnit (approved):
some minor comments, but otherwise looks good. lets get another approval before merging though.

parachain/pallets/incentivized-channel/src/inbound/mod.rs
vgeddes: Just to explain this function a bit more (and I'll add a comment to the code as well):

* If the fee is zero, do nothing
* Otherwise, withdraw the fee amount from the Dot-App module account, returning a negative imbalance
* Figure out the fraction of the fee amount that should be paid to the relayer
* Pay the relayer if their account exists.
* Adjust the negative imbalance by offsetting the amount paid to the relayer
* Resolve the negative imbalance by depositing it into the treasury account

Here's a primer on imbalances to explain that aspect a bit more: https://blog.polymath.network/substrate-deep-dive-imbalances-8dfa89cc1d1

parachain/pallets/incentivized-channel/src/inbound/test.rs
```diff
@@ -259,3 +261,21 @@ fn test_submit_with_invalid_nonce() {
...
 	});
 }
+
+#[test]
+fn test_handle_fee() {
+	new_tester(SOURCE_CHANNEL_ADDR.into()).execute_with(|| {
+		let relayer: AccountId = Keyring::Bob.into();
+
+		let _ = Balances::deposit_creating(&SourceAccount::get(), 100000000000); // 10 DOT
+		let _ = Balances::deposit_creating(&TreasuryAccount::get(), Balances::minimum_balance());
```
Rizziepit, line 271: You use `resolve_creating` in `handle_fee`. So it should create this account if it doesn't exist, right? I'd add a test for that

parachain/pallets/incentivized-channel/src/inbound/mod.rs
Rizziepit: This looks fine to me. Perhaps remove the `RewardRelayer` stuff from core though

parachain/pallets/incentivized-channel/src/inbound/mod.rs
```diff
@@ -106,3 +113,37 @@ decl_module! {
...
+		if amount.is_zero() {
+			return;
+		}
+
+		let imbalance = match T::Currency::withdraw(&T::SourceAccount::get(), amount, WithdrawReasons::TRANSFER, KeepAlive) {
+			Ok(imbalance) => imbalance,
+			Err(err) => {
+				debug::error!("Unable to withdraw from source account: {:?}", err);
```
Rizziepit, line 127: Heads up that `debug` becomes `log` in https://github.com/Snowfork/polkadot-ethereum/pull/331

Rizziepit (commented):
I reviewed the Substrate code. lgtm

test/test/test.js
```diff
@@ -37,6 +37,80 @@ describe('Bridge', function () {
...
+      let amountWrapped = BigNumber(Web3.utils.toWei('100', "ether")); // 100 SnowDOT (18 decimal places)
+
+      const account = ethClient.accounts[1];
+
+      let beforeEthBalance = await ethClient.getDotBalance(account);
+      let beforeSubBalance = await subClient.queryAccountBalance(polkadotSenderSS58);
+
+      // lock DOT using basic channel
+      await subClient.lockDOT(subClient.alice, account, amount.toFixed(), 0)
+      await sleep(70000);
```
philipstanislaus, line 54: Is there a better way to wait for the event arriving on Ethereum than sleeping?

philipstanislaus (approved):
LGTM, few minor comments

ethereum/contracts/ChannelAccess.sol
```diff
@@ -0,0 +1,51 @@
...
+    );
+
+    // Authorize an operator to submit messages for *all* users.
+    function _authorizeDefaultOperator(address operator) internal {
+        defaultOperators[operator] = true;
+    }
+
+    // Revoke authorization for a default operator.
+    function _revokeDefaultOperator(address operator) internal {
+        require(msg.sender != operator, "Authorizing self as operator");
```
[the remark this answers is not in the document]
vgeddes: I adapted this from the ERC-777 [operators](https://eips.ethereum.org/EIPS/eip-777#operators-1) feature, so I mostly went with their design.

Can always change this if required. But for now its fine if only users and governance contracts can revoke.

ethereum/contracts/ChannelAccess.sol
vgeddes: added docs in latest upload.

ethereum/migrations/2_next.js
vgeddes: That's actually just stub code until I develop the governance contract and deploy it. No need for config

ethereum/contracts/IncentivizedOutboundChannel.sol
vgeddes: Will keep it mind when I flesh out governance on the Ethereum side.

parachain/pallets/incentivized-channel/src/inbound/mod.rs
```diff
@@ -106,3 +113,37 @@ decl_module! {
...
+				return;
+			}
+		};
+
+		let reward_fraction: Perbill = RewardFraction::get();
```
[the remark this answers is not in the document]
vgeddes: Nah, [Perbill](https://substrate.dev/rustdocs/v3.0.0/sp_runtime/struct.Perbill.html) is always a fraction in the range `[0, 1]`.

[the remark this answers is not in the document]
philipstanislaus: Okay, that‘s good to know!

parachain/pallets/incentivized-channel/src/inbound/mod.rs
```diff
@@ -106,3 +113,37 @@ decl_module! {
...
 }
+
+impl<T: Config> Module<T> {
+
+	fn handle_fee(amount: BalanceOf<T>, relayer: &T::AccountId) {
+		if amount.is_zero() {
+			return;
+		}
+
+		let imbalance = match T::Currency::withdraw(&T::SourceAccount::get(), amount, WithdrawReasons::TRANSFER, KeepAlive) {
```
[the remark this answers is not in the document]
vgeddes: I don't think that actually has any effect at the moment for the balances pallet, looking at https://github.com/paritytech/substrate/blob/master/frame/balances/src/lib.rs#L532

The only reason that seems to change any withdrawal logic is `TRANSACTION_PAYMENT`. Presumably other implementations of `Currency` may do something different.

In any case, `T::SourceAccount` actually maps to the DOTApp account, rather than the account of a specific user. So I don't see any semantic benefit in marking the withdrawal as a fee. We're not really charging a fee from the DOTAccount in exchange for some action.

parachain/pallets/incentivized-channel/src/inbound/mod.rs
```diff
@@ -106,3 +113,37 @@ decl_module! {
...
+			}
+		};
+
+		let reward_fraction: Perbill = RewardFraction::get();
+		let reward_amount = reward_fraction.mul_ceil(amount);
+
+		let rewarded = T::Currency::deposit_into_existing(relayer, reward_amount)
+			.unwrap_or_else(|_| PositiveImbalanceOf::<T>::zero());
+
+		let adjusted_imbalance = match imbalance.offset(rewarded) {
```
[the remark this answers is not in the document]
vgeddes: `rewarded` will always be less than `imbalance`, ignoring negative vs positive.

parachain/pallets/incentivized-channel/src/inbound/mod.rs
```diff
@@ -106,3 +113,37 @@ decl_module! {
...
+			Ok(imbalance) => imbalance,
+			Err(_) => {
+				debug::error!("Unable to offset imbalance");
+				return;
+			}
+		};
+
+		T::Currency::resolve_creating(&T::TreasuryAccount::get(), adjusted_imbalance);
```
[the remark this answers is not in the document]
vgeddes: Ah, `adjusted_imbalance` is always a `NegativeImbalance`. It will always compile.

test/test/test.js
vgeddes: Definitely, we can always update the E2E tests to watch for events. Its just very low-hanging fruit in the scheme of things.

parachain/pallets/incentivized-channel/src/inbound/test.rs
vgeddes: Yeah the E2E tests do verify that the treasury account is created and has the right amount deposited.

parachain/pallets/incentivized-channel/src/inbound/mod.rs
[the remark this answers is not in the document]
philipstanislaus: I think the main difference is how it will show up in block explorer, if at all. A transfer has a sender and a recipient, while a fee is not "sent" to anyone. This is hypothetical though, because block explorers currently use events rather than withdraw reasons. 

No strong opinion, but to not mess up explorers, I guess changing to `FEE` might make sense.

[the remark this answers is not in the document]
vgeddes: Hey yeah I remember we previously chatted about that. I agree. That's why I moved to withdraw & deposit instead of transfer, as events are only emitted for the latter (because its a high level operation). So we're fine in that regard.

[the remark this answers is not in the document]
vgeddes: For reference, no events are emitted in `Currency::withdraw`: https://github.com/paritytech/substrate/blob/master/frame/balances/src/lib.rs#L1473

PR #328, opened 2021-03-31
Benchmarks for incentivized channel & constant time `verify`

Rizziepit:
This PR adds benchmarks for the incentivized channel's 'submit' call. It also modifies the lightclient pallet to make the `Verifier::verify` call an O(1) operation. It does this by checking the new flag `StoredHeader.finalized` instead of iterating over the header chain to check finality (see discussion at https://github.com/Snowfork/polkadot-ethereum/pull/325#discussion_r604443621).

Finally, I'm also generating weights and configuring `WeightInfo` for the runtimes.


parachain/pallets/verifier-lightclient/src/lib.rs
```diff
@@ -331,6 +334,13 @@ impl<T: Config> Module<T> {
 			)?;
 			if new_finalized_block_id != finalized_block_id {
 				FinalizedBlock::put(new_finalized_block_id);
+				Headers::<T>::mutate(new_finalized_block_id.hash, |option| -> DispatchResult {
+					if let Some(header) = option {
+						header.finalized = true;
+						return Ok(());
+					}
+					Err(Error::<T>::Unknown.into())
+				})?;
```
philipstanislaus, line 343: Do we also have to iterate through previous headers to mark them as finalized, up until the highest previous finalized header?

philipstanislaus (approved):
Nice, thanks 👍

parachain/pallets/verifier-lightclient/src/lib.rs
Rizziepit: No. This finalized header's parent would have been marked as finalized when `header.number - 1` was imported, and so forth for each ancestor

PR #329, opened 2021-04-01
Limit number of messages in commitment queues

Rizziepit:
This PR sets an upper bound on the number of messages that can be queued / committed per channel in the commitments pallet. This resolves https://github.com/Snowfork/polkadot-ethereum/issues/226

### DIFF
--- a/parachain/pallets/commitments/src/lib.rs
+++ b/parachain/pallets/commitments/src/lib.rs
@@ -5,6 +5,7 @@ use frame_support::{
 	decl_module, decl_storage, decl_event, decl_error,
 	weights::Weight,
 	dispatch::DispatchResult,
+	traits::Get,
 };
 use sp_io::offchain_index;
 use sp_core::{H160, H256, RuntimeDebug};
@@ -55,6 +56,9 @@ pub trait Config: frame_system::Config {
 	type Hashing: Hash<Output = H256>;
 
 	type Event: From<Event> + Into<<Self as frame_system::Config>::Event>;
+
+	/// Max number of messages that can be queued and committed in one go for a given channel.
+	type MaxMessagesPerCommit: Get<usize>;
 }
 
 decl_storage! {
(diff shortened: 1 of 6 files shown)


parachain/pallets/commitments/src/lib.rs
```diff
@@ -150,17 +157,21 @@ impl<T: Config> Module<T> {
 impl<T: Config> MessageCommitment for Module<T> {
 
 	// Add a message for eventual inclusion in a commitment
-	// TODO (Security): Limit number of messages per commitment
-	//   https://github.com/Snowfork/polkadot-ethereum/issues/226
 	fn add(channel_id: ChannelId, target: H160, nonce: u64, payload: &[u8]) -> DispatchResult {
-		MessageQueues::append(
```
vgeddes, line 156: We do need to use [StorageValue::append](https://substrate.dev/rustdocs/v3.0.0/frame_support/storage/types/struct.StorageValue.html#method.append). Its an optimization that allows us to append an item without having to read and decode the entire vector.

I suggest ditching `try_mutate`, and instead use [StorageValue::decode_len](https://substrate.dev/rustdocs/v3.0.0/frame_support/storage/types/struct.StorageValue.html#method.decode_len) to retrieve the length of the vector (its also optimized, and is able read the length without decoding the entire vector).

parachain/runtime/local/src/lib.rs
```diff
@@ -441,12 +441,14 @@ impl verifier_lightclient::Config for Runtime {
 
 parameter_types! {
 	pub const CommitInterval: BlockNumber = 5;
+	pub const MaxMessagesPerCommit: usize = 100;
```
vgeddes, line 444: I'd be conservative and lower the limit to 20. Producing the commitment is very intensive (lots of hashing) and may consume the weight for the entire block.

musnit: same for concerns on the ethereum side - all messages need to be able to be processed within 1 etherum block. 20 is probably fine.

parachain/pallets/commitments/src/lib.rs
Rizziepit: Nice. I'll update

philipstanislaus (commented):
LGTM

PR #332, opened 2021-04-06
System benchmarks & switch to WASM

Rizziepit:
This PR adds generated weights for `frame_system` and regenerates all weights using WASM (it's much slower - don't be alarmed). This depends on the upgrades in https://github.com/Snowfork/polkadot-ethereum/pull/331


vgeddes (approved):
Thanks for catching that pallet name clash 🙂

philipstanislaus (approved):
LGTM 👍

PR #333, opened 2021-04-07
Commitments for incentivized channel

vgeddes:
This backports @alvatar's work on the incentivized channel commitments, in order to unblock @Rizziepit's benchmarking work.

Passes unit & E2E tests.


philipstanislaus (commented):
LGTM 👌

PR #337, opened 2021-04-12
Benchmark parachain's DB, extrinsic and block overheads

Rizziepit:
This PR adds the benchmarking logic and weights for the remaining configurable weights in our runtime:
* base block weight & base extrinsic weight (configured via `BlockWeights`)
* DB read & write weights (configured via `DbWeight`)

These weights are computed using the `node-bench` Substrate crate and inserted into <runtime>/src/weights/constants.rs using a Handlebars template.

Note that I added the `artemis-runtime-common` crate. I'm hoping we can use that crate to consolidate more of the duplicated runtime logic.

### DIFF
--- a/parachain/scripts/benchmark.sh
+++ b/parachain/scripts/benchmark.sh
@@ -39,6 +39,9 @@ benchmark_pallets()
 
     echo "Generating weights module for $RUNTIME_DIR with pallets $PALLETS"
 
+    echo "pub mod constants;" >> $TMP_DIR/mod.rs
+    echo "" >> $TMP_DIR/mod.rs
+
     for pallet in $PALLETS
     do
         MODULE_NAME="$(tr -s [:] _ <<< $pallet)_weights"
@@ -61,7 +64,7 @@ benchmark_pallets()
 
 benchmark_node()
 {
-    echo "Benchmarking node. This will take a few minutes..."
+    echo "Benchmarking node. This can take 1 to 2 hours"
 
     cargo install node-bench --version 0.8.0
     yarn global add handlebars-cmd@0.1.4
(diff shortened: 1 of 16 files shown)


parachain/scripts/benchmark.sh
```diff
@@ -16,41 +16,73 @@ else
...
+    PALLETS="assets dot_app erc20_app eth_app frame_system pallet_balances pallet_timestamp verifier_lightclient"
+
+    echo "Generating weights module for $RUNTIME_DIR with pallets $PALLETS"
+
+    echo "pub mod constants;" >> $TMP_DIR/mod.rs
+    echo "" >> $TMP_DIR/mod.rs
+
+    for pallet in $PALLETS
+    do
+        target/release/artemis benchmark \
```
vgeddes, line 47: On the m5d/c5d instances, is the SSD mounted at `/`? Otherwise may we need to add the `--base-path DIR` option to ensure that the DB is being stored on the SSD.

Rizziepit: Good point. On c5d it was mounted at /. Presumably m5d is the same. I'll double check

vgeddes (approved):

parachain/scripts/helpers/weight-constants-template.hbs
```diff
@@ -0,0 +1,17 @@
+use frame_support::{
```
philipstanislaus, line 1: Can we add a comment on top of this template that tells later viewers of the file how it is generated? 

Something like: 

// This file is auto generated through ..., refer to README.md#benchmark for details how to generate it

philipstanislaus (commented):
This is great, left just a small comment. 🎉

PR #342, opened 2021-04-14
Benchmarks for commitments logic & Utility pallet weights

Rizziepit:
This PR implements benchmarks for `on_initialize` in the incentivized channel, i.e. the commitments storage logic. I split `on_initialize` up into 3 benchmarks so that we can return appropriately low weights for the cheap cases.

I also took this opportunity to integrate weights for the Utility pallet since the pallet was added recently.


parachain/runtime/rococo/src/weights/incentivized_channel_outbound_weights.rs
```diff
@@ -0,0 +1,53 @@
...
+#![allow(unused_parens)]
+#![allow(unused_imports)]
+
+use frame_support::{traits::Get, weights::Weight};
+use sp_std::marker::PhantomData;
+
+/// Weight functions for incentivized_channel::outbound.
+pub struct WeightInfo<T>(PhantomData<T>);
+impl<T: frame_system::Config> incentivized_channel::outbound::WeightInfo for WeightInfo<T> {
+	fn on_initialize(m: u32, ) -> Weight {
```
vgeddes, line 38: Just curious, how does the benchmarking weight generator know what to do with `m`/`num_messages` (i.e the saturating_mul)? I don't see some kind of template being used.

parachain/pallets/incentivized-channel/src/outbound/benchmarking.rs
```diff
@@ -0,0 +1,65 @@
...
+	// Benchmark `on_initialize` under worst case conditions, i.e. messages
+	// in queue are committed.
+	on_initialize {
+		let m in 1 .. T::MaxMessagesPerCommit::get() as u32;
+
+		for _ in 0 .. m {
+			MessageQueue::append(Message {
+				target: H160::zero(),
+				nonce: 0u64,
+				payload: vec![1, 2, 3],
```
vgeddes, line 23: We should benchmark using a bigger payload, as the payload size does affect the weight. As part of the commitment generation process, the payload is hashed and that is fairly expensive.

Looking over our current set of apps, they all generate payloads roughly less than 128 bytes. For example:  https://github.com/Snowfork/polkadot-ethereum/blob/1978dfb5b49b68618ee87172500c0e567697cce7/parachain/pallets/dot-app/src/lib.rs#L127

So may be try a payload like `vec![1u8; 128]`.

parachain/runtime/rococo/src/weights/incentivized_channel_outbound_weights.rs
Rizziepit: This line in the benchmark is special macro syntax:
```
let m in 1 .. T::MaxMessagesPerCommit::get() as u32;
```

It declares `m` as a variable for the purpose of linear regression. This results in the benchmark being run for various values of `m` (in our case it will be run 21 times for each value of `m` due to the benchmark args `--steps 50 --repeat 20`). And then the linear regression analysis automatically figures out the multiplier for `m`.

Major caveat here: this means `frame-benchmarking` can only be used to benchmark O(1) and O(n) complexity. Logarithmic / superlinear complexity is off the table.

parachain/pallets/incentivized-channel/src/outbound/benchmarking.rs
Rizziepit: Good point. I'm going to try make the payload size a variable and see how it goes

Rizziepit:
@vgeddes I added the average payload size as a weight parameter. That allows the size to go over 128 bytes without us underestimating the weight

parachain/runtime/rococo/src/weights/incentivized_channel_outbound_weights.rs
vgeddes: awesome, very cool :)

vgeddes (approved):
+1

parachain/pallets/incentivized-channel/src/outbound/mod.rs
```diff
@@ -132,22 +150,27 @@ impl<T: Config> Module<T> {
 		let key = offchain_key(T::INDEXING_PREFIX, commitment_hash);
 		offchain_index::set(&*key, &messages.encode());
 
-		// TODO: calculate the real weight instead of returning max block weight
-		T::BlockWeights::get().max_block
+		let message_count = messages.len();
+		T::WeightInfo::on_initialize(
+			message_count as u32,
+			(payload_byte_count / message_count).saturating_add(1) as u32,
```
philipstanislaus, line 156: Please add a comment why we add 1 byte here

philipstanislaus (commented):
Awesome, just one minor nitpick 😄

PR #345, opened 2021-04-15
Improve pallet ordering

vgeddes:
In preparation for my governance PR, which adds a bunch of pallets, we need to improve pallet ordering. The goal is to ensure that when we add new pallets, we don't cause the pallet indices for bridge apps to change. This would break the bridge as the Solidity apps hardcode pallet indices.

Pallet indices from `[64, 256)` are now reserved for bridge apps. Anything else can use `[0, 63]`.


musnit:
@vgeddes plz add to parachain README/docs

philipstanislaus (commented):
LGTM

PR #347, opened 2021-04-17
Relayer js test

musnit:
instead of trying to fix the current relayer's reliability, i thought i'd try write a new simpler one.

it uses a stateless, idempotent, functional approach and is intended to be run, completed and then restarted every few seconds or so.

i'm fairly optimistic that this is probably the right thing to pursue for parachain -> ethereum channel relays, even after we add the data for MMRs and parachain merkle proofs. (though could continue in a golang script rather than JS)

thoughts? should we consider pursuing this path further?

to test it out:
 - checkout this branch
 -  run everything as usual (the original relayer is now limited to eth -> parachain mode only)
 - wait for everything to start, including parachain block production
 - run this new relayer with run.sh
 - run the e2e tests
 - watch the new relayer do its thing
 - all e2e tests should pass


vgeddes:
> instead of trying to fix the current relayer's reliability, i thought i'd try write a new simpler one.
> 
> it uses a stateless, idempotent, functional approach and is intended to be run, completed and then restarted every few seconds or so.

I'm all for discussing new algorithm designs for the substrate listener, and we should judge them on their individual merits.

That said, I think there are many good reasons to implement those ideas in the existing relayer. For one, its better to think of the relayer as a monotholic service container. In that role, it handles cross-cutting concerns such as configuration, logging, metrics, service lifecycle. If we split out parts of the relayer into tasks/microservices, those concerns would need to be re-implemented, and that's something we should not consider lightly. Especially at this late stage in the project.

The relayer is warty yes, and that's due to technical debt. To address that, in each milestone we should strive to address a small amount of that debt. My own view is that we need to troubleshoot and understand the reliability issues, and then improve the relayer iteratively. This is what I've done with https://github.com/Snowfork/polkadot-ethereum/pull/349.

musnit:
> 
> I'm all for discussing new algorithm designs for the substrate listener, and we should judge them on their individual merits.
> 

ofc, agree


> That said, I think there are many good reasons to implement those ideas in the existing relayer. For one, its better to think of the relayer as a monotholic service container. In that role, it handles cross-cutting concerns such as configuration, logging, metrics, and service lifecycle. If we split out parts of the relayer into tasks/microservices, those concerns would need to be re-implemented, and that's something we should not consider lightly. Especially at this late stage in the project.
> 

yes agree, i wrote the js as a proof of concept, not to imply that it should replace the existing relayer. i think the service container approach is great.

having said that, i do think the following goals should apply to the service container as well as each service being run by the container:
 - idempotent
 - stateless
 - functional
 -  capable of crashing and restarting and recovering gracefully

Additionally, I think having both the service container, and all services running together all on a single thread could be a good idea. The NodeJS event loop is great for this which is why I wrote the proof of concept, though agree it likely makes most sense to continue in our Golang service and to consider that in the context of our existing service.

> The relayer is warty yes, and that's due to technical debt. To address that, in each milestone we should strive to address a small amount of that debt. My own view is that we need to troubleshoot and understand the reliability issues, and then improve the relayer iteratively. This is what I've done with #349.

i think good error detection/crashing, restart and graceful recovery is the most important thing, and we should get that in asap, with further reliability improvements/debt being less important. even if there are lots of weird edge cases where the relayer fails that we leave out, that's totally fine as long as we can keep up graceful recovery.

PR #348, opened 2021-04-17
Governance V1

vgeddes:
This PR introduces our first cut at governance, and should be sufficient for our V1 release on Rococo. I've gone with a model  fleshed out after discussions with Interlay as well as W3F/Parity devs working on common-good parachain governance (For the new Statemint parachain).

Governance consists of two bodies:
1. A privileged root origin. On our staging and rococo environments, this will be a normal sudo account. On Kusama and Polkadot, this will _evolve_ to be the [relay chain council](https://wiki.polkadot.network/docs/en/learn-governance#council) (yes, governance via XMCP).
1. A local [council](https://substrate.dev/rustdocs/v3.0.0/pallet_collective/index.html), consisting of [members](https://substrate.dev/rustdocs/v3.0.0/pallet_membership/index.html) who are collectively authorized to issue and vote on new proposals.

The privileged root has total control of the local council, being able to change its membership or cancel any of its proposals. In our dev environments this is a convenience. In production, this allows the relay chain council to oversee governance of  common-good parachains.

The initial members of the local council are set via genesis configuration / the chain spec. Once running, the council can change its membership.

For now there are really only 2 privileged actions these bodies can perform, other than council changes:
1) Updating the runtime code
2) Updating reward ratio for relayers/treasury with this new dispatchable `IncentivizedInboundChannel.set_reward_fraction(origin, fraction)`

NOTE: On Kusama/Polkadot, it is a possibility that the local council will disappear, to be replaced entirely by governance-via-xcmp.


parachain/runtime/snowbridge/src/lib.rs
```diff
@@ -357,6 +360,49 @@ impl cumulus_pallet_xcm_handler::Config for Runtime {
...
 }
 
+// Governance
+
+impl pallet_sudo::Config for Runtime {
+	type Event = Event;
+	type Call = Call;
+}
+
+type EnsureRootOrHalfLocalCouncil = EnsureOneOf<
```
musnit, line 370: so this basically ensures either sudo (ie, alice on local/rococo) or one half of council, right?

are all 3 lines part of this check?

vgeddes: yeah, exactly. Either sudo or at least half of council. We can always fine-tune the proportion later.

musnit:
@vgeddes plz make sure you get a review from @Rizziepit and/or @philipstanislaus on this

Rizziepit:
Lgtm. The only issue is that benchmarks haven't been added, both for the new pallets and the new callable in the incentivized channel pallet. I'll put up a PR to add those. I assume it was skipped because all these callables have near-zero weight for non-privileged origins?

I don't understand the governance requirements well. So can't comment on the strategy but I trust you @vgeddes

vgeddes:
Hey @Rizziepit no need to add a benchmark for that new dispatchable. Besides being having a weight of near zero, its not really exploitable since only authorized council members can call it.

Rizziepit:
> Besides being having a weight of near zero, its not really exploitable since only authorized council members can call it.

It looks like an unauthorized origin can call it but the function will return at `ensure_origin`. Is that right? My understanding is that any underestimate of weight, regardless of magnitude, is a security hole. So if the weight of the `ensure_origin` call alone is, say, 1 instead of 0, that could be a problem. No idea if this is practically exploitable. But it does strike me as similar to extrinsics in the sudo pallet and they have non-zero weights: https://github.com/paritytech/substrate/blob/master/frame/sudo/src/lib.rs#L148-L152

Looks like a quick one to implement. Might as well?

vgeddes:
Oops yeah, you're right. I didn't think that weight issue through properly.

Will definitely take up your offer to write a benchmark for it. 

I'll start doing benchmarks for any new dispatchables I add, to get aquainted with the weights system.

PR #351, opened 2021-04-20
New docs on relayer

musnit:
docs: https://deploy-preview-351--snowbridge-docs.netlify.app/relayer-architecture/

### DIFF
--- a/docs/markdown/using-snowchain/validating-and-relaying/relayer-architecture.md
+++ b/docs/markdown/using-snowchain/validating-and-relaying/relayer-architecture.md
@@ -1,0 +1,36 @@
+---
+layout: default
+title: Relayer Architecture
+parent: Validating and Relaying
+permalink: /relayer-architecture/
+nav_order: 3
+---
+
+# Relayer Architecture
+The relayer is a single golang binary that acts as a service container for various workers that are involved in the relaying process. As our bridge requires a mix of different kinds of data types and proofs to be relayed at different times, co-ordinated in different ways, we split up these responsibilities into different workers.
+
+## Worker Goals
+ - Each worker is expected to be stateless and functional
+ - All transactions sent by workers should be idempotent (the on-chain logic does guarantee this in most scenarios, so usually the worker does not need to worry about this, but in the BEEFY relay process there may be some considerations)
+ - Workers should not communicate with eachother directly. If they have any dependency on eachother's work, they should co-ordinate via [Stigmergy](https://en.wikipedia.org/wiki/Stigmergy) through querying on-chain data.
+ - Workers should fail and recover gracefully. This means that they should not deadlock and should shut down cleanly upon any error. On startup, they should refresh their knowledge of the state of the world through on-chain queries, resume from that state, and catch themselves and the on-chain state up to date before any other behaviour.
+
+# Current Workers
+
+## Ethereum -> Parachain Feeder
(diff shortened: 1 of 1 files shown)


docs/markdown/using-snowchain/validating-and-relaying/relayer-architecture.md
```diff
@@ -0,0 +1,36 @@
+---
+layout: default
+title: Relayer Architecture
+parent: Validating and Relaying
+permalink: /relayer-architecture/
+nav_order: 3
+---
+
+# Relayer Architecture
+The relayer is a single golang binary that acts as a worker container for various workers that are involved in the relaying process. As our bridge requires a mix of different kinds of data types and proofs to be relayed at different times, co-ordinated in different ways, we split up these responsibilities into different workers.
```
philipstanislaus, line 10:
```suggestion
The relayer is a single Go binary that acts as a container for various workers that are involved in the relaying process. As our bridge requires a mix of different kinds of data types and proofs to be relayed at different times, co-ordinated in different ways, we split up these responsibilities into different workers that separate concerns.
```

docs/markdown/using-snowchain/validating-and-relaying/relayer-architecture.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Worker Goals
+ - Each worker is expected to be stateless and functional
```
philipstanislaus, line 13: This seems to be wrong – workers are not functional in the "no side effect" sense, they interact through networking and a database and maintain internal state. 

Maybe I'm misunderstanding your point?

docs/markdown/using-snowchain/validating-and-relaying/relayer-architecture.md
```diff
@@ -0,0 +1,36 @@
...
+ - All transactions sent by workers should be idempotent (the on-chain logic does guarantee this in most scenarios, so usually the worker does not need to worry about this, but in the BEEFY relay process there may be some considerations)
+ - Workers should not communicate with eachother directly. If they have any dependency on eachother's work, they should co-ordinate via [Stigmergy](https://en.wikipedia.org/wiki/Stigmergy) through querying on-chain data.
+ - Workers should fail and recover gracefully. This means that they should not deadlock and should shut down cleanly upon any error. On startup, they should refresh their knowledge of the state of the world through on-chain queries, resume from that state, and catch themselves and the on-chain state up to date before any other behaviour.
```
philipstanislaus, line 16:
```suggestion
 - Workers should recover gracefully in case of failures. This means that they should shut down cleanly upon any error, deadlocks should never happen. On startup, they should refresh their knowledge of the state of the world through on-chain queries, resume from that state, and catch themselves and the on-chain state up to date before any other behaviour.
```

philipstanislaus, line 16: Add:
```- Each worker can be run independent of other workers in the binary.```

philipstanislaus (approved):
LGTM, just a few minor comments

docs/markdown/using-snowchain/validating-and-relaying/relayer-architecture.md
musnit: yeh, fixing..

musnit:
@philipstanislaus fixed

merging it, @vgeddes @Rizziepit if you have any suggestions/changes, can do in follow up PR

PR #354, opened 2021-04-21
Relayer worker pool re-architecture

Rizziepit:


relayer/workers/worker.go
```diff
@@ -0,0 +1,89 @@
...
+
+	Start(ctx context.Context, eg *errgroup.Group) error
+}
+
+
+type WorkerPool struct {
+	workers []Worker
+}
+
+func NewWorkerPool(workers []Worker) *WorkerPool {
```
vgeddes, line 29: I know this still a first-cut WIP, but just adding my thoughts as we flesh this out:

From the looks of it, each worker has persistent state that is re-initialized or mutated by `worker.Start` and `worker.Teardown` between worker restarts. I think it would better if the lifetime of a worker object does not persist between restarts. This would reduce unintended state management bugs, as its a functional approach.

To achieve this, we could add a `pool.RegisterWorker(name, factory)` API, where  `factory` is a function that creates a new `Worker` object on-demand.

Then from the pool's perspective, the worker lifecycle would look something like this:

1. Create worker using factory function
2. Start worker
3. Stop worker if it dies
4. Go back to 1

The factory function could look roughly this:
```
type WorkerFactory func(arg1, arg1) (*Worker, error)
```

Thoughts?

relayer/workers/worker.go
```diff
@@ -0,0 +1,89 @@
...
+	"os/signal"
+	"syscall"
+
+	"golang.org/x/sync/errgroup"
+)
+
+type Worker interface {
+	Name() string
+
+	SetUp() error
```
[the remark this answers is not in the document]
musnit: yeh i think i agree

relayer/workers/worker.go
musnit: yeh sounds nice, though haven't thought about it too deeply

relayer/core/relayV2.go
```diff
@@ -0,0 +1,34 @@
...
+	"github.com/sirupsen/logrus"
+
+	"github.com/snowfork/polkadot-ethereum/relayer/workers"
+	"github.com/snowfork/polkadot-ethereum/relayer/workers/ethrelayer"
+)
+
+type RelayV2 struct {}
+
+func (re *RelayV2) Run() error {
+	config, err := LoadConfig()
```
vgeddes, line 16: Inline with the new independent worker approach, we should probably reorganise our configuration hierarchy to look like this:

```toml
[core]
foo = 1
bar = 2

[workers]
    [workers.EthRelay]
        [workers.EthRelay.ethereum]
        address = "http://foo.bar"

        [workers.EthRelay.substrate]
        address = "http://foo.bar"

    [workers.ParaRelay]
        [workers.ParaRelay.ethereum]
        address = "http://foo.bar"

        [workers.ParaRelay.substrate]
        address = "http://foo.bar"
```

Then the config for an individual worker gets parsed into a `Config` object and passed to the worker initializer or factory function.

relayer/core/relayV2.go
```diff
@@ -0,0 +1,34 @@
...
+	if err != nil {
+		return err
+	}
+	
+	ethrelayer := ethrelayer.NewWorker(
+		&config.Eth,
+		&config.Parachain,
+		logrus.WithField("worker", ethrelayer.Name),
```
vgeddes, line 24: Yeh, this should be sufficient for logging :+1:

relayer/workers/ethrelayer/worker.go
```diff
@@ -0,0 +1,144 @@
...
+
+type Worker struct {
+	ethconfig  *ethereum.Config
+	ethconn    *ethereum.Connection
+	paraconfig *parachain.Config
+	paraconn   *parachain.Connection
+	log        *logrus.Entry
+}
+
+const Name = "EthereumRelayer"
```
vgeddes, line 29: We should load this from config as I suggested in https://github.com/Snowfork/polkadot-ethereum/pull/354/files#r617686493

Also a naming guideline, a name like `relay:ethereum➞parachain` would be more descriptive.

musnit: i'm gonna migrate this worker in a new PR, that keeps to existing conventions. should have PR up in an hour or so

then in a future PR we can migrate all workers to use new conventions around logging, metrics, config, etc

relayer/workers/worker.go
```diff
@@ -0,0 +1,89 @@
...
+	childEg, childCtx := errgroup.WithContext(ctx)
+	err = worker.Start(childCtx, childEg)
+	if err != nil {
+		return err
+	}
+
+	return childEg.Wait()
+}
+
+func (wp *WorkerPool) Run() error {
```
vgeddes, line 53: nit: Would prefer `Start` to be consistent with `worker.Start`

relayer/workers/worker.go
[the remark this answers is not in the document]
Rizziepit: We could. It's mostly here to mirror `TearDown` so that the need to clean up after yourself is explicit. But it's not necessary

relayer/workers/worker.go
```diff
@@ -0,0 +1,89 @@
...
+	"syscall"
+
+	"golang.org/x/sync/errgroup"
+)
+
+type Worker interface {
+	Name() string
+
+	SetUp() error
+	TearDown()
```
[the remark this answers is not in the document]
Rizziepit: `TearDown` wouldn't actually stop the worker. I'm fine to get rid of `TearDown`, but I think it's misleading to have a `Stop` function if it only does cleanup but doesn't actually shut down goroutines. The goroutines are terminated using contexts.

relayer/workers/worker.go
Rizziepit: A factory sounds good, yes. I'll do that

relayer/core/relayV2.go
Rizziepit: I think config reorganization should go in another PR because it will require changes in listener / writer logic, etc. This PR is just for the worker wrappers

relayer/workers/worker.go
Rizziepit: I named it differently because the behaviour is different:
* `Start`: starts goroutines and returns
* `Run`: starts goroutines and blocks until `eg.Wait` returns and the worker is done

Yay or nay?

relayer/core/relayV2.go
vgeddes: That's fine. For now perhaps, I guess you could probably pass the entire config down to the worker factory and then the worker can pick out the parts it wants.

relayer/workers/worker.go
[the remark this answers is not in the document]
vgeddes: That makes total sense. What about removing the `Teardown` function entirely? That will force us to write workers which gracefully clean up after themselves.

relayer/workers/worker.go
vgeddes: :+1:

relayer/workers/worker.go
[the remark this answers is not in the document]
vgeddes: to clarify further, the worker can have an internal `tearDown` that is not part of the public worker API.

relayer/workers/ethrelayer/main.go
```diff
@@ -1,160 +1,147 @@
...
-		}
-
-		return nil
-	})
+	return uint64(finalizedHeader.Number), nil
+}
 
-	err = worker.parachainConn.Connect(ctx)
+func (w *Worker) connect(ctx context.Context) error {
+	kpForEth, err := secp256k1.NewKeypairFromString(w.ethconfig.PrivateKey)
```
musnit, line 111: dont need this, do we?

relayer/workers/ethrelayer/main.go
```diff
@@ -1,160 +1,147 @@
...
+	if w.ethconn != nil {
+		w.ethconn.Close()
+		w.ethconn = nil
 	}
 
-	var headerID ethereum.HeaderID
-	_, err = worker.parachainConn.GetAPI().RPC.State.GetStorageLatest(storageKey, &headerID)
-	if err != nil {
-		return nil, err
+	if w.ethconn != nil {
```
musnit, line 143: paraconn

relayer/chain/ethereum/listener.go
```diff
@@ -28,16 +28,19 @@ type Listener struct {
 	basicOutboundChannel        *outbound.BasicOutboundChannel
 	incentivizedOutboundChannel *outbound.IncentivizedOutboundChannel
 	mapping                     map[common.Address]string
-	address                     common.Address
```
musnit, line 31: can delete this file right?

relayer/workers/ethrelayer/main.go
Rizziepit: It's used below. Apologies for the messy diff. I took the opportunity to simplify this worker a lot

relayer/chain/ethereum/listener.go
Rizziepit: Yeah. I'll do it in a follow-up

relayer/workers/ethrelayer/main.go
musnit: yeh, i mean it can be removed and then just use nil below, as the eth connection never sends any transactions, only queries state?

Rizziepit:
This is ready for review now. Stuff in this PR:
* `WorkerPool` and `Worker` abstractions <- this is the main stuff to review. Check out worker.go
* Making @musnit's ethrelayer worker conform to `Worker` and removing code that was intended for `Chain`
* Ability to run `Relay` or `RelayV2` by passing the `--v2` flag. This is just a convenience while we migrate

@vgeddes I implemented the factory pattern in a very minimalist way. `WorkerPool` and `WorkerFactory` could pretty much be dropped into any codebase. I think this simplicity will serve us well down the line

relayer/workers/ethrelayer/main.go
Rizziepit: ah yes. i'll remove

vgeddes (approved):
+1

relayer/workers/worker.go
```diff
@@ -0,0 +1,126 @@
...
+					return err
+				}
+
+				log.WithField("worker", worker.Name()).Debug("Starting worker")
+				err = wp.runWorker(ctx, worker)
+				log.WithField("worker", worker.Name()).Debug("Worker terminated")
+
+				select {
+				case <-ctx.Done():
+					return ctx.Err()
```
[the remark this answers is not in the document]
Rizziepit: Yes. That will happen at the TODO below. But the auto-restart won't happen in case of `ctx->Done()`. `ctx` here is the worker pool context that gets canceled by sigterm, etc. It doesn't get canceled by individual worker contexts. See `childCtx` in `runWorker`

relayer/chain/ethereum/listener.go
Rizziepit: Deleting it in https://github.com/Snowfork/polkadot-ethereum/pull/367

philipstanislaus (commented):
LGTM 🥳

PR #355, opened 2021-04-21
Patch 1

yrong:


test/scripts/start-services.sh
```diff
@@ -7,13 +7,16 @@ configdir=$(mktemp -d -t artemis-config-XXX)
 start_ganache()
 {
     echo "Starting Ganache"
+    
+    pushd ../ethereum
```
vgeddes, line 11: This isn't necessary. Only the `test` sub-project needs `ganache-cli`. The `ethereum` subproject relies only on truffle.

PR #356, opened 2021-04-21
Worker improvements

musnit:
Misc improvements bundled together with new workers

### DIFF
--- a/relayer/magefile.go
+++ b/relayer/magefile.go
@@ -26,3 +26,13 @@ func Lint() error {
 func Install() error {
 	return sh.Run("go", "build", "-o", "$GOPATH/bin/artemis-relay", "main.go")
 }
+
+func Dev() error {
+	cmd := "go"
+	env := map[string]string{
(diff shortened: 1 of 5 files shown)


relayer/magefile.go
```diff
@@ -26,3 +26,13 @@ func Lint() error {
...
 }
+
+func Dev() error {
+	cmd := "go"
+	env := map[string]string{
+		"ARTEMIS_ETHEREUM_KEY":   "0x935b65c833ced92c43ef9de6bff30703d941bd92a2637cb00cfad389f5862109",
+		"ARTEMIS_PARACHAIN_KEY":  "//Relay",
+		"ARTEMIS_RELAYCHAIN_KEY": "//Alice",
+	}
+	return sh.RunWithV(env, cmd, "run", "./main.go", "run", "--config", "$configdir/config.toml")
```
vgeddes, line 37: Is `$config` going to be set somewhere?

vgeddes (approved):

relayer/magefile.go
musnit: atm im just doing it manually, because it changes in the start-services script everytime, but we can improve this in future too.

PR #357, opened 2021-04-21
Ethrelayer worker migration

musnit:
Ready for review

single run of e2e tests all pass, though haven't tested thoroughly with different edge cases (eg: crash/catchup)


relayer/workers/ethrelayer/main.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+
+	err := worker.ethereumConn.Connect(ctx)
+	if err != nil {
+		return err
+	}
+
+	// Short-lived channels that communicate initialization parameters
+	// between the two chains. The chains close them after startup.
+	subInit := make(chan chain.Init)
```
musnit, line 85: This seems unused. Was it ever used or is it safe to delete? @Rizziepit @denalimarsh

Rizziepit: Yip get rid of it. It was only relevant in the `Chain` world where one chain doesn't have access to the connection of another chain. I got rid of `ethInit` in https://github.com/Snowfork/polkadot-ethereum/pull/354/ too

relayer/workers/ethrelayer/main.go
```diff
@@ -0,0 +1,165 @@
...
+	ethInit := make(chan chain.Init)
+
+	// If the Substrate chain needs init params from Ethereum,
+	// retrieve them here and send to subInit before closing.
+	close(subInit)
+
+	eg.Go(func() error {
```
Rizziepit, line 92: You can get rid of this goroutine and simply use the value returned from `queryEthereumInitParams()` directly. See implementation in https://github.com/Snowfork/polkadot-ethereum/pull/354. Feel free to leave it to me as well. I understand if you don't want to change logic in these migrations

Rizziepit (approved):
LGTM. Some minor comments inline. Feel free to defer logic changes to later

relayer/workers/ethrelayer/main.go
musnit: yeh, will leave it to you, thanks!

relayer/workers/ethrelayer/main.go
```diff
@@ -0,0 +1,160 @@
...
+	parachainKeypair, err := sr25519.NewKeypairFromSeed(parachainConfig.PrivateKey, "")
+	if err != nil {
+		return nil, err
+	}
+	parachainConn := parachain.NewConnection(parachainConfig.Endpoint, parachainKeypair.AsKeyringPair(), log)
+
+	// channel for messages from ethereum
+	ethMessages := make(chan []chain.Message, 1)
+	// channel for headers from ethereum (it's a blocking channel so that we
+	// can guarantee that a header is forwarded before we send dependent messages)
```
philipstanislaus, line 46: How exactly do you mean this? The guarantee only holds if there is one messages array received before the header, not if there are multiple arrays of messages received.

PR #360, opened 2021-04-21
Fix Ethereum unit tests

vgeddes:


musnit:
@vgeddes i have another local branch where i'm trying to parellelize the e2e tests btw. may resume and push that in a few days once current relayer stuff is done

PR #361, opened 2021-04-21
Minor fixes to startscript

musnit:
some highlights:
 - clear and use same tmp folder each time so that can easily run relayer or other parts separately without having to copy paste folder
 - autocleaning of leftover processes
 - start ganache with 0 blocktime during contract deployment for quick deployment, then restart it for slower blocks
 - clean up config replacement stuff

### DIFF
--- a/test/scripts/start-services.sh
+++ b/test/scripts/start-services.sh
@@ -2,21 +2,43 @@
 set -eu
 
 # config directory
-configdir=$(mktemp -d -t artemis-config-XXX)
+configdir=/tmp/snowbridge-e2e-config
+rm -rf $configdir
+mkdir $configdir
+
+# kill all potentially old processes
+kill $(ps -aux | grep -e polkadot/target -e ganache-cli -e release/artemis | awk '{print $2}') || true
 
 start_ganache()
 {
     echo "Starting Ganache"
-    
-    pushd ../ethereum
+
     yarn run ganache-cli \
         --port=8545 \
-        --blockTime=4 \
         --networkId=344 \
         --deterministic \
+        --db $configdir/ganachedb \
         --mnemonic='stone speak what ritual switch pigeon weird dutch burst shaft nature shove' \
         >ganache.log 2>&1 &
-    popd    
+
+    scripts/wait-for-it.sh -t 32 localhost:8545
+    sleep 5
+}
+
+restart_ganache()
+{
+    echo "Restarting Ganache with a slower block time"
+
+    kill $(ps -aux | grep -e ganache-cli | awk '{print $2}') || true
+
+    yarn run ganache-cli \
+        --port=8545 \
+        --blockTime=6 \
+        --networkId=344 \
+        --deterministic \
+        --db $configdir/ganachedb \
+        --mnemonic='stone speak what ritual switch pigeon weird dutch burst shaft nature shove' \
+        >ganache.log 2>&1 &
 
     scripts/wait-for-it.sh -t 32 localhost:8545
     sleep 5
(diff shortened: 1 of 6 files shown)


test/scripts/start-services.sh
```diff
@@ -62,32 +88,15 @@ start_parachain()
...
-            \"--enable-offchain-indexing=true\",
-            \"--\",
-            \"--execution=wasm\"
-        ],
-        \"chain\": \"$configdir/spec.json\"
-    }"
     node ../test/scripts/helpers/overrideParachainSpec.js $configdir/polkadotConfig.json \
-        parachains.0 "$parachain_conf"
+        parachains.0.bin "$bin" \
+        parachains.0.flags.0 "--execution=native"
```
Rizziepit, line 93: Hmm.... so I would prefer the test services explicitly stating the ports and log files they use here ("id" field is used as log name). Otherwise these details are scattered all over the codebase. It's easy to break this startup script by changing a port in what looks like an entirely independent project

musnit: well i suppose now they just use whatever is default in config.json - this file is only really used for e2e tests now too, isn't it? so if we move it to this test folder, then that would be explicit?

musnit: though i'm not sure - is this used for anything else?

parachain/config.json
```diff
@@ -27,11 +27,12 @@
 		{
 			"bin": "target/release/artemis",
 			"id": "200",
-			"wsPort": 9966,
-			"port": 30666,
```
vgeddes, line 31: Can we rather just customize the config for E2E tests in the setup script? This config has uses outside of the E2E tests.

test/scripts/start-services.sh
vgeddes: Agree with Riz, the config in `parachain/config.json` has other uses besides the E2E tests (like for testing XCMP, which isn't yet testable as part of the E2E tests).

test/scripts/start-services.sh
```diff
@@ -2,21 +2,47 @@
 set -eu
 
 # config directory
-configdir=$(mktemp -d -t artemis-config-XXX)
+rm -rf tmp
+mkdir tmp
+configdir=$(pwd)"/tmp"
```
vgeddes, line 7: I'd rather not keep temporary state in our checkout. It clutters up things. Can we stick to to a static folder under `/tmp` like `/tmp/artemis-config`? That's the place where temporary files go.

test/scripts/start-services.sh
```diff
@@ -2,21 +2,47 @@
...
+
+# kill all potentially old processes
+kill $(ps -aux | grep -e polkadot/target -e ganache-cli -e release/artemis | awk '{print $2}') || true
 
 start_ganache()
 {
     echo "Starting Ganache"
-    
+
     pushd ../ethereum
```
vgeddes, line 16: This will use the wrong version of `ganache-cli`. See my comment in https://github.com/Snowfork/polkadot-ethereum/pull/355#discussion_r617755754.

test/scripts/start-services.sh
```diff
@@ -62,32 +88,16 @@ start_parachain()
...
-        \"flags\": [
-            \"--execution=native\",
-            \"-lruntime=debug,import_header=trace,bridge=trace\",
-            \"--rpc-cors=all\",
-            \"--offchain-worker=Always\",
-            \"--enable-offchain-indexing=true\",
-            \"--\",
-            \"--execution=wasm\"
-        ],
-        \"chain\": \"$configdir/spec.json\"
```
vgeddes, line 79: This line is important. Will break if not overriden.

musnit: has been fixed

parachain/config.json
musnit: sure, coming in next commit

test/scripts/start-services.sh
musnit: sure, coming in next commit

test/scripts/start-services.sh
musnit: fixed

test/scripts/start-services.sh
musnit: fixes

musnit:
@vgeddes @Rizziepit all comments addressed

PR #364, opened 2021-04-22
Fix start script

yrong:


musnit:
These should not be necessary - ganache should be run under yarn, in the test folder, not in ethereum folder

yrong:
Suggest you can check it a litter carefully, it's not only fix for ganache, but to parachain start script, I dont think transformEthHeader script can be executed in parachain folder without the npm depencies. And I would suggest the test script also strictly tested  

```
start_parachain()
{
    echo "Starting Parachain"
    pushd ../parachain
    ...
    node ../test/scripts/helpers/transformEthHeader.js
    ...
}
```

musnit:
yeh, parachain pop makes sense

PR #365, opened 2021-04-22
Clean up e2e tests

musnit:
A bunch of improvements including:
 - some refactoring of shared constants/fixtures into own file
 - tests in their own files
 - tests no longer wait based on time, they all wait and watch for the actual change they expect
 - overall, full test suite is 2-3x quicker
 - tests kinda work running in parallel, just some minor failures due to gas price imbalance checks when multiple tests run at a time. may do a follow up PR to fix that, in which case we'll be able to run them all in parallel


vgeddes (approved):
Looks great, these improvements are definitely welcome!

philipstanislaus (commented):
Awesome work 🎉

PR #366, opened 2021-04-22
[WIP] Sync relayer from historical block

denalimarsh:
**Goal**: sync Relayer from a historical block by populating database state without executing state transitions (e.g. sending transactions) until we're fully synced to the latest block.

**Problem**: the Relayer moves items through the BEEFY relay lifecycle by matching incoming events against database items' `InitialVerificationTx.Hash()` and `CompleteVerificationTx.Hash()`. When the Relayer is in sync mode it isn't sending transactions, so we can't search for items in the database by transaction hashes.

**Solution**: use the emitted ID field in the event to query the smart contract for the item's payload stored within the smart contract. Then simulate payload generation _as if_ we had sent the transaction and match the payloads. This alternative strategy could also be a valid approach in live mode.

**Implementation Status**:
- Database integrated into every step of the lifecycle so we can use it to hold exact state as needed by sync mode
- Simulated payload generation + matching against database payload is implemented

There's a potential problem with this approach - if we didn't witness the original BEEFY justification then the database won't have the `ValidatorAddresses` and `SignedCommitment` required for building the completion tx. But we can resolve this by starting the historical Relaychain subscriptions before the Ethereum subscriptions, then we'll have all the relevant BEEFY justifications in our database to match against (unless a tx comes in for a very old BEEFY justification).

Note: this PR needs to be rebased to `main` before being merged.


denalimarsh:
Closed in favor of https://github.com/Snowfork/polkadot-ethereum/pull/384

PR #367, opened 2021-04-22
Delete old Eth listener, move Parachain writer test

Rizziepit:
As per title


Rizziepit:
Heads up for merge conflict @denalimarsh

PR #368, opened 2021-04-23
Parachaincommitmentrelayer catches up past lost commitments on startup

musnit:
For now, this just does a brute force search through old blocks to gather all commitments needed to catch the channels up, and processes them before continuing as usual. We can improve later if needed, or remove the catchup if we implement a better way of detecting the starting block.

Note: catchup.go is a bit messy/complex due to having to track basic and incentivised channel commitments separately during the search - could improve later if needed, but it's isolated to this file so I think its fine.


vgeddes:
As an _interim_ solution I may be OK with merging this in. How much testing has it had?

Then the other thing. Over last week or 2, I've been working intermittently on a more comprehensive solution in https://github.com/Snowfork/polkadot-ethereum/tree/vincent/commitment-block-number, one that allows us to catch up in both directions by recording the last processed blocks. Its actually quite far along and also finalizes our commitments implementation for V1. Work on it is dormant while the relayer is being refactored, but I intend to fully resume development,  after other hi-priority work is completed first.

musnit:
> As an _interim_ solution I may be OK with merging this in. How much testing has it had?
> 
> Then the other thing. Over last week or 2, I've been working intermittently on a more comprehensive solution in https://github.com/Snowfork/polkadot-ethereum/tree/vincent/commitment-block-number, one that allows us to catch up in both directions by recording the last processed blocks. Its actually quite far along and also finalizes our commitments implementation for V1. Work on it is dormant while the relayer is being refactored, but I intend to fully resume development, after other hi-priority work is completed first.

tested a bunch locally and working fine

your solution looks good too! though we should prioritize hi-prio work first as you say then discuss. Some solution like in this PR will still be necessary for the future basic channel irrespective of whether we add block numbers or not too

musnit:
```
time="2021-04-26T17:47:09Z" level=error msg="Failed to submit transaction" error="the tx doesn't have the correct nonce. account has nonce of: 15 tx has nonce of: 14" worker=parachain-commitment-relayer
time="2021-04-26T17:47:09Z" level=error msg="Error submitting message to ethereum" error="the tx doesn't have the correct nonce. account has nonce of: 15 tx has nonce of: 14" worker=parachain-commitment-relayer
```

i do notice this error still occassionally, unrelated to this PR though, which currently results in the worker and relayers shutting down. If you restart the relayer/workers, it catches up them all and recovers and works fine. Once @Rizziepit implements auto-restart on workers, it should recover fine automatically too.

PR #370, opened 2021-04-24
Fix test script-2

yrong:


test/scripts/start-services.sh
```diff
@@ -4,14 +4,16 @@ set -eu
...
+mkdir -p $configdir
 
 # kill all potentially old processes
-kill $(ps -aux | grep -e polkadot/target -e ganache-cli -e release/artemis | awk '{print $2}') || true
+kill $(ps -aux | grep -e polkadot/target -e ganache-cli -e release/artemis | grep -v grep |awk '{print $2}') || true
 
 start_ganache()
 {
     echo "Starting Ganache"
+    pushd ../ethereum
```
vgeddes, line 15: @yrong It shouldn't be necessary to do this. `ganache-cli` is already installed under the `test` directory if you run `yarn install` in that dir. Are you doing that?

.gitignore
```diff
@@ -3,3 +3,4 @@
 
 # VSCode Workspace
 workspace.code-workspace
+.idea
```
vgeddes, line 6: We aren't accepting any more editor-specific ignore directives. There is a much better approach which involves using a global `gitignore` file: https://sebastiandedeyne.com/setting-up-a-global-gitignore-file/

test/scripts/start-services.sh
```diff
@@ -61,40 +64,51 @@ deploy_contracts()
 
 start_parachain()
 {
-    echo "Starting Parachain"
+    echo "Build Relaychain"
```
vgeddes, line 67: I'd prefer that this be done in a higher-level "bootstrap" script that does the following:
1. Installs polkadot-launch
2. Installs & builds polkadot
3. Runs the start-services script

test/scripts/start-services.sh
yrong: ok, gotcha. but i would still suggest to start ganache from ethereum

test/scripts/start-services.sh
yrong: yes make sense.  notice that the polkadot binary is a forked release and seems currently polkadot enable beefy by default Have you guys made some other customization or can we just use the official release?

```
git clone https://github.com/snowfork/polkadot.git /tmp/polkadot
pushd /tmp/polkadot
git checkout enable_beefy_on_rococo
cargo build --release
```

philipstanislaus (approved):
This is nice, can we get the open issues fixed to merge this @yrong? Many thanks!

vgeddes:
Closing. Thanks for coming up with these improvements @yrong, but some of them aren't in the direction we want to go. Also this branch  is very out-of-date compared to our current start-services script.

PR #372, opened 2021-04-26
Fix default location of polkadot binary

vgeddes:
This was causing start-services to fail unless an `.env` file was in place. That should be optional.

Also misc updates for README and polkadot-launch


musnit (approved):

vgeddes:
@musnit fyi latest changes in upstream polkadot-launch mean we actually need to use an older version like `89e970` until we update our configs. I've put `89e970` into our README for now.

Tested and works.

PR #375, opened 2021-04-27
fix: MMR proof decoding in subscribeBeefyJustifications script

philipstanislaus:


blablacio (approved):
Looks good!

Didn't know about `GenerateMMRProofResponse`, which seems to do the magic here.

PR #377, opened 2021-04-27
Restart workers & improved deadlock handling

Rizziepit:
Changes in this PR:
* Workers now restart, either immediately or after `restart-delay` if configured
* A deadlock handler can be passed to `WorkerPool.RunWithContext`. The default is still to cancel the context -> wait a bit -> kill process. But we can now trivially change that in future

### DIFF
--- a/relayer/core/relayV2.go
+++ b/relayer/core/relayV2.go
@@ -18,12 +18,12 @@ func (re *RelayV2) Run() error {
 		return err
 	}
 
-	ethrelayerFactory := func() (workers.Worker, error) {
+	ethrelayerFactory := func() (workers.Worker, *workers.WorkerConfig, error) {
 		return ethrelayer.NewWorker(
 			&config.Eth,
 			&config.Parachain,
 			logrus.WithField("worker", ethrelayer.Name),
-		), nil
+		), &config.Workers.EthRelayer, nil
 	}
 
 	// TODO: add all workers
--- a/relayer/workers/worker.go
+++ b/relayer/workers/worker.go
@@ -20,7 +20,7 @@ type Worker interface {
 	Start(ctx context.Context, eg *errgroup.Group) error
 }
 
-type WorkerFactory func() (Worker, error)
+type WorkerFactory func() (Worker, *WorkerConfig, error)
 
 type WorkerPool []WorkerFactory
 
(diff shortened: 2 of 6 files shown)


relayer/workers/worker.go
```diff
@@ -99,23 +108,41 @@ func (wp WorkerPool) run(ctx context.Context, log *logrus.Entry) error {
...
+					log.WithError(err).WithField("worker", worker.Name()).Debug("Worker terminated")
+				}
 
 				select {
 				case <-ctx.Done():
 					return ctx.Err()
 				default:
-					// TODO: instead retry with backoff up to X retries
-					return err
+					<-time.After(time.Duration(config.RestartDelay) * time.Second)
```
Rizziepit, line 144: I figured this is good enough for now. We can do backoffs and limit restarts if we see we need it

relayer/core/relayV2.go
```diff
@@ -18,12 +18,12 @@ func (re *RelayV2) Run() error {
 		return err
 	}
 
-	ethrelayerFactory := func() (workers.Worker, error) {
+	ethrelayerFactory := func() (workers.Worker, *workers.WorkerConfig, error) {
```
vgeddes, line 21: I'd prefer if we pass config as a parameter to the factory function, instead of capturing it from the environment. This allows us to move the definition of these factories to to individual worker modules later down the line (that was always my intent when suggesting the factory idea)

vgeddes (approved):
🚀

relayer/core/relayV2.go
Rizziepit: I tried that route at first but didn't want the `WorkerPool` to do any config parsing, i.e. when it selects a config entry to pass to a worker factory, it's "parsing" the config. I want to make config parsing the exclusive responsibility of relay.go.

An alternative that avoids config parsing in `WorkerPool` while also allowing the config to be passed to the factory: change `type WorkerPool []WorkerFactory` to `type WorkerPool [](WorkerFactory, WorkerConfig)`. What do you think?

Something worth noting: my initial instinct was also that the factory should accept the config as argument, but I changed my mind, specifically for `WorkerConfig` (other configs might be fine). Usage of params like `RestartDelay` and `Enabled` happen outside the worker and it's probably best if the worker itself has no knowledge of these concepts.

musnit: yeh theres maybe 2 different kinds of config, one for how the pool should control the specific worker, and another internal to the worker

PR #378, opened 2021-04-27
Integrate all parachain and relaychain proof and leaf/header related endpoints into relayer

musnit:
This is a test script, meant to replicate the same functionality as in scripts/helpers/subscribeBeefyJustifications.js in Golang and to get us completed integrations with all APIs and data we'll need for working verification.

There are still some TODOs but they are noncritical/nonblocking and can be fixed in follow PR where we integrate this with the listener.


relayer/cmd/subscribe_beefy.go
```diff
@@ -0,0 +1,204 @@
...
+		"Paras",
+		"Heads", encoded, nil)
+	if err != nil {
+		log.WithError(err).Error("Failed to create parachain header storage key")
+	}
+
+	// TODO1 - in polkadotjs webapp it seems possible to query this state via paras.heads without specifying any option
+	// and get all parachain heads as a response. eg: trying it on Rococo returns a lot of paraheads.
+	// I think we'll need this (to be able to check against the parachain headers hash
+	// of all headers in MMRLeaf.ParachainHeads), so we should get this working too.
```
philipstanislaus: This is partially resolved in 95964a81adbe57e248040793ef9a2793dd650fb3 . I added the right key by investigating the JSONRPC command sent by polkadot.js apps. This i a bit hacky, since a hardcoded key won't work if metadata ever updates. We should figure out why the computed key is different from the one polkadot.js apps uses.

relayer/cmd/subscribe_beefy.go
```diff
@@ -0,0 +1,204 @@
...
+	response, err := relaychainConn.GetAPI().RPC.State.GetStorageRaw(allParaHeadsStorageKey, blockHash)
+	if err != nil {
+		log.WithError(err).Error("Failed to get parachain header")
+	}
+
+	// TODO2 - the above query returns some extra bytes, related the the HeadData type (try this state query in polkadotjs
+	// webapp for example). These extra bytes I think are the parachain ID, so the response type needs to account for
+	// this properly. the below is just a hack to get the actual header out. It's also not clear to me if the response
+	// contains the entire header, or just a hash of the header, or some truncated header? If it's the entire header,
+	// then great we can use it entirely instead of querying for it in a follow up call
```
philipstanislaus, line 134: It should be the full parachain header, as you can see from this check: https://github.com/paritytech/polkadot/blob/master/node/core/candidate-validation/src/lib.rs#L418

philipstanislaus: In order to decode that, we need the types of our parachain's header

relayer/cmd/subscribe_beefy.go
```diff
@@ -0,0 +1,204 @@
...
+
+	log.WithFields(logrus.Fields{
+		"hash": headerHash.Hex(),
+	}).Info("Querying for parachain header")
+	headerData, err := parachainConn.Api().RPC.Chain.GetHeader(headerHash)
+	if err != nil {
+		log.WithError(err).Error("Failed to get parachain header")
+		// TODO3: This usually errors with "json: cannot unmarshal string into Go struct field Header.number of type uint32"
+		// Seems maybe related to the block number encoding thing on parachains vs relay chains, or maybe because our parachain
+		// header has a different type to the relay chain header and we need to setup that type?
```
philipstanislaus: Fixed in bb013dcf16863df9bb8aedf0ce7779c30f1a654f , reversion of block number type did not work.

philipstanislaus (commented):
LGTM, added a few hints and resolved a few TODOs.

relayer/cmd/subscribe_beefy.go
```diff
@@ -123,10 +122,6 @@ func GetParaheads(blockHash types.Hash, relaychainConn *relaychain.Connection) (
 		log.WithError(err).Error("Failed to create parachain header storage key")
 	}
 
-	// TODO1 - in polkadotjs webapp it seems possible to query this state via paras.heads without specifying any option
```
musnit: @philipstanislaus still seems to be an issue for me - going to re-add this TODO

relayer/cmd/subscribe_beefy.go
musnit: I did notice polkadot.js apps websocket connection makes an extra request for this, something like getKeys before the query for the actual data. Maybe its querying all keys for the mapping first, then appending the key to the base storage key before sending out the RPC?

musnit: yeh if you try on live Rococo, where there are lots of parachains, it first does state_getKeysPaged then does state_queryStorageAt with a lotttttt of extra params, one for each parachain, basically querying them all together manually. i guess we'll need to do something similar?

musnit: have a partial fix, will push.

relayer/cmd/subscribe_beefy.go
musnit: updated code with our parachain header decoding and it works, but it still requires removing the extra prefix bytes as in https://github.com/Snowfork/polkadot-ethereum/pull/378/files#diff-6bc10d8bbb796a09a8e3772904d4f0950335534c6a6b04d647c24fc66c04b6c8R139 - not sure what that prefix is for. maybe related to the Option<HeadData> type?

musnit: if you query paraheads on rococo you can see that every header response, even from different chains, is prefixed with some bytes too in the response, though in the UI the prefix is removed

relayer/cmd/subscribe_beefy.go
musnit: working now and gets all headers, though still a small hack to remove some extra bytes from the base key as in https://github.com/Snowfork/polkadot-ethereum/pull/378/files#diff-6bc10d8bbb796a09a8e3772904d4f0950335534c6a6b04d647c24fc66c04b6c8R164 - not sure why

relayer/cmd/subscribe_beefy.go
philipstanislaus: @musnit those bytes are coming from the SCALE codec. The first byte signifies `Some` (`0x01`) or `None` (`0x00`), the next bytes are a compact encoded length prefix, in our case 4, but the prefix could have less bytes. 

Let me properly decode those later.

relayer/cmd/subscribe_beefy.go
philipstanislaus: Oh, great find! Thanks for diving into this. I'll fix the extra bytes.

musnit:
closing this as it is integrated into https://github.com/Snowfork/polkadot-ethereum/pull/380

PR #384, opened 2021-04-30
Sync Relayer from previous Ethereum block

denalimarsh:
This PR includes the Ethereum-side functionality required for the Relayer to sync from a previous block. Syncing process steps:
- [ ] On initialization, query previous BEEFY justifications from the relay chain, storing them in the BEEFY justification database
- [x] Query all previous `InitialVerificationSuccessful` events from Ethereum, use them to update the status of items in the database
- [x] Query all previous `FinalVerificationSuccessful` events, update the status of database items
- [x] Iterate over the database and send appropriate transactions to the blockchain
- [x] Transition to live mode by starting subscriptions

I'm currently blocked by step 1, as it's not possible to query BEEFY justifications by block yet.


musnit:
@denalimarsh i know you're blocked on parity, but is this existing functionality something we could complete/merge in and address the initialization extras when the apis are ready?

relayer/workers/beefyrelayer/beefy-ethereum-listener.go
```diff
@@ -153,26 +151,88 @@ func (li *BeefyEthereumListener) queryLightClientEvents(ctx context.Context, sta
...
+		// Fetch validation data from contract using event.ID
+		validationData, err := li.lightClientBridge.ContractCaller.ValidationData(nil, event.Id)
+		if err != nil {
+			li.log.WithError(err).Error(fmt.Sprintf("Error querying validation data for ID %d", event.Id))
+		}
+
+		// Attempt to match items in database based on their payload
+		itemFoundInDatabase := false
+		items := li.beefyDB.GetItemsByStatus(store.CommitmentWitnessed)
+		for _, item := range items {
```
philipstanislaus, line 180: Can we not query commitments directly, e. g. by hash? This could get super slow, depending on how many blocks we need to catch up on and how many commitments are in the database and how many blocks we need to catch up on.

relayer/workers/beefyrelayer/beefy-ethereum-listener.go
```diff
@@ -181,3 +241,152 @@ func (li *BeefyEthereumListener) processLightClientEvents(ctx context.Context, e
...
+		// Fetch validation data from contract using event.ID
+		validationData, err := li.lightClientBridge.ContractCaller.ValidationData(nil, event.Id)
+		if err != nil {
+			li.log.WithError(err).Error(fmt.Sprintf("Error querying validation data for ID %d", event.Id))
+		}
+
+		// Attempt to match items in database based on their payload
+		itemFoundInDatabase := false
+		items := li.beefyDB.GetItemsByStatus(store.InitialVerificationTxConfirmed)
+		for _, item := range items {
```
philipstanislaus, line 296: Same as above, query by hash?

relayer/workers/beefyrelayer/beefy-ethereum-listener.go
```diff
@@ -181,3 +241,152 @@ func (li *BeefyEthereumListener) processLightClientEvents(ctx context.Context, e
...
+			generatedPayload := li.simulatePayloadGeneration(*item)
+			if generatedPayload == validationData.Payload {
+				li.log.Info("Deleting finalized item from the database'")
+				deleteCmd := store.NewDatabaseCmd(item, store.Delete, nil)
+				li.dbMessages <- deleteCmd
+
+				itemFoundInDatabase = true
+				break
+			}
```
philipstanislaus, line 305: Do we really need to delete those items? It might be worthwhile to keep them with a `Done` status for audit/debugging purposes?

relayer/workers/beefyrelayer/beefy-ethereum-listener.go
```diff
@@ -63,19 +63,48 @@ func (li *BeefyEthereumListener) Start(cxt context.Context, eg *errgroup.Group,
...
 	return nil
 }
 
-func (li *BeefyEthereumListener) pollEventsAndHeaders(
-	ctx context.Context,
-	descendantsUntilFinal uint64,
-) error {
+func (li *BeefyEthereumListener) pollHistoricEventsAndHeaders(ctx context.Context, descendantsUntilFinal uint64) error {
+	// Load starting block number and latest block number
+	blockNumber := li.ethereumConfig.StartBlock
```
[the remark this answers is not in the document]
denalimarsh: We can't - the database doesn't persist state between runs, it's always empty upon initialization.

relayer/workers/beefyrelayer/beefy-ethereum-listener.go
denalimarsh: Which hash are you referring to here?

relayer/workers/beefyrelayer/beefy-ethereum-listener.go
denalimarsh: Here we're using [soft delete](https://gorm.io/docs/delete.html#Soft-Delete) by default, which sets the `DeletedAt` field and filters the object out of normal database queries (the item can still be found with special query types).

relayer/workers/beefyrelayer/beefy-ethereum-listener.go
```diff
@@ -63,19 +63,48 @@ func (li *BeefyEthereumListener) Start(cxt context.Context, eg *errgroup.Group,
 		return err
 	}
 	li.blockWaitPeriod = blockWaitPeriod.Uint64()
+
+	// If starting block < latest block, sync the Relayer to the latest block
+	blockNumber, err := li.ethereumConn.GetClient().BlockNumber(ctx)
+	if err != nil {
+		return err
+	}
+	if uint64(li.ethereumConfig.StartBlock) < blockNumber {
```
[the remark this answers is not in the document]
denalimarsh: Yeah, that's true. A better solution might be if we automatically skip syncing mode and start the relayer from the current block height if the config's start block is null or set to 0.

relayer/workers/beefyrelayer/beefy-ethereum-listener.go
philipstanislaus: This current logic is getting all witnessed commitments from the DB, and then we iterate over them. If there are millions of commitments in the DB, this will be super expensive. Could we instead query the commitment directly by ID/hash/index?

relayer/workers/beefyrelayer/beefy-ethereum-listener.go
philipstanislaus: Awesome!

relayer/workers/beefyrelayer/beefy-ethereum-listener.go
denalimarsh: - Data from the BEEFY commitment: `ValidatorAddresses`, `SignedCommitment` (we can generate the `Payload` from this)
- Data from the initial verification tx: on-contract commitment ID (we can use this to query `Payload` from contract)
- Data from the second verification tx: on-contract commitment ID (and therefore `Payload`)

In live mode we're doing these lookups using hashes from txs we send, but in sync mode in order to associate a BEEFY commitment with an initial verification tx we need to compare the `Payload` because we don't have any other information available. We could adopt the contract's commitment ID and use it to refer to BEEFY commitments in the relayer database, but even then we'd need to compare the `Payload` to make the initial association.

Maybe in sync mode we could subscribe to our own address' sent transactions so that they're available to make these lookups? Open to ideas on this one.

denalimarsh: @philipstanislaus given the scope of these (potential) changes are you okay if I make an issue for this so we can get the functional current version merged into master?

philipstanislaus: Yes, definitely 👍

relayer/workers/beefyrelayer/beefy-ethereum-listener.go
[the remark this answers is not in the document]
philipstanislaus: Sounds good, could you just add a comment to the code, explaining that rationale?

relayer/workers/beefyrelayer/beefy-ethereum-listener.go
denalimarsh: Created issue https://github.com/Snowfork/polkadot-ethereum/issues/414

relayer/workers/beefyrelayer/beefy-ethereum-listener.go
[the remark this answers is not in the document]
denalimarsh: Added!

PR #386, opened 2021-05-04
Benchmark `set_reward_fraction` + weights for pallet-collective

Rizziepit:
As per title.

@vgeddes added pallets (pallet-collective, pallet-membership, pallet-sudo) and the `set_reward_fraction` dispatchable in #348. pallet-membership's benchmarks aren't ready to be added - it's tracked in https://github.com/Snowfork/polkadot-ethereum/issues/383


parachain/runtime/rococo/src/lib.rs
```diff
@@ -376,7 +376,7 @@ type EnsureRootOrHalfLocalCouncil = EnsureOneOf<
 parameter_types! {
 	pub const LocalCouncilMotionDuration: BlockNumber = 7 * DAYS;
 	pub const LocalCouncilMaxProposals: u32 = 100;
-	pub const LocalCouncilMaxMembers: u32 = 4;
+	pub const LocalCouncilMaxMembers: u32 = 8;
```
Rizziepit, line 379: @vgeddes pallet-collective's benchmarks don't work if this value is <= 6 because of things like https://github.com/paritytech/substrate/blob/7b4d55be81c4f35ef9812dad8d0349f43dcce105/frame/collective/src/benchmarking.rs#L221. Is it a problem to increase this?

vgeddes: Should be fine. :+1:

PR #390, opened 2021-05-07
Limit bytes in outbound message payload

Rizziepit:
This PR addresses https://github.com/Snowfork/polkadot-ethereum/issues/149. We're no longer storing unbounded message payloads in offchain storage.

### DIFF
--- a/parachain/pallets/incentivized-channel/src/outbound/mod.rs
+++ b/parachain/pallets/incentivized-channel/src/outbound/mod.rs
@@ -54,6 +54,9 @@ pub trait Config: system::Config {
 
 	type Hashing: Hash<Output = H256>;
 
+	// Max bytes in a message payload
+	type MaxMessagePayloadSize: Get<usize>;
+
 	/// Max number of messages that can be queued and committed in one go for a given channel.
 	type MaxMessagesPerCommit: Get<usize>;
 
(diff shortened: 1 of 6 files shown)


parachain/pallets/incentivized-channel/src/outbound/mod.rs
```diff
@@ -153,6 +164,8 @@ impl<T: Config> Module<T> {
 		let message_count = messages.len();
 		T::WeightInfo::on_initialize(
 			message_count as u32,
+			// We overestimate message payload size rather than underestimate.
+			// So add 1 here to account for integer division truncation.
```
Rizziepit, line 168: Adding this in response to https://github.com/Snowfork/polkadot-ethereum/pull/342#discussion_r616110876

musnit:
looks good! i presume 256 is enough for all our current apps including the upcoming NFTApp?

vgeddes (approved):
Looks good!

Rizziepit:
> looks good! i presume 256 is enough for all our current apps including the upcoming NFTApp?

Yes. I had a look at https://github.com/Snowfork/polkadot-ethereum/pull/338/ and the NFT `OutboundPayload` should be under 100 bytes.

Heads up @vgeddes @musnit - I didn't add this to the basic channel. Do we want to maintain parity? Should we get rid of the basic channel if we don't intend to use it?

PR #391, opened 2021-05-08
Fees and Rewards for Sub->Eth

vgeddes:
Covers fees/rewards for the Sub->Eth direction

The fee charged for including a message is fixed and can be updated via governance. To keep this PR relatively simple for now, the entire fee gets rewarded to relayers, without the treasury getting a cut. This of course can be changed in a follow-up PR.

Changes:
* Parachain
  * The incentivized outbound channel now charges fees for including a message in commitment. The fee amount is included in the outbound message.
  * Fixed bug where nonce would be incremented but message would not be included in commitment due to queue length constraints.
* Ethereum
  * Dropped the abstract InboundChannel contract since the basic and incentivized inbound message payloads are now different
  * Incentivized inbound channel now rewards relayers based on the fee amount included in the inbound message
  * Major refactoring of unit tests to accommodate the above, and the following:
    * Unit tests for individual apps now use mock channels. The previous approach of bootstrapping all the channels for app tests was getting gnarly, especially as the channels are becoming more complex. And we haven't even integrated the Grandpa light client in them yet
    * Improved compatibility for Hardhat when we switch over (I've backported changes from an old hardhat branch I've kept around).
    * The tests no longer rely on the truffle migration script. So tests instantiate their own contracts, and our migration scripts can focus on production deployments
* Relay
  * Dropping the abstract InboundChannel contract resulted in a lot of surgery in the commitment relayer unfortunately.


ethereum/test/test_basic_inbound_channel.js
```diff
@@ -0,0 +1,123 @@
+require("chai")
+  .use(require("chai-as-promised"))
+  .should();
+
+const BasicInboundChannel = artifacts.require("BasicInboundChannel");
+const MockApp = artifacts.require("MockApp");
+
+const { ethers } = require("ethers");
+
+const makeCommitment = (messages) => {
```
vgeddes, line 10: @denalimarsh thought you may be interested in this - A much simpler way to generate commitments on the JS side for unit tests.

ethereum/contracts/RewardSource.sol
```diff
@@ -0,0 +1,8 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.7.6;
+pragma experimental ABIEncoderV2;
+
+// Something that can reward a relayer
+interface RewardSource {
```
vgeddes, line 6: The `ETHApp` contract implements this interface, so it can dispense rewards.

relayer/workers/parachaincommitmentrelayer/ethereum-channel-writer.go
```diff
@@ -14,44 +14,45 @@ import (
...
 type EthereumChannelWriter struct {
-	config    *ethereum.Config
-	conn      *ethereum.Connection
-	contracts map[substrate.ChannelID]*inbound.Contract
-	messages  <-chan []chain.Message
-	log       *logrus.Entry
+	config                     *ethereum.Config
+	conn                       *ethereum.Connection
+	basicInboundChannel        *inbound.BasicInboundChannel
+	incentivizedInboundChannel *inbound.IncentivizedInboundChannel
```
philipstanislaus, line 23: This makes the writer aware of specific channels – not sure that's a good design choice. Probably fine for now, having the writer agnostic to concrete channels might be a case of [YAGNI](https://xkcd.com/974/).

relayer/workers/parachaincommitmentrelayer/parachaincommitment/catchup.go
```diff
@@ -196,3 +200,84 @@ func (li *Listener) checkIfDigestItemContainsNonce(
...
+	if err != nil {
+		li.log.WithError(err).Error("Failed to decode commitment messages")
+		return nil, err
+	}
+
+	return messages, nil
+}
+
+func (li *Listener) getIncentivizedMessages(commitment types.H256) ([]chainTypes.IncentivizedOutboundChannelMessage, error) {
+	storageKey, err := parachain.MakeStorageKey(substrate.ChannelID{IsBasic: true}, commitment)
```
philipstanislaus, line 254: This should be `IsIncentivized: true`, right?

relayer/workers/parachaincommitmentrelayer/ethereum-channel-writer.go
vgeddes: This is very true. The previous version of the writer was agnostic to the concrete channels, because both channels conformed to the abstract `InboundChannel` contract and its `Message` struct. However this PR added a `fee` field to the `Message` struct of the incentivized inbound channel, and that meant the inbound channel could no longer conform to the `InboundChannel` interface.

This problem isn't isolated to the writer. The listener is also fully aware of the concrete channels (predating this PR).

So in future, I think what we need to explore is to add a generic `bytes data` to our Message structs. Then the incentivized channels can add have their fee info there, while still conforming to abstract `InboundChannel` and `OutboundChannel` contracts.

ethereum/contracts/IncentivizedInboundChannel.sol
```diff
@@ -74,8 +91,15 @@ contract IncentivizedInboundChannel is InboundChannel {
...
 
+            _rewardAmount = _rewardAmount + _messages[i].fee;
             emit MessageDispatched(_messages[i].nonce, success);
         }
+
+        // Attempt to reward the relayer
+        try rewardSource.reward(_relayer, _rewardAmount) { }
+        catch {
+            emit RelayerNotRewarded(_relayer, _rewardAmount);
+        }
```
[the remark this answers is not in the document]
vgeddes: Relayers may not want others to know that they are being rewarded certain sums though? I think we discussed that when I implemented the other direction (eth->sub fees/rewards). I can add an event if you think otherwise though.

relayer/workers/parachaincommitmentrelayer/ethereum-channel-writer.go
philipstanislaus: @vgeddes makes sense. I'm also not sure this abstraction is necessary at this point, my inclination is rather to keep it simple until we need a multitude of channels in the future.

ethereum/contracts/IncentivizedInboundChannel.sol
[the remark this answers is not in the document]
philipstanislaus: It is transparent anyways, e. g. through call tracing. 

I don't really see a strong use case for having an event, and since it costs gas, it would be fine to leave it as is.

ethereum/contracts/BasicInboundChannel.sol
```diff
@@ -28,25 +32,6 @@ contract BasicInboundChannel is InboundChannel {
         view
         returns (bool success)
     {
-        // Prove we can get the MMRLeaf that is claimed to contain our Parachain Block Header
```
musnit, line 31: these comments are still important as they are changes we need to make - plz keep them in

ethereum/contracts/IncentivizedInboundChannel.sol
```diff
@@ -28,25 +62,6 @@ contract IncentivizedInboundChannel is InboundChannel {
         view
         returns (bool success)
     {
-        // Prove we can get the MMRLeaf that is claimed to contain our Parachain Block Header
```
musnit, line 31: plz keep

parachain/pallets/incentivized-channel/src/outbound/mod.rs
```diff
@@ -116,34 +136,33 @@ pub fn offchain_key(prefix: &[u8], hash: H256) -> Vec<u8> {
...
-		MessageQueue::append(
-			Message {
-				target,
-				nonce,
-				payload: payload.to_vec(),
-			},
-		);
-		Ok(())
+		let fee = Self::fee();
+		T::FeeCurrency::withdraw(who, fee).map_err(|_| Error::<T>::NoFunds)?;
```
musnit, line 151: where does the fee go? is it burned?

musnit: assume this works fine too if user is burning eth, then the total burned will be the base message burned + fee burned?

ethereum/contracts/IncentivizedInboundChannel.sol
```diff
@@ -74,8 +91,15 @@ contract IncentivizedInboundChannel is InboundChannel {
             (bool success, ) =
                 _messages[i].target.call{value: 0, gas: MAX_GAS_PER_MESSAGE}(_messages[i].payload);
 
+            _rewardAmount = _rewardAmount + _messages[i].fee;
```
musnit, line 94: so, the reward is based on adding up all message fees. it would be worth double checking/auditing exactly how this fee variable can be controlled/changed.

so far it looks like it can only be controlled/changed by submitting a tx to the parachain to send a message through the channel, and only when burning the equivalent amount of eth on the parachain side equal to the fee that has been preset in config. looks good, though this is a critical field/code path, so plz audit yourself again too @philipstanislaus @vgeddes cc @Rizziepit

ethereum/contracts/IncentivizedInboundChannel.sol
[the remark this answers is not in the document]
musnit: when would this case ever occur? if it does we have a serious problem and should basically pause/shut down the bridge to allow for recovery from governance?

relayer/workers/parachaincommitmentrelayer/ethereum-channel-writer.go
musnit: ye sounds fine for now

test/test/ethapp.js
```diff
@@ -7,9 +7,13 @@ const { expect } = require("chai")
 
 const { polkadotRecipientSS58, polkadotRecipient, bootstrap } = require('../src/fixtures');
 
+const { ChannelId } = require("../src/helpers");
+
 describe('Bridge', function () {
 
   let ethClient, subClient;
+  let relayerAccount = "0x87D987206180B8f3807Dd90455606eEa85cdB87a";
```
musnit, line 15: can this be taken from accounts[xx] rather than hardcoded?

test/test/ethapp.js
```diff
@@ -40,26 +44,38 @@ describe('Bridge', function () {
...
-      expect(beforeSubBalance.minus(afterSubBalance)).to.be.bignumber.equal(amount);
+      expect(beforeSubBalance.minus(afterSubBalance)).to.be.bignumber.equal(amount.plus(fee));
       // conservation of value
-      expect(beforeEthBalance.plus(beforeSubBalance)).to.be.bignumber.equal(afterEthBalance.plus(afterSubBalance));
+      expect(beforeEthBalance.plus(beforeSubBalance))
+        .to.be.bignumber
+        .equal(afterEthBalance.plus(afterSubBalance).plus(fee));
+
+      // relayer should be rewarded fee
+      console.log(beforeRelayerBalance.toString());
```
musnit, line 76: can remove logs if tests are good

relayer/workers/parachaincommitmentrelayer/parachaincommitment/catchup.go
vgeddes: fixed

parachain/pallets/incentivized-channel/src/outbound/mod.rs
vgeddes: yeah, under the covers the amount is withdrawn from the account, and deducted from total issuance, so in effect is burned.

ethereum/contracts/IncentivizedInboundChannel.sol
vgeddes: The default per-message fee is set on the parachain side via genesis storage for the incentivized outbound channel. This fee can be updated though via the `set_fee` extrinsic (which is locked down via governance).

ethereum/contracts/BasicInboundChannel.sol
vgeddes: Sure ok. But in future, lets rather track this stuff via tickets or git branches. No reason to have commented-out code in main when people can track their in-progress work in side branches.

musnit: agree in general but in this case this is security-critical code that is essential for us to get done and it is a good thing for it to be messing all over the place and drawing attention/disrupting clean code :)

ethereum/contracts/IncentivizedInboundChannel.sol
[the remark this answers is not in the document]
vgeddes: > when would this case ever occur? if it does we have a serious problem and should basically pause/shut down the bridge to allow for recovery from governance?

The `RelayerNotRewarded` event is just so I can verify the unhappy path in unit tests. It may go away in future.

Agree that we should somehow shutdown the bridge when we hit this unhappy path, but I think that's better to implement in another PR.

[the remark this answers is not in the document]
musnit: @vgeddes otherwise we can just remove the try catch as this should never happen?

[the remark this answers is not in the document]
musnit: can also remove that test, as its not really an unhappy path test, it's testing something that should never happen

PR #393, opened 2021-05-11
WIP: Inbound channel verification rebased

blablacio:
This is a WIP implementing inbound channel verification through light client bridge.

Current blocker is that relay chain returns beefy commitment hash hashed with blake2b, which for performance reasons needs to be keccak256 instead.


ethereum/package-lock.json
```diff
@@ -0,0 +1,12486 @@
+{
```
philipstanislaus, line 1: Is this file committed intentionally? I believe we should choose to either use npm or yarn, but not both, since that could lead to undeterministic builds, depending on the build system.

relayer/workers/parachaincommitmentrelayer/parachaincommitment/listener.go
```diff
@@ -257,3 +367,93 @@ func (li *Listener) getMessagesForDigestItem(digestItem *chainTypes.AuxiliaryDig
...
+
+	for _, message := range messages {
+		if message.Nonce <= nonceToFind {
+			return true, nil
+		}
+	}
+	return false, nil
+}
+=======
+>>>>>>> 5b26e87... fixz
```
philipstanislaus, line 459: Merge markers?

philipstanislaus (commented):
Few comments, in addition to open notes on #374

musnit:
continued on https://github.com/Snowfork/polkadot-ethereum/pull/406

PR #400, opened 2021-05-21
Final channel implementations (Substrate)

vgeddes:
The basic channel, which has lagged a bit, now matches the incentivized channel in terms of implementation and error-handling. 

Additionally, the basic channel can only be used by an authorized account (the "principal"). Messages sent by other accounts will not be accepted for submission. This is just a stop-gap until we finish implementing V2 of the basic channel.

The principal account can be changed via governance using the `basicOutboundChannel.set_principal` extrinsic.


parachain/pallets/basic-channel/src/outbound/mod.rs
```diff
@@ -33,23 +92,128 @@ decl_event! {
...
+			if (now % Self::interval()).is_zero() {
+				Self::commit()
+			} else {
+				T::WeightInfo::on_initialize_non_interval()
+			}
+		}
+
+		#[weight = T::WeightInfo::set_principal()]
+		pub fn set_principal(origin, principal: <T::Lookup as StaticLookup>::Source) -> DispatchResult {
+			T::SetPrincipalOrigin::ensure_origin(origin)?;
```
musnit, line 125: does this check enforce that the original is the current principal?

musnit:
do e2e tests still pass with these changes? how do the accounts in the e2e tests that test the incentivized channels get their initial SnowDOT/SnowETH to pay for incentivized transfers?

vgeddes:
> do e2e tests still pass with these changes? how do the accounts in the e2e tests that test the incentivized channels get their initial SnowDOT/SnowETH to pay for incentivized transfers?

Running unit & E2E tests is standard procedure before I cut a PR, unless it's in draft mode 🙂

As for your latter question, that relates more to https://github.com/Snowfork/polkadot-ethereum/pull/391 which was merged in a few days ago. Anyway, the individual tests run in a specific sequence which ensures that the incentivised channel is full bootstrapped and tested.

parachain/pallets/basic-channel/src/outbound/mod.rs
vgeddes: Not really possible technically. The authorised origin (root or council) can never be the principal.

However the principal can be a member of the council.

musnit: what does this line/check do?

vgeddes: ensures that the dispatch origin is either root or a quorum of council members.

parachain/runtime/rococo/Cargo.toml
```diff
@@ -137,6 +135,7 @@ runtime-benchmarks = [
     "dispatch/runtime-benchmarks",
     "erc20-app/runtime-benchmarks",
     "eth-app/runtime-benchmarks",
+    "basic-channel/runtime-benchmarks",
```
Rizziepit, line 138: This is duplicated. See above

parachain/runtime/local/Cargo.toml
```diff
@@ -137,6 +135,7 @@ runtime-benchmarks = [
     "dispatch/runtime-benchmarks",
     "erc20-app/runtime-benchmarks",
     "eth-app/runtime-benchmarks",
+    "basic-channel/runtime-benchmarks",
```
Rizziepit, line 138: This is a duplicate. See above. It's alphabetical

Rizziepit (approved):
This looks good

Remember to re-generate weights on the benchmarking instance once runtime configuration has stabilized. Off the top of my head, the LC's `DescendantsUntilFinalized` value affects weights a lot. In future we might want a Github hook for this.

parachain/pallets/basic-channel/src/outbound/mod.rs
philipstanislaus: @vgeddes I guess we could also allow the principal to change the principal – would that make sense? 

Regarding implementation, did you see this: https://github.com/paritytech/substrate/blob/master/frame/assets/src/lib.rs#L479-L482

philipstanislaus (approved):
Nice, thanks 🎉

parachain/runtime/rococo/src/weights/basic_channel_outbound_weights.rs
```diff
@@ -0,0 +1,59 @@
+
```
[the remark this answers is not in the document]
vgeddes: Nah, I just copied the benches over from the incentivized channel, just to make the code build.  

Figured we'd rerun the benchmarks anyway at some point.

parachain/pallets/basic-channel/src/outbound/mod.rs
```diff
@@ -33,23 +92,128 @@ decl_event! {
...
-	pub fn submit(_: &T::AccountId, target: H160, payload: &[u8]) -> DispatchResult {
+
+	/// Submit message on the outbound channel
+	pub fn submit(who: &T::AccountId, target: H160, payload: &[u8]) -> DispatchResult {
+		ensure!(
+			*who == Self::principal(),
+			Error::<T>::NotAuthorized,
+		);
+		ensure!(
+			MessageQueue::decode_len().unwrap_or(0) < T::MaxMessagesPerCommit::get(),
```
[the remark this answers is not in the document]
vgeddes: The current code is correct - At the end of every commitment cycle, the vector stored under `MessageQueue` is deleted from storage. So when a new cycle begins, and a message is about to be added, `MessageQueue::decode_len` will duly fail and so we must transmute that failure to "the message queue is empty".

parachain/pallets/basic-channel/src/outbound/mod.rs
```diff
@@ -1,26 +1,85 @@
...
+	fn on_initialize_no_messages() -> Weight;
+	fn set_principal() -> Weight;
+}
+
+impl WeightInfo for () {
+	fn on_initialize(_: u32, _: u32) -> Weight { 0 }
+	fn on_initialize_non_interval() -> Weight { 0 }
+	fn on_initialize_no_messages() -> Weight { 0 }
+	fn set_principal() -> Weight { 0 }
+}
```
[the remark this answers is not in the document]
vgeddes: The "default" implementation doesn't really work like that. Its not a last-resort default-case in case an extrinsic isn't benchmarked.

Each runtime explicitly configures the `WeightInfo` implementations it wants. Here, the default implementation is necessary to configure a working mock runtime for unit tests, which don't really require accurate weight info.

parachain/pallets/basic-channel/src/outbound/mod.rs
vgeddes: I'd rather keep things simple for now and only authorize government agents (root or council). In practice, the principal is likely going to be on the council anyway, where they can propose a new principal or influence the vote.

parachain/runtime/snowbridge/Cargo.toml
```diff
@@ -137,6 +135,7 @@ runtime-benchmarks = [
     "dispatch/runtime-benchmarks",
     "erc20-app/runtime-benchmarks",
     "eth-app/runtime-benchmarks",
+    "basic-channel/runtime-benchmarks",
```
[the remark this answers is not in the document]
vgeddes: All fixed. Nice catch!

PR #401, opened 2021-05-22
Add limited support for hardhat

vgeddes:
This adds support for compiling and unit testing contracts using Hardhat and its built-in EVM. The existing Truffle configuration is left untouched, so all legacy contract development workflows still operate as normal.

Note that the E2E tests and the production migration script still use Truffle as before.

To use hardhat for running unit tests:
1. `yarn install`
2. `npx hardhat test`

### DIFF
--- a/ethereum/.gitignore
+++ b/ethereum/.gitignore
@@ -1,7 +1,6 @@
-build**
-
+build/
 .env
-
 node_modules/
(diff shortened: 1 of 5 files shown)


philipstanislaus (commented):
Thanks 💪

PR #403, opened 2021-05-25
migrate all workers to relay v2

musnit:


relayer/core/relay.go
```diff
@@ -1,222 +1,66 @@
...
-		beefyRelayer, err = beefyrelayer.NewWorker(&config.Relaychain, &config.Eth, &config.BeefyRelayerDatabase)
-		if err != nil {
-			return nil, err
-		}
-	}
-
-	var ethRelayer *ethrelayer.Worker
+	var pool workers.WorkerPool
 
 	if config.Workers.EthRelayer.Enabled {
```
Rizziepit, line 22: The `Enabled` checks are duplicative. The worker pool will discard workers that are disabled [here](https://github.com/Snowfork/polkadot-ethereum/blob/ceea5878f722f23906521ba39cdb8aa9f07c63dc/relayer/workers/worker.go#L120-L122). So, if you prefer, you can instantiate the pool without appending like:
```
pool := workers.WorkerPool{
    ethrelayerFactory,
    beefyrelayerFactory,
    parachaincommitmentrelayerFactory,
}
```

Rizziepit (approved):
Looks good. Yay for restarts!

philipstanislaus (approved):
LGTM

PR #405, opened 2021-05-25
Seperate relayer keys

musnit:
Solves https://github.com/Snowfork/polkadot-ethereum/issues/404

Note: Both keys are placed under the .Eth config, even though they are worker-specific. Ideally we have separate configs for each worker, but given atm they all have the same type/structure, it doesn't seem worth implementing this yet and I think using the global .Eth config is fine

FYI - this PR seems to make quite a big difference to reliability. I don't see any failure/restart in the workers now, even when running ~12 e2e tests across both channels in parallel :)


relayer/cmd/fetch_messages.go
```diff
@@ -83,13 +82,9 @@ func getEthContractEventsAndTrie(
...
 ) ([]*gethTypes.Log, *gethTrie.Trie, error) {
 	ctx := context.Background()
-	kp, err := secp256k1.NewKeypairFromString(config.PrivateKey)
-	if err != nil {
-		return nil, nil, err
-	}
 
-	conn := ethereum.NewConnection(config.Endpoint, kp, logrus.WithField("chain", "Ethereum"))
-	err = conn.Connect(ctx)
+	conn := ethereum.NewConnection(config.Endpoint, nil, logrus.WithField("chain", "Ethereum"))
```
philipstanislaus, line 86: Can we remove the kp argument from this function, if it's not used anymore?

philipstanislaus (approved):
LGTM 🎉

PR #407, opened 2021-05-27
change beefy writer to use keccak

musnit:
Now that the real lightclient checks have been enabled, the beefy writer is failing its transactions. we need to start fixing the data/proofs it sends to ensure it passes all checks.

the first failure/revert is due to the validatorRegistry.checkValidatorInSet check, which was failing because:
 - writer was using blake2b not keccak
 - writer was hashing validator addresses twice, rather than once before putting into merkle tree

This PR fixes those.

Next, the ECDSA.recover signature check is failing, but will address that in a future PR.


philipstanislaus (approved):
LGTM

denalimarsh (approved):
Awesome, thanks for hooking this up!

PR #411, opened 2021-05-31
Fix MMR peak bagging, add tests cases

philipstanislaus:
- Changes peak bagging to expect all right peaks to be bagged into one hash in the proof
- Copies @vgeddes' hardhat config to enable logging from Solidity (thanks 🎉 ) 
- MMR verification works now with the same data that substrate uses in https://github.com/paritytech/substrate/blob/87e63fa7c70be29383cb95ff52d71874f1db908a/frame/merkle-mountain-range/src/tests.rs#L263 
- Adds further test fixtures generated with substrate code


musnit:
@philipstanislaus is this based onto the right branch? or should it be onto main?

ethereum/contracts/MMRVerification.sol
```diff
@@ -92,17 +93,7 @@ contract MMRVerification {
         // TODO: Make this more elegant
         bytes32 rightPeaksHash;
         if (numRightPeaks > 0) {
-            if (numRightPeaks == 1) {
```
musnit, line 95: so the only real logic change is removing these 2 cases and just doing a simple one-line bag for both cases?

philipstanislaus: Exactly

philipstanislaus:
It is based on the correct branch since @denalimarsh has made further changes to fix MMR bagging.

philipstanislaus:
This is ready to be merged @musnit

musnit:
@philipstanislaus ok - can we get a PR targeting main? either this one or merge into dm-update-peak-bagging and then PR that?

philipstanislaus:
Replaced by #418

PR #415, opened 2021-06-01
minor bugfix

musnit:


musnit:
@denalimarsh going to merge this - plz review/lmk if something is wrong though

PR #418, opened 2021-06-06
Fix MMR peak bagging, add tests cases, rebased

philipstanislaus:
Replaces #411 

- Changes peak bagging to expect all right peaks to be bagged into one hash in the proof
- Copies @vgeddes' hardhat config to enable logging from Solidity (thanks 🎉 ) 
- MMR verification works now with the same data that substrate uses in https://github.com/paritytech/substrate/blob/87e63fa7c70be29383cb95ff52d71874f1db908a/frame/merkle-mountain-range/src/tests.rs#L263 
- Adds further test fixtures generated with substrate code


philipstanislaus:
@musnit this one is now ready to merge

ethereum/test/test_mmr_verification.js
```diff
@@ -52,7 +53,8 @@ describe("MMRVerification Contract", function () {
                 items: [
                     "0xff5d891b28463a3440e1b650984685efdf260e482cb3807d53c49090841e755f",
                     "0x00b0046bd2d63fcb760cf50a262448bb2bbf9a264b0b0950d8744044edf00dc3",
-                    "0x16de0900b57bf359a0733674ebfbba0f494e95a8391b4bfeae850019399f3ec0"
+                    mergeKeccak256("0xaf3327deed0515c8d1902c9b5cd375942d42f388f3bfe3d1cd6e1b86f9cc456c", // bag right hand side peaks keccak(right, left)
```
denalimarsh, line 56: +1 for this approach, much clearer how these hashes are generated now

ethereum/test/test_mmr_verification.js
```diff
@@ -392,4 +415,32 @@ describe("MMRVerification Contract", function () {
             expect(await mmrVerification.verifyInclusionProof.call(root, leafNodeHash, proof.leaf_index, proof.leaf_count, proof.items)).to.be.true
         });
     })
-})
+
+    context("7-leaf MMR from fixture", function () {
```
denalimarsh, line 419: nit: seems like the fixture tests can replace the 7-leaf and 15-leaf tests above? Fine to leave as is for this PR though

denalimarsh, line 419: nit: seems like the fixture test cases can now replace the 7-leaf and 15-leaf test cases above? No need to change in this PR though

denalimarsh (approved):
tACK, thanks again for cherry-picking commits as it makes the review process much easier

ethereum/test/test_mmr_verification.js
philipstanislaus: Yeah, good point – but more tests can't really hurt, can they?

philipstanislaus: ;) I'm leaving them in for now, since the data is different

PR #420, opened 2021-06-07
feat: remove no longer needed commitmentHash from LightClientBridge c…

philipstanislaus:
…ontract


ethereum/contracts/LightClientBridge.sol
```diff
@@ -291,10 +289,8 @@ contract LightClientBridge {
                 requiredNumOfSignatures
             );
 
-        /**
-         * @dev Encode and hash the commitment
-         */
```
philipstanislaus, line 296: This created an error, probably a bug in the Solidity compiler: 

```
./ethereum/contracts/LightClientBridge.sol:295:17: ParserError: Only state variables or file-level variables can have a docstring.
        bytes32 commitmentHash =
                ^------------^

Compilation failed. See above.
Truffle v5.3.8 (core: 5.3.8)
Node v14.16.1
```

musnit: weird

denalimarsh: Looks related to https://github.com/ethereum/solidity/issues/9139

musnit:
deprecating - no longer relevant

ethereum/contracts/LightClientBridge.sol
Vishwas1: same here coming with v0.8.0 how to solve this ?

PR #421, opened 2021-06-07
Confirm successful Eth header import in parachain

Rizziepit:
This PR makes the Eth -> Parachain relayer worker crash if we cannot confirm that a given header (and thus also the batched messages) were successfully imported.

The way I'm doing this is by adding a callback, `onProcessed`, to the outgoing extrinsic pool logic. `onProcessed` is called after we receive confirmation from the parachain that the extrinsic has been processed. `ParachainWriter` then queries the parachain to check that the header exists in storage.

I also restructured the `Header` type to make stuff like the header hash easier to access.

### DIFF
--- a/relayer/chain/ethereum/header.go
+++ b/relayer/chain/ethereum/header.go
@@ -12,6 +12,7 @@ import (
 	"github.com/sirupsen/logrus"
 	"github.com/snowfork/ethashproof"
 	"github.com/snowfork/ethashproof/ethash"
+	"github.com/snowfork/go-substrate-rpc-client/v2/scale"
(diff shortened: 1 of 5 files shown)


musnit:
looks good - if you've tested this by running it for a while and ensured things are running fine and e2e tests all pass, then i think we can merge

Rizziepit:
e2e tests all pass + it ran without a hitch over 100s of headers. It's possible that the header confirmation logic isn't airtight, but I don't think that's a problem. It will fix itself by restarting the worker.

In prod we should monitor logs for messages like "Header import failed for header..."

philipstanislaus (commented):
LGTM, thanks 👍

PR #417, opened 2021-06-04
Update polkadot dependencies

vgeddes:
Updated Polkadot/Cumulus dependencies to polkadot v0.9.4. This also required us to update GSRPC to get metadata V13 changes.

Note that XCM verification & testing is pushed out to a follow-up PR given the scope-creep that has arisen so far.

E2E tests mostly pass. For the tests that fail, there is an unrelated & pre-existing issue in the commitment relayer which results in some commitments getting ignored and are not being passed on to Ethereum.

Changes:
* Parachain
  * Update deps, runtime code, and chain specs to work with polkadot v0.9.4
  * XCM: Remove obsolete `transfer` pallet. We no longer need a custom pallet to allow users to initiate XCM transfers.
  * Use Aura for block authoring
  * Block time increased to 12 secs to match reality around current relay chain performance.
* Relayer:
  * Update GSRPC and go-ethereum deps
  * Update receipt trie generation to work with go-ethereum 1.10.3
* GSRPC (branch https://github.com/Snowfork/go-substrate-rpc-client/tree/centrifuge-rebased):
  * Contains latest upstream changes
  * Various bug & polkadot v0.9.4 compatibility fixes for the following issues:
    *  https://github.com/paritytech/substrate/issues/9137
    * https://github.com/centrifuge/go-substrate-rpc-client/pull/171
    * https://github.com/centrifuge/go-substrate-rpc-client/issues/168
 * E2E Stack
   * Requires latest polkadot-launch.
   * Need to run 2 collator nodes (alice & bob) as a minimum.
   * Expose unsafe RPC methods (required for offchain DB queries in Polkadot v0.9.4)

PR #422, opened 2021-06-10
run 2 parachains for xcm testing

musnit:
e2e tests all still pass

### DIFF
--- a/parachain/config-dup.json
+++ b/parachain/config-dup.json
@@ -20,6 +20,15 @@
 					"--enable-offchain-indexing=true",
 					"--offchain-worker=Always"
 				]
+			},
+			{
+				"name": "eve",
+				"wsPort": 9956,
+				"port": 30556,
+				"flags": [
+					"--enable-offchain-indexing=true",
+					"--offchain-worker=Always"
+				]
 			}
 		]
 	},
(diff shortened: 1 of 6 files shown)


philipstanislaus (commented):
Nice! Do you plan to merge this? How big is the overhead of running this locally? Just wondering how much more complicated the dev setup gets.

musnit:
overhead is minimal/nothing. waiting on @vgeddes to share comments before merging though.

vgeddes:
We'll likely want something like this in future. Right now though, I think its a bit premature to add this functionality to our E2E stack.

In the short term (milestone V1), this testing feature will likely not be used at all, as its more straightforward to test simple XCM asset transfers using `polkadot-launch parachain/config-xcmp.json`. No need to run the relayer, and deploy contracts, etc. It also gives us more flexibility to customize parachain code/config for testing different XCM scenarios, instead of the structure mandated by the E2E stack.

Then in the longer term, as we evolve XCM and bridge messaging, the solution implemented in this PR won't be sufficient. Instead of a running an duplicate of our parachain, we'll need to run a custom parachain with pallets and config that differs from snowbridge.

musnit:
@vgeddes we need to be able to test xcm e2e, including from the frontend. i don't really see much issue with adding this to our e2e stack. it will be used and the additional is pretty minimal.

musnit:
can put it in a different script if u prefer but i think an e2e script with both is still important

musnit:
@vgeddes this is actually needed to test basic functionality of the grandpa light client properly when testing the parachain merkle proofs as the merkle tree should be tested with more than 1 leaf.

i've also updated the script to take an argument "duplicate" such that it will run 1 parachain without the arg and 2 duplicated parachains with it, so should still be compatible with your needs for xcm testing/manually running a second parachain :)

musnit:
e2e tests all pass in both configurations, with 2 or 3 chains running fine in each configuration, so going to merge

vgeddes:
Cool, sounds good. On a related note, in #417 we're also gonna have to run at least 2 collators per parachain due to a change in block authoring logic in Cumulus.

PR #423, opened 2021-06-10
Use contract's BEEFY commitment ID in relayer when possible

denalimarsh:
Refactored the relayer BEEFY store's test suite as part of this PR after finding a bug in the store's `GetItemByCompleteVerificationTxHash` method.

### DIFF
--- a/ethereum/contracts/LightClientBridge.sol
+++ b/ethereum/contracts/LightClientBridge.sol
@@ -83,7 +83,7 @@ contract LightClientBridge {
 
     uint256 public constant THRESHOLD_NUMERATOR = 2;
     uint256 public constant THRESHOLD_DENOMINATOR = 3;
-    uint256 public constant BLOCK_WAIT_PERIOD = 45;
+    uint256 public constant BLOCK_WAIT_PERIOD = 3;
 
     /**
      * @notice Deploys the LightClientBridge contract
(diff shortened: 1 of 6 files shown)


musnit:
added some commits with minor fixes and to change the complete tx to actually use the nonce. i think this is good to merge now.

ethereum/contracts/LightClientBridge.sol
```diff
@@ -83,7 +83,7 @@ contract LightClientBridge {
 
     uint256 public constant THRESHOLD_NUMERATOR = 2;
     uint256 public constant THRESHOLD_DENOMINATOR = 3;
-    uint256 public constant BLOCK_WAIT_PERIOD = 45;
+    uint256 public constant BLOCK_WAIT_PERIOD = 3;
```
philipstanislaus, line 86: Shall we create a ticket to reset this to a production valu? Or did we discuss with Web3 Foundation that this should be 3?

philipstanislaus (commented):
Nice, thanks @denalimarsh 💪

ethereum/contracts/LightClientBridge.sol
musnit: https://github.com/Snowfork/polkadot-ethereum/issues/426

PR #424, opened 2021-06-10
Update beefy relayer to send validator positions, signatures, pubkeys and pubkey merkle proofs correctly

musnit:
Ready for review. The beefy-relayer is working now and completing the full initial+complete flow with all proofs/checks passing 🎉 🎉 🎉 

Some changes to smart contracts for review - plz review and audit these properly:
 - requiredNumberOfSignatures now moved to its own function and made to round up the calculation (eg: 2 validators with 2/3 now rounds up to 2 sigs required instead of 1)
 - createRandomBitfield added to smart contracts to generate a brand new bitfield with length based on rounding up to the nearest 256
 - validatorClaimsBitfield.countSetBits is now >= 2/3, rather than > 2/3 because the requiredNumberOfSignatures is rounded up. seems fine?

### DIFF
--- a/ethereum/contracts/utils/Bitfield.sol
+++ b/ethereum/contracts/utils/Bitfield.sol
@@ -79,7 +79,7 @@ library Bitfield {
         bitfield = new uint256[](arrayLength);
 
         for (uint256 i = 0; i < bitsToSet.length; i++) {
-            set(bitfield, i);
+            set(bitfield, bitsToSet[i]);
         }
 
         return bitfield;
--- a/ethereum/contracts/utils/Bitfield.sol
+++ b/ethereum/contracts/utils/Bitfield.sol
@@ -68,33 +68,18 @@ library Bitfield {
         return bitfield;
     }
 
-    /**
-     * @notice Draws a random number, derives an index in the bitfield, and sets the bit if it is in the `prior` and not
-     * yet set. Repeats that `n` times.
-     */
-    function randomNBits(uint256 seed, uint256 n)
+    function createBitfield(uint256[] calldata bitsToSet, uint256 length)
         public
         pure
         returns (uint256[] memory bitfield)
     {
         // Calculate length of uint256 array based on rounding up to number of uint256 needed
-        uint256 arrayLength = (n + 255) / 256;
+        uint256 arrayLength = (length + 255) / 256;
 
         bitfield = new uint256[](arrayLength);
-        uint256 found = 0;
 
-        for (uint256 i = 0; found < n; i++) {
-            bytes32 randomness = keccak256(abi.encode(seed + i));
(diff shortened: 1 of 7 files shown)


relayer/workers/beefyrelayer/beefy-ethereum-writer.go
```diff
@@ -99,14 +101,30 @@ func (wr *BeefyEthereumWriter) WriteNewSignatureCommitment(ctx context.Context,
...
-	msg, err := beefyJustification.BuildNewSignatureCommitmentMessage(valIndex)
+	contract := wr.lightClientBridge
+	if contract == nil {
+		return fmt.Errorf("Unknown contract")
+	}
+
+	signedValidators := []*big.Int{}
+	for i := range beefyJustification.SignedCommitment.Signatures {
+		// TODO: skip over empty/missing signatures
+		// if signature.Option.IsSome() {
```
musnit, line 112: @philipstanislaus not sure, this could be a GSRPC decode bug or potentially a bug in the beefy code that generates this data onchain, but when decoding, i’m getting entries that have a correct value/signature but that have the Option bool set to false, ie, signature.Option.IsSome() is always false

ethereum/contracts/utils/Bitfield.sol
```diff
@@ -68,6 +68,23 @@ library Bitfield {
...
+        pure
+        returns (uint256[] memory bitfield)
+    {
+        // Calculate length of uint256 array based on rounding up to number of uint256 needed
+        uint256 arrayLength = (length + 255) / 256;
+
+        bitfield = new uint256[](arrayLength);
+
+        for (uint256 i = 0; i < bitsToSet.length; i++) {
+            set(bitfield, i);
```
philipstanislaus, line 82: This should be: 

```suggestion
            set(bitfield, bitsToSet[i]);
```

relayer/workers/beefyrelayer/beefy-ethereum-writer.go
philipstanislaus: @musnit I'll have a look at that tomorrow, but sounds like it's an issue with Beefy.

philipstanislaus (changes requested):
Awesome, just a few minor notes 🎉

denalimarsh (commented):
LGTM pending the revisions from @philipstanislaus

relayer/workers/beefyrelayer/beefy-ethereum-writer.go
musnit: cool, created a task

ethereum/contracts/utils/Bitfield.sol
musnit: nice catch thx

PR #430, opened 2021-06-17
Malicious dot app and gas controls test

musnit:
This PR creates a malicious smart contract and tests that it does not affect the bridge as a whole and does not impact other apps. See README for more.

still WIP


musnit:
@vgeddes when i lock up an erc20 i get the message dispatched across into the erc20 pallet but the mint function returns this:

image

I guess the erc20 app address is not being set properly? though it looks correct:
```bash
➜  polkadot-ethereum git:(malicious-gas) ✗ jq '.genesis.runtime.erc20App.address' /tmp/snowbridge-e2e-config/spec.json
"0x6495675F661E208CC34d8e163CF1f5b2d6d05129"
```

any thoughts what could be wrong?

vgeddes:
> any thoughts what could be wrong? does it need to be all uppercase/lowercase or anything like that?

Can you perform a storage query via the polkadot UI to verify that the new address was in place?

The capitalization of the hexadecimal shouldn't matter I think.

musnit:
@philipstanislaus @vgeddes this is ready for review

philipstanislaus (approved):
LGTM, nice!

PR #433, opened 2021-06-18
Channel relay chain proof integration aidan

musnit:
@philipstanislaus starting to work on improvements/fixes in here in a different PR in case i mess things up too much - will reconcile things once code is in a better state with e2e tests working again


ethereum/contracts/BasicInboundChannel.sol
```diff
@@ -22,60 +22,39 @@ contract BasicInboundChannel {
...
-            beefyLightClient.verifyBeefyMerkleLeaf(
-                beefyMMRLeaf,
-                _beefyMMRLeafIndex,
-                _beefyMMRLeafCount,
-                _beefyMMRLeafProof
-            ),
-            "Invalid proof"
+            gasleft() >= (_messages.length * MAX_GAS_PER_MESSAGE) + GAS_BUFFER,
+            "insufficient gas for delivery of all messages"
         );
```
philipstanislaus, line 58: I did actually move this up because the check is cheap and we could fail quicker that way

ethereum/contracts/DOTApp.sol
```diff
@@ -45,24 +47,38 @@ contract DOTApp is FeeSource, AccessControl {
         c2.outbound = _incentivized.outbound;
 
         _setupRole(FEE_BURNER_ROLE, feeBurner);
+        _setupRole(INBOUND_CHANNEL_ROLE, _basic.inbound);
+        _setupRole(INBOUND_CHANNEL_ROLE, _incentivized.inbound);
```
philipstanislaus, line 51: Nice

ethereum/contracts/IncentivizedInboundChannel.sol
```diff
@@ -54,66 +55,37 @@ contract IncentivizedInboundChannel is AccessControl {
...
-        bytes32 _commitment,
-        bytes32 _beefyMMRLeaf,
+        ParachainLightClient.OwnParachainHeadPartial
+            calldata _ownParachainHeadPartial,
+        bytes32[] calldata _parachainHeadsProof,
+        ParachainLightClient.BeefyMMRLeafPartial calldata _beefyMMRLeafPartial,
         uint256 _beefyMMRLeafIndex,
         uint256 _beefyMMRLeafCount,
-        bytes32[] memory _beefyMMRLeafProof
+        bytes32[] calldata _beefyMMRLeafProof
```
philipstanislaus, line 66: Shall we wrap all of these types in a simple `ParachainLightClient.CommitmentProof` struct? That would simplify the API here a bit.

relayer/workers/parachaincommitmentrelayer/beefy-listener.go
```diff
@@ -269,96 +260,15 @@ func (li *BeefyListener) GetAllParaheads(blockHash types.Hash, ourParachainId ui
...
-	return &Keccak256{}
-}
-
-// Hash generates a Keccak256 hash from a byte array
-func (h *Keccak256) Hash(data []byte) []byte {
-	hash := crypto.Keccak256(data)
-	return hash[:]
+func createParachainHeaderProof(allParaHeads []types.Header, ourParaHead types.Header) string {
+	//TODO: implement
+	return ""
```
philipstanislaus, line 265: Is the code above you removed no longer applicable?

philipstanislaus (commented):
Great 👍

philipstanislaus (approved):

ethereum/contracts/BasicInboundChannel.sol
musnit: ah but then we'd need to make sure there is also a buffer to check for doing all the proof checks and everything, so more complex to analyse securely, so i thought just safer to put it at the end and add a small simple buffer

ethereum/contracts/IncentivizedInboundChannel.sol
musnit: yeh probably a good idea to do in a follow up PR

relayer/workers/parachaincommitmentrelayer/beefy-listener.go
musnit: it is in another open PR, so I was planning to merge this one to main and let it continue in that PR

musnit:
e2e tests work and code is in a good position now for smaller easier PRs, so going to merge this to main and continue in smaller PRs.

PR #436, opened 2021-06-22
First draft of Beefy Light client encoding and hashing mmr leaf and checking mmr proof

musnit:
It works! Full beefy commitments with mmr verification of leaves and type construction :).

ethereum tests pass, e2e tests pass

There were a lot of changes to the smart contracts in this too, including updating to the latest solidity, so a bunch to review.

Also, note that the Scale compact uint encoding has not yet been implemented and is just hardcoded - will possibly need to do this based on https://substrate.dev/docs/en/knowledgebase/advanced/codec#compactgeneral-integers

Beefy relaying in realtime still seems to have some issue though - will debug next


vgeddes (commented):
Its hard to review this effectively as there's a lot of whitespace churn. I suspect you've got `editor.formatOnSave` enabled in VS Code.

While format-on-save is a really good practice :+1:, we need to agree on the formatting style and config/plugins. Otherwise we'll have whitespace churn in our history due to different formatting styles, which makes historical analysis more difficult (git blame would be less useful).

So I suggest we leave the reformatting for a separate dedicated PR. Once that's merged, we can collectively start using format-on-save. This is something I've been wanting to do for a while.

Anyhow, thanks for upgrading us to Solidity 0.8. Been wanting to do that for a while :)

ethereum/contracts/BeefyLightClient.sol
```diff
@@ -260,7 +259,12 @@ contract BeefyLightClient {
         bytes32[] calldata mmrProofItems
     ) public {
         verifyCommitment(id, commitment, validatorProof);
-
+        verifyNewestMMRLeaf(
+            latestMMRLeaf,
+            mmrProofItems,
+            commitment.payload,
+            commitment.blockNumber
+        );
```
philipstanislaus, line 267: This change implies that any commitment verification needs the full leaf, even if no validator set change is required. That increases gas consumption quite a bit I assume – do you plan to separate the two types of commitments in a later PR?

ethereum/contracts/ScaleCodec.sol
```diff
@@ -11,6 +11,11 @@ library ScaleCodec {
         return number;
     }
 
+    function encodeUintCompact(uint16 number) public pure returns (bytes2) {
+        // TODO - remove hardcoded implementation
+        return bytes2(uint16(0xc101));
```
philipstanislaus, line 16: If we leave the function in here, we should indeed use a working generic implementation. 

For the MMR leaf, this is not necessary though, since the length of that type is fixed. The easiest workaround is to just use the fixed length prefix as a constant in the `encodeMMRLeaf` function in `ethereum/contracts/BeefyLightClient.sol`.

ethereum/test/test_beefy_light_client.js
```diff
@@ -28,29 +28,35 @@ describe("Beefy Light Client", function () {
...
 
-    await this.beefyLightClient.newSignatureCommitment(
-      fixture.commitmentHash,
+    const commitmentHash = await this.beefyLightClient.createCommitmentHash(fixture.commitment);
+
+    const tx = await this.beefyLightClient.newSignatureCommitment(
+      commitmentHash,
       initialBitfield,
-      signatureSubstrateToEthereum(fixture.signature0),
+      fixture.signatures[0],
```
philipstanislaus, line 55: If you do the signature conversion when creating the fixture now, shall we add a section to the README on how to update the fixture?

relayer/workers/beefyrelayer/fixture-data-logger.go
```diff
@@ -0,0 +1,136 @@
...
+		"Gas":      tx.Gas(),
+		"GasPrice": tx.GasPrice(),
+		"Value":    tx.Value(),
+		"Data":     hex.EncodeToString(tx.Data()),
+	}).Info("Call info")
+
+	res, err := client.CallContract(context.Background(), msg, nil)
+	if err != nil {
+		return "", err
+	}
```
philipstanislaus, line 133: What is the purpose of this function? It's in a logger file name, but does an actual call?

philipstanislaus (approved):
Awesome to see this coming together and working! 🍾 

Left a few comments

ethereum/contracts/ScaleCodec.sol
```diff
@@ -71,20 +76,40 @@ library ScaleCodec {
         v = input;
 
         // swap bytes
-        v = ((v & 0xFF00FF00FF00FF00FF00FF00FF00FF00FF00FF00FF00FF00FF00FF00FF00FF00) >> 8) |
-            ((v & 0x00FF00FF00FF00FF00FF00FF00FF00FF00FF00FF00FF00FF00FF00FF00FF00FF) << 8);
+        v =
```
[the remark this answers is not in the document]
musnit: https://github.com/Snowfork/polkadot-ethereum/pull/442/commits/3e66b4d2116c2f10f692f068577d8c1ecd3648a6

ethereum/contracts/BeefyLightClient.sol
musnit: Well we need to either:
 - always get the full leaf so that we can confirm the validator set id has not changed
 - check that the new change is within the same epoch so that we can assume there was no validator set change

adding a separate type would involve needing to implement the latter. we currently just track a MAXIMUM_BLOCK_GAP rather than tracking epochs/timestamps specifically, so it may not be that easy to check for an epoch change (eg: if 2 justifications come in a row at block 900 and block 902 but there was an epoch change/validator set change at 901) and would presumably add gas consumption to do.

there could be a cheaper/easier way, but i haven't thought about it too deeply yet and also given that the gas consumption addition for always sending the leaf is constant and not that significant, i'd say it's premature optimisation that we only need to worry about in future when we past the lower hanging fruit.

ethereum/contracts/ScaleCodec.sol
musnit: https://github.com/paritytech/polkadot/blob/7ab51d455d11bdaeacd923b279b912b92f0bc580/runtime/common/src/mmr.rs#L65 yeh looks like it is

musnit: https://github.com/Snowfork/polkadot-ethereum/pull/442/commits/2f82fc9240be1128bbbce0807e5b1f76fdd303be

ethereum/test/test_beefy_light_client.js
musnit: https://github.com/Snowfork/polkadot-ethereum/pull/442/commits/6919266bcae1c8cdb7318e3bdf8e7a99e9f045fc

relayer/workers/beefyrelayer/fixture-data-logger.go
musnit: yeh, it's just a call (basically simulates the tx without actually sending it), not an actual tx. added some info:
https://github.com/Snowfork/polkadot-ethereum/pull/442/commits/d33920b8b1feee369b1f0c26bef46e2c829e9c02

ethereum/contracts/BeefyLightClient.sol
philipstanislaus: Fully agree with you, my point here is that we already have that information in the commitment (from [here](https://github.com/Snowfork/polkadot-ethereum/blob/main/ethereum/contracts/BeefyLightClient.sol#L63-L67)): 

```Solidity
    struct Commitment {
        bytes32 payload;
        uint64 blockNumber;
        uint32 validatorSetId;
    }
```

So we only need the Beefy leaf once for the set change.

But agree, can optimize for that at a later point.

musnit: oh right, ofc. yep makes sense :)

PR #438, opened 2021-06-24
Process validator set change

musnit:
Beefy processes the validator set change now. A bunch of smart contract code modified here - plz review.

Future Todo: Add a test which actually tests the validator set change


parachain/src/chain_spec/local.rs
```diff
@@ -1,23 +1,20 @@
 use cumulus_primitives_core::ParaId;
 use hex_literal::hex;
 use local_runtime::{
-	AccountId, EthereumHeader,
```
vgeddes, line 4: Woah :astonished:

Please revert. You're reformatting the entire file. Its also not the correct formatting style. We use tabs instead of spaces for Rust code.

I suspect you've got format-on-save enabled in VS Code for both Rust and Solidity code. While that's actually good practice, we need to agree on a common formatting style, and then make sure our editors all use the same config. Otherwise we're going to have lots of whitespace/formatting churn throughout our commits that make it harder to use `git blame` for historical analysis.

So I suggest you disable format-on-save for now.

parachain/src/chain_spec/local.rs
```diff
@@ -26,84 +23,81 @@ use artemis_core::AssetId;
...
+        Some(props),
+        Extensions {
+            relay_chain: "local_testnet".into(),
+            para_id: para_id.into(),
+        },
+    )
 }
 
 /// Configure initial storage state for FRAME modules.
-fn testnet_genesis(
```
vgeddes, line 102: ditto. See comment above. This formatting change is unnecessary.

parachain/src/chain_spec/rococo.rs
```diff
@@ -1,23 +1,20 @@
 use cumulus_primitives_core::ParaId;
 use hex_literal::hex;
 use rococo_runtime::{
-	AccountId, EthereumHeader,
```
vgeddes, line 4: ditto. Seem comment above about formatting.

parachain/src/chain_spec/snowbridge.rs
```diff
@@ -1,23 +1,20 @@
 use cumulus_primitives_core::ParaId;
 use hex_literal::hex;
-use snowbridge_runtime::{
```
vgeddes, line 3: ditto. See comment above about formatting.

ethereum/contracts/BeefyLightClient.sol
```diff
@@ -395,12 +408,11 @@ contract BeefyLightClient {
             "Error: Block wait period not over"
         );
 
-        uint256[] memory randomBitfield =
-            Bitfield.randomNBitsWithPriorCheck(
-                getSeed(data),
-                data.validatorClaimsBitfield,
-                requiredNumOfSignatures
-            );
+        uint256[] memory randomBitfield = Bitfield.randomNBitsWithPriorCheck(
```
vgeddes, line 411: Lets rather agree on & perform formatting changes in a separate PR.

ethereum/contracts/BeefyLightClient.sol
```diff
@@ -528,15 +540,14 @@ contract BeefyLightClient {
         pure
         returns (bytes memory)
     {
-        bytes memory scaleEncodedMMRLeaf =
```
vgeddes, line 531: ditto. Lets rather agree on & perform formatting changes in a separate PR.

ethereum/contracts/BeefyLightClient.sol
```diff
@@ -93,32 +93,38 @@ contract BeefyLightClient {
     Blake2b public blake2b;
     uint256 public currentId;
     bytes32 public latestMMRRoot;
+    uint64 public latestBeefyBlock;
```
philipstanislaus, line 96:
```suggestion
    uint64 public latestPolkadotBlockNumber;
```

philipstanislaus: Since Beefy has no blocks.

philipstanislaus (commented):
Cool 👏

philipstanislaus (approved):

parachain/src/chain_spec/local.rs
musnit: https://github.com/Snowfork/polkadot-ethereum/pull/442/commits/a31d23b3ae9b5d3f5a697ddb62b138c0c3225ad5 and https://github.com/Snowfork/polkadot-ethereum/pull/442/commits/8f9accfdf79d7b5bf931ae623a8eb55fbfc8af3c

parachain/src/chain_spec/rococo.rs
musnit: https://github.com/Snowfork/polkadot-ethereum/pull/442/commits/a31d23b3ae9b5d3f5a697ddb62b138c0c3225ad5 and https://github.com/Snowfork/polkadot-ethereum/pull/442/commits/8f9accfdf79d7b5bf931ae623a8eb55fbfc8af3c

parachain/src/chain_spec/snowbridge.rs
musnit: https://github.com/Snowfork/polkadot-ethereum/pull/442/commits/a31d23b3ae9b5d3f5a697ddb62b138c0c3225ad5 and https://github.com/Snowfork/polkadot-ethereum/pull/442/commits/8f9accfdf79d7b5bf931ae623a8eb55fbfc8af3c

ethereum/contracts/BeefyLightClient.sol
musnit: this file has been changed all over anyway, so it's not feasible nor important to worry about whitespace mismatch for it i think.

ethereum/contracts/BeefyLightClient.sol
musnit: this file has been changed all over anyway, so it's not feasible nor important to worry about whitespace mismatch for it i think.

ethereum/contracts/BeefyLightClient.sol
musnit: well it refers to the latest block signed by beefy, so think it's still fine/makes sense. also this is a beefy light client not a polkadot light client so maybe better to keep it technically correct. anyway doesn't really matter.

philipstanislaus: Fair point. A better name might be `latestRelayBlockNumber`, but agree, let's leave it as is ;)

PR #439, opened 2021-06-25
Get real root working

musnit:
This is a draft PR experimenting with getting real merkle roots from real MMRLeaves working. Seems like they are generated via base-16 modified merkle patricia tries, which is different to the normal merkle tree's we've had working so far, so we may be blocked unless we can change them or rewrite all of the solidity merkle tree verifier AND js proof creation AND golang proof creation

Not intended to be merged


musnit:
deprecated for https://github.com/Snowfork/snowbridge/pull/481

PR #441, opened 2021-06-25
Parachain commitments catchup

musnit:
This PR is fixing/implementing the parachain commitment relayer catchup mechanism. This will be used both for graceful restarting of the worker, as well as for handling beefy commitments that skip a few parachain blocks.

Still draft/WIP


relayer/workers/parachaincommitmentrelayer/beefy-listener.go
```diff
@@ -156,7 +164,7 @@ func (li *BeefyListener) processBeefyLightClientEvents(ctx context.Context, even
 		// the first parachain block that has not yet been fully processed on ethereum,
 		// and then package and relay all newer heads/commitments together with their corresponding leaf
 		mmrProof := li.relaychainConn.GetMMRLeafForBlock(uint64(relayChainBlockNumber-1), blockHash)
-		allParaHeads, ourParaHead := li.GetAllParaheads(blockHash, OUR_PARACHAIN_ID)
+		allParaHeads, ourParaHead := li.relaychainConn.GetAllParaheadsWithOwn(blockHash, OUR_PARACHAIN_ID)
```
vgeddes, line 167: There's still the fundamental issue in that we're only analysing a single MMR leaf in between justifications, meaning we're skipping over para heads that could contain commitments.

Its not clear to me how this would be mitigated by fixing catchup. How would the worker know when to restart and catchup? Even if we overcome that, it just doesn't seem right for the worker to crash and restart so very often. This perhaps points to a need to redesign the listener again, and decoupling commitment listening from justification listening.

Anyhow, the situation with Polkadot v0.9.4 is that justifications get emitted periodically (not sure if its regular or not). For example, suppose justifications get emitted at relay chain block 0, 10, 20. Furthermore, our parachain produces new blocks on the same interval as the relay chain. This code, which seems limited to discovering a single para head in between justifications, would only discover 3 of our parachain heads, and ignore 17 other heads. So we're potentially ignoring 17 commitments.

So it seems we possibly do need to be analyzing `Paras.Heads` over a range of relay chain blocks `[M, N]`, where M and N are the relay chain blocks for two successive justifications?

vgeddes (commented):
Added some comments inline. While we certainly should fix catchup, I'm not sure how effective its going to be in fixing the changed nature of beefy subscriptions in my polkadot 0.9.4 branch: #417

relayer/workers/parachaincommitmentrelayer/catchup.go
```diff
@@ -185,7 +187,7 @@ func (li *Listener) searchForLostCommitments(lastBlockNumber uint64, basicNonceT
 	return nil
```
philipstanislaus, line 187: Lost commitments are not returned/processed here, will that be part of this PR? Or do we have a task for that?

relayer/workers/parachaincommitmentrelayer/beefy-listener.go
```diff
@@ -77,7 +77,15 @@ func (li *BeefyListener) Start(ctx context.Context, eg *errgroup.Group) error {
...
-		err := li.subBeefyJustifications(ctx)
+
+		blockNumber, hash, err := li.fetchLatestBlockAndHash()
+		if err != nil {
+			return nil
+		}
+
+		li.catchupMissedCommitments(ctx, blockNumber, hash)
+
+		err = li.subBeefyJustifications(ctx)
```
philipstanislaus, line 88: Catchup could take longer than one justification, so we could miss additional commitments. We might need to change the catchup logic to run within every justification we receive.

relayer/workers/parachaincommitmentrelayer/beefy-listener.go
philipstanislaus: Good observation – the original architecture foresaw to have them decoupled. The idea was to listen to any new commitments and put them in a queue. Whenever a new beefy MMR root is observed, check the queue for commitments that can be proven now, and submit them one by one.

musnit:
replacing with https://github.com/Snowfork/polkadot-ethereum/pull/444

relayer/chain/relaychain/connection.go
```diff
@@ -101,3 +102,79 @@ func (co *Connection) GetMMRLeafForBlock(
...
+		for _, change := range headerResponse.Changes {
+
+			// TODO fix this manual slice with a proper type decode. only the last few bytes are for the ParaId,
+			// not sure what the early ones are for.
+			key := change.StorageKey[40:]
+			var parachainID types.U32
+			if err := types.DecodeFromBytes(key, &parachainID); err != nil {
+				co.log.WithError(err).Error("Failed to decode parachain ID")
+			}
+
```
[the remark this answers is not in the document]
musnit: makes sense, will skip over when its not our ID

relayer/workers/parachaincommitmentrelayer/beefy-listener.go
musnit: > So it seems we possibly do need to be analyzing `Paras.Heads` over a range of relay chain blocks `[M, N]`, where M and N are the relay chain blocks for two successive justifications?

@vgeddes this PR is still WIP, sorry - should have mentioned that the review is a bit premature. the intention is to implement exactly this and to use it for both the catchup and for the normal relaying flow

relayer/workers/parachaincommitmentrelayer/catchup.go
musnit: yep in progress in this PR

relayer/workers/parachaincommitmentrelayer/beefy-listener.go
musnit: yeh planning to implement in this PR.

PR #444, opened 2021-06-28
Parachain commitments catchup rebased

musnit:
This PR is fixing/implementing the parachain commitment relayer catchup mechanism. This will be used both for graceful restarting of the worker, as well as for handling beefy commitments that skip a few parachain blocks.

Still draft/WIP


musnit:
replacing with https://github.com/Snowfork/polkadot-ethereum/pull/446

PR #448, opened 2021-06-30
Parachain proof

musnit:
Creating merkle proof of our parachain's header within the relayer.

e2e tests pass

This is untested (because the proofs are not actually checked properly yet) and will likely need some modification when the proof production on the parachain has been fixed (eg: may need to SCALE encode again and should add some tests) but I think is ready for review/merge now to get the core pieces in place.


ethereum/contracts/utils/MerkleProof.sol
```diff
@@ -61,4 +61,53 @@ library MerkleProof {
 
         return computedHash == root;
     }
+
+    function computeMerkleLeafAtPosition(
```
musnit, line 65: @philipstanislaus i presume we still need the position to compute the root?

if so, can we just use this code below but revert instead of "return false;" ?

ethereum/contracts/utils/MerkleProof.sol
```diff
@@ -61,4 +61,48 @@ library MerkleProof {
...
+        bytes32 leaf,
+        uint256 pos,
+        uint256 width,
+        bytes32[] calldata proof
+    ) public pure returns (bytes32) {
+        bytes32 computedHash = leaf;
+
+        require(pos + 1 > width, "Merkle position is too high");
```
philipstanislaus, line 73:
```suggestion
        require(pos < width, "Merkle position is too high");
```

ethereum/contracts/utils/MerkleProof.sol
```diff
@@ -61,4 +61,48 @@ library MerkleProof {
...
+
+            // check if at rightmost branch and whether the computedHash is left
+            if (pos + 1 == width && computedHashLeft) {
+                // there is no sibling and also no element in proofs, so we just go up one layer in the tree
+                pos /= 2;
+                width = ((width - 1) / 2) + 1;
+                continue;
+            }
+
+            require(i >= proof.length, "Merkle proof is too short");
```
philipstanislaus, line 87:
```suggestion
            require(i < proof.length, "Merkle proof is too short");
```

ethereum/contracts/utils/MerkleProof.sol
philipstanislaus: Yes, definitely. Fixed just two minor issues. Also should name this function: 

```suggestion
    function computeRootFromProofAtPosition(
```

philipstanislaus: Can we use this function in the `verifyMerkleLeafAtPosition` above?

philipstanislaus (commented):
Only checked out the MerkleProof.sol file so far

musnit:
@philipstanislaus suggested fixes have been made

relayer/chain/parachain/header_proof.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+
+	// Create the tree
+	paraMerkleTree, err := merkletree.NewUsing(paraTreeData, &Keccak256{}, nil)
+	if err != nil {
+		return [][32]byte{}, err
+	}
+
+	// Generate Merkle Proof for our parachain's head
+	proof, err := paraMerkleTree.GenerateProof(ourParaHeadBytes)
```
philipstanislaus, line 29: The library used here may create a different merkle tree from our solidity implementation, see comment above. Let’s see what Thomasz implements and then adjust.

philipstanislaus (approved):
Great 👍

PR #449, opened 2021-07-05
Improve logging and create better test fixtures

musnit:
Unit tests now use real data with all the details of all the proofs, especially the channel tests that now test real mmr/beefy proof connection.

This may not add much yet, as the integration tests cover the above already, but as we extend the unit tests to test changes in the proofs and different steps of the verification process this will be required

This also removes the mocks in our channel tests. For further testing of actual verification functionality, we'd need either more comprehensive mocks that mock out the beefy light client too with real-world fixture data, or we could just deploy real contracts and set them up for the unit tests. I went with the latter as this seemed quicker/easier, though it's likely more thorough too.

Fixture data is currently produced and logged by the relayer consistently, though not in the exact JSON format the tests expect, so a little bit of manual copying from logs to JSON is still needed. ```full-flow-template.json``` is provided to help with this, as I expect to need to update the fixtures/tests in future as the relay chain/merkle proofs change with expected incoming fixes from parity.


philipstanislaus (approved):
Very nice changes 💪

PR #450, opened 2021-07-05
Constrain basic channel to only owner

musnit:
This PR constrains the basic channel to an owner which should be the bootstrap user.

In test/ropsten, we can setup a proxy contract to allow anyone to still use it for testing/QA purposes.


ethereum/contracts/TestBasicInboundChannelProxy.sol
```diff
@@ -0,0 +1,35 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.6;
+
+import "@openzeppelin/contracts/access/Ownable.sol";
+import "./ParachainLightClient.sol";
+import "./BasicInboundChannel.sol";
+
+contract TestBasicInboundChannelProxy is Ownable {
```
philipstanislaus, line 8: Could you add a comment describing what the idea of this contract is? Why do we test this case?

philipstanislaus (approved):

ethereum/contracts/TestBasicInboundChannelProxy.sol
musnit: renamed and commented

musnit:
@vgeddes will wait for your review before merging on this one

musnit:
@vgeddes going to merge this - plz review and share any comments if necessary after merge

PR #437, opened 2021-06-24
Running some gas tests

musnit:
Putting up draft PR for reference - not ready for merge

PR #451, opened 2021-07-06
Beefy gas tests rebased

musnit:
replacing https://github.com/Snowfork/snowbridge/pull/437

WIP - currently the merkletree verification is failing with 200 validators - need to look into it.

### DIFF
--- a/ethereum/test/beefy-helpers.js
+++ b/ethereum/test/beefy-helpers.js
@@ -55,13 +55,6 @@ async function createRandomPositions(numberOfPositions, numberOfValidators) {
   return shuffled.slice(0, numberOfPositions)
 }
 
-async function createFullPositions(numberOfValidators) {
-  const positions = [];
-  for (let i = 0; i < numberOfValidators; i++) {
-    positions.push(i);
-  }
-  return positions.sort((a, b) => b - a)
-}
 
 const runBeefyLightClientFlow = async (beefyLightClient, beefyFixture, totalNumberOfSignatures, totalNumberOfValidators) => {
   const initialBitfieldPositions = await createRandomPositions(totalNumberOfSignatures, totalNumberOfValidators)
@@ -146,7 +139,6 @@ async function createCompleteValidatorProofs(id, beefyLightClient, allValidatorP
 module.exports = {
   createBeefyValidatorFixture,
   createRandomPositions,
-  createFullPositions,
   createAllValidatorProofs,
   runBeefyLightClientFlow,
   createCompleteValidatorProofs,
--- a/ethereum/test/beefy-helpers.js
+++ b/ethereum/test/beefy-helpers.js
@@ -1,9 +1,12 @@
 const { ethers } = require("ethers");
 const _ = require("lodash");
+const secp256k1 = require('secp256k1');
 
 const {
-  createMerkleTree
+  createMerkleTree, mine, printBitfield
 } = require("./helpers");
+const realWorldFixture = require('./fixtures/full-flow.json');
+
 const { keccakFromHexString } = require("ethereumjs-util");
 
 async function createBeefyValidatorFixture(numberOfValidators) {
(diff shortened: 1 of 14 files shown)


ethereum/test/beefy-helpers.js
```diff
@@ -0,0 +1,56 @@
...
+    const wallet = ethers.Wallet.createRandom();
+    wallets.push(wallet);
+  }
+  const walletsByLeaf = wallets.reduce((accum, wallet) => {
+    const leaf = '0x' + keccakFromHexString(wallet.address).toString('hex')
+    accum[leaf] = wallet
+    return accum
+  }, {})
+
+  wallets = wallets.sort((a, b) => a.address < b.address)
```
philipstanislaus, line 21: Why sort the wallets?

philipstanislaus: That should not be needed, validator set on Polkadot is also not sorted by the beefy session key, but rather uses a deterministic sorting (maybe by stash key?). The beefy session key which is used for signing should be unsorted though.

ethereum/test/helpers.js
```diff
@@ -133,16 +138,18 @@ function signatureSubstrateToEthereum(sig) {
...
 }
 
-function createMerkleTree(leavesHex) {
-  const leavesHashed = leavesHex.map(leaf => keccakFromHexString(leaf));
-  const merkleTree = new MerkleTree(leavesHashed, keccak, { sort: true });
-
+function createMerkleTree(values) {
+  const leaves = values.map(value => keccakFromHexString(value));
+  const merkleTree = new MerkleTree(leaves, keccak, {
+    sortLeaves: true,
```
philipstanislaus, line 144: Should not be necessary. I think it makes no difference here since we do also sort the wallets, see my comment above. Let's remove this here to not create assumptions that won't hold in the actual production setup.

ethereum/test/beefy-helpers.js
philipstanislaus: For reference: 

> //! Inner nodes are created by concatenating child hashes and hashing again. The implementation
> //! does not perform any sorting of the input data (leaves) nor when inner nodes are created.

https://github.com/paritytech/grandpa-bridge-gadget/pull/225/files#diff-0d5537e8e591fa5d42cb1e23f47fa81950441ce7d9a9eca4cdd4621f08cf02d0R28-R29

philipstanislaus: Also, here is the code that concats and hashes the pairs, without sorting: 

https://github.com/paritytech/grandpa-bridge-gadget/pull/225/files#diff-0d5537e8e591fa5d42cb1e23f47fa81950441ce7d9a9eca4cdd4621f08cf02d0R351-R356

philipstanislaus (approved):
LGTM, thanks @musnit !

ethereum/test/beefy-helpers.js
```diff
@@ -40,17 +44,110 @@ async function createBeefyValidatorFixture(numberOfValidators) {
...
+async function createFullPositions(numberOfValidators) {
   const positions = [];
-  while (positions.length < numberOfPositions) {
-    const position = Math.floor(Math.random() * numberOfValidators);
-    if (positions.indexOf(position) === -1) {
-      positions.push(position);
+  for (let i = 0; i < numberOfValidators; i++) {
+    positions.push(i);
+  }
+  return positions.sort((a, b) => b - a)
```
philipstanislaus, line 63: Why do we sort those positions?

ethereum/test/test_beefy_light_client_real_fixture_data.js
```diff
@@ -59,12 +60,33 @@ describe("Beefy Light Client", function () {
     await mine(45);
 
     const bitfield = await this.beefyLightClient.createRandomBitfield(lastId);
-    expect(printBitfield(bitfield)).to.eq('11')
+    const bitfieldString = printBitfield(bitfield);
+    const bitfieldIndex = [Number(bitfield)] - 1
```
philipstanislaus, line 64: What's the idea of subtracting a number from an array here?

philipstanislaus: I just checked and it works due to some JavaScript conversion magic, but is there an actual reason for this?

philipstanislaus (commented):
Nice! Left just 2 minor comments 👍

philipstanislaus (approved):

ethereum/test/beefy-helpers.js
musnit: ah cool - well, then it shouldn't matter. will leave as is and change if needed when integrating real proofs.

ethereum/test/beefy-helpers.js
musnit: this is no longer used, will remove in next PR: https://github.com/Snowfork/snowbridge/pull/454

ethereum/test/helpers.js
musnit: will do when integrating real production data - added to trello

ethereum/test/test_beefy_light_client_real_fixture_data.js
musnit: err i actually have no idea what i was doing here, probably a typo/mistake that didn't actually result in a failure

musnit: fixed here: https://github.com/Snowfork/snowbridge/pull/454/files#diff-55c3a46404a460a09847d296e249c0e87d0a9cd4741eccda7838c4b0adaae78fR64

PR #457, opened 2021-07-14
Make system ready for deployment

vgeddes:
Changes:
* Relayer
  * Added `global.data-dir` config field to set data directory for any storage
  * ethash cache location is now in the above data dir (also depends on changes in the ethashproof repo)
  * Renamed environment variables for secrets, for clarity
  * Misc improvements for handling & configuration of beefy sqlite DB
  * Parachain ID is queried from chain rather than being hardcoded.
* Parachain
  * Misc simplifications and improvements to chain spec code.
* Ethereum
  * Truffle migration script now dumps all deployed contract addresses to `/tmp/addresses.json`. This file is required by tooling in the [snowbridge-infra](https://github.com/Snowfork/snowbridge-infra) repo.
* E2E tests
  * Parachain ID now defaults to `1000` in line with guidelines (non-system parachains begin at 1000).
* Infra
  * Deleted all obsolete Ansible stuff. The new Nomad tooling lives in the [snowbridge-infra](https://github.com/Snowfork/snowbridge-infra) repo.


ethereum/migrations/4_dump_address.js
```diff
@@ -0,0 +1,31 @@
+require("dotenv").config();
```
musnit, line 1: re filename: 4_ is used, can we make this 6_ or keep it as 4_ and bump the others up by one?

parachain/scripts/make-snowbridge-image.sh
```diff
@@ -0,0 +1,57 @@
...
+        echo "error: unknown parameter"
+        exit 1
+        ;;
+  esac
+done
+
+version=$(git describe --tags $(git rev-list --tags --max-count=1) | sed s/v//)
+revision=$(git rev-parse HEAD)
+
+c=$(buildah from ubuntu:20.04)
```
musnit, line 23: is buildah nativey available on ubuntu AND macos? do we need to add it to installation requirements in README?

musnit: or is this script not expected to be used locally?

musnit: also which runtime will this image be built for?

parachain/src/chain_spec/rococo.rs
```diff
@@ -61,7 +26,7 @@ where
 
 pub fn get_chain_spec(para_id: ParaId) -> ChainSpec {
 	let mut props = Properties::new();
-	props.insert("tokenSymbol".into(), "DEV".into());
+	props.insert("tokenSymbol".into(), "ROC".into());
```
musnit, line 29: is this creating an additional ROC token on our parachain? or refering to the relay chain's token? i would rather call it SNOW or smth if its a new token

parachain/src/chain_spec/snowbridge.rs
```diff
@@ -61,7 +26,7 @@ where
 
 pub fn get_chain_spec(para_id: ParaId) -> ChainSpec {
 	let mut props = Properties::new();
-	props.insert("tokenSymbol".into(), "DEV".into());
+	props.insert("tokenSymbol".into(), "DOT".into());
```
musnit, line 29: same as above comment re ROC

parachain/src/chain_spec/snowbridge.rs
```diff
@@ -151,46 +121,30 @@ fn testnet_genesis(
 				)
 			]
 		},
-		nft: NFTConfig {
+		nft: snowbridge_runtime::NFTConfig {
 			tokens: vec![]
 		},
-		verifier_lightclient: VerifierLightclientConfig {
```
musnit, line 157: how does this get set now?

relayer/workers/beefyrelayer/store/store.go
```diff
@@ -96,71 +100,78 @@ func NewDatabaseCmd(info *BeefyRelayInfo, cmdType CmdType, instructions map[stri
...
 	}
 }
 
-func PrepareDatabase(config *Config) (*gorm.DB, error) {
-	if len(config.DBPath) == 0 {
-		return nil, fmt.Errorf("invalid database path: %s", config.DBPath)
-	}
-	tmpDBFile, err := ioutil.TempFile("", config.DBPath)
+func (d *Database) Initialize() error {
+	tmpfile, err := ioutil.TempFile("", "beefy.*.db")
```
musnit, line 119: 👍

relayer/workers/beefyrelayer/store/store.go
```diff
@@ -96,71 +100,78 @@ func NewDatabaseCmd(info *BeefyRelayInfo, cmdType CmdType, instructions map[stri
...
+
+		d.log.Info("Shutting down DB")
+		sqlDB := d.DB.DB()
+		if sqlDB != nil {
+			err2 = sqlDB.Close()
+			if err2 != nil {
+				d.log.WithError(err2).Error("Unable to close DB connection")
+			}
+
+			err2 = os.Remove(d.Path)
```
musnit, line 156: nice 👍 - if the relayer crashes and restarts without this, i presume it just erases/overwrites from scratch into a new file when it restarts?

relayer/workers/beefyrelayer/main.go
```diff
@@ -135,18 +139,6 @@ func (worker *Worker) Start(ctx context.Context, eg *errgroup.Group) error {
 	return nil
 }
 
-func (worker *Worker) Stop() {
```
musnit, line 138: we don't need this anymore?

relayer/workers/parachaincommitmentrelayer/beefy-listener.go
```diff
@@ -55,6 +54,26 @@ func (li *BeefyListener) Start(ctx context.Context, eg *errgroup.Group) error {
 	}
 	li.beefyLightClient = beefyLightClientContract
 
+	// Fetch ParaId
```
musnit, line 57: nice improvement - better than putting it in config as you say as we already have the endpoint in config

musnit:
@vgeddes looks great, thanks! just some questions and minor comments to resolve and then can merge

ethereum/migrations/4_dump_address.js
vgeddes: fixed. Renamed to 99_dump_address.js

relayer/workers/beefyrelayer/main.go
vgeddes: Nope. was never being called. Riz and I decided that workers shouldn't have a Stop function, and instead they should just clean up gracefully themselves after encountering an error.

PR #458, opened 2021-07-14
Constrain basic outbound channel

vgeddes:
Hey @musnit, in reference to #450 I didn't do a good job of clarifying which basic channel needed to be constrained. It was the outbound channel, not the inbound one 🙂 

We need to ensure that a single authorized end user (the "principal") can send messages to Substrate via the basic outbound channel. This is what is already implemented for the Substrate -> Ethereum direction.

Changes:
* Constrain outbound channel to authorized apps (exactly like the incentivized outbound channel)
* Constrain outbound channel to authorized principal (exactly like the outbound channel on the Substrate side). Authorized apps can only send messages on behalf of the authorized principal.

Since the authorized apps (our apps) are trusted, they can only ever send messages on behalf of the authorized principal.

Please add the following to your `ethereum/.env`, this ensures that the E2E tests & UI tests will pass:
```
# Basic Channel
BASIC_CHANNEL_PRINCIPAL=0x0000000000000000000000000000000000000000
```

E2E tests all pass.


musnit:
ah yeh ofcourse - i should have realized that.

thanks for catching this and fixing!

ethereum/migrations/2_next.js
```diff
@@ -148,13 +148,22 @@ module.exports = function (deployer, network, accounts) {
     // TODO: deploy the contract in this migration and use its address
     let administrator = accounts[0];
 
-    // Fee for incentivized channel
+    // Post-construction initialization for Basic outbound channel
+    if (!("BASIC_CHANNEL_PRINCIPAL" in process.env)) {
```
musnit, line 152: is this set anywhere like in start-services script?

vgeddes: ah, as per the PR description you need to set that in `ethereum/.env`. That's where other contract config goes (such as fee amounts for the incentivized channel).

Though in the near future (like next week) we need to make this configuration per chain, and ideally persisted somewhere other than env files.

musnit: ah ok plz add to README and smth like .env.example

vgeddes: Fixed in https://github.com/Snowfork/snowbridge/pull/458/commits/6105943822ad64ae0eab03bff4e2a33e12aea207

musnit (approved):

PR #460, opened 2021-07-19
Use hardhat for deployments

vgeddes:
This is required for production deployments. Truffle migrations are very error-prone and slow with Infura/Ropsten.

We're now using [Hardhat Deploy](https://github.com/wighawag/hardhat-deploy) for deployments. The Ethereum readme has been updated with the new deployment particulars.

I've retained `truffle-config.js` in the source tree, as truffle is still useful for debugging transactions.


musnit (approved):
looks great! can merge if e2e tests pass

ethereum/scripts/make-relay-config.ts
```diff
@@ -0,0 +1,68 @@
+import { deployments } from "hardhat";
+import TOML from '@iarna/toml';
+import fs from 'fs';
+import path from 'path';
+
+const main = async () => {
```
philipstanislaus, line 6: Very nice 🎉

philipstanislaus (commented):
Nice 💪

PR #463, opened 2021-07-21
Merkle Proof tests

musnit:
This PR adds tests that match the rust tests in https://github.com/paritytech/grandpa-bridge-gadget/blob/33705917c51165e813e54f7985823a8b02dc6a9d/beefy-merkle-tree/src/lib.rs to ensure that our JS merkle tree tests and libs as well as our solidity proof verification all match and work with the same real-world data

ready for review. test passes.


philipstanislaus (commented):
LGTM 👍

PR #464, opened 2021-07-21
Beefy full client

musnit:
Quick test of a beefy client that does full signature verification instead of 2-step interactive random protocol, just for interest sake - likely won't merge

Results:
 - Running flow with 200 validators: Tx successful - gas used: 2343853
 - Running flow with 400 validators: Tx successful - gas used: 4235531
 - Running flow with 667 validators: fail, out of gas
 - Running flow with 1000 validators: fail, out of gas


vgeddes:
Very obsolete, closing this off.

PR #466, opened 2021-07-21
Replace Artemis codename with Snowbridge

vgeddes:
Now that we finally seem to have settled on a name.

The output binaries are now named as follows:
* snowbridge (parachain)
* snowbridge-relay (relayer)

E2E tests all pass

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
-# Polkadot Ethereum Bridge
+# Snowbridge
 
-Components for a Polkadot Ethereum Bridge
+A trust-minimized bridge between Polkadot and Ethereum.
 
 ## Development
 
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Snowbridge
 
-A trust-minimized bridge between Polkadot and Ethereum.
+A trustless bridge between Polkadot and Ethereum.
 
 ## Development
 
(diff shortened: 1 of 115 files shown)


musnit:
ah awesome thanks

musnit (approved):

README.md
```diff
@@ -1,6 +1,6 @@
-# Polkadot Ethereum Bridge
+# Snowbridge
 
-Components for a Polkadot Ethereum Bridge
+A trust-minimized bridge between Polkadot and Ethereum.
```
philipstanislaus, line 3: Can we say „trustless“ instead? 

I know that there is always smart contract risk and rising the consensus protocols and light client protocols involved, but even if not technically correct, the term “trustless” seems to have become the accepted way to describe that.

parachain/runtime/snowbridge/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "snowbridge-runtime"
-description = "Artemis Snowbridge Runtime"
+description = "Snowbridge Snowbridge Runtime"
```
philipstanislaus, line 3: Dupe

philipstanislaus: Might be correct though? Can we rename the runtime/testnet?

philipstanislaus (commented):
Thanks!

philipstanislaus (approved):

README.md
vgeddes: sure, fixed in latest commit.

parachain/runtime/snowbridge/Cargo.toml
vgeddes: I still need to rename the runtime pallets appropriately, so will fix the dupe along with that change.

PR #467, opened 2021-07-23
Rename pallet VerifierLightclient to EthereumLightClient

vgeddes:
Is a more correct name, aiding readers & reviewers of our code.


vgeddes:
Merging in as this is a pretty common-sense change.

philipstanislaus (commented):
Makes sense, thanks!

PR #468, opened 2021-07-23
Increase validator session length to 4 hours in beefy light client

vgeddes:
Necessary to have a working testnet, as deploying everything in a 10 minute span (100 blocks) is difficult to accomplish. And neither is such a short _from-scratch_ deployment cycle a goal of our production environment.

Passed E2E tests.

As an aside, the session length on Polkadot is 4 hours, and 1 hour on Kusama.


philipstanislaus (commented):
LGTM

PR #469, opened 2021-07-24
Remove unsafe RPC calls

vgeddes:
This is necessary for our relay to be interoperable with our staging testnet, Kusama, and Polkadot. The beefy errors we saw in our staging env earlier this week are are directly related to this.

We can't use the `state_queryStorage` RPC in production. Its marked as unsafe and is blocked on public networks, as it can easily be exploited to kill a node. Instead we need to use `state_queryStorageAt`, which will only retrieve changes for a single block. I've added support for this newer RPC in GSRPC (see companion PR below).

The affected code paths involve fetching beefy authorities and parachain heads. Some of this code I've cleaned up, and pushed logic down into GSRPC where appropriate.

Companion PR for GSRPC: https://github.com/Snowfork/go-substrate-rpc-client/pull/13

E2E tests are passing

### DIFF
--- a/relayer/chain/relaychain/connection.go
+++ b/relayer/chain/relaychain/connection.go
@@ -6,6 +6,7 @@ package relaychain
 import (
 	"context"
 	"fmt"
+	"sort"
 
 	"github.com/sirupsen/logrus"
 
@@ -104,13 +105,19 @@ func (co *Connection) GetMMRLeafForBlock(
 	return proofResponse, nil
 }
 
-type Head struct {
+type ParaHead struct {
 	LeafIndex  int // order in which this head was returned from the storage query
 	ParaID uint32
-	Data   []byte
+	Data   types.Bytes
 }
 
-func (co *Connection) FetchParaHeads(blockHash types.Hash) (map[uint32]Head, error) {
+// Offset of encoded para id in storage key.
+// The key is of this format:
+//   ParaId: u32
+//   Key: hash_twox_128("Paras") + hash_twox_128("Heads") + hash_twox_64(ParaId) + Encode(ParaId)
+const ParaIDOffset = 16 + 16 + 8
+
+func (co *Connection) FetchParaHeads(blockHash types.Hash) (map[uint32]ParaHead, error) {
 
 	keyPrefix := types.CreateStorageKeyPrefix("Paras", "Heads")
 
@@ -126,7 +133,7 @@ func (co *Connection) FetchParaHeads(blockHash types.Hash) (map[uint32]Head, err
 		return nil, err
 	}
 
-	heads := make(map[uint32]Head)
+	heads := make(map[uint32]ParaHead)
 
 	for _, changeSet := range changeSets {
 		for index, change := range changeSet.Changes {
@@ -135,8 +142,7 @@ func (co *Connection) FetchParaHeads(blockHash types.Hash) (map[uint32]Head, err
 			}
 
 			var paraID uint32
-
-			if err := types.DecodeFromBytes(change.StorageKey[40:], &paraID); err != nil {
+			if err := types.DecodeFromBytes(change.StorageKey[ParaIDOffset:], &paraID); err != nil {
 				co.log.WithError(err).Error("Failed to decode parachain ID")
 				return nil, err
 			}
@@ -149,7 +155,7 @@ func (co *Connection) FetchParaHeads(blockHash types.Hash) (map[uint32]Head, err
 				return nil, err
 			}
 
-			heads[paraID] = Head{
+			heads[paraID] = ParaHead{
 				LeafIndex: index,
 				ParaID:    paraID,
 				Data:      headData,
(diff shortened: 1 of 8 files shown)


relayer/workers/parachaincommitmentrelayer/catchup.go
```diff
@@ -183,7 +206,7 @@ func (li *BeefyListener) parablocksWithProofs(blocks []ParaBlockWithDigest, late
 			MMRProofResponse: mmrProof,
 			Header:           ownParaHead,
 			HeaderProof:      ownParaHeadProof,
-			HeaderProofPos:   ownParaHeadProofPos,
+			HeaderProofPos:   0,
```
vgeddes, line 209: @musnit, The index of our parachain header in the response for the `Paras.Heads` storage query likely isn't relevant for any purpose involving the Merkle Tree of Para Heads. They are two different trees. So I've just set this to 0 for now.

musnit: @vgeddes I think we need the index in order to verify the merkle proof: https://github.com/Snowfork/snowbridge/blob/main/ethereum/contracts/utils/MerkleProof.sol#L34

vgeddes: The `Paras.Head` storage map is part of the base-16 modified merkle-patricia tree that is used for storage on all substrate chains. That's surely different from the plain binary Merkle tree as encoded in the MMR?

musnit: so afaik the plain binary tree is created from all parachain heads based on the order that the Paras.Heads query returns, so that query returns the leaves which are put into the binary tree. the position in the array there should correspond to the leaf position in the binary merkle tree

vgeddes: I see. I've updated the code as necessary.

relayer/chain/relaychain/connection.go
```diff
@@ -104,102 +104,79 @@ func (co *Connection) GetMMRLeafForBlock(
...
+				continue
 			}
 
-			var headerBytes types.Bytes
-			if err := types.DecodeFromBytes(change.StorageData, &headerBytes); err != nil {
-				co.log.WithError(err).Error("Failed to decode MMREncodableOpaqueLeaf")
-				return nil, 0, types.Header{}, err
+			var paraID uint32
+
+			if err := types.DecodeFromBytes(change.StorageKey[40:], &paraID); err != nil {
```
musnit, line 139: any idea the last 40 bytes are the paraID/what the full type of the StorageKey actually is?

relayer/chain/relaychain/connection.go
```diff
@@ -104,102 +104,79 @@ func (co *Connection) GetMMRLeafForBlock(
...
+			}
 		}
 	}
-	var ownParachainHeader types.Header
-	if err := types.DecodeFromBytes(headers[ownParachainHeaderPos], &ownParachainHeader); err != nil {
-		co.log.WithError(err).Error("Failed to decode Header")
-		return nil, 0, types.Header{}, err
-	}
 
-	co.log.WithField("parachainId", ownParachainId).Info("Decoding header for own parachain")
```
musnit, line 179: please keep all this excessive logging in at least until we complete e2e verification. it's likely going to all be needed for debugging/completion purposes once we activate the verification in the smart contracts, as i don't expect it to just work.

also, it seems fine to even just keep the excessive logging in indefinitely?

relayer/workers/parachaincommitmentrelayer/beefy-listener.go
```diff
@@ -76,26 +76,27 @@ func (li *BeefyListener) Start(ctx context.Context, eg *errgroup.Group) error {
 
 	eg.Go(func() error {
 
-		verifiedBeefyBlockNumber, verifiedBeefyBlockHash, err := li.fetchLatestVerifiedBeefyBlock(ctx)
+		relayBlockNumber, relayBlockHash, err := li.fetchLatestVerifiedBeefyBlock(ctx)
```
musnit, line 79: prefer to keep these variables named as is. relay is a terribly ambiguous word.

relayer/workers/parachaincommitmentrelayer/catchup.go
```diff
@@ -155,11 +157,31 @@ func (li *BeefyListener) parablocksWithProofs(blocks []ParaBlockWithDigest, late
...
 			li.log.WithField("relayBlockHash", relayBlockHash.Hex()).Info("Got relay chain blockhash")
-			allParaHeads, ownParaHeadProofPos, ownParaHead, err = li.relaychainConn.GetAllParaheadsWithOwn(relayBlockHash, li.paraID)
+			heads, err := li.relaychainConn.FetchParaHeads(relayBlockHash)
 			if err != nil {
 				li.log.WithError(err).Error("Failed to get paraheads")
 				return nil, err
 			}
+
+			if _, ok := heads[li.paraID]; !ok {
+				return nil, fmt.Errorf("chain is not a registered parachain")
```
musnit, line 167: should this panic instead?

relayer/workers/parachaincommitmentrelayer/catchup.go
```diff
@@ -155,11 +157,31 @@ func (li *BeefyListener) parablocksWithProofs(blocks []ParaBlockWithDigest, late
...
+			}
+
+			var header types.Header
+			if err := types.DecodeFromBytes(heads[li.paraID].Data, &header); err != nil {
+				li.log.WithError(err).Error("Failed to decode Header")
+				return nil, err
+			}
+
+			tmp := make([]types.Bytes, 0, len(heads))
+			for _, v := range heads {
```
musnit, line 177: does this guarantee ordering, ie, guarantee that the order of heads in the byte array will be ascending based on LeadIndex?

vgeddes: ah good catch. Will need to order by LeafIndex.

relayer/workers/parachaincommitmentrelayer/catchup.go
vgeddes: Code in the relayer should never panic, as it will cause all workers to die. The idea with error-handling here is that the error will bubble up, the worker will self-terminate and be restarted.

relayer/workers/parachaincommitmentrelayer/beefy-listener.go
vgeddes: Sure, I agree regarding beefy vs relay and will make the change. Though I'd drop the `verified` prefix. The signal to noise ratio of `verifiedBeefyBlockNumber` and `verifiedParaBlockNumber` is quite low, so dropping `verified` makes it easier to discern between the two things.

relayer/chain/relaychain/connection.go
vgeddes: Will add it back :+1:

relayer/workers/parachaincommitmentrelayer/catchup.go
musnit: yeh - i guess in this case we know for sure the worker should not be restarted as there has most likely been a config issue, so it would be nice to be able to distinguish between error-that-should-restart and error-that-should-not but seems fine for now

relayer/chain/relaychain/connection.go
vgeddes: The full key type is this:
```
ParaId: u32
Key: hash_twox_128("Paras") + hash_twox_128("Heads") + hash_twox_64(ParaId) + ParaId
```
The length of the first 3 hashes is 40 bytes which lines up with what our code is doing.

This article explains it well:
https://www.shawntabrizi.com/substrate/transparent-keys-in-substrate/

relayer/chain/relaychain/connection.go
vgeddes: Yeah, excessive logging is good. Added a bunch of extra logging in the latest upload.

relayer/workers/parachaincommitmentrelayer/beefy-listener.go
vgeddes: fixed in latest upload

relayer/chain/relaychain/connection.go
```diff
@@ -104,13 +105,19 @@ func (co *Connection) GetMMRLeafForBlock(
...
-	Data   []byte
+	Data   types.Bytes
 }
 
-func (co *Connection) FetchParaHeads(blockHash types.Hash) (map[uint32]Head, error) {
+// Offset of encoded para id in storage key.
+// The key is of this format:
+//   ParaId: u32
+//   Key: hash_twox_128("Paras") + hash_twox_128("Heads") + hash_twox_64(ParaId) + Encode(ParaId)
+const ParaIDOffset = 16 + 16 + 8
```
musnit, line 118: nice

relayer/chain/relaychain/connection.go
```diff
@@ -160,23 +166,52 @@ func (co *Connection) FetchParaHeads(blockHash types.Hash) (map[uint32]Head, err
 	return heads, nil
 }
 
-// Fetch the latest block of a parachain that has been finalized at a relay chain block hash
+// ByLeafIndex implements sort.Interface based on the LeafIndex field.
+type ByLeafIndex []ParaHead
```
musnit, line 182: fancy ;o

musnit (approved):
looks good now thx! feel free to merge if e2e tests pass and will address any leftovers in the coming week

relayer/workers/parachaincommitmentrelayer/catchup.go
vgeddes: I've come up with the following remedy:

* `FetchParaHeads` remains unchanged:
* Added a method `AsProofInput` which transforms the map of heads into a form suitable for proof generation

I think its fine to transition from an ordered array to map and then back to array if it allows us to break down complexity into manageable, composable parts. Its kinda like a data pipeline. The extra overhead of this pipeline and sorting is minimal since N (number of parachains) is gonna be less than 200 which is practically O(1) for a computer.

musnit: yep agree, awesome thx

PR #470, opened 2021-07-24
Misc cleanup: Golang style fixes

vgeddes:
`conn.GetAPI` should be `conn.API()`, and so on. Should only use `GetX` or `FetchX` when making an actual API call.

E2E test all pass.


musnit (approved):

vgeddes:
Too many merge conflicts. Gonna cut a new PR.

PR #479, opened 2021-07-29
Deduplicate difficulty config

vgeddes:
Values were duplicated all over the codebase. Also the production `snowbridge` runtime should be using values for mainnet.


philipstanislaus (commented):
LGTM

philipstanislaus (approved):

PR #483, opened 2021-07-30
Eip 2718 fix

musnit:
Eip 2718 changed the transaction receipt encoding with multiple types of transaction receipts, so the trie generation needs to accomodate those types. (https://github.com/ethereum/EIPs/blob/master/EIPS/eip-2718.md#receipts)

I noticed this when the eth-relayer crashed on a block that had the new kind of eip-2718 transaction in it. After applying this fix, the relayer resumed and continued past that block gracefully, so I'm assuming this fix is sufficient and works due to that.

all e2e tests pass

### DIFF
--- a/relayer/chain/ethereum/trie.go
+++ b/relayer/chain/ethereum/trie.go
@@ -2,26 +2,65 @@ package ethereum
 
 import (
 	"bytes"
+	"sync"
 
+	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/rlp"
 	"github.com/ethereum/go-ethereum/trie"
 )
 
 func MakeTrie(items types.Receipts) (*trie.Trie, error) {
-	keyBuf := new(bytes.Buffer)
+
 	trie := new(trie.Trie)
 
-	for i := 0; i < items.Len(); i++ {
-		keyBuf.Reset()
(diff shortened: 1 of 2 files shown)


relayer/chain/ethereum/trie.go
```diff
@@ -2,26 +2,65 @@ package ethereum
...
 
-		value, err := rlp.EncodeToBytes(items[i])
-		if err != nil {
-			return nil, err
-		}
+	valueBuf := encodeBufferPool.Get().(*bytes.Buffer)
+	defer encodeBufferPool.Put(valueBuf)
 
-		trie.Update(keyBuf.Bytes(), value)
+	// StackTrie requires values to be inserted in increasing hash order, which is not the
```
musnit, line 32: I'm using a normal Trie instead of a StackTrie because we need Trie.Prove() later on, which StackTrie does not have. I think StackTrie is just an optimization that prunes, and is not actually structured any differently, but I'm not 100% sure - https://github.com/ethereum/go-ethereum/pull/21407/files#diff-450c04b40c4e2f8c12ed10930cd0a8b8a6c25b77b1f3469e7205b43b5fad058aR46 has the implementation/further details to confirm

philipstanislaus: Quickly glanced through the StackTrie code, looks indeed like it's only used as an optimization.

philipstanislaus (commented):
LGTM 👍

PR #484, opened 2021-07-30
Increase robustness of extrinsics monitoring

vgeddes:
I found this bug by accident and it was causing the ethereum relay to crash under certain conditions (when a monitored extrinsic is usurped by another transaction). Should only consider an extrinsic to be executed if its actually included in a block.


musnit (approved):

philipstanislaus (commented):
LGTM

PR #485, opened 2021-07-31
Fix E2E tests

vgeddes:
For some reason `chai-as-promised` and `chai-bignumber` were removed as deps in a previous commits.

Also marked the test for malicious dot app to be skipped as the test implementation is still WIP it seems.


musnit:
@vgeddes its not WIP - it works but is only intended to work when you run the malicious e2e setup

philipstanislaus (commented):
LGTM

PR #486, opened 2021-08-01
Productionise relay services

vgeddes:
Various fixes and improvements to get the relay services working robustly in production.

The main change however is to split the relayer into separate processes, a design which has better error discovery and robustness (The previous monolithic container design was a dead end). It is also a better fit for service orchestrators such as Nomad, K8s, Docker-Compose etc.

Changes:
* Service management
  * Relay services now run in independent processes
  * Simplified logging (no need to pass loggers down into services now that there is only 1 service per process)
  * Eliminated buggy (and obsolete) deadlock detector heuristic
* Configuration
  * Configuration now reorganised to better reflect the fields for listeners ("Sources"), and writers ("Sinks")
  * EIP-1554 FeeCap, TipCap, and GasLimit are now read from configuration
* Bug fixes
  * Don't overpay for Ethereum EIP-1554 transactions. FeeCap & TipCap are automatically calculated. Only need to provide GasLimit to prevent flaky "eth_estimateGas` calls.
  * Listeners & Writers should not return an error if their main loops are terminated due to context cancellation (i.e via signals). This ensures that `errgroup.Wait()` returns the real error cause, not a random "context cancelled".
  * On listener/writer teardown, channel draining was flawed and could cause deadlock, leaving the process hanging where it could only be killed using SIGKILL.
  * BeefyEthereumWriter was not returning potential errors from its main loop, causing the service not to terminate as required.

Example UX changes:
```
snowbridge-relay run beefy --config /path/to/config.toml --ethereum.private-key-file /path/to/key
snowbridge-relay run ethereum --config /path/to/config.toml --substrate.private-key-file /path/to/key
````

Example config for ethereum relay:
```json
{
  "source": {
    "data-dir": "/tmp/snowbridge-e2e-config",
    "ethereum": {
      "endpoint": "ws://localhost:8546"
    },
    "descendants-until-final": 3,
    "contracts": {
      "BasicOutboundChannel": "0xdeadbeef",
      "IncentivizedOutboundChannel": "0xdeadbeef"
    }
  },
  "sink": {
    "parachain": {
      "endpoint": "ws://localhost:11144"
    }
  }
}
```

E2E tests pass cleanly.


relayer/relays/parachain/main.go
```diff
@@ -2,139 +2,93 @@ package parachain
...
-	if err != nil {
-		return nil, err
-	}
+	parachainConn := parachain.NewConnection(config.Source.Parachain.Endpoint, nil)
+	relaychainConn := relaychain.NewConnection(config.Source.Polkadot.Endpoint)
 
-	parachainConn := parachain.NewConnection(parachainConfig.Endpoint, nil, log)
-	relaychainConn := relaychain.NewConnection(relaychainConfig.Endpoint, log)
-	ethereumConn := ethereum.NewConnection(ethereumConfig.Endpoint, ethereumKp, log)
+	// FIXME: This is used by both the source & sink. They should use separate connections
```
musnit, line 31: plz use TODO or put in github issues. don't think we've used FIXME anywhere before so it could get lost

musnit: although if things work reliably with a shared connection, then maybe we don't even have to worry?

test/scripts/start-services.sh
```diff
@@ -76,10 +75,11 @@ start_polkadot_launch()
     echo "Writing Polkadot configuration"
     polkadotbinary=/tmp/polkadot/target/release/polkadot
     if [[ -f ../test/.env ]]; then
+        # shellcheck disable=SC1091
         source ../test/.env
     fi
 
-    if [ $# -eq 1 ] && [ $1 = "duplicate" ];
+    if [ $# -eq 1 ] && [ "$1" = "duplicate" ];
```
musnit, line 82: does this still work? did u test?

test/scripts/start-services.sh
```diff
@@ -113,36 +113,101 @@ start_polkadot_launch()
...
+    (
+        : > beefy-relay.log
+        while :
+        do
+            echo "Starting beefy relay"
+            ${relay} run beefy \
+                --config $configdir/beefy-relay.json \
+                --ethereum.private-key "0x935b65c833ced92c43ef9de6bff30703d941bd92a2637cb00cfad389f5862109" \
+                >>beefy-relay.log 2>&1 || true
+            sleep 20
```
musnit, line 165: ah nice, so restart is handled automatically now by start-services? how is it handled in the staging/prod context?

musnit:
this all looks great, thx a lot! context cancelled fix will help a ton too :).

a few minor comments but otherwise looks good to merge if e2e tests pass.

do the duplicate and malicious configurations still work as expected?

test/scripts/start-services.sh
vgeddes: Yeah the start script will automatically restart the relay if it failes. In staging Nomad automatically restarts failed services, subject to the configuration you provide it.

For example, in the nomad output below, you can see the ethereum relay service failed 3 times, but still has 1 running instance - so its healthy.
```
$ nomad status ethereum-relay
ID            = ethereum-relay
Name          = ethereum-relay
Submit Date   = 2021-08-02T12:34:44Z
Type          = service
Priority      = 50
Datacenters   = dc1
Namespace     = default
Status        = running
Periodic      = false
Parameterized = false

Summary
Task Group  Queued  Starting  Running  Failed  Complete  Lost
main        0       0         1        3       0         0

Allocations
ID        Node ID   Task Group  Version  Desired  Status   Created     Modified
ad1b6f16  1d7b49bc  main        0        run      running  16m18s ago  16m5s ago
2c3c6887  1d7b49bc  main        0        stop     failed   5h6m ago    16m18s ago
```

test/scripts/start-services.sh
vgeddes: The only effect of this change is that if you call `start-services.sh "foo bar"`, then it will compare "foo bar" to "duplicate", instead of "foo". This is more correct behavior. Basically individual command line args should be quoted in scripts as a good practice.

vgeddes: anyway, I've fixed the duplicate thingy and it all works now.

relayer/cmd/run/parachain/command.go
```diff
@@ -0,0 +1,125 @@
...
+		cleanedKey = strings.TrimPrefix(privateKey, "0x")
+	}
+
+	keypair, err := secp256k1.NewKeypairFromString(cleanedKey)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse private key: %w", err)
+	}
+
+	return keypair, nil
+}
```
philipstanislaus, line 125: Repetitive code, can we merge this with the functions in relayer/cmd/run/beefy/command.go and relayer/cmd/run/ethereum/command.go?

test/config/launchConfigOverrides.json
```diff
@@ -25,7 +25,7 @@
           "port": 31201,
           "name": "bob",
           "flags": [
-            "-lruntime=debug",
+            "-lruntime=trace,sc_basic_authorship=trace,txpool=trace",
```
philipstanislaus, line 28: Is this intended to be on `main`?

test/config/launchConfigOverrides.json
```diff
@@ -8,7 +8,7 @@
           "port": 31200,
           "name": "alice",
           "flags": [
-            "-lruntime=debug",
+            "-lruntime=trace,sc_basic_authorship=trace,txpool=trace",
```
philipstanislaus, line 11: Is this intended to be on `main`?

test/config/launchConfigOverridesDup.json
```diff
@@ -20,6 +20,23 @@
             "--",
             "--execution=wasm"
           ]
+        },
+        {
+          "wsPort": 11155,
+          "port": 31201,
+          "name": "bob",
+          "flags": [
+            "-lruntime=trace,sc_basic_authorship=trace,txpool=trace",
```
philipstanislaus, line 29: Is this intended to be on `main`?

philipstanislaus (commented):
Nice work 👍

PR #487, opened 2021-08-03
Parachain merkle proofs

musnit:
This PR enables parachain merkle proofs, and some other improvements including:
 - Ensuring the relayer always queries for MMR proofs to the latest confirmed mmr root, rather than old mmr roots
 - Ensuring both beefy and parachain relayer track the length/leaf count of the MMR properly from the relay chain and send it explicitly (with beefy-relayer sending latest, and parachain-relayer sending the count corresponding to the latest confirmed mmr root on ethereum)
 - Ensuring the parachain relayer tracks the leaf index of each mmr leaf/proof correctly, based on that leaf's index in the latest confirmed mmr with some assumptions
 - Dynamic bootstrap of validator registry contract instead of hardcoded
 - Redo parachain light client to avoid building/scale encoding entire header

e2e tests pass! ready for review


relayer/relays/parachain/catchup.go
```diff
@@ -189,22 +189,23 @@ func (li *BeefyListener) parablocksWithProofs(blocks []ParaBlockWithDigest, late
...
 		}
 
-		// Note - relayChainBlockNumber will be one less than the block number discovered
+		// Note - relayChainBlockNumber will be one less than the actual block number we want
 		// due to the decrement at the end of the loop, but the mmr leaves are 0 indexed whereas
 		// block numbers start from 1, so we actually do want to query for the leaf at
-		// one less than the block number
-		mmrProof, err := li.relaychainConn.GetMMRLeafForBlock(uint64(relayChainBlockNumber), relayBlockHash)
+		// relayChainBlockNumber
+		mmrProof, err := li.relaychainConn.GetMMRLeafForBlock(uint64(relayChainBlockNumber), latestRelaychainBlockHash)
```
musnit, line 203: fyi: this is the critical change - querying for proofs to the same latestRelaychainBlockHash, irrespective of whether the proof is for a new or older relayChainBlockNumber

ethereum/scripts/configure-beefy.ts
```diff
@@ -0,0 +1,53 @@
+require("dotenv").config();
+const hre = require("hardhat");
+
+const SubClient = require('../../test/src/subclient').SubClient;
```
vgeddes, line 4: I'd rather not pull in E2E test fixture stuff. Its also not necessary.

Instead can you create a Polkadot-API instance using the example at: https://polkadot.js.org/docs/api/start/create.

For determining the API endpoint you can pass an environment variable to the script. This will allow the script to work both for local and staging environments.

ethereum/scripts/configure-beefy.ts
```diff
@@ -0,0 +1,53 @@
+require("dotenv").config();
+const hre = require("hardhat");
```
vgeddes, line 2: Should use ES6 `import` instead of `require`.

musnit:
@philipstanislaus @vgeddes e2e tests pass! ready for review. have confirmed that the merkle proofs are working properly and generating the correct parachain merkle root too, though they are not being actively used yet as the mmr proof check is still disabled. that will hopefully come in next PR.

ethereum/scripts/configure-beefy.ts
musnit: fixed https://github.com/Snowfork/snowbridge/pull/487/commits/ceeefc1b38af8e0510d0244fb44dedc747482523

ethereum/scripts/configure-beefy.ts
musnit: getting annoying typescript errors if I do so, gonna ignore this.

ethereum/scripts/configure-beefy.ts
```diff
@@ -0,0 +1,55 @@
...
+  const validatorRegistryDeployment = await hre.deployments.get("ValidatorRegistry");
+  const validatorRegistryContract = await new hre.ethers.Contract(validatorRegistryDeployment.address,
+    validatorRegistryDeployment.abi);
+
+  const validatorRegistry = await validatorRegistryContract.connect(signer)
+
+  const relayChainProvider = new WsProvider(relaychainEndpoint);
+  const relaychainAPI = await ApiPromise.create({
+    provider: relayChainProvider,
+    typesBundle: bundle
```
vgeddes, line 22: Its not necessary to provide snowbridge types if you're only querying the relay chain.

ethereum/package.json
```diff
@@ -40,6 +41,7 @@
     "hardhat-gas-reporter": "^1.0.4",
     "ts-node": "^10.0.0",
     "typescript": "^4.3.2",
-    "web3": "^1.3.6"
+    "web3": "^1.3.6",
+    "@snowfork/snowbridge-types": "link:../types"
```
vgeddes, line 45: Our types aren't necessary in this context.

relayer/chain/relaychain/connection.go
```diff
@@ -233,3 +209,26 @@ func (co *Connection) FetchFinalizedParaHead(relayBlockhash types.Hash, paraID u
...
+		return 0, err
+	}
+	var mmrLeafCount uint64
+
+	ok, err := co.API().RPC.State.GetStorage(mmrLeafCountKey, &mmrLeafCount, relayBlockhash)
+	if err != nil {
+		return 0, err
+	}
+
+	if !ok || mmrLeafCount == 0 {
```
vgeddes, line 225: There's actually two distinct unrelated error cases here, so I would not lump them in the with same log message as it could cause confusion. There's "MMR Leaf Count Not Found" and "MMR is empty and has no leaves"

relayer/relays/beefy/beefy-relaychain-listener.go
```diff
@@ -106,11 +106,18 @@ func (li *BeefyRelaychainListener) subBeefyJustifications(ctx context.Context) e
...
-			latestMMRProof, err := li.relaychainConn.GetMMRLeafForBlock(blockNumber-1, blockHash)
+			latestMMRProof, err := li.relaychainConn.GetMMRLeafForBlock(blockNumber-1, blockHash, li.config.Source.Polkadot.BeefyStartingBlock)
 			if err != nil {
 				log.WithError(err).Error("Failed get MMR Leaf")
 				return err
 			}
+
+			mmrLeafCount, err := li.relaychainConn.FetchMMRLeafCount(blockHash)
+			if err != nil {
+				log.WithError(err).Error("Failed get MMR Leaf Count")
```
vgeddes, line 117: See my other comment about the error handling in `FetchMMRLeafCount`. Strictly speaking the function should not return an error if the retrieved leaf count is zero, The validation of the leaf count should be hoisted up here.

ethereum/contracts/ParachainLightClient.sol
```diff
@@ -36,47 +29,49 @@ library ParachainLightClient {
         bytes32 nextAuthoritySetRoot;
     }
 
+    bytes4 public constant PARACHAIN_ID_SCALE = 0xe8030000;
```
vgeddes, line 32: Could you rather add the ParaId as constructor param, and then in the constructor use `ScaleCodec.encode32` to get a copy of the encoded bytes. This helps us when the ParaId is no longer 1000.

musnit: This contract is a library, so has no constructor/state. Going forward I think either we can wrap it in a contract for our specific parachain and put the ParaId there as you say, or put the id into the channel contract, or just deploy a new library for new ParaIds.

Also, I believe using a constant may result in cheaper gas costs not just on the original deployment, but on every future calls into the contract that read that constant, because that constant is stored in the contract code rather than in storage. If so, we should keep it as a constant. I'll double check if this is the case or not.

ethereum/scripts/configure-beefy.ts
musnit: fixed c4420d0

ethereum/package.json
musnit: fixed 5186f86

relayer/relays/beefy/beefy-relaychain-listener.go
musnit: fixed 06a2d8d

relayer/chain/relaychain/connection.go
musnit: fixed 06a2d8d

philipstanislaus (approved):
This is great 🚀

ethereum/contracts/BeefyLightClient.sol
```diff
@@ -342,16 +345,17 @@ contract BeefyLightClient {
...
+        uint64 leafIndex,
+        uint64 leafCount
     ) public {
         bytes memory encodedLeaf = encodeMMRLeaf(leaf);
         bytes32 hashedLeaf = hashMMRLeaf(encodedLeaf);
 
         mmrVerification.verifyInclusionProof(
             root,
             hashedLeaf,
-            length - 1,
```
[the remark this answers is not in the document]
musnit: one of many :P. the harder one was discovering that the mmr hashes the actual leaf, not the opaqueencodedleaf

ethereum/contracts/utils/MerkleProof.sol
```diff
@@ -29,6 +29,29 @@ library MerkleProof {
...
+     * @notice Compute the root of a MMR from a leaf and proof
+     *
+     * @param leaf the leaf we want to prove
+     * @param proof an array of nodes to be hashed in order that they should be hashed
+     * @param side an array of booleans signalling whether the corresponding node should be hashed on the left side or the right side of the current hash
+     */
+    function createMMRRoot(
+        bytes32 leaf,
+        bytes32[] calldata proof,
+        bool[] calldata side
```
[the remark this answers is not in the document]
musnit: yep, i think many checks can be skipped when we only care about preventing fraudulent inclusion

test/config/launchConfigOverridesDup.json
```diff
@@ -8,13 +8,12 @@
           "port": 31200,
           "name": "alice",
           "flags": [
-            "-lruntime=debug",
+            "-lruntime=trace,sc_basic_authorship=trace,txpool=trace",
```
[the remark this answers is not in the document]
musnit: this file is unused - will remove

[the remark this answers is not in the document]
musnit: 90752e0

relayer/relays/parachain/ethereum-channel-writer.go
```diff
@@ -213,12 +230,39 @@ func (wr *EthereumChannelWriter) WriteIncentivizedChannel(
...
+	suffix, err := hex.DecodeString(prefixSuffix[1])
+	if err != nil {
+		return err
+	}
+
+	paraVerifyInput := incentivized.ParachainLightClientParachainVerifyInput{
+		OwnParachainHeadPrefixBytes: prefix,
+		OwnParachainHeadSuffixBytes: suffix,
+		ParachainHeadProof:          paraHeadProof,
 	}
```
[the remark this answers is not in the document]
musnit: problem is they are different types, ie, incentivized.ParachainLightClientParachainVerifyInput and basic.ParachainLightClientParachainVerifyInput so is a bit of a mission to do this cleanly atm

relayer/relays/parachain/catchup.go
```diff
@@ -192,8 +190,10 @@ func (li *BeefyListener) parablocksWithProofs(blocks []ParaBlockWithDigest,
 		}
 
 		// Note - relayChainBlockNumber will be one less than the actual block number we want
-		// due to the decrement at the end of the loop, so we add back 1
-		mmrProof, err := li.relaychainConn.GetMMRLeafForBlock(uint64(relayChainBlockNumber+1), latestRelaychainBlockHash, li.config.Polkadot.BeefyStartingBlock)
+		// due to the decrement at the end of the loop, so we increment by 1. Additionally,
+		// parachain merkle roots are created 1 block later than the actual parachain headers,
+		// so we increment twice.
```
philipstanislaus, line 195: Are you sure this is stable? Beefy runs async between validators, and IIUC it could lag more than 1 block behind parachain heads being committed.

philipstanislaus (commented):
Nice! 😎

relayer/relays/parachain/catchup.go
musnit: @philipstanislaus Beefy runs async, so could lag behind, but once it gets done it should still produce one leaf for every block and so the function that maps from mmrLeafIndex -> mmrLeaf -> paraHeadsRoot -> Relaychain block number which commits to those Paraheads should still be constant. If not, then it means the MMR could be missing some para heads?

Also, if beefy is lagging far behind, the call to GenerateProof will fail and the relayer will just continue and try again later just fine :).

Having said that, I'm not sure of the above - it includes a bunch of assumptions that we need to dig in to verify, including:
 - The polkadot relay chain does finalize every single parachain head, and include every head of every parachain in Paras.Head over time. It never skips a parachain head or finalizes multiple heads from a single parachain in a row in a single relay chain block
 - The same for beefy - Beefy MMR leaves are 1:1 with relaychain blocks (besides for blocks created before the existence of beefy) and there is one beefy leaf for every relaychain block and therefore one beefy leaf for the Paras.Heads value at every relaychain block. There is thus effectively an isomorphism from beefy leaf -> Paras.Heads
 - This isomorphism is a straightforward linear function, specifically:
Paras.Heads(relayChainBlockNumber) = Paras.Heads(mmrLeafIndex)

philipstanislaus: Did just dive into the parachain consensus code (https://github.com/paritytech/cumulus/blob/master/client/consensus/common/src/parachain_consensus.rs) a bit. 

Collators do not create blocks, they create block candidates. Actual block production and finalization is done by the assigned validators of the relay chain. I'm still not sure whether there is any guarantee that every single finalized parachain block is included in a MMR leaf's parachain root. It seems possible that parachain blocks are skipped and never included in that root. 

I'll try to find out whether the hypothesis that parachain blocks can be skipped holds or not.

I can confirm though that there is a 1:1 relationship between beefy leaves and relaychain blocks.

musnit: @philipstanislaus cc @vgeddes hmm ok. if it is possible that parachain blocks are skipped and never included in that root, then we have a bigger problem as commitments will get lost. We'd need to move to a commitment mechanism which keeps a buffer of old commitments and/or messages over a few blocks to ensure there are no gaps. Let us know when you get deeper.

rphmeier: I think it's important to differentiate between the parachain consensus provided by the relay chain and the way that framework is used in Cumulus.

What the relay chain consensus provides is that maximum one state transition is processed per relay chain block. Therefore, the head-data output by each the state transition is present in the post-state of at least one relay chain block.

The implementation of Cumulus chains means one block is one state transition. So nothing is skipped, as you would like to have as a property. But this is a Cumulus implementation detail which could change in the future by combining multiple blocks into one state transition

philipstanislaus: @rphmeier thanks for the clarification.

What does a para head in polkadot’s the paras pallet (https://github.com/paritytech/polkadot/blob/4165c7f83a85e12c7ec56366a9559bf0eba25f8d/runtime/parachains/src/paras.rs#L328) refer to? One state transition?

Also if cumulus changes in the future, we could stay on an implementation that associates one block with one state transition, right?

PR #489, opened 2021-08-04
XCMP Script & Improvements to start-services

vgeddes:
XCM
===
Added a script `scripts/make-xcm-transfer.ts` which initiates a reserve-backed asset transfer by constructing and sending an XCM message to the parachain. This script replaces the obsolete/non-working `transfer` pallet, and so XCM messages are now always constructed offchain (much easier to maintain and test).

Eventually, if our UI wants to do XCM transfers it should integrate the code in the script.

As for XCM testing, our chain is able to process the message succesfully, and emit another message for delivery to the recipient parachain. Getting the message processed on the recipient side is gonna take more work however. We can't just use a copy of our parachain.

Start-Services
===
I'm not testing XCM using the `start-services duplicate` feature as the turn-around time between runs is too long. Instead its just to more convenient to use `polkadot-launch` directly. I'm using pre-funded accounts (specified by chain spec) from which to withdraw/deposit amounts.

So I've updated start-services such that it always runs a second parachain, though this is purely for testing Beefy/MMR data structures.

Please look for through the updated `test/README.md` for details. A new version of polkadot-launch is required, plus a few other tools.


philipstanislaus (commented):
👍

PR #490, opened 2021-08-04
add enumerable test ERC721

Geimaj:
We wanted to test transfers though the UI with multiple different contracts so this deploys another ERC721 contract. This contract also implements IERC721Enumerable to enable to UI to display all the tokens owned by the user

### DIFF
--- a/ethereum/deploy/12-erc20-app.ts
+++ b/ethereum/deploy/12-erc20-app.ts
@@ -47,10 +47,4 @@ module.exports = async ({
     log: true,
   });
 
-  await deployments.deploy("TestToken721Enumerable", {
-    from: deployer,
-    args: ["Test Enum", "TESTE"],
-    log: true,
-  });
-
 };
--- a/ethereum/deploy/12-erc20-app.ts
+++ b/ethereum/deploy/12-erc20-app.ts
@@ -1,11 +1,11 @@
 require("dotenv").config();
 
-import {HardhatRuntimeEnvironment} from "hardhat/types";
+import { HardhatRuntimeEnvironment } from "hardhat/types";
 
 module.exports = async ({
-    deployments,
-    getUnnamedAccounts,
-    network,
+  deployments,
+  getUnnamedAccounts,
+  network,
 }: HardhatRuntimeEnvironment) => {
   let [deployer] = await getUnnamedAccounts();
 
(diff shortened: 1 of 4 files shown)


ethereum/deploy/12-erc20-app.ts
```diff
@@ -35,15 +35,21 @@ module.exports = async ({
...
   });
 
   await deployments.deploy("TestToken", {
     from: deployer,
-    args:["Test Token", "TEST"],
+    args: ["Test Token", "TEST"],
+    log: true,
+  });
+
+  await deployments.deploy("TestToken721Enumerable", {
```
musnit, line 50: i think this should go in /ethereum/deploy/13-erc721-app.ts rather?

musnit: could you add the loop to mint some nfts to the deployer account too in 13-erc721-app.ts  as well?

vgeddes: I'd rather leave minting and any other custom setup for the dashboard to be done in a separate hardhat script that can be run after the deployments when required. For two reasons:
1) The deployment script already takes a long time to complete, especially with geth.
2) At least for E2E testing purposes, our ERC20 and ERC721 tests are already doing the minting without relying on the deployment script (See https://github.com/Snowfork/snowbridge/blob/main/test/test/erc721app.js#L32).

ethereum/scripts/mint-nft.ts
```diff
@@ -0,0 +1,20 @@
+const hre = require("hardhat");
```
vgeddes, line 1: Should use ES6 imports:

```suggestion
import { ethers, deployments, getUnnamedAccounts } from "hardhat";
```

vgeddes (approved):
Looks, good. Added a comment about ES6 imports, but otherwise looks good.

philipstanislaus (commented):
LGTM – could you add a bit context what the reason for this addition is? What is it used for?

musnit:
@philipstanislaus ERC721Enumerable advantages here: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC721/extensions/ERC721Enumerable.sol#L9

basically, our frontend can easily query for all tokens owned by a particular user with this, whereas it can't always in non-enumerable tokens. i believe most nft dapps have extra frontend logic on a case-by-case basis to handle non-enumerable tokens, or some other workaround.

philipstanislaus:
Okay, got it. Makes sense for the demo - for a production app we may want to keep the contracts as efficient as possible and use The Graph or another indexing layer for aggregating data.

PR #495, opened 2021-08-16
Use simplified MMR Proof

ParthDesai:
### Description
`SimplifiedMMRProof` is pre-processed MMR proof format that is very easy to be verifiable by Solidity code while maintaining same security guarantee provided by regular MMR Proof.

### How does it work?
It works by pre-calculating order of the merkle tree proof so that we don't have to use mathematic operation to determine the same on solidity side. It also takes advantage of nervos library optimization where the last proof item is already bagged from right hand side. Combining all of the above means that solidity code does not have to do any mathematical operation instead just iterating through both peaks and proof and hash them.

TODO:

- [x] Writing go code to convert normal mmr proof to simplified version
- [x] Writing test for the conversion function
- [x] Writing solidity code to verify simplified mmr proof
- [x] Polishing solidity code
- [x] Integrating  code
- [x] Make sure e2e passes
- [x] Make ethereum tests pass


musnit:
Awesome, looking forward to see contract + relayer integration and getting tests passing.

I put our code from the Slack discussion here: https://github.com/Snowfork/snowbridge/blob/main/ethereum/contracts/utils/MerkleProof.sol#L39 - given this is no longer relevant and not 100% usable, you should be able to replace/remove the createMMRRoot function in this PR too and replace it with a SimplifiedMMRVerifier contract

ParthDesai:
> Awesome, looking forward to see contract + relayer integration and getting tests passing.
> 
> I put our code from the Slack discussion here: https://github.com/Snowfork/snowbridge/blob/main/ethereum/contracts/utils/MerkleProof.sol#L39 - given this is no longer relevant and not 100% usable, you should be able to replace/remove the createMMRRoot function in this PR too and replace it with a SimplifiedMMRVerifier contract

Added unpolished solidity code. Do check it out.

ethereum/contracts/SimplifiedMMRVerification.sol
```diff
@@ -0,0 +1,81 @@
+pragma solidity ^0.8.5;
+
+contract  SimplifiedMMRVerification {
+    function verifyInclusionProof(
+        bytes32 root,
+        bytes32 leafNodeHash,
+        bytes32[] memory restOfThePeaks,
+        bool hasRightBaggedPeak,
```
musnit, line 8: can we remove this bool and just always give the rightBaggedPeak? (ie, if the peak is actually the most right, then rightBaggedPeak = 0x0000000000000000)

ethereum/contracts/SimplifiedMMRVerification.sol
```diff
@@ -0,0 +1,81 @@
...
+        }
+
+        bytes32[] memory reversedPeaks = new bytes32[](numberOfPeaks);
+        uint peakInsertionPointer = 0;
+
+        if (hasRightBaggedPeak) {
+            reversedPeaks[peakInsertionPointer++] = rightBaggedPeak;
+        }
+
+        bytes32 merkleRootPeak = calculateMerkleRoot(leafNodeHash, merkleProofItems, merkleProofOrder);
```
musnit, line 28: You should be able to user our existing MerkleProof.sol library for this

ethereum/contracts/SimplifiedMMRVerification.sol
```diff
@@ -0,0 +1,81 @@
...
+        bytes32 root,
+        bytes32 leafNodeHash,
+        bytes32[] memory restOfThePeaks,
+        bool hasRightBaggedPeak,
+        bytes32 rightBaggedPeak,
+        bytes32[] memory merkleProofItems,
+        bool[] memory merkleProofOrder
+    ) public pure returns (bool) {
+        require(merkleProofOrder.length == merkleProofItems.length);
+        // TODO: Do validation of other input parameters
```
musnit, line 14: i don't think we need validation? if they are wrong, the proof will fail anyway

ethereum/contracts/SimplifiedMMRVerification.sol
ParthDesai: @musnit The function `computeRootFromProofAtPosition` in the library is less optimized. It still uses pos and width, while we do not need it.

ethereum/contracts/SimplifiedMMRVerification.sol
ParthDesai: @musnit Updated

ethereum/contracts/BeefyLightClient.sol
```diff
@@ -288,17 +290,19 @@ contract BeefyLightClient {
         Commitment calldata commitment,
         ValidatorProof calldata validatorProof,
         BeefyMMRLeaf calldata latestMMRLeaf,
-        uint64 leafIndex,
-        uint64 leafCount,
-        bytes32[] calldata mmrProofItems
+        bytes32[] calldata restOfThePeaks,
```
musnit, line 293: we should wrap these up into a single struct like BeefyMMRProof

ethereum/contracts/SimplifiedMMRVerification.sol
musnit: Yeh - though pos and width may be less bytes than merkleProofOrder, so maybe the input calldata is slightly smaller.

I'm not sure which results in an effectively cheaper gas price, though I suspect the extra compute in the library is cheaper than the extra calldata that may be used for the merkleProofOrder.

Anyway, don't worry about this, I think the difference is probably fairly insignificant anyway.

ParthDesai: @musnit What do you think about bitfield? i.e we can have a uint256

musnit: yeh true, maybe even uint16 or uint8 will be fine for a MMR for the first year or two while it has less than a billion ?

ParthDesai: I am using 64 bit bitfield now and it is tested. The reason behind 64 bit is rust code defines maximum height possible to be 63.

ParthDesai:
@musnit Let's merge this. Tests are passing now. And I have tested the implementation thoroughly.

ethereum/contracts/SimplifiedMMRVerification.sol
```diff
@@ -0,0 +1,96 @@
...
+
+        bytes32 merkleRootPeak = calculateMerkleRoot(leafNodeHash, proof.merkleProofItems, proof.merkleProofOrderBitField);
+        reversedPeaks[peakInsertionPointer++] = merkleRootPeak;
+
+        if (proof.restOfThePeaks.length > 0) {
+            for (uint i = 0; i < proof.restOfThePeaks.length; i++) {
+                reversedPeaks[peakInsertionPointer] = proof.restOfThePeaks[proof.restOfThePeaks.length - i - 1];
+                peakInsertionPointer++;
+            }
+        }
```
philipstanislaus, line 41: I still don't understand why we cannot convert this whole logic into a simple merkle proof. I made an example that shows how a MMR can be transformed into a simple merkle proof – could you point out where I am wrong or provide a counterexample? 

Screenshot 2021-09-16 at 23 08 52

Thanks!

relayer/crypto/merkle/simplified_mmr_proof_test.go
```diff
@@ -0,0 +1,110 @@
...
+		fmt.Printf("\nRoot: %#x\n", testData[i].ReferenceMMRRoot)
+		fmt.Printf("Leaf Node hash: %#x\n", testData[i].LeafHash)
+
+		// Printing solidity fixture
+		fixture := convertSimplifiedProofToFixture(simplifiedProof)
+		prettyOut, err := json.MarshalIndent(fixture, "", "\t")
+		assert.NoError(t, err)
+		fmt.Println(string(prettyOut))
+	}
+}
```
philipstanislaus, line 110: For better test coverage, could you please add the following two tests? 

1. Programmatically generate a 15 leaf MMR and then create and validate a proof for each of the 15 leaves, as well as validate proof failure with a leaf that's not in the MMR. 
2. Verify that the code works on a 1 leaf MMR, as well as validate proof failure with a leaf that's not in the MMR.

philipstanislaus (commented):
This is great! 

Left a few comments. 

Can we also remove the previous MMR verification code – that should no longer be needed, right?

ethereum/contracts/SimplifiedMMRVerification.sol
```diff
@@ -0,0 +1,96 @@
...
+
+contract  SimplifiedMMRVerification {
+    function verifyInclusionProof(
+        bytes32 root,
+        bytes32 leafNodeHash,
+        SimplifiedMMRProof memory proof
+    ) public pure returns (bool) {
+        require(proof.merkleProofItems.length < 64);
+
+        bool hasRightBaggedPeak = proof.rightBaggedPeak != 0x0;
```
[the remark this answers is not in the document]
ParthDesai: @philipstanislaus :) That was the original design. @musnit asked me to change it this way to save on gas costs.

ethereum/contracts/SimplifiedMMRVerification.sol
```diff
@@ -0,0 +1,96 @@
+// "SPDX-License-Identifier: UNLICENSED"
+pragma solidity ^0.8.5;
+
+struct SimplifiedMMRProof {
+    bytes32[] restOfThePeaks;
+    bytes32 rightBaggedPeak;
```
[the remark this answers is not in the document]
ParthDesai: https://github.com/Snowfork/snowbridge/pull/495#discussion_r710662363

ethereum/contracts/SimplifiedMMRVerification.sol
ParthDesai: Part of the reason is the nodes just below the root have different handling then in binary merkle proof. In binary merkle tree those nodes can only be two, so the merkle proof generation/verification code can handle them easily using the recursive algorithm. 

But in MMR the nodes just below root can be any number and the algorithm to calculate root is different.

ethereum/contracts/SimplifiedMMRVerification.sol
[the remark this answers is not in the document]
philipstanislaus: @musnit what are your thoughts here, are we fine with a small likelihood that this will fail? It's probably not relevant in practice, but may shed a negative light on us in audits or completely break the protocol (same probability that I will guess your private key, and same probability that someone actually owns the private key for the 0x0 address on Ethereum).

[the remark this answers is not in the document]
ParthDesai: @philipstanislaus The peak is hash, so technically they would have to find a node which hashes to 0x0. This node has following structure: 
```
struct BeefyMMRLeafPartial {
        uint8 version;
        uint32 parentNumber;
        bytes32 parentHash;
        uint64 nextAuthoritySetId;
        uint32 nextAuthoritySetLen;
        bytes32 nextAuthoritySetRoot;
}
```

So, you will have to generate a node with all of this field valid and then that node need to hash 0x0 in order to do this.
I would consider probability of this extremely low.

ethereum/contracts/SimplifiedMMRVerification.sol
philipstanislaus: Could you give me an example where the peaks are not handled in a binary way? 

The code you implemented does pairwise (binary) hashing: https://github.com/Snowfork/snowbridge/pull/495/files/ec5a2a4cccf4e17998a9dacf1ac6ba2f3c9308ab#diff-ccff880ac8ad6b758b2b22c2c6ed830cb524733bd4a05d38d3ad37165f6333e1R55 
Also the example you give `Hash(B, Hash(C, D))` is pairwise, and can be represented as a binary merkle tree.

Do you have an example where the parent is composed of more than two children?

philipstanislaus: Here is a graphic for how the MMR above looks as a binary merkle tree: 

Screenshot 2021-09-17 at 08 43 04

ethereum/contracts/SimplifiedMMRVerification.sol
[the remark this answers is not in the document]
musnit: i don't think we need to be secure against someone finding the preimage of 0x0 - even with a [hypercomputer](https://en.wikipedia.org/wiki/Hypercomputation), that preimage is likely not even valid input given the constraints @ParthDesai mentions above :P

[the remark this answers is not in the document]
philipstanislaus: Agree, but I’m less concerned about an attack, but rather about a failure if the actual/valid hash of the right bag is 0x0. In that case, a valid MMR proof will fail. 

But it’s not a real concern, because we could simply wait for the next beefy commitment, which changes the right bagged peaks hash, resolving the issue.

All good to keep it 👌👍

ethereum/contracts/SimplifiedMMRVerification.sol
ParthDesai: Ok. Simplified the calculation.

relayer/crypto/merkle/simplified_mmr_proof_test.go
ParthDesai: @philipstanislaus 

1. The go code does not contain the logic to create and validate a proof for each of the 15 leaves. This test exists on the Rust side.
2. Same as above.

ParthDesai:
@philipstanislaus @musnit Everything is fixed now. I am keeping the normal mmr contract since @musnit has written a test to show the gas cost difference. Let's merge it.

vgeddes:
Thanks Parth! Merging...

musnit:
Looks good. merged now so we can start testing it on staging but @philipstanislaus plz check over the last new commita still too, thanks!

philipstanislaus (commented):
Nice, thanks 🎉

PR #502, opened 2021-08-19
confirm the ethereum tx receipts proof works before sending

musnit:


musnit:
@vgeddes do u think we should add this?

vgeddes:
Nah, seems a bit unnecessary. All its doing is veryifing that go-ethereum can prove messages with proofs generated by itself. It doesn't say anything about the correctness of our ethereum-light-client.

PR #507, opened 2021-08-30
Improve relayer reliability

vgeddes:
Fixed a bunch of reliability issues that were causing problems:

* All relays: Prevent deadlocks by ensuring all channel writes are non-blocking
* All relays: Prevent deadlocks by cancelling all spawned goroutines if relay did not initialize entirely successfully   
* Beefy Relay / Ethereum Listener: Add error handling that was previously non-existent.
* Ethereum Relay / Parachain Writer: Simplified the scope/design of the outgoing extrinsics pool to reduce the number of spurious failures.
* Misc refactorings for conceptual clarity


relayer/relays/beefy/beefy-ethereum-listener.go
```diff
@@ -203,63 +272,79 @@ func (li *BeefyEthereumListener) processHistoricalInitialVerificationSuccessfulE
...
+
+				found = true
 				break
 			}
 		}
-		if !itemFoundInDatabase {
+		if !found {
 			// Don't have an existing item to update, therefore we won't be able to build the completion tx
-			log.Error("BEEFY justification data not found in database for InitialVerificationSuccessful event. Ignoring event.")
+			return fmt.Errorf("item not found in database for InitialVerificationSuccessful event")
```
musnit, line 288: i think we just want to ignore if there is nothing found, so maybe should not return an error?

relayer/relays/beefy/beefy-ethereum-listener.go
```diff
@@ -203,63 +272,79 @@ func (li *BeefyEthereumListener) processHistoricalInitialVerificationSuccessfulE
...
 			continue
 		}
 
-		item := li.beefyDB.GetItemByInitialVerificationTxHash(event.Raw.TxHash)
-		if item.Status != store.InitialVerificationTxSent {
-			continue
+		item, err := li.beefyDB.GetItemByInitialVerificationTxHash(event.Raw.TxHash)
+		if err != nil {
+			log.Error("Failed to retrieve item from Beefy DB")
+			return err
```
musnit, line 331: again, not sure if we want to actually have an error here, or just ignore and continue?

relayer/relays/beefy/beefy-ethereum-listener.go
```diff
@@ -289,67 +374,55 @@ func (li *BeefyEthereumListener) queryFinalVerificationSuccessfulEvents(ctx cont
...
 		}
 
-		log.Infof(
-			"6: Deleting finalized item %s from the database",
-			event.Id,
-		)
+		item, err := li.beefyDB.GetItemByID(event.Id.Int64())
+		if err != nil {
+			log.Error("Failed to retrieve item from Beefy DB")
+			return err
```
musnit, line 415: again, not sure if we want to actually return error, or just ignore this?

relayer/relays/ethereum/parachain-writer.go
```diff
@@ -212,20 +188,7 @@ func (wr *ParachainWriter) WritePayload(ctx context.Context, payload *ParachainP
 		return err
 	}
 
-	onProcessed := func() error {
-		// Confirm that the header import was successful
-		header := payload.Header.HeaderData.(ethereum.Header)
-		hash := header.ID().Hash
-		imported, err := wr.queryImportedHeaderExists(hash)
```
musnit, line 219: you don't think this is needed anymore? so its basically fire and forget?

vgeddes: Yeah good question. The problem is that it wasn't really working 30% of time. In the sense that the extrinsic was successfully included in a block, but the query failed. Was causing a huge number of spurious restarts.

So I want to go back to basics and then potentially add it back at a later stage.

relayer/relays/beefy/beefy-ethereum-listener.go
vgeddes: I think as a general principle we want the relay to fail-fast on any unexpected error, 1) to be able to reset things to a known good state, 2) to make the error visible so we can address any flaws.

The fetches from the underlying SQL db could fail for many reasons, including IO errors, or a bug in a previous step which needed to initialize the item in the DB. If we don't handle the error, the underlying error could manifest in other problems down the line.

relayer/relays/beefy/beefy-ethereum-listener.go
vgeddes: Yeah, OK this one I understand better now. Will remove the error check here.

alistair-singh (approved):

relayer/relays/beefy/beefy-ethereum-listener.go
vgeddes: I've updated the error handling to handle both yours and mine's concerns. The new code will just skip if no record is found, and fail fast on any other error.

relayer/relays/ethereum/parachain-writer.go
vgeddes: Looks like the import check is needed for healthy operation. So I've added it back, but with tweaks - we only perform the check if the extrinsic was finalized, rather than merely being included in a block. This seems to prevent the spurious import check errors.

relayer/relays/beefy/beefy-ethereum-listener.go
vgeddes: fixed

musnit (approved):

PR #509, opened 2021-09-01
Get starting block from beefy light client

alistair-singh:
### Info
This change is a draft PR for getting the latest block from the beefy light client to get some feedback.

### Testing
e2e tests pass.


musnit:
looks good

alistair-singh:
Tested it with the contract changed as suggested. The relayer caught up from the oldest block. After catch up I ran the e2e tests and they passed.

vgeddes (approved):
Looks good!

PR #510, opened 2021-09-02
Sync historical beefy justifications on relayer start up

alistair-singh:
This is a continuation from PR https://github.com/Snowfork/snowbridge/pull/497. That PR is closed because we have switched back to hybrid and the `main` branch has had updates which need to be included.

We had to switch back to a hybrid model where we poll for synchronization and subscribe for new justifications as the poll code misses justifications. It seems like the beefy justifications are sometimes added after block finalization.

### DIFF
--- a/relayer/relays/beefy/config.go
+++ b/relayer/relays/beefy/config.go
@@ -10,8 +10,8 @@ type Config struct {
 }
 
 type SourceConfig struct {
-	Polkadot           config.PolkadotConfig `mapstructure:"polkadot"`
-	SyncSkipBlockCount uint64                `mapstructure:"sync-skip-block-count"`
+	Polkadot        config.PolkadotConfig `mapstructure:"polkadot"`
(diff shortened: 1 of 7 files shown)


relayer/relays/beefy/config.go
```diff
@@ -10,7 +10,8 @@ type Config struct {
 }
 
 type SourceConfig struct {
-	Polkadot config.PolkadotConfig `mapstructure:"polkadot"`
+	Polkadot           config.PolkadotConfig `mapstructure:"polkadot"`
+	SyncSkipBlockCount uint64                `mapstructure:"sync-skip-block-count"`
```
vgeddes, line 14: Just need to change this to `beefy-skip-period` or something like that.

vgeddes (approved):
Approved! Added one comment about `sync-skip-block-count`, but otherwise looks good!

PR #517, opened 2021-09-14
The beefy relay no longer tries to process historical verification events

vgeddes:
The historical processing actually never really worked anyway because the sqlite DB is reset between restarts. Now that we have beefy catchup in place, the DB will be refilled with _new_ events for _old_ beefy blocks. so we should be safe.

Also reduced the chances of submitting ethereum transactions with the same nonce. Not a perfect solution, can be improved later on.

E2E tests pass


musnit (approved):
lgtm - if e2e tests pass can merge

PR #518, opened 2021-09-16
Experiments fetching MMR Leafs

vgeddes:
Looks like the E2E test still pass repeatedly when we don't decrement the block number when calling `GetMMRLeafForBlock`.

Also some cleanups for the configuration of when Beefy was activated on the the relay chain.

Leaving in draft I while I attend to other pressing matters.


musnit:
Code looks good - would be nice to resolve 100% the index understanding though

alistair-singh (approved):

musnit:
I don't know if we should merge this unless we know it actually fixes/improves something. Seems to just keep functionality the same? May be worth trying against staging though if you run it locally against staging?

alistair-singh (approved):

PR #519, opened 2021-09-16
Changed Parachain relayer to use state_getKeysPaged

alistair-singh:
Using state_getKeysPaged to get para heads. The state_getKeysPaged RPC call has been observed to be faster than state_getKeys in staging.

https://linear.app/snowfork/issue/SNO-51/parachain-relay-is-unlikely-to-ever-catch-up-in-production

Todo
- [x] Move page size to config
- [x] e2e testing.

### DIFF
--- a/relayer/config/config.go
+++ b/relayer/config/config.go
@@ -6,8 +6,7 @@ type PolkadotConfig struct {
 }
 
 type ParachainConfig struct {
-	Endpoint          string `mapstructure:"endpoint"`
-	KeyFetchBlockSize uint32 `mapstructure:"key-fetch-block-size"`
+	Endpoint string `mapstructure:"endpoint"`
 }
 
 type EthereumConfig struct {
--- a/relayer/config/config.go
+++ b/relayer/config/config.go
@@ -6,7 +6,8 @@ type PolkadotConfig struct {
 }
 
 type ParachainConfig struct {
-	Endpoint string `mapstructure:"endpoint"`
+	Endpoint          string `mapstructure:"endpoint"`
+	KeyFetchBlockSize uint32 `mapstructure:"key-fetch-block-size"`
 }
 
 type EthereumConfig struct {
--- a/relayer/go.mod
+++ b/relayer/go.mod
@@ -23,7 +23,7 @@ require (
 	github.com/rs/cors v1.8.0 // indirect
 	github.com/sirupsen/logrus v1.7.0
 	github.com/snowfork/ethashproof v0.0.0-20210729080250-93b61cd82454
-	github.com/snowfork/go-substrate-rpc-client/v3 v3.0.7
+	github.com/snowfork/go-substrate-rpc-client/v3 v3.0.8
 	github.com/spf13/afero v1.5.1 // indirect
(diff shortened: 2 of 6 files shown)


relayer/go.mod
```diff
@@ -37,3 +37,5 @@ require (
 	gopkg.in/ini.v1 v1.62.0 // indirect
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 )
+
+replace github.com/snowfork/go-substrate-rpc-client/v3 => /home/alistair/go-substrate-rpc-client
```
alistair-singh, line 41: Will remove this once related PR is merged. https://github.com/Snowfork/go-substrate-rpc-client/pull/20

alistair-singh:
Ran e2e tests and it passes. Also tested a catch up case where I run the tests with the parachain relayer stopped to generate some backlog. Then start the relayer to confirm a catch up is done and the new code path is hit. Then run the e2e tests again while the parachain relayer is running and the tests pass.

relayer/config/config.go
```diff
@@ -6,7 +6,8 @@ type PolkadotConfig struct {
 }
 
 type ParachainConfig struct {
-	Endpoint string `mapstructure:"endpoint"`
+	Endpoint          string `mapstructure:"endpoint"`
+	KeyFetchBlockSize uint32 `mapstructure:"key-fetch-block-size"`
```
vgeddes, line 10: I'm not sure its worth making this configurable. Its more a constant that is unlikely to change much.

Also for future reference, this `ParachainConfig` is purely for client and connection-related config, such as endpoints, timeouts, and so on. The struct is shared between three different relayers and so `KeyFetchBlockSize` will be irrelevent for the other relays (ethereum relay and the beefy relay). 

So if you want add a config key next time, rather put it into one of these structs: 

https://github.com/Snowfork/snowbridge/blob/2c61adf8e3eecf419ca238a5b6e6ba1cb6cde03c/relayer/relays/parachain/config.go#L10

https://github.com/Snowfork/snowbridge/blob/2c61adf8e3eecf419ca238a5b6e6ba1cb6cde03c/relayer/relays/parachain/config.go#L23

In the terminology I've used to organise configuration (for the parachain relay), A source is from where events originate (i.e the parachain and the relaychain), and the sink is where we need to write events (Ethereum).

The configuration for the other relays is modelled the same.

test/config/parachain-relay.json
```diff
@@ -8,7 +8,8 @@
             "beefy-starting-block": 0
         },
         "parachain": {
-            "endpoint": "ws://localhost:11144"
+            "endpoint": "ws://localhost:11144",
+            "key-fetch-block-size": 50
```
vgeddes, line 12: See above comment, I don't think this is worth making configurable.

vgeddes (approved):
Looks great! Approving, with one comment about configuration though.

relayer/config/config.go
alistair-singh: Is it fine do leave this as a `const` in the code? Or should I move it to the appropriate Source/Sink Config?

vgeddes: A const in the code should be fine

PR #521, opened 2021-09-22
Increase confirmation time for Ropsten

vgeddes:
On staging, the parachain accepted a message from a non-canonical fork, which means later messages from the canonical chain were rejected due to an invalid nonce. This is the only interpretation we can make for the invalid nonce errors.

The ethereum light client is currently configured with a very short confirmation time - 3 blocks. Bumping this up to 8 blocks for now. Though we may need to increase it even more after further evaluation.


vgeddes:
Merging as I need to rebuild staging this morning.

PR #524, opened 2021-09-29
private->public vars

musnit:


vgeddes (approved):
Looks good. Ideally we'd expose the fee via a getter, but that's unimportant for now.

musnit:
> Looks good. Ideally we'd expose the fee via a getter, but that's unimportant for now.

solidity public variables automatically create getters :)

PR #525, opened 2021-10-01
Relay should submit correct MMR proofs for historical commitments

vgeddes:
An early WIP solution that makes the following changes:

* Generation of MMR proofs happens right before transaction is submitted to inbound channels on Ethereum
* Simplifications & potential speedup for the code that scans relay chain blocks.

Resolves SNO-112

### DIFF
--- a/relayer/relays/parachain/beefy-listener.go
+++ b/relayer/relays/parachain/beefy-listener.go
@@ -26,7 +26,7 @@ type BeefyListener struct {
 	relaychainConn      *relaychain.Connection
 	parachainConnection *parachain.Connection
 	paraID              uint32
-	messages            chan<- MessagePackage
+	messages            chan<- *Task
 }
 
 func NewBeefyListener(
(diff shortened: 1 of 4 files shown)


relayer/chain/relaychain/connection.go
```diff
@@ -68,26 +68,43 @@ func (co *Connection) Close() {
...
-	// blocks produced before beefy was activated. We subtract the block in which beefy was started on the
+	// blocks produced before beefy was activated. We subtract the block in which beefy was activated on the
 	// chain to account for this.
-	leafIndex := blockNumber - beefyStartingBlock - 1
+	//
+	// LeafIndex(currentBlock, activationBlock) := currentBlock - Max(activationBlock, 1)
+	//
+	// Example: LeafIndex(5, 3) = 2
+	//
+	// Block Number: 1 -> 2 -> 3 -> 4 -> 5
```
musnit, line 91: very nice thx :)

PR #533, opened 2021-11-01
SNO-104 Add destination parachain id as parameter.

alistair-singh:
Resolves: https://linear.app/snowfork/issue/SNO-104/ethereum-apps-should-specify-destination-parachain

Some plumbing to pass parachain id when we 

e2e tests pass.

TODO:
- [x] Build and send Xcm packet
- [x] Do the same for ERC20App and ERC721App
- [x] Run e2e tests


parachain/pallets/eth-app/src/lib.rs
```diff
@@ -141,13 +142,19 @@ pub mod pallet {
 			sender: H160,
 			recipient: <T::Lookup as StaticLookup>::Source,
 			amount: U256,
+			para_id: u32,
```
vgeddes, line 145: Lets rather make `para_id` an `Option<u32>`, so we explicitly know if the Solidity side passed a para_id or not. See my equivalent comments on the solidity side.

parachain/primitives/core/src/assets.rs
```diff
@@ -42,3 +44,7 @@ pub trait SingleAsset<AccountId> {
 
 	fn deposit(who: &AccountId, amount: U256) -> DispatchResult;
 }
+
+pub trait XcmTransactAsset<AccountId> {
```
vgeddes: I think `XcmReserveTransfer` would be a more appropriate name.

vgeddes (commented):
Looks good so far. Added a few comments (even though this is still a draft).

alistair-singh:
e2e tests pass.

I have made quite a few assumptions about how an XCM packet for a reserve transfer is supposed to be used. So we need to address SNO-47 and add new XCM transfer e2e tests.

I have also not added any unit tests for eth-app and erc20-app. We can mock the XCM so that it fails but that would cause the mint call to outright fail. We do not do any logic based on the error cases. e2e tests are more important at this stage.

ethereum/contracts/ERC20App.sol
```diff
@@ -110,16 +113,30 @@ contract ERC20App is AccessControl {
         address _token,
         address _sender,
         bytes32 _recipient,
-        uint256 _amount
+        uint256 _amount,
+        uint32 _paraId
     ) private pure returns (bytes memory) {
-        return
-            abi.encodePacked(
+        if(_paraId == 0) {
```
musnit, line 119: question - should we handle these 2 cases differently at the application level here, or rather just at the dispatch level when received by our parachain?

musnit: i guess it's cheaper to handle it here?

musnit:
for the reverse direction, ie, third-party-parachain to ethereum, i suppose this can be handled purely through composing xcm messages, without any new logic needing to be added on our chain/applications. is that correct? if so, we should have a script+test for the reverse direction once we've got SNO-47 sorted

ethereum/contracts/ERC20App.sol
alistair-singh: I am not sure if it would be cheaper. My guess is that handling this in rust/parachain would be cheaper since rust has more opportunity for optimization given its static nature (this would depend on benchmarks to calculate weight). I think using optional types on the rust end buys us cleaner API(and code) for the mint call on the parachain side.

alistair-singh:
> for the reverse direction, ie, third-party-parachain to ethereum, i suppose this can be handled purely through composing xcm messages, without any new logic needing to be added on our chain/applications. is that correct? if so, we should have a script+test for the reverse direction once we've got [SNO-47](https://linear.app/snowfork/issue/SNO-47) sorted

I think we would have to add more code to interpret the xcm received and push a mint/burn message onto the outbound channel. 
We have code to do this for the assets pallet so I assume it would need to be the similar for the reverse direction. https://github.com/Snowfork/snowbridge/blob/main/parachain/primitives/xcm-support/src/lib.rs#L53-L79

The code would have to skip interacting with the assets pallet and build a message to mint on the dotapp contract.

musnit:
> > for the reverse direction, ie, third-party-parachain to ethereum, i suppose this can be handled purely through composing xcm messages, without any new logic needing to be added on our chain/applications. is that correct? if so, we should have a script+test for the reverse direction once we've got [SNO-47](https://linear.app/snowfork/issue/SNO-47) sorted
> 
> I think we would have to add more code to interpret the xcm received and push a mint/burn message onto the outbound channel. We have code to do this for the assets pallet so I assume it would need to be the similar for the reverse direction. https://github.com/Snowfork/snowbridge/blob/main/parachain/primitives/xcm-support/src/lib.rs#L53-L79
> 
> The code would have to skip interacting with the assets pallet and build a message to mint on the dotapp contract.

yeh - i imagine for the transfer-dot-to-ethereum flow the xcm would just directly call mint on the dotapp, rather than interacting with the outbound channel directly

ethereum/contracts/ERC20App.sol
vgeddes: yeah, Its not really a matter of performance/cost, just about a cleaner API as Alistair says.

vgeddes (commented):
Looks good overall! Added a first round of comments.

ethereum/contracts/ETHApp.sol
```diff
@@ -93,13 +121,31 @@ contract ETHApp is RewardSource, AccessControl {
...
+                _sender,
+                bytes1(0x00), // Encode recipient as MultiAddress::Id
+                _recipient,
+                _amount.encode256(),
+                bytes1(0x00)
+            );
+    }
+
+    // SCALE-encode payload with parachain Id
+    function encodeCall(
```
vgeddes, line 135: To reduce the chance of bugs being introduced in future, rather explicitly give these `encodeCall` functions different names, such as `encodeCall` and `encodeCallWithParaId`, etc.

vgeddes (approved):
Approving. Added 1 comment which should be addressed before merging.

PR #534, opened 2021-11-04
Sno 136 v0.9.12 upgrade

alistair-singh:
Resolves: https://linear.app/snowfork/issue/SNO-136/parachain-xcm-v3-support

* Added new constants. To choose these I just eyeballed common settings for non test/mock code using github code search.
    1. `OperationalFeeMultiplier` is 5.
    2. `MaxInstructions` is 100.
    3. `VERSION_DISCOVERY_QUEUE_SIZE` is 100.
    4. `AdvertisedXcmVersion` is currently set to 3 and probably should be 1.
    5. `MaxAuthorities` is 32.
    6. `force_unreserve` default weight is 54_482_000.
* I chose to ignore some change sets from the parachain template:
  1. Removing `pallet-randomness-collective-flip` crate.
  2. Adding `sc-transaction-pool-api` crate. [code](https://github.com/substrate-developer-hub/substrate-parachain-template/blob/eb9d228e332260fe0e3e1a8319764f33945a77b5/node/Cargo.toml#L62)
  3. Adding `ss58Format` to chain_spec properties. [code](https://github.com/substrate-developer-hub/substrate-parachain-template/blob/eb9d228e332260fe0e3e1a8319764f33945a77b5/node/src/chain_spec.rs#L66)
  4. Choosing not to handle `AssetTrap` and `AssetClaims` (Leaving it set to `()`). Template uses `PolkadotXcm`. [code](https://github.com/substrate-developer-hub/substrate-parachain-template/blob/eb9d228e332260fe0e3e1a8319764f33945a77b5/runtime/src/lib.rs#L471-L474)
  5. New constraints on `start_node_impl`. [code](https://github.com/substrate-developer-hub/substrate-parachain-template/blob/eb9d228e332260fe0e3e1a8319764f33945a77b5/node/src/service.rs#L192-L193)
```
		+ pallet_transaction_payment_rpc::TransactionPaymentRuntimeApi<Block, Balance>
		+ substrate_frame_rpc_system::AccountNonceApi<Block, AccountId, Nonce>,
```
* `start-services.sh` works after updating @polkadot/api to 6.8.1 because of new metadata format.
  1. Relayers fail with ”unsupported metadata version 14" because GRPC support is TDB. This means e2e tests cannot be run yet.

I am going to upgrade XCM in a follow up commit as this one is a bit large.


vgeddes (commented):
Looks good so far. I've reviewed and found a few things, besides my other comments:

* The `#[pallet::metadata()]` macro has been deprecated and should no longer be used to decorate `pub enum Event` in pallets.
* When implementing the `frame_system::Config` trait in the runtimes and mocks for tests, the `BaseCallFilter` type should be `frame_support::traits::Everything` instead of `()`.
* The `MessageId` type in the `primitives/core` library should have `TypeInfo` implemented.
* In all crates that depend on `serde` and `codec`, bump the versions as follows to match Substrate:
```toml
serde = { version = "1.0.130", optional = true }
codec = { version = "2.2.0", package = "parity-scale-codec", default-features = false, features = ["derive"] }
```

parachain/pallets/ethereum-light-client/Cargo.toml
```diff
@@ -15,22 +15,23 @@ codec = { version = "2.0.0", package = "parity-scale-codec", default-features =
 hex = { package = "rustc-hex", version = "2.1.0", default-features = false }
 hex-literal = { version = "0.3.1", optional = true }
 rlp = { version = "0.5", default-features = false }
+scale-info = { version = "1.0", default-features = false, features = ["derive"] }
```
vgeddes, line 18: Can we order this import so that it comes right after `codec`? Serdec, codec, and scale-info seem to be needed in most pallets so its good to group them together.

parachain/pallets/assets/Cargo.toml
```diff
@@ -8,22 +8,23 @@ edition = "2018"
 targets = ["x86_64-unknown-linux-gnu"]
 
 [dependencies]
-codec = { version = "2.0.0", package = "parity-scale-codec", default-features = false, features = ["derive"] }
+serde = { version = "1.0.130", optional = true }
+codec = { version = "2.2.0", package = "parity-scale-codec", default-features = false, features = ["derive"] }
+scale-info = { version = "1.0", default-features = false, features = ["derive"] }
```
vgeddes, line 13: Wherever you add `scale-info` as a dependency, you also need to include its `std` [feature](https://github.com/paritytech/scale-info/blob/d13be2a6437493a0ef8ed5a471c16abcc3e424a5/Cargo.toml#L26) in the `std` feature of  the crates where it is used (See further below in this file). This is all to ensure that conditional compilation continues to work. See https://doc.rust-lang.org/cargo/reference/features.html for more details.

The `std` feature is used by Parity to compile code to work with Rust's standard library. However parts of a Substrate blockchain, like its runtime code are also [compiled to WASM](https://docs.substrate.io/v3/advanced/executor/#wasm-execution) where the standard library is not available.

vgeddes (approved):
Looks great! Lets get this merged in.

I'm OK with E2E tests not working until GSRPC has been upgraded with Metadata 14. This doesn't block us in any way though, since we can still build previous versions of the blockchain for local and staging.

PR #545, opened 2021-12-06
SNO-142 ETHApp.sol: should not use transfer() for transferring Ether

sumitsnk:
This PR resolves the issue https://linear.app/snowfork/issue/SNO-142/ethappsol-should-not-use-transfer-for-transferring-ether

### DIFF
--- a/ethereum/contracts/ETHApp.sol
+++ b/ethereum/contracts/ETHApp.sol
@@ -3,13 +3,14 @@ pragma solidity ^0.8.5;
 pragma experimental ABIEncoderV2;
 
 import "@openzeppelin/contracts/access/AccessControl.sol";
+import "@openzeppelin/contracts/security/ReentrancyGuard.sol";
 import "./RewardSource.sol";
 import "./ScaleCodec.sol";
 import "./OutboundChannel.sol";
 
 enum ChannelId {Basic, Incentivized}
 
-contract ETHApp is RewardSource, AccessControl {
+contract ETHApp is RewardSource, AccessControl, ReentrancyGuard  {
     using ScaleCodec for uint256;
 
     uint256 public balance;
@@ -75,7 +76,7 @@ contract ETHApp is RewardSource, AccessControl {
         bytes32 _sender,
         address payable _recipient,
         uint256 _amount
-    ) public onlyRole(INBOUND_CHANNEL_ROLE) {
+    ) public onlyRole(INBOUND_CHANNEL_ROLE) nonReentrant(){
         require(_amount > 0, "Must unlock a positive amount");
         require(
             balance >= _amount,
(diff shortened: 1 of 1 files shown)


ethereum/contracts/ETHApp.sol
```diff
@@ -75,15 +76,16 @@ contract ETHApp is RewardSource, AccessControl {
         bytes32 _sender,
         address payable _recipient,
         uint256 _amount
-    ) public onlyRole(INBOUND_CHANNEL_ROLE) {
+    ) public onlyRole(INBOUND_CHANNEL_ROLE) nonReentrant(){
```
musnit, line 79: why is nonReentrant used/necessary here?

ethereum/contracts/ETHApp.sol
```diff
@@ -107,7 +109,9 @@ contract ETHApp is RewardSource, AccessControl {
         external
         override
         onlyRole(REWARD_ROLE)
+        nonReentrant()
```
musnit, line 112: why is nonReentrant used/necessary here?

sumitsnk: The usage of the call function forwards all remaining gas. So this might be vulnerable to re-entrancy attack.
Please refer link for more info : https://medium.com/coinmonks/protect-your-solidity-smart-contracts-from-reentrancy-attacks-9972c3af7c21

musnit: yeh, adding this makes sense as a general default practice, however I don't think we should blindly follow practices without investigating and confirming 100% whether or not they are applicable to our code. Similarily, whenever there is some uncertainty, ie _"this might be vulnerable"_, we should flag that and resolve the uncertainty to be sure exactly whether or not there is an issue, rather than just making an assumption and using a practice

In this case, I don't believe there is a re-entrancy attack vector, specifically because:
 - the function is permissioned - only a specific contract which we already know the code for can call this, and that contract cannot re-enter
 - even if for some reason that contract could re-enter, or another contract was used that could, there is still no attack vector as this function uses a check-effects-interactions pattern which already protects against re-entrancy: https://docs.soliditylang.org/en/v0.6.11/security-considerations.html#use-the-checks-effects-interactions-pattern

vgeddes (commented):
Sumit, lets remove all the re-entrancy guards and try get this merged in.

Make sure to run `npx hardhat test` to verify everything works.

ethereum/contracts/ETHApp.sol
```diff
@@ -83,7 +83,8 @@ contract ETHApp is RewardSource, AccessControl {
         );
 
         balance = balance - _amount;
-        _recipient.transfer(_amount);
+        (bool success, ) = _recipient.call{value: _amount}("");
+        require(success, "Unable to send Ether");
```
musnit, line 87: I think we need to just double check on this line:
 - if the send fails, and the message is within a batch of 10 other messages, will this message break the whole batch or not?

I think not and this is likely fine - I do recall creating a test for this scenario too.

Can you double check @sumitsnk ?

musnit (approved):

vgeddes (approved):

PR #546, opened 2021-12-06
 SNO-143 ERC20 tokens could get locked up permanently

sumitsnk:
This PR resolves the issue https://linear.app/snowfork/issue/SNO-143/erc20appsol-tokens-could-get-locked-up-permanently

### DIFF
--- a/ethereum/contracts/ERC20App.sol
+++ b/ethereum/contracts/ERC20App.sol
@@ -81,7 +81,10 @@ contract ERC20App is AccessControl {
         );
         channel.submit(msg.sender, call);
 
-        IERC20(_token).safeTransferFrom(msg.sender, address(this), _amount);
+        require(
+            IERC20(_token).transferFrom(msg.sender, address(this), _amount),
+            "Contract token allowances insufficient to complete this lock request"
+        );
     }
 
     function unlock(
--- a/ethereum/contracts/ERC20App.sol
+++ b/ethereum/contracts/ERC20App.sol
@@ -4,6 +4,7 @@ pragma experimental ABIEncoderV2;
 
 import "@openzeppelin/contracts/access/AccessControl.sol";
 import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
 import "./ScaleCodec.sol";
 import "./OutboundChannel.sol";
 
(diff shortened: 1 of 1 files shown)


musnit:
I'm also not sure about if we should implement this or not. Consider the following: https://samczsun.com/the-dangers-of-surprising-code/

Even considering the above link though, it may still be a good idea to keep using safeTransfer to protect against permanent lockups. Let's think about this more before merging.

ethereum/contracts/ERC20App.sol
```diff
@@ -79,10 +81,7 @@ contract ERC20App is AccessControl {
         );
         channel.submit(msg.sender, call);
 
-        require(
-            IERC20(_token).transferFrom(msg.sender, address(this), _amount),
-            "Contract token allowances insufficient to complete this lock request"
-        );
+        IERC20(_token).safeTransferFrom(msg.sender, address(this), _amount);
```
musnit, line 84: I'm not sure this is relevant for us - this line transfers tokens from other contracts into ours, and so permanent lockup may not make sense? I think as is this may actually break the contract because it does not implement IERC721Receiver, which is a requirement for accepting a safeTransferFrom.

ethereum/contracts/ERC20App.sol
```diff
@@ -98,10 +97,7 @@ contract ERC20App is AccessControl {
         );
 
         balances[_token] = balances[_token] - _amount;
-        require(
-            IERC20(_token).transfer(_recipient, _amount),
-            "ERC20 token transfer failed"
-        );
+        IERC20(_token).safeTransfer(_recipient, _amount);
```
musnit, line 100: This makes sense in the unlock function, but I'm also still not sure about if we should implement this or not. Plz read through and make sure you understand the following: https://samczsun.com/the-dangers-of-surprising-code/

Basically, safeTransfer introduces the potential for a re-entrancy attack. Having said that though, it is still nice to have permanent lockup protection. In our case I don't think there is any risk of a re-entrancy attack as this function is permissioned and uses check-effects-interaction, and so I think it's fine to keep safeTransfer.

ethereum/contracts/ERC20App.sol
sumitsnk: On looking around the code for the SafeERC20 token which basically uses the low-level call to perform the function call. Also there is no requirement for any IERC20Receiver while using SafeERC20.sol.

musnit: OK - having said that, I don't think safeTransferFrom adds any security here in the lock function, so is likely wasted gas. I don't think we should use it in this function.

musnit (approved):

vgeddes (approved):
LGMT.

PR #540, opened 2021-11-23
Allow Ethereum apps to call dispatchables on any parachain

vgeddes:
This change allows the inbound channel in the snowbridge parachain to forward dispatchable calls to third-party parachains.

Changes:
* The dispatch pallet is now responsible for either executing the message locally or forwarding it to a third-party parachain
* The `submit()` dispatchable on the inbound channel now has a much more complicated weight calculation, since messages can now have 2 different execution paths.
* At the protocol layer, messages from ethereum now include an optional parachain id and minimum dispatch weight.

Todo:
- [ ] Upgrade from XcmV0 to XcmV1 / 0.9.12
- [ ] Final consolidation/cleanups on the above changes.
- [ ] Need to refactor / fix some things with our benchmarking setup, in order to easily benchmark the `submit()` dispatchable
- [ ] Add benchmarks for `submit()`
- [ ] Add tests

Resolves: SNO-138

PR #547, opened 2021-12-06
Update benchmarks & weights

vgeddes:
I've fixed a few issues around benchmarking & weights. Overall, maintaining & updating weights is simpler. Also unblocks #540 which depends heavily on the changes here.

* Removed benchmarks & calculated weights for channel submit methods in preparation for #540 
* Our runtimes now use the stock substrate weights for FRAME pallets, similar to what other parachains are now doing.
* Fixed broken benchmarks for the EthApp, Erc20App, and DOT pallets
* A customized chainspec is no longer required to run benchmarks

Resolves SNO-139


parachain/pallets/assets/src/lib.rs
```diff
@@ -16,7 +16,8 @@
 //!
 //! - [`MultiAsset`](../snowbridge_core/assets/trait.MultiAsset.html): Functions for dealing with a
 //! multiple fungible assets.
-//! - [`SingleAsset`](../snowbridge_core/assets/trait.SingleAsset.html): Functions for dealing with a
+//! - [`SingleAsset`](../snowbridge_core/assets/trait.SingleAsset.html): Functions for dealing with
+//!   a
```
alistair-singh, line 20: Nit: Comment formatting.

alistair-singh (approved):
Looks good. Approving.

I see we have removed benchmarks for BasicInboundChannel and are using default weights. Why wouldn't we need specialized weights for BasicInboundChannel?

vgeddes:
> I see we have removed benchmarks for BasicInboundChannel and are using default weights. Why wouldn't we need specialized weights for BasicInboundChannel?

The benchmarks and weights for the inbound channels will be completely redeveloped in #540. So part of what this PR does is emoves the old to make way for the new.

vgeddes:
> > I see we have removed benchmarks for BasicInboundChannel and are using default weights. Why wouldn't we need specialized weights for BasicInboundChannel?
> 
> The benchmarks and weights for the inbound channels will be completely redeveloped in #540. So part of what this PR does is removes the old to make way for the new.

@alistair-singh Just to add some clarity, the previous benchmarks/weights for the inbound channel `submit()` calls assumed that the dispatched message would either call `DotApp.lock`, `EthApp.mint`, or `Erc20App.mint`. So the benchmarks for `submit()` simulated the call to `DotApp.lock`, as it had the highest weight/cost.

With the new message dispatch mechanism in #540, the above no longer works, since there is greater flexibility in how messages can either be dispatched. Either locally (to our apps) or remotely on apps living on another parachain. So instead, when users/apps submit a message to an outbound channel (for delivery to the other side), they also need to submit the maximum weight of the message. Then on the other side, in the inbound channel, the decoded message will be validated to ensure that its actual pre-dispatch weight is less than the declared maximum weight. If that condition does not hold, the message will not be dispatched anywhere.

The channels require the maximum weight to be declared upfront, as a message's actual pre-dispatch weight is only available much later in the validation & processing pipeline.

PR #548, opened 2021-12-09
Optimize nonce validation in inbound channel

sumitsnk:
This PR resolves the issue SNO-147

### DIFF
--- a/ethereum/contracts/BasicInboundChannel.sol
+++ b/ethereum/contracts/BasicInboundChannel.sol
@@ -56,14 +56,14 @@ contract BasicInboundChannel {
     }
 
     function processMessages(Message[] calldata _messages) internal {
-        // Using tempNonce variable to store nonce value
-        uint64 tempNonce = nonce;
+        // Caching nonce for gas optimization
+        uint64 cachedNonce = nonce;
 
         for (uint256 i = 0; i < _messages.length; i++) {
             // Check message nonce is correct and increment nonce for replay protection
-            require(_messages[i].nonce ==  tempNonce + 1, "invalid nonce");
+            require(_messages[i].nonce ==  cachedNonce + 1, "invalid nonce");
 
-            tempNonce = tempNonce + 1;
+            cachedNonce = cachedNonce + 1;
 
             // Deliver the message to the target
             (bool success, ) = _messages[i].target.call{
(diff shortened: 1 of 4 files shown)


ethereum/contracts/BasicInboundChannel.sol
```diff
@@ -56,11 +56,14 @@ contract BasicInboundChannel {
     }
 
     function processMessages(Message[] calldata _messages) internal {
+        // Using tempNonce variable to store nonce value
```
vgeddes, line 59: This comment doesn't really explain anything. Instead, maybe explain that we are caching the nonce in a memory variable for gas optimization, or something like that.

ethereum/contracts/BasicInboundChannel.sol
```diff
@@ -56,11 +56,14 @@ contract BasicInboundChannel {
...
+        uint64 tempNonce = nonce;
```
vgeddes, line 60: I would rather rename `tempNonce` to `cachedNonce`.

ethereum/contracts/IncentivizedInboundChannel.sol
```diff
@@ -82,12 +82,13 @@ contract IncentivizedInboundChannel is AccessControl {
         Message[] calldata _messages
     ) internal {
         uint256 _rewardAmount = 0;
+        // Using tempNonce variable to store nonce value
+        uint64 tempNonce = nonce;
```
vgeddes, line 86: See my comments about this same line in the basic channel.

vgeddes (commented):
Looks good so far Sumit.

I think we need to add unit tests to guard against any future regressions in the cached nonce behavior. Could you update the following tests (or add new ones) to verify that the cached nonce was saved back to persistent storage.

https://github.com/Snowfork/snowbridge/blob/05fb58b7a2ed9d41884f2c4546cacdf87fb548c2/ethereum/test/test_basic_inbound_channel.js#L40

https://github.com/Snowfork/snowbridge/blob/05fb58b7a2ed9d41884f2c4546cacdf87fb548c2/ethereum/test/test_incentivized_inbound_channel.js#L47

vgeddes:
Resolves SNO-147

ethereum/contracts/BasicInboundChannel.sol
musnit: Yeh, in general comments should explain "Why" something is done, not "what" is being done. Variable names should explain "what" is being done.

musnit (approved):

vgeddes (approved):
Looks great! Go ahead and merge in.

PR #550, opened 2021-12-14
Implement BasicInboundChannelV2 contract 

sumitsnk:


ethereum/contracts/BasicInboundChannelV2.sol
```diff
@@ -0,0 +1,90 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity ^0.8.5;
+pragma experimental ABIEncoderV2;
+
+import "./ParachainLightClient.sol";
+import "./BeefyLightClient.sol";
+import "./SimplifiedMMRVerification.sol";
+
+contract BasicInboundChannel {
```
vgeddes, line 9: Should be named BasicInboundChannelV2 so as not to clash with the previous version.

vgeddes (commented):
looks good so far.

ethereum/contracts/BasicInboundChannelV2.sol
musnit: or rather MultiUserChannel

ethereum/contracts/BasicInboundChannelV2.sol
```diff
@@ -0,0 +1,94 @@
...
+        bytes payload;
+    }
+
+    event MessageDispatched(uint64 nonce, bool result);
+
+    constructor(BeefyLightClient _beefyLightClient) {
+        beefyLightClient = _beefyLightClient;
+    }
+
+    function generateCommitmentHash(Leaf calldata _leaf,bytes32[] calldata _leafProof)
```
musnit, line 34: we already have and use a MerkleTree library. Would we be able to use that instead of re-implementing this?

sumitsnk: updated

ethereum/contracts/BasicInboundChannelV2.sol
```diff
@@ -0,0 +1,91 @@
...
+        );
+
+        processMessages(_leaf);
+    }
+
+    function processMessages(Leaf calldata _leaf) public { 
+        // User nonce for replay protection
+        uint64 usercachedNonce =  userNonce[_leaf.account];
+        
+        require(usercachedNonce + 1  == _leaf.nonce , "invalid nonce");
```
musnit, line 75: i think we only need 1 nonce per leaf now, not 1 nonce per message? so we can just increment the nonce once directly and then check it.

@vgeddes does this make sense?

vgeddes: Indeed, that is correct. 👍 

In fact that also applies with the existing flat commitment, but we never optimized that case away.

vgeddes: @sumitsnk yeah to clarify, after this `require` call, you no longer need to check or reference the nonce anymore. And there is no need to cache the nonce.

vgeddes: actually, though you do need to increment the user nonce before dispatching the message.

sumitsnk: When only one nonce is used per leaf, there is no need to cache the nonce and also update the user nonce before dispatching the message.

ethereum/contracts/BasicInboundChannelV2.sol
```diff
@@ -0,0 +1,87 @@
...
+       Leaf calldata _leaf,
+       bytes32[] calldata leafProof,
+        bool[] calldata nodeSide,
+        ParachainLightClient.ParachainVerifyInput
+            calldata _parachainVerifyInput,
+        ParachainLightClient.BeefyMMRLeafPartial calldata _beefyMMRLeafPartial,
+        SimplifiedMMRProof calldata proof
+    ) public {
+
+        bytes32 commitment = generateCommitmentHash(_leaf,leafProof,nodeSide);
```
vgeddes, line 52: Nit: formatting, need spaces after commas.

ethereum/contracts/BasicInboundChannelV2.sol
```diff
@@ -0,0 +1,87 @@
...
+    {
+        bytes32 leafNodeHash = keccak256(abi.encode(_leaf));
+
+        return MerkleProof.computeRootFromProofAndSide(leafNodeHash,_leafProof,nodeSide);
+    }
+
+    function submit(
+       Leaf calldata _leaf,
+       bytes32[] calldata leafProof,
+        bool[] calldata nodeSide,
```
vgeddes, line 45: nit: indent is wrong.

ethereum/contracts/BasicInboundChannelV2.sol
```diff
@@ -0,0 +1,87 @@
...
+            // Deliver the message to the target
+            (bool success, ) = _leaf.messages[i].target.call{
+                value: 0,
+                gas: MAX_GAS_PER_MESSAGE
+            }(_leaf.messages[i].payload);
+ 
+            emit MessageDispatched(userNonce[_leaf.account], success);
+        }
+    }
+}
```
vgeddes, line 87: nit: please add a newline at the end of your files.

If you use VS code, the setting `files.insertFinalNewline` will do this automatically for you.

ethereum/test/test_basic_inbound_channelV2.js
```diff
@@ -0,0 +1,106 @@
...
+    const beefyFixture = await createBeefyValidatorFixture(
+      totalNumberOfValidatorSigs
+    )
+    this.beefyLightClient = await deployBeefyLightClient(beefyFixture.root,
+      totalNumberOfValidatorSigs);
+
+    await runBeefyLightClientFlow(fixture, this.beefyLightClient, beefyFixture, totalNumberOfValidatorSigs, totalNumberOfValidatorSigs)
+  });
+
+  describe("submit", function () {
```
vgeddes, line 37: Lets mark these tests to be skipped, since we don't have valid fixture data to make them pass. We can revisit the tests again later once the fixture data is available.

```suggestion
  describe.skip("submit", function () {
```

vgeddes:
Obsoleted by https://github.com/Snowfork/snowbridge/pull/637

PR #554, opened 2021-12-20
ETH app upgradeability

wildmolasses:
Partially resolves https://linear.app/snowfork/issue/SNO-27/ethereum-contract-upgradibility

I've edited ETHApp.sol to add upgradeability as was outlined [here](https://www.notion.so/snowfork/Snowbridge-contract-upgradeability-pt-2-4f1bc631db574becb8920b9ebf747ba6#3473c4c471ae449ea1fa15ce2e450297) and discussed on slack.

The `CHANNEL_UPGRADE_ROLE` (set in constructor to `msg.sender`, but can change to constructor arg if you'd prefer) is able to call `upgrade` passing 2 channel tuples (basic, incentivized).

Upgradeability can be permanently disabled if `CHANNEL_UPGRADE_ROLE` calls `renounceRole` (see tests).

An "important note": By setting `CHANNEL_UPGRADE_ROLE` as the roleAdmin of `INBOUND_CHANNEL_ROLE`, the `CHANNEL_UPGRADE_ROLE` can arbitrarily call `grantRole` and `revokeRole` on the `INBOUND_CHANNEL_ROLE`. Since the `CHANNEL_UPGRADE_ROLE` is assumed to be trustworthy, this seemed like an OK tradeoff to make, but if not I'm happy to tighten the access implementation more.

Actions needed:
- Please confirm that the assumption of "upgrading" both basic and incentivized channels at once makes sense
- Please confirm that it's ok to set `CHANNEL_ADMIN_ROLE` to `msg.sender` in constructor
- Please confirm that the access management pattern is appropriate (see "important note" above)
- Please give OK to implement same pattern in rest of apps and then channels
- Let's also chat more about `encodeCall`


ethereum/contracts/ETHApp.sol
```diff
@@ -47,6 +50,8 @@ contract ETHApp is RewardSource, AccessControl {
         c2.inbound = _incentivized.inbound;
         c2.outbound = _incentivized.outbound;
 
+        _setupRole(CHANNEL_UPGRADE_ROLE, msg.sender);
+        _setRoleAdmin(INBOUND_CHANNEL_ROLE, CHANNEL_UPGRADE_ROLE);
```
musnit, line 54: still need to look into openzep code properly, but it seems like _setRoleAdmin basically makes it that only those with CHANNEL_UPGRADE_ROLE can change the INBOUND_CHANNEL_ROLE in future?

without this, does it default to letting anyone change it, or letting noone change it?

wildmolasses: defaults to "let no one change it"

vgeddes:
Looks good!

> Please confirm that the assumption of "upgrading" both basic and incentivized channels at once makes sense.

> Please confirm that it's ok to set CHANNEL_ADMIN_ROLE to msg.sender in constructor
Yeah this fine for now.

> Please confirm that the access management pattern is appropriate (see "important note" above)
Looks good! I have been following roughly the same pattern in some other contracts for allowing permission to update configuration: https://github.com/Snowfork/snowbridge/blob/main/ethereum/contracts/IncentivizedInboundChannel.sol#L34


> Please give OK to implement same pattern in rest of apps and then channels

👍 


> Let's also chat more about encodeCall

Looking over this issue some more...

vgeddes:
OK, lets assume that `encodeCall` won't be upgradable, and that we're restricted to handling any changes to the protocol on the polkadot/parachain side.

wildmolasses:
Hi @vgeddes @musnit , this PR is now ready for review.

- As the deployer you can call upgrade on all 4 apps to change the channels. Tests are included.
- I changed the README and package.json to prefer yarn over npx.
- I changed the README to suggest running a local node via `yarn hardhat node`. I also changed the local node chainId to the hardhat default, 31337. Please let me know if this messes something up e2e and I will revert immediately.
- Two tasks are added, one called `upgrade` and one called `renounce`. You can see the README for a usage guide, let me know if more detail would be helpful.

ethereum/contracts/DOTApp.sol
```diff
@@ -97,4 +108,23 @@ contract DOTApp is FeeSource, AccessControl {
                 _amount.encode256()
             );
     }
+
+    function upgrade(
```
musnit, line 112: @wildmolasses 

should we consider moving all this duplicate code into something like a shared ChannelApp.sol contract that these can inherit from? maybe in a follow up PR though if so

musnit:
@wildmolasses is there a way for a current upgrader with CHANNEL_UPGRADE_ROLE to give access to a new upgrader (eg: if we want to upgrade governance from centralized account -> voting smart contract>)

ethereum/contracts/DOTApp.sol
wildmolasses: Sure, that sounds good.

wildmolasses:
> @wildmolasses For an end-user of an application, is there an easy way for them to confirm who controls upgradability for the app and whether or not a full revoke of upgradability has been done? (eg: a simple getter function they can call on etherscan like owner() or upgrader() or isUpgradable())

Hmm, currently this information can be discovered via events. AccessControl methods each emit events, so an end-user can check whether there was a `RoleRevoked` event emitted with role CHANNEL_UPGRADE_ROLE. Is this enough, or should I look into making it more explicit?

musnit:
> Hmm, currently this information can be discovered via events. AccessControl methods each emit events, so an end-user can check whether there was a `RoleRevoked` event emitted with role CHANNEL_UPGRADE_ROLE. Is this enough, or should I look into making it more explicit?
> 
> >

The use-case that comes to mind (one i often find myself in) is I'm looking at a contract on Etherscan and determining whether to trust and use it for my own needs, and I want to check for any risks like a rugpull. Events are not really great for this, as you'd need to search across and compare a bunch of events to see the current state, but it's not too bad. I see this is basically a limitation of the OpenZeppelin implementation and I'm not sure we could fix this without adding extra gas costs to track this explicitly. I suppose we could add a counter that increments on _setupRole and revokeRole without modifying the underlying libraries and wouldn't impact the main burn/mint functions in any way, but seems like overkill.

I think it's fine as is then and we can just put an event query into our UI on our status page.



> I'm happy you asked. We could set `CHANNEL_UPGRADE_ROLE` to be the roleAdmin of `CHANNEL_UPGRADE_ROLE`. Then this would be possible by using `CHANNEL_UPGRADE_ROLE` to call `grantRole`. With your go ahead I can implement and add tests.

sounds good! the existing renounce script would still then effectively renounce complete control of both channel upgrades and governance upgrades?

wildmolasses:
> sounds good! the existing renounce script would still then effectively renounce complete control of both channel upgrades and governance upgrades?

Yes!

> Events are not really great for this, as you'd need to search across and compare a bunch of events to see the current state, but it's not too bad. I see this is basically a limitation of the OpenZeppelin implementation and I'm not sure we could fix this without adding extra gas costs to track this explicitly.

We could consider the option where we remove OZ AccessControl in favor of a simpler one. The access roles seem simple enough. Instead of role mappings which are hard to query, we could make the key roles normal contract variables. Besides simplifying the contracts and contract state readability, i suspect it will incur gas savings on storage and retrieval too (i should run a benchmark). But if we are happy with the existing access control pattern and associated gas costs, then that's ok too.

musnit:
looks good to merge from me.

@vgeddes please do final review and security audit.

ethereum/contracts/ERC20App.sol
```diff
@@ -35,6 +35,12 @@ contract ERC20App is AccessControl {
         uint256 amount
     );
 
+    event Upgraded(
+        address ugprader,
```
vgeddes, line 39: Nit: spelling.

ethereum/contracts/ETHApp.sol
```diff
@@ -20,6 +20,12 @@ contract ETHApp is RewardSource, AccessControl {
 
     event Unlocked(bytes32 sender, address recipient, uint256 amount);
 
+    event Upgraded(
+        address ugprader,
```
vgeddes, line 24: nit: spelling

ethereum/contracts/ERC721App.sol
```diff
@@ -33,6 +33,12 @@ contract ERC721App is AccessControl {
         address recipient
     );
 
+    event Upgraded(
+        address ugprader,
```
vgeddes, line 37: nit: spelling

vgeddes (approved):
Added some minor nits but otherwise looks very good.

Perhaps in a follow-up PR we can factor out some of the upgradability logic. Not important for now though.

ethereum/contracts/ERC20App.sol
wildmolasses: Thanks for catching these!

wildmolasses:
Feedback addressed, rebased onto latest main

wildmolasses:
Feedback addressed, rebased onto latest main

PR #555, opened 2021-12-21
Sparse merkle tree multi-proof

hujw77:
For gas saving, we could use the sparse merkle tree multi-proof to validate the commitment signed by beefy authority set.
More detail can see https://github.com/darwinia-network/darwinia-bridge-sol/pull/104

### DIFF
--- a/ethereum/contracts/BeefyLightClient.sol
+++ b/ethereum/contracts/BeefyLightClient.sol
@@ -8,6 +8,7 @@ import "./utils/Bitfield.sol";
 import "./ValidatorRegistry.sol";
 import "./SimplifiedMMRVerification.sol";
 import "./ScaleCodec.sol";
+import "./SparseMerkleMultiProof.sol";
 
 /**
  * @title A entry contract for the Ethereum light client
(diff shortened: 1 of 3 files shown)


vgeddes:
Hey @hujw77 thanks for the contribution. I'll be able to review this properly next week, unless my colleagues get to it first.

musnit:
@hujw77 Taking another quick look - Using a multiproof instead of an array of individual proofs looks like a great idea and optimization!

We'll need to build the proof creation logic into our relayer and do a deeper team audit and review before merging this. At the moment we're focused on some other maintenance and new feature work, so it may take a few weeks before we schedule this work in, but will leave this open as it's definitely something we'll want to merge in before this code is used in production.

Thanks again!

hujw77:
@musnit For the proof creation logic, there is research about it at https://github.com/hujw77/sparse-merkle-tree
It may be some help for audits and seminars.

vgeddes:
Very obsolete. Now that https://github.com/paritytech/substrate/pull/12857 has been merged, we can use the OpenZeppelin multi-proof verifier for this:

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/5dbde1a5c954a79338959d4eb9c7e716a217d6a1/contracts/utils/cryptography/MerkleProof.sol#L77

PR #559, opened 2021-12-29
Solidity ERC20App V2

sumitsnk:
This PR resolves the ticket : https://linear.app/snowfork/issue/SNO-158/solidity-erc20app-v2
It contains the assumption that only address with INBOUND_CHANNEL_ROLE  is allowed to add wrapped token address to the list.

### DIFF
--- a/ethereum/contracts/ERC20App.sol
+++ b/ethereum/contracts/ERC20App.sol
@@ -79,12 +79,6 @@ contract ERC20App is AccessControl {
             "Invalid channel ID"
         );
 
-        uint8 _decimals;
-        string memory _name;
-        string memory _symbol;
-
-        (_name, _symbol, _decimals) = tokenDetails(_token);
-
         balances[_token] = balances[_token] + _amount;
 
         emit Locked(_token, msg.sender, _recipient, _amount, _paraId);
@@ -93,14 +87,8 @@ contract ERC20App is AccessControl {
             channels[_channelId].outbound
         );
 
-        bytes memory createCall;
         if (!wrappedTokenList[_token]) {
-            createCall = encodeCreateTokenCall(
-                _token,
-                _name,
-                _symbol,
-                _decimals
-            );
+            bytes memory createCall = encodeToken(_token);
             wrappedTokenList[_token] = true;
             channel.submit(msg.sender, createCall);
         }
@@ -225,5 +213,16 @@ contract ERC20App is AccessControl {
             _decimals = 0;
         }
         return (_name, _symbol, _decimals);
+    }
+
+    function encodeToken(address _token) private view returns (bytes memory) {
+        uint8 _decimals;
+        string memory _name;
+        string memory _symbol;
+
+        (_name, _symbol, _decimals) = tokenDetails(_token);
+        bytes memory createCall;
+        createCall = encodeCreateTokenCall(_token, _name, _symbol, _decimals);
+        return (createCall);
     }
 }
--- a/ethereum/contracts/ERC20App.sol
+++ b/ethereum/contracts/ERC20App.sol
@@ -16,6 +16,7 @@ enum ChannelId {
 contract ERC20App is AccessControl {
     using ScaleCodec for uint256;
     using ScaleCodec for uint32;
+    using ScaleCodec for uint8;
     using SafeERC20 for IERC20;
 
     mapping(address => uint256) public balances;
@@ -23,6 +24,10 @@ contract ERC20App is AccessControl {
     mapping(ChannelId => Channel) public channels;
 
     bytes2 constant MINT_CALL = 0x4201;
+
+    bytes2 constant CREATE_CALL = 0x4202;
+
+    mapping(address => bool) public wrappedTokenList;
 
     event Locked(
         address token,
@@ -38,6 +43,8 @@ contract ERC20App is AccessControl {
         address recipient,
         uint256 amount
     );
+
+    event TokenWrapped(address token, address operator);
 
     struct Channel {
         address inbound;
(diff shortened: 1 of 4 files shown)


musnit:
@vgeddes @sumitsnk could it not be better for this to be handled on the parachain side, rather than adding in this complexity with extra messages, checks and gas on the ethereum side?

alternatively, having a separate transaction and flow for create rather than batching them together may be a better ux too?

musnit:
> @vgeddes @sumitsnk could it not be better for this to be handled on the parachain side, rather than adding in this complexity with extra messages, checks and gas on the ethereum side?
> 
> alternatively, having a separate transaction and flow for create rather than batching them together may be a better ux too?

@sumitsnk it's also not clear if/why the create flow needs to be permissioned and only accessable to the INBOUND_CHANNEL_ROLE, and not clear if the flow even works with the INBOUND_CHANNEL_ROLE being required. Can't it be open to anyone, with addWrapppedToken being called automatically as part a create message send?

this setup also places a message-ordering requirement on the application, which could lead to edge case scenarios like 2 users both sending a new token through the basic channel in their own lane and both trying to create a token with redundant messages that would need to be handled.

vgeddes:
> @vgeddes @sumitsnk could it not be better for this to be handled on the parachain side, rather than adding in this complexity with extra messages, checks and gas on the ethereum side?
> 

This was my initial thinking (having a single message handler on parachain side), but it would have lead to quite a gross API, and an overly complicated weight accounting on the parachain side). It also introduced tight coupling between the logically separate processes of creating an asset and minting account balances. This made me feel uneasy.

As for the added gas costs, when users want to transfer balances for an existing asset, its only an extra ~800 gas to check the registry. Thats not too bad I think.

> alternatively, having a separate transaction and flow for create rather than batching them together may be a better ux too?

I like this approach. It is cleaner. ​Though it does create some extra work (updating E2E tests, webapp, etc). And we'll need to write scripts to bootstrap the most popular tokens for our Staging/Kusama testnets.

Since the change is mostly UX-driven I'm thinking we should rather tackle this extra work in a follow-up PR. I want to get the core of Assets V2 into master soon. Let me know your thoughts.

vgeddes (commented):
Looks very good so far! Added a few comments.

ethereum/contracts/ERC20App.sol
```diff
@@ -60,8 +67,21 @@ contract ERC20App is AccessControl {
...
+    {
+        require(!wrappedTokenList[_token], "Wrapped token already exists");
+        wrappedTokenList[_token] = true;
+
+        emit TokenWrapped(_token, msg.sender);
+    }
+
     function lock(
         address _token,
+        string calldata _name,
```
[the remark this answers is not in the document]
sumitsnk: The name, symbol, and decimal methods are optional in the EIP-20 standard and are not available in the IERC20.sol.
Using the _token.name() might cause problems.
https://github.com/ethereum/EIPs/blob/master/EIPS/eip-20.md

sumitsnk:
> > @vgeddes @sumitsnk could it not be better for this to be handled on the parachain side, rather than adding in this complexity with extra messages, checks and gas on the ethereum side?
> > alternatively, having a separate transaction and flow for create rather than batching them together may be a better ux too?
> 
> @sumitsnk it's also not clear if/why the create flow needs to be permissioned and only accessable to the INBOUND_CHANNEL_ROLE, and not clear if the flow even works with the INBOUND_CHANNEL_ROLE being required. Can't it be open to anyone, with addWrapppedToken being called automatically as part a create message send?
> 
> this setup also places a message-ordering requirement on the application, which could lead to edge case scenarios like 2 users both sending a new token through the basic channel in their own lane and both trying to create a token with redundant messages that would need to be handled.

@musnit  The initial thought behind the permission for addWrapppedToken is that any malicious user attempting to update incorrect information can be prevented. 
According to the recommendation, I have updated the code so that the wrappedTokenList registry is only updated within the lock function.

ethereum/contracts/ERC20App.sol
```diff
@@ -73,20 +79,39 @@ contract ERC20App is AccessControl {
             "Invalid channel ID"
         );
 
+        uint8 _decimals;
+        string memory _name;
+        string memory _symbol;
+
+        (_name, _symbol, _decimals) = tokenDetails(_token);
```
musnit, line 86: this looks like it could be within the if statement below, so that it only runs the first time, right?

ethereum/contracts/ERC20App.sol
```diff
@@ -73,20 +79,39 @@ contract ERC20App is AccessControl {
...
+
         balances[_token] = balances[_token] + _amount;
 
         emit Locked(_token, msg.sender, _recipient, _amount, _paraId);
 
+        OutboundChannel channel = OutboundChannel(
+            channels[_channelId].outbound
+        );
+
+        bytes memory createCall;
```
musnit, line 96: this could also be within the if statement right?

ethereum/contracts/ERC20App.sol
```diff
@@ -73,20 +79,39 @@ contract ERC20App is AccessControl {
...
+        if (!wrappedTokenList[_token]) {
+            createCall = encodeCreateTokenCall(
```
musnit, line 98: should we move all this logic within the if statement into its own separate function?

musnit:
@sumitsnk looks good! a few comments on gas optimization and code readability, but otherwise functionality looks great.

can we also get a test that tests that the case with no name/symbol/decimals works as expected?

vgeddes:
@sumitsnk, can we change the base branch for this PR to `assets-v2` please.

ethereum/contracts/ERC20App.sol
sumitsnk: Yes,exactly it should be within the if condition only.

vgeddes (approved):
Looks great! Am gonna merge this into `assets-v2` now.

PR #562, opened 2022-01-11
Assets V2

vgeddes:
This changes improves the story around asset handling for the Ether and ERC20 applications. The goal is finalize our assets API and ease XCM integrations with other teams. 

- Amounts bridged over from Ethereum are now capped at 128 bits, and represented using `u128` instead of `U256`.
- Replaced our custom assets pallet and XCM support code with drop-in equivalents from upstream Substrate.
- Metadata for ERC20 tokens is now stored in the Assets pallet 
- ERC20 tokens bridged from Ethereum now have a derived 128-bit asset id on the parachain side.

The changes break our existing API, and all clients will need to be updated. Particularly, since wrapped ERC20 tokens are no longer identified by their ethereum contract address, clients will need to perform a lookup in the erc20App pallet to find the derived 128-bit asset ids. 

Resolves: SNO-17, SNO-183


musnit:
> The changes break our existing API, and all clients will need to be updated. Particularly, since wrapped ERC20 tokens are no longer identified by their ethereum contract address, clients will need to perform a lookup in the erc20App pallet to find the derived 128-bit asset ids.

Could/should we have a single function call/api on the erc20App pallet (or assets pallet) to support clients querying things like balances and such via contract address easily?

parachain/pallets/erc20-app/src/mock.rs
```diff
@@ -61,22 +63,63 @@ impl system::Config for Test {
...
 	type OnKilledAccount = ();
 	type SystemWeightInfo = ();
 	type SS58Prefix = ();
 	type OnSetCode = ();
 }
 
+impl pallet_randomness_collective_flip::Config for Test {}
+
+parameter_types! {
+		pub const ExistentialDeposit: u64 = 1;
```
alistair-singh, line 77: Nit: Whitespace.

parachain/pallets/erc20-app/src/lib.rs
```diff
@@ -167,5 +188,42 @@ pub mod pallet {
...
+			T::Assets::create(asset_id, T::PalletId::get().into_account(), true, 1)?;
+
+			<AssetId<T>>::insert(token, asset_id);
+
+			Ok(())
+		}
+	}
+
+	impl<T: Config> Pallet<T> {
+		fn make_asset_id(address: H160) -> u128 {
```
vgeddes, line 210: @musnit @alistair-singh Just want to raise a specific design implication here for discussion:

The new assets pallet uses a 128-bit number for asset identifiers. So for each new wrapped asset, we have to generate a unique 128-bit ID. The options for this generation are to either keep an incrementing counter in storage, or to generate random numbers with a very low probability of collision. I'm opting for the latter as its stateless.

So the goal of this function to generate a random 128-bit number using various sources of entropy (The contract address, last K block hashes, etc).

The probability of collision is exactly the same as the scenario where random V4 UUIDs (also 128 bits) are used as database keys. According to this [wikipage](https://en.wikipedia.org/wiki/Universally_unique_identifier#Collisions):

> For example, the number of random version-4 UUIDs which need to be generated in order to have a 50% probability of at least one collision is 2.71 quintillion. This number is equivalent to generating 1 billion UUIDs per second for about 85 years. A file containing this many UUIDs, at 16 bytes per UUID, would be about 45 exabytes.

And

> The probability to find a duplicate within 103 trillion version-4 UUIDs is one in a billion.

Given that we'll only expect to see maybe a few dozen new assets per week, our probability of collision would be several magnitudes less than the above. Practically negligible. Like we'd need to operate the bridge for hundreds or thousands of years before seeing a collision.

Currently the worst case possibility is that the transferred funds would be lost on collision.

musnit:
can't u just add a check for collision too?

parachain/pallets/erc20-app/src/lib.rs
alistair-singh: We could store the original H160 as asset metadata since we are using assets pallet. We could then look up and use it for collision detection so that we can fail gracefully. Clients can retry the transaction and it would successful if we have different entropy.

vgeddes: Thanks Alistair, that could work. Though I've now moved to simpler mechanism with less potential for negative scenarios. I've introduced a new pallet called `AssetRegistry`, which contains a counter in storage. Every time we need a new asset id, we just read that counter and then increment it.

alistair-singh (approved):

PR #563, opened 2022-01-14
Added e2e tests for One-click Parachain

alistair-singh:
Adds one click e2e tests for both ETHApp and ERC20App in the Ethereum to Substrate direction.

Fixes: SNO-167

### DIFF
--- a/ethereum/contracts/ERC20App.sol
+++ b/ethereum/contracts/ERC20App.sol
@@ -34,7 +34,8 @@ contract ERC20App is AccessControl {
         address sender,
         bytes32 recipient,
         uint128 amount,
-        uint32 paraId
+        uint32 paraId,
+        uint128 fee
     );
 
     event Unlocked(
@@ -83,16 +84,18 @@ contract ERC20App is AccessControl {
         bytes32 _recipient,
         uint128 _amount,
         ChannelId _channelId,
-        uint32 _paraId
+        uint32 _paraId,
+        uint128 _fee
     ) public {
         require(
             _channelId == ChannelId.Basic ||
                 _channelId == ChannelId.Incentivized,
             "Invalid channel ID"
         );
+
         balances[_token] = balances[_token] + _amount;
 
-        emit Locked(_token, msg.sender, _recipient, _amount, _paraId);
+        emit Locked(_token, msg.sender, _recipient, _amount, _paraId, _fee);
 
         OutboundChannel channel = OutboundChannel(
             channels[_channelId].outbound
@@ -108,13 +111,7 @@ contract ERC20App is AccessControl {
         if (_paraId == 0) {
             call = encodeCall(_token, msg.sender, _recipient, _amount);
         } else {
-            call = encodeCallWithParaId(
-                _token,
-                msg.sender,
-                _recipient,
-                _amount,
-                _paraId
-            );
+            call = encodeCallWithParaId(_token, msg.sender, _recipient, _amount, _paraId, _fee);
         }
 
         channel.submit(msg.sender, call);
@@ -166,7 +163,8 @@ contract ERC20App is AccessControl {
         address _sender,
         bytes32 _recipient,
         uint128 _amount,
-        uint32 _paraId
+        uint32 _paraId,
+        uint128 _fee
     ) private pure returns (bytes memory) {
         return bytes.concat(
                 MINT_CALL,
@@ -176,7 +174,8 @@ contract ERC20App is AccessControl {
                 _recipient,
                 _amount.encode128(),
                 bytes1(0x01),
-                _paraId.encode32()
+                _paraId.encode32(),
+                _fee.encode128()
             );
     }
 
--- a/ethereum/contracts/ETHApp.sol
+++ b/ethereum/contracts/ETHApp.sol
@@ -24,7 +24,8 @@ contract ETHApp is RewardSource, AccessControl {
         address sender,
         bytes32 recipient,
         uint128 amount,
-        uint32 paraId
+        uint32 paraId,
+        uint128 fee
     );
 
     event Unlocked(bytes32 sender, address recipient, uint128 amount);
@@ -75,7 +76,8 @@ contract ETHApp is RewardSource, AccessControl {
     function lock(
         bytes32 _recipient,
         ChannelId _channelId,
-        uint32 _paraId
+        uint32 _paraId,
+        uint128 _fee
     ) public payable {
         require(msg.value > 0, "Value of transaction must be positive");
         require(
@@ -84,16 +86,16 @@ contract ETHApp is RewardSource, AccessControl {
             "Invalid channel ID"
         );
 
-        // evert in case of overflow.
+        // revert in case of overflow.
         uint128 value = (msg.value).toUint128();
 
-        emit Locked(msg.sender, _recipient, value, _paraId);
+        emit Locked(msg.sender, _recipient, value, _paraId, _fee);
 
         bytes memory call;
         if (_paraId == 0) {
             call = encodeCall(msg.sender, _recipient, value);
         } else {
-            call = encodeCallWithParaId(msg.sender, _recipient, value, _paraId);
+            call = encodeCallWithParaId(msg.sender, _recipient, value, _paraId, _fee);
         }
 
         OutboundChannel channel = OutboundChannel(
@@ -135,7 +137,8 @@ contract ETHApp is RewardSource, AccessControl {
         address _sender,
         bytes32 _recipient,
         uint128 _amount,
-        uint32 _paraId
+        uint32 _paraId,
+        uint128 _fee
     ) private pure returns (bytes memory) {
         return bytes.concat(
                 MINT_CALL,
@@ -144,7 +147,8 @@ contract ETHApp is RewardSource, AccessControl {
                 _recipient,
                 _amount.encode128(),
                 bytes1(0x01),
-                _paraId.encode32()
+                _paraId.encode32(),
+                _fee.encode128()
             );
     }
 
--- a/ethereum/test/test_eth_app.js
+++ b/ethereum/test/test_eth_app.js
@@ -17,11 +17,12 @@ const { expect } = require("chai");
 const ETHApp = artifacts.require("ETHApp");
 const ScaleCodec = artifacts.require("ScaleCodec");
 
-const lockupFunds = (contract, sender, recipient, amount, channel) => {
+const lockupFunds = (contract, sender, recipient, amount, channel, paraId, fee) => {
   return contract.lock(
     addressBytes(recipient),
     channel,
-    0, // paraId
+    paraId,
+    fee,
     {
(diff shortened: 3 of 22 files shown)


parachain/primitives/xcm-support/src/lib.rs
```diff
@@ -91,39 +90,37 @@ where
...
+					fun: Fungible(4_000_000),
 				},
+				weight_limit: Limited(4_000_000),
 			},
+			DepositAsset { assets: Wild(All), max_assets: 2, beneficiary: origin_location.clone() },
 		]);
 
-		let remote_weight: u64 = T::Weigher::weight(&mut remote_message)
-			.map_err(|_| DispatchError::Other("Unweighable message."))?;
+		const WEIGHT_MULTIPLIER: u64 = 2;
```
alistair-singh, line 119: Need to figure out a better way of calculating remote weight. 

We cannot use the local weight of the message because they might be different per parachain. For instance the test parachain uses double the weight. This is probably because we have tuned weights on the snowbridge parachain vs the default weights in the test parachain.

parachain/utils/test-parachain/runtime/src/lib.rs
```diff
@@ -204,6 +208,11 @@ pub const UNIT: Balance = 1_000_000_000_000;
 pub const MILLIUNIT: Balance = 1_000_000_000;
 pub const MICROUNIT: Balance = 1_000_000;
 
+// The asset id for SnowEth
+pub const SNOW_ETH_ASSET_ID: u32 = 1000_0001;
```
vgeddes, line 212: Can just use `1` as the asset id. Simplifies things a bit for clients that need to query the asset.

```suggestion
pub const SNOW_ETH_ASSET_ID: u32 = 1;
```

parachain/utils/test-parachain/runtime/src/lib.rs
```diff
@@ -204,6 +208,11 @@ pub const UNIT: Balance = 1_000_000_000_000;
...
+// The asset for the Test Token used by e2e tests
+pub const TEST_TOKEN_ASSET_ID: u32 = 1000_0002;
```
vgeddes, line 214:
```suggestion
pub const TEST_TOKEN_ASSET_ID: u32 = 2;
```

parachain/primitives/xcm-support/src/lib.rs
vgeddes: Indeed, yeah this problematic. What I think we need to do is have the user/client pass the fee amount (in DOT) for `BuyExecution` on the Ethereum side, as a parameter to `EthApp.lock`. And then we use `WeightLimit::Unlimited` when constructing the `BuyExecution` instruction.

That way we bypass the problem, and make it the client's responsibility to figure out the necessary fees for remote execution.

Hopefully there will be some sort of improved negotiation protocol in XCM so that we can introspect the weight of remote chains.

vgeddes (commented):
Awesome! Great work. Very glad that everything has come together and that we now have messaging in place.

Only concern is the bit about calculating remote weight. Added my thinking in the relevant thread.

Happy for this to be merged once that particular issue is resolved.

parachain/primitives/xcm-support/src/lib.rs
musnit: yep agree, should be client's responsibility

ethereum/contracts/ERC20App.sol
```diff
@@ -78,23 +79,30 @@ contract ERC20App is AccessControl {
         bytes32 _recipient,
         uint256 _amount,
         ChannelId _channelId,
-        uint32 _paraId
+        uint32 _paraId,
+        uint256 _weight
```
vgeddes, line 83: Two things - 

1. Naming. We're actually asking the user to specifying the maximum DOT fee the recipient should pay to buy XCM execution weight. Fees and xcm weight are in different units so its important to make that distinction. So I would rename this to `_fee`/`fee` here and everywhere else.
2. Should use 16-byte`uint128`, not `uint256`, since we are specifying a quantity that only exists on the remote parachain, where DOT balances are recorded using `u128`. Using uint256 complicates things.

parachain/primitives/core/src/assets.rs
```diff
@@ -14,6 +14,13 @@ pub enum AssetId {
 	Token(H160),
 }
 
+#[derive(Encode, Decode, Copy, Clone, PartialEq, Eq, PartialOrd, RuntimeDebug, TypeInfo)]
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+pub struct RemoteParachain {
+	pub para_id: u32,
+	pub weight: U256,
```
vgeddes, line 21: Excellent, glad that you've introduced this struct!

ethereum/contracts/ETHApp.sol
```diff
@@ -82,7 +85,13 @@ contract ETHApp is RewardSource, AccessControl {
             "Invalid channel ID"
         );
 
-        emit Locked(msg.sender, _recipient, msg.value, _paraId);
+        require(
```
vgeddes, line 88: I would remove this check, as I don't think it helps much. Even if the check passes, the XCM transfer could still fail further down the line if the fee isn't enough to buy the necessary weight. And there is no way for the contract to know those fee/weight numbers.

The user or its proxy (a client such as our web frontend) should take responsibility for  querying the destination parachains to determine how much the fee is, and only then issue a `lock` call to this contract.

ethereum/test/test_eth_app.js
```diff
@@ -69,14 +70,57 @@ describe("ETHApp", function () {
...
+      // Confirm contract's balance has increased
+      const afterBalance = await web3.eth.getBalance(this.app.address);
+      afterBalance.should.be.bignumber.equal(amount);
+
+      // Confirm contract's locked balance state has increased by amount locked
+      const afterBalanceState = BigNumber(await web3.eth.getBalance(this.app.address));
+      afterBalanceState.should.be.bignumber.equal(beforeBalance.plus(amount));
+    });
+
+    it("should not lock funds to destination parachain without fee", async function () {
```
vgeddes, line 110: ditto, as per my comment above, the contract shouldn't really do any verification of the fee.

parachain/primitives/core/src/assets.rs
```diff
@@ -14,6 +14,13 @@ pub enum AssetId {
 	Token(H160),
 }
 
+#[derive(Encode, Decode, Copy, Clone, PartialEq, Eq, PartialOrd, RuntimeDebug, TypeInfo)]
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+pub struct RemoteParachain {
+	pub para_id: u32,
```
vgeddes, line 20: This struct is part of the public API, so it and its fields needs doc comments.

parachain/pallets/eth-app/src/lib.rs
```diff
@@ -152,13 +155,12 @@ pub mod pallet {
 			let recipient = T::Lookup::lookup(recipient)?;
 			T::Asset::deposit(&recipient, amount)?;
 
-			if let Some(id) = para_id {
+			if let Some(destination) = destination {
 				T::XcmReserveTransfer::reserve_transfer(
```
vgeddes, line 159: We're going to lose the user's funds here. The reserve transfer should be attempted last, after the `deposit_event`. If the function returns an error, it should be ignored (though we should log the error at the very least).

If we don't do the above, then a failed reserve transfer would cause `mint` to fail and the storage transaction for `mint` to be rolled back (as denoted by that `#[transactional]` annotation). This would rollback the `T::Asset::deposit()` and the user would lose their funds.

parachain/pallets/erc20-app/src/lib.rs
```diff
@@ -159,8 +162,8 @@ pub mod pallet {
 			let asset_id = AssetId::Token(token);
 			T::Assets::deposit(asset_id, &recipient, amount)?;
 
-			if let Some(id) = para_id {
-				T::XcmReserveTransfer::reserve_transfer(origin, asset_id, id, &recipient, amount)?;
+			if let Some(destination) = destination {
+				T::XcmReserveTransfer::reserve_transfer(asset_id, &recipient, amount, destination)?;
```
vgeddes, line 166: My comment in `eth-app/lib.rs` also applies here. We should be ignoring any errors returned by `reserve_transfer`.

ethereum/contracts/ETHApp.sol
musnit: yeh - agree. It would be great if you can address this in the doc you're writing up with all the potential edge cases @alistair-singh and then we can work confirm whether or not it makes sense to check the fee here at all based on a more comprehensive review of all the edge cases

parachain/pallets/eth-app/src/lib.rs
musnit: yeh - again making sure we have a document that covers every possible scenario/edge case and ensures we have both an acceptable on-chain outcome as well as ui/ux outcome that preserves user funds irrespective of the scenario is important.

parachain/pallets/erc20-app/src/lib.rs
musnit: ye though probably need some log/event that can be detected by the frontend or some third party for clear ux for the end user

ethereum/contracts/ETHApp.sol
vgeddes: Just to make sure we're all aligned here - Lets get PR merged in first (after addressing my last batch of comments), and then circle back with a doc & final PR to round out any rough edges regarding fees & XCM error-handling. And by XCM error-handling I mean the concepts introduced in this doc: https://medium.com/polkadot-network/xcm-part-iii-execution-and-error-management-ceb8155dd166, which mostly detail how to manage xcm script errors on the remote parachain.

I'm also going through the linear project at https://linear.app/snowfork/project/one-click-tofrom-any-parachain-49a0031f4a9b/SNO, cutting tickets to track the last remaining work.

musnit: @vgeddes yep! just make sure to include all these above tasks around documentation, edge-case exploration and fixes/implementation as linear tasks before merging

vgeddes: Done: https://linear.app/snowfork/project/one-click-to-any-parachain-49a0031f4a9b/SNO

ethereum/test/test_eth_app.js
alistair-singh: Removed tests.

ethereum/contracts/ETHApp.sol
alistair-singh: Removed checks.

parachain/pallets/eth-app/src/lib.rs
alistair-singh: Changed this to log error on failure and then continue.

vgeddes (approved):
Lets merge it in 🚀

PR #564, opened 2022-01-16
Update READMEs

vgeddes:
Some documentation was outdated, obsolete or incorrect.

Also removed obsolete XCM configs & docs in parachain dir, since we're able to test XCM using E2E tests.


test/README.md
```diff
@@ -18,11 +18,10 @@ The E2E tests run against local deployments of the parachain, relayer and ganach
   (cd ../ethereum && yarn install)
   ```
 
-* Development environment for the relay services. See relayer [requirements](../relayer/README.md#development).
-* `timeout` - native package on Ubuntu, on macOS try ```brew install coreutils```
+* Development environment for the relay services. See setup [instructions](../relayer/README.md#development).
 * `jq` - https://stedolan.github.io/jq/download/
 * geth - https://geth.ethereum.org/docs/install-and-build/installing-geth
-* sponge - Is available in the moreutils package. On Mac see https://formulae.brew.sh/formula/moreutils. On Linux:
```
musnit, line 25: i'd leave this in still - if ppl wanna run on mac then it's useful/helpful even though we may not put much resources into mac support

PR #576, opened 2022-02-04
Use bytes.concat instead of abi.encodePacked

vgeddes:
Minor improvement: https://github.com/ethereum/solidity/issues/10903. Though still using `abi.encodePacked` to convert `address` to `bytes`.

Passed E2E tests.

Fixes: SNO-195


alistair-singh (approved):

sumitsnk (approved):

musnit:
@vgeddes does this have gas cost implications? u sure it's cheaper with the extra abiEncode calls?

vgeddes:
> @vgeddes does this have gas cost implications? u sure it's cheaper with the extra abiEncode calls?

Unlikely to be a problem, I'd say. At a compiler level, abi.encodePacked and bytes.concat are more like primitives/macros that get expanded to specialised assembly depending on the arguments.

If there is an extra gas cost it's probably a rounding error, and an OK bargain for the increased clarity in our scale encoders.

musnit:
> > @vgeddes does this have gas cost implications? u sure it's cheaper with the extra abiEncode calls?
> 
> Unlikely to be a problem, I'd say. At a compiler level, abi.encodePacked and bytes.concat are more like primitives/macros that get expanded to specialised assembly depending on the arguments.
> 
> If there is an extra gas cost it's probably a rounding error, and an OK bargain for the increased clarity in our scale encoders.

this all sounds reasonable but always better to just read the docs and check opcode costs and calculate to check for sure, rather than making assumptions - would be good if you can check that nonetheless

PR #579, opened 2022-02-08
XCM Failure Cases Fixes

alistair-singh:
Fixes: SNO-191.

### DIFF
--- a/parachain/pallets/erc20-app/src/lib.rs
+++ b/parachain/pallets/erc20-app/src/lib.rs
@@ -30,7 +30,6 @@ mod tests;
 use frame_support::{
 	dispatch::{DispatchError, DispatchResult},
 	log,
-	storage::{with_transaction, TransactionOutcome},
 	traits::{
 		tokens::fungibles::{Create, Mutate},
 		EnsureOrigin,
--- a/parachain/pallets/erc20-app/src/lib.rs
+++ b/parachain/pallets/erc20-app/src/lib.rs
@@ -30,6 +30,7 @@ mod tests;
 use frame_support::{
 	dispatch::{DispatchError, DispatchResult},
 	log,
+	storage::{with_transaction, TransactionOutcome},
 	traits::{
 		tokens::fungibles::{Create, Mutate},
 		EnsureOrigin,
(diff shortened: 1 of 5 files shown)


parachain/primitives/xcm-support/src/lib.rs
```diff
@@ -85,10 +88,18 @@ where
...
+			let result =
+				T::XcmExecutor::execute_xcm_in_credit(origin_location, message, weight, weight)
+					.ensure_complete();
 
-		Ok(())
+			match result {
+				Ok(()) => TransactionOutcome::Commit(Ok(())),
+				Err(_) =>
+					return TransactionOutcome::Rollback(Err(DispatchError::Other(
+						"Xcm execution failed.",
```
vgeddes, line 100: If `result` is an error we should be passing it to `TransactionOutcome::Rollback` so that it can handled or logged higher up the stack.

alistair-singh: The issue with this is that `DispatchResult` does not provide a way to accept an arbitrary Error type or an error message string that is built dynamically at runtime. It only accepts `&'static str`.

I have considered adding an enum type to represent errors returned from `transfer_reserve`.
```rust
pub enum TransferError {
    ZeroFees,
    UnweighableMessage,
    XcmExecution(xcm::v2::traits::Error),
}
```

The error here is of type `xcm::v2::traits::Error` and to pass this higher up the stack we would need to add it to the public interface of `reserve_transfer`. This would add a dependency on the `xcm` crate for `primitives/core` crate as well as `eth-app` and `erc20-app` pallets. Currently only `xcm-support` crate depends on the `xcm` crate.

Should we be concerned about adding these dependencies? We could potentially just log the error here and return `Err(DispatchError::Other("Xcm execution failed.")` as per the existing implementation.

vgeddes: Cool, understood. Anyway, with the other changes you made (like hoisting with_transaction), I guess my earlier comment is obsolete.

parachain/pallets/erc20-app/src/lib.rs
```diff
@@ -186,19 +186,23 @@ pub mod pallet {
...
+					if let Err(err) = result {
+						log::error!(
+							"Failed to execute xcm transfer to parachain {} - {:?}.",
+							destination.para_id,
+							err
+						);
+						return TransactionOutcome::Rollback(());
+					}
+					TransactionOutcome::Commit(())
+				});
```
vgeddes, line 205: Since we can't return an error if the xcm transfer failed, we could emit an event instead, something like `XcmTransferFailed { error: XcmError }`.

alistair-singh: I have created a ticket for tracking XCM transfer state with events. SNO-227

PR #581, opened 2022-02-11
Add dispatchable to handle long range forks.

alistair-singh:
Resolves: SNO-186

### DIFF
--- a/parachain/pallets/ethereum-light-client/src/lib.rs
+++ b/parachain/pallets/ethereum-light-client/src/lib.rs
@@ -276,10 +276,7 @@ pub mod pallet {
 		/// Requires sudo user.
 		#[pallet::weight(T::WeightInfo::import_header())]
 		#[transactional]
-		pub fn handle_long_range_fork(
-			origin: OriginFor<T>,
-			forked_at: H256,
-		) -> DispatchResult {
+		pub fn handle_long_range_fork(origin: OriginFor<T>, forked_at: H256) -> DispatchResult {
 			Ok(())
 		}
 	}
--- a/parachain/pallets/ethereum-light-client/src/lib.rs
+++ b/parachain/pallets/ethereum-light-client/src/lib.rs
@@ -265,6 +265,23 @@ pub mod pallet {
 
 			Ok(())
 		}
+
+		/// Reset's the internal state of the light client.
+		///
+		/// `forked_at` - The hash of the header where the fork originated. The common ancestor.
+		///
+		/// This is only meant to be used in staging environments where long range forks
+		/// can happen.
+		///
+		/// Requires sudo user.
+		#[pallet::weight(T::WeightInfo::import_header())]
+		#[transactional]
+		pub fn handle_long_range_fork(
+			origin: OriginFor<T>,
+			forked_at: H256,
+		) -> DispatchResult {
+			Ok(())
+		}
 	}
 
 	impl<T: Config> Pallet<T> {
--- a/parachain/pallets/ethereum-light-client/src/lib.rs
+++ b/parachain/pallets/ethereum-light-client/src/lib.rs
@@ -274,9 +274,10 @@ pub mod pallet {
 		/// can happen.
 		///
 		/// Requires sudo user.
-		#[pallet::weight(T::WeightInfo::import_header())]
+		#[pallet::weight(1_000_000)]
 		#[transactional]
 		pub fn handle_long_range_fork(origin: OriginFor<T>, forked_at: H256) -> DispatchResult {
+			let _sender = ensure_root(origin)?;
 			Ok(())
 		}
 	}
--- a/parachain/pallets/ethereum-light-client/src/lib.rs
+++ b/parachain/pallets/ethereum-light-client/src/lib.rs
@@ -290,15 +290,13 @@ pub mod pallet {
 				return Err("Cannot reset to fork after the current finalized block.".into())
 			}
 
-			let (new_finalized_hash, _) =
+			let (new_finalized_hash, finalized_header) =
 				Self::find_finalized_ancestor(forked_at, required_descendants).ok_or(
 					"Cannot reset to fork if it does not have the required number of decendants.",
 				)?;
 
-			let finalized_stored_header =
-				<Headers<T>>::get(new_finalized_hash).ok_or(Error::<T>::MissingHeader)?;
 			let finalized_block_id = EthereumHeaderId {
-				number: finalized_stored_header.header.number,
+				number: finalized_header.number,
 				hash: new_finalized_hash,
 			};
 
--- a/parachain/pallets/ethereum-light-client/src/lib.rs
+++ b/parachain/pallets/ethereum-light-client/src/lib.rs
@@ -266,7 +266,7 @@ pub mod pallet {
 			Ok(())
 		}
 
-		/// Reset's the internal state of the light client.
+		/// Reset's the internal state of the light client to a previously imported header.
 		///
 		/// `forked_at` - The hash of the header where the fork originated. The common ancestor.
 		///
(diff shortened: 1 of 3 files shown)


musnit:
@alistair-singh so with ensure_root, if there is no sudo user (eg: in production), this function can't be called?

@vgeddes plz review

@claravanstaden probably would be valuable for you to review too as part of becoming more familiar with rust and the ethereum light client!

parachain/pallets/ethereum-light-client/src/tests.rs
```diff
@@ -526,3 +526,246 @@ fn it_can_only_import_max_headers_worth_of_headers() {
...
+		//        |   |
+		// 6      B6  |     Current Best Block
+		//            |
+		// 7          B4
+		//            |
+		// 8          B5    Expected Finalized Block
+		//            |
+		// 9          B6
+		//            |
+		// 10         B7    Expected Best Block
```
claravanstaden, line 625: This is great, thanks for the picture - worth a thousand words 😉

parachain/pallets/ethereum-light-client/src/mock.rs
```diff
@@ -157,14 +157,17 @@ pub fn genesis_ethereum_block_hash() -> H256 {
...
 
-pub fn child_of_header(header: &EthereumHeader) -> EthereumHeader {
+pub fn child_of_header(header: &EthereumHeader, timestamp: Option<u64>) -> EthereumHeader {
 	let mut child: EthereumHeader = Default::default();
 	child.difficulty = 1.into();
 	child.parent_hash = header.compute_hash();
 	child.number = header.number + 1;
+	if let Some(timestamp) = timestamp {
+		child.timestamp = timestamp
+	}
```
claravanstaden, line 170: Is this added for test purposes only? I see the timestamps being set in the tests: https://github.com/Snowfork/snowbridge/blob/f5e2ea1c9ad29a61db31b4898e3983cfd3442c46/parachain/pallets/ethereum-light-client/src/tests.rs#L550

What does adding the timestamp accomplish?

parachain/pallets/ethereum-light-client/src/lib.rs
```diff
@@ -415,29 +457,38 @@ pub mod pallet {
...
+		) -> Option<(H256, EthereumHeader)> {
+			ancestry::<T>(best_block_id.hash).enumerate().find_map(|(i, pair)| {
+				if i < required_descendants {
+					None
+				} else {
+					Some(pair)
+				}
+			})
+		}
+
```
claravanstaden, line 472: Great work splitting this out, I see the logic was used in a few places.

parachain/pallets/ethereum-light-client/src/lib.rs
```diff
@@ -265,6 +265,48 @@ pub mod pallet {
...
+			let stored_header = <Headers<T>>::get(forked_at).ok_or(Error::<T>::MissingHeader)?;
+			let header = stored_header.header;
+
+			let required_descendants = T::DescendantsUntilFinalized::get() as usize;
+			let best_block_id = EthereumHeaderId { number: header.number, hash: forked_at };
+
+			// Return error if the fork occurs after the current finalized block
+			let current_finalized = <FinalizedBlock<T>>::get();
+			if best_block_id.number >= current_finalized.number {
+				return Err(Error::<T>::Unknown.into())
```
claravanstaden, line 291: Is an `Unknown` error sufficient because we only use this dispatchable during testing?

claravanstaden (approved):
Awesome work! 🥇  Thanks for the tests and the diagrams in the tests. 😁  I have a few questions, otherwise great!

alistair-singh:
> @alistair-singh so with ensure_root, if there is no sudo user (eg: in production), this function can't be called?

Yes this is correct.

parachain/pallets/ethereum-light-client/src/mock.rs
alistair-singh: So I needed two different blocks with the same number to have different hashes. So I needed to pick some field on the header type and alter it in some way. I chose timestamp because to make the tests easier to digest in the sense of two blocks with the same number coming in at different points in time. But I could of chosen any field really.

parachain/pallets/ethereum-light-client/src/lib.rs
alistair-singh: Yeah so I also did not feel too great about overloading the `Unknown` error here, but this dispatchable is only going to be used in staging environments by one of us following a runbook to recover from a hard fork. So I did not want to pollute the interface by adding custom error messages for this one method.

vgeddes: You can also just return an untyped string message instead of unknown:
```suggestion
				return Err("new best block is not valid".into())
```

That gets converted into `DispatchError::Other(some_string)`.

parachain/pallets/ethereum-light-client/src/mock.rs
claravanstaden: Great, makes sense, thanks for the explanation.

parachain/pallets/ethereum-light-client/src/lib.rs
```diff
@@ -265,6 +265,48 @@ pub mod pallet {
...
+		/// can happen.
+		///
+		/// Requires sudo user.
+		#[pallet::weight(1_000_000)]
+		#[transactional]
+		pub fn handle_long_range_fork(origin: OriginFor<T>, forked_at: H256) -> DispatchResult {
+			ensure_root(origin)?;
+
+			let stored_header = <Headers<T>>::get(forked_at).ok_or(Error::<T>::MissingHeader)?;
+			let header = stored_header.header;
```
vgeddes, line 283: Nit: Should be able to collapse to:

```suggestion
			let header = <Headers<T>>::get(forked_at).ok_or(Error::<T>::MissingHeader)?.header;
```

Should compile

parachain/pallets/ethereum-light-client/src/lib.rs
```diff
@@ -265,6 +265,48 @@ pub mod pallet {
...
+		///
+		/// `forked_at` - The hash of the header where the fork originated. The common ancestor.
+		///
+		/// This is only meant to be used in staging environments where long range forks
+		/// can happen.
+		///
+		/// Requires sudo user.
+		#[pallet::weight(1_000_000)]
+		#[transactional]
+		pub fn handle_long_range_fork(origin: OriginFor<T>, forked_at: H256) -> DispatchResult {
```
vgeddes, line 279: Nit: I prefer this name, which is slightly more descriptive while still being short enough:

```suggestion
		pub fn force_reset_to_fork(origin: OriginFor<T>, forked_at: H256) -> DispatchResult {
```

parachain/pallets/ethereum-light-client/src/lib.rs
```diff
@@ -415,29 +457,38 @@ pub mod pallet {
 			Ok(())
 		}
 
+		fn find_finalized_decendant(
```
vgeddes, line 460: Actually, this should be named `find_finalized_ancestor`, since we are looking for an ancestor of the best_block.

parachain/pallets/ethereum-light-client/src/lib.rs
```diff
@@ -415,29 +457,38 @@ pub mod pallet {
...
+			best_block_id: &EthereumHeaderId,
```
vgeddes, line 461: You should probably just pass a `H256` instead of `EthereumHeaderId` here, since the function is only interested in the block hash.

parachain/pallets/ethereum-light-client/src/lib.rs
```diff
@@ -265,6 +265,48 @@ pub mod pallet {
...
+					.ok_or(Error::<T>::Unknown)?;
+
+			let finalized_stored_header =
+				<Headers<T>>::get(new_finalized_hash).ok_or(Error::<T>::MissingHeader)?;
+			let header = finalized_stored_header.header;
+			let finalized_block_id =
+				EthereumHeaderId { number: header.number, hash: new_finalized_hash };
+
+			<FinalizedBlock<T>>::put(finalized_block_id);
+			<BestBlock<T>>::put((best_block_id, header.difficulty));
```
vgeddes, line 306: Looks like a bug as `header.difficulty` is the difficulty of the new finalized block, not the new best block.

alistair-singh: Is it the sum of difficulty between the best and finalized block?

vgeddes: Actually looks like its the total difficulty (sum of all individual block difficulties since genesis) of the chain. So you're gonna need to use some of the logic from import_validated_header: https://github.com/Snowfork/snowbridge/blob/main/parachain/pallets/ethereum-light-client/src/lib.rs#L361

vgeddes: Awesome 👍

parachain/pallets/ethereum-light-client/src/lib.rs
```diff
@@ -265,6 +265,48 @@ pub mod pallet {
...
+			if best_block_id.number >= current_finalized.number {
+				return Err("Cannot reset to fork after the current finalized block.".into())
+			}
+
+			let (new_finalized_hash, _) =
+				Self::find_finalized_ancestor(forked_at, required_descendants).ok_or(
+					"Cannot reset to fork if it does not have the required number of decendants.",
+				)?;
+
+			let finalized_stored_header =
```
vgeddes, line 298: This retrieval is redundant, since `find_finalized_ancestor` already returns an `EthereumHeader`, which you're ignoring. And you can get the block number from the `EthereumHeader`, so need to retrieve the StoredHeader.

vgeddes (approved):
Looks good! Approved.

PR #582, opened 2022-02-14
Eth2 Light Client

claravanstaden:
Starts adding the Eth2 light client scaffolding in a new pallet.


parachain/pallets/ethereum2-light-client/src/mock.rs
```diff
@@ -0,0 +1,50 @@
...
+	type Block = frame_system::mocking::MockBlock<Test>;
+
+	frame_support::construct_runtime!(
+		pub enum Test where
+			Block = Block,
+			NodeBlock = Block,
+			UncheckedExtrinsic = UncheckedExtrinsic,
+		{
+			System: frame_system::{Pallet, Call, Storage, Event<T>},
+			Verifier: verifier::{Pallet, Call, Config, Storage, Event<T>},
```
vgeddes, line 25: Obsolete naming. In our production runtimes we refer to the module as `EthereumLightClient` instead of `Verifier`

```suggestion
			EthereumLightClient: crate::{Pallet, Call, Config, Storage, Event<T>},
```

parachain/pallets/ethereum2-light-client/src/mock.rs
```diff
@@ -0,0 +1,50 @@
...
+		}
+	);
+
+	impl frame_system::Config for Test {
+		type Origin = Origin;
+		type AccountId = AccountId;
+	}
+
+	impl verifier::Config for Test {}
```
vgeddes, line 34:
```suggestion
	impl crate::Config for Test {}
```

parachain/pallets/ethereum2-light-client/src/weights.rs
```diff
@@ -0,0 +1,59 @@
+//! Autogenerated weights for ethereum_light_client
```
vgeddes, line 1: Benchmarks won't be important for quite a while. We can add them much later. So I would advise just removing this file and any other benchmarking stuff,

parachain/pallets/ethereum2-light-client/src/lib.rs
```diff
@@ -0,0 +1,165 @@
...
+	#[pallet::genesis_build]
+	impl<T: Config> GenesisBuild<T> for GenesisConfig {
+		fn build(&self) {
+
+		}
+	}
+
+	#[pallet::call]
+	impl<T: Config> Pallet<T> {
+		#[pallet::weight(T::WeightInfo::import_header())]
```
vgeddes, line 148: For now, I'd just ignore benchmarks & weights, and use a dummy constant:

```suggestion
		#[pallet::weight(1_000_000)]
```

vgeddes (commented):
Added some minor tips & advise. :)

parachain/pallets/ethereum2-light-client/src/lib.rs
```diff
@@ -0,0 +1,402 @@
...
+	pub finalized_header: Option<BeaconBlockHeader>,
+	/// Vector[Bytes32, floorlog2(FINALIZED_ROOT_INDEX)]
+	pub finality_branch: Vec<H256>,
+	///  Sync committee aggregate signature
+	pub sync_aggregate: SyncAggregate,
+	///  Fork version for the aggregate signature
+	pub pubfork_version: Version,
+}
+
+pub use frame_system::pallet::*;
```
vgeddes, line 93: You're importing frame_system's pallet contents instead of the contents of this pallet

parachain/pallets/ethereum2-light-client/src/lib.rs
```diff
@@ -0,0 +1,447 @@
...
+/// Sync committee as it is stored in the runtime storage.
+#[derive(Clone, Default, Encode, Decode, PartialEq, RuntimeDebug, TypeInfo)]
+pub struct SyncCommittee {
+	pub pubkeys: Vec<String>, // TODO most likely not a string
+}
+
+#[derive(Clone, Default, Encode, Decode, PartialEq, RuntimeDebug, TypeInfo)]
+pub struct SyncAggregate {
+	// 1 or 0 bit, indicates whether a sync committee participated in a vote
+	pub sync_committee_bits: Vec<u64>,
```
vgeddes, line 75: Should rather use `u8` instead of `u64` for this :)

```suggestion
	pub sync_committee_bits: Vec<u8>,
```

parachain/pallets/ethereum2-light-client/src/lib.rs
```diff
@@ -0,0 +1,447 @@
...
+	pub state_root: H256,
+	// BLS Signature of the block by the block proposer, TODO type isn't right yet
+	pub signature: String,
+	proposer_index: ValidatorIndex,
+}
+
+/// Sync committee as it is stored in the runtime storage.
+#[derive(Clone, Default, Encode, Decode, PartialEq, RuntimeDebug, TypeInfo)]
+pub struct SyncCommittee {
+	pub pubkeys: Vec<String>, // TODO most likely not a string
```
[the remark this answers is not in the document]
claravanstaden: Thanks for the heads-up, @vgeddes!

parachain/pallets/ethereum-beacon-light-client/src/lib.rs
```diff
@@ -0,0 +1,562 @@
...
+				genesis_validators_root: genesis_validators_root.as_bytes().to_vec(),
+			})
+		}
+
+		fn compute_signing_root(attested_header: BeaconBlockHeader, domain: H256) -> Vec<u8> {
+			todo!()
+		}
+
+		fn bls_fast_aggregate_verify(participant_pubkeys: Vec<Vec<u8>>, signing_root: Vec<u8>, sync_committee_signature: Vec<u8>) -> bool {
+			todo!()
```
vgeddes, line 536: For BLS signature verification, the reference eth2 implementation (also rust) suggests that the [milagro](https://github.com/apache/incubator-milagro-crypto-rust
) crypto library could be a good fit: https://github.com/sigp/lighthouse/blob/90d5ab15660e62e16390731d92c74b640a70bed9/crypto/bls/src/lib.rs#L12
There may be other good candidates we can work with. However we have a strict constraints to ensure that the code can be compiled down to WASM:
1. Must be written in pure Rust (ie no bindings to C/C++ libraries, inline assembly, etc).
2. Must compile without the standard library available: https://docs.rust-embedded.org/book/intro/no-std.html

If we go ahead with Milagro, we know it supports (1), so we will still need to fork it to support (2).

vgeddes: Ah we may be in luck, in this variant/subset of the milagro library supports `no_std` out of the box: https://github.com/sigp/milagro_bls

Big warnings about the lack of security audits though. But I think we can address that later on.

parachain/pallets/ethereum-beacon-light-client/src/lib.rs
```diff
@@ -0,0 +1,562 @@
...
+	pub(super) type CurrentSyncCommittee<T: Config> = StorageValue<_, SyncCommittee, ValueQuery>;
+
+	/// Next sync committee corresponding to the active header
+	#[pallet::storage]
+	pub(super) type NextSyncCommittee<T: Config> = StorageValue<_, SyncCommittee, ValueQuery>;
+
+	/// Best available header to switch finalized head to if we see nothing else
+	#[pallet::storage]
+	pub(super) type BestValidUpdate<T: Config> =
+		StorageValue<_, Option<LightClientUpdate>, ValueQuery>; // TODO: Maybe I should use OptionQuery here instead of Option?
```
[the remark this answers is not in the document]
claravanstaden: I'll change this when I call the light from the relayer, if that's OK? I'd like to see how it makes a difference, practically, by calling the light client from the relayer with `ValueQuery` first and then see how `OptionQuery` simplifies it.

claravanstaden:
@vgeddes @alistair-singh @musnit I think this is ready for review, even if there are still some unfinished parts. I would like to open smaller PRs from here on out.

parachain/pallets/ethereum-beacon-light-client/Cargo.toml
```diff
@@ -0,0 +1,64 @@
...
+scale-info = { version = "1.0", default-features = false, features = ["derive"] }
+hex = { package = "rustc-hex", version = "2.1.0", default-features = false }
+hex-literal = { version = "0.3.1", optional = true }
+rlp = { version = "0.5", default-features = false }
+eth2_ssz = { version = "0.4.0" }
+eth2_ssz_derive = { version = "0.3.0" }
+eth2_ssz_types = { version = "0.2.1" }
+tree_hash = { version = "0.4.0" }
+tree_hash_derive = { version = "0.4.0" }
+milagro_bls = { git = "https://github.com/sigp/milagro_bls", tag = "v1.4.2" }
```
vgeddes, line 24: This lib has an [std](https://github.com/sigp/milagro_bls/blob/16655aa033175a90c10ef02aa144e2835de23aec/Cargo.toml#L25) feature, which we should not enable by default.
```suggestion
milagro_bls = { git = "https://github.com/sigp/milagro_bls", tag = "v1.4.2", default-features = false }
```

parachain/pallets/ethereum-beacon-light-client/Cargo.toml
```diff
@@ -0,0 +1,64 @@
...
+    "frame-support/std",
+    "frame-system/std",
+    "rlp/std",
+    "sp-core/std",
+    "sp-io/std",
+    "sp-runtime/std",
+    "sp-std/std",
+    "snowbridge-core/std",
+    "snowbridge-ethereum/std",
+]
```
vgeddes, line 58: Need to add the std feature of milagro_bls here. So that when we want to compile the code to native machine code (not WASM) we can enable std feature and allow linking to the core std library.

parachain/pallets/ethereum-beacon-light-client/Cargo.toml
```diff
@@ -0,0 +1,57 @@
+[package]
+name = "snowbridge-ethereum-beacon-light-client"
+description = "Snowbridge Beacon Light Client Pallet"
+version = "0.0.1"
+edition = "2021"
```
alistair-singh, line 5: The reason why the build is still failing after adding edition 2021 is because the github workflow doing the build is using an older version of the rust toolchain. 

https://github.com/Snowfork/snowbridge/blob/main/.github/workflows/parachain.yml#L37-L39

This commit is the version I used for edition 2021 in the `v0.9.17` PR that is not yet merged in.
https://github.com/Snowfork/snowbridge/pull/591/commits/7167fdcb67e2d83b00a92944aab17b5f9eed2425

claravanstaden: Thanks Alistair! 👍  So the edition 2021 should work after the rust toolchain upgrade? I've started updating the other crate's editions too: https://github.com/Snowfork/snowbridge/pull/605/files

claravanstaden:
Dependent on https://github.com/Snowfork/snowbridge/pull/591 to update Rust toolchain for the Cargo edition `2021` to work.

claravanstaden:
Closing in favour of https://github.com/Snowfork/snowbridge/pull/610.

PR #583, opened 2022-02-14
Improve configuration of contract addresses

sumitsnk:
This PR resolves the ticket SNO-182.


vgeddes:
This is a good start, but I think it would better to make the script a hardhat task that pulls the addresses from the https://github.com/wighawag/hardhat-deploy plugin. Will chat more about it tomorrow during the day.

ethereum/tasks/contractAddress.ts
```diff
@@ -0,0 +1,33 @@
...
+
+        console.log("BasicInboundChannel", BasicInboundChannel.address)
+        console.log("BasicOutboundChannel", BasicOutboundChannel.address)
+        console.log("IncentivizedInboundChannel", IncentivizedInboundChannel.address)
+        console.log("IncentivizedOutboundChannel", IncentivizedOutboundChannel.address)
+        console.log("ETHApp", ETHApp.address)
+        console.log("ERC20App", ERC20App.address)
+        console.log("DOTApp", DOTApp.address)
+        console.log("ERC721App", ERC721App.address)
+        console.log("SnowDOTAddress", SnowDOTAddress)
```
alistair-singh, line 31: Just a suggestion but we should use an output format that can be easily copy/pasted to configure the webui. 

https://github.com/Snowfork/snowbridge-web/blob/main/src/contractAddresses.json

Or even output json and the web ui contracts could be updated with the following command:
```
yarn hardhat  contractAddressList --network localhost > $WEB_UI_REPO/src/contractAddresses.json
```

vgeddes: Totally agree, this was what I was expecting. @sumitsnk can you implement the above?

test/scripts/helpers/contractAddresses.js
```diff
@@ -0,0 +1,50 @@
+const fs = require("fs");
+const Web3 = require("web3");
```
vgeddes, line 2: This script isn't necessary anymore, since we have the hardhat task in place.

test/scripts/start-services.sh
```diff
@@ -5,6 +5,7 @@ root_dir="$(realpath ..)"
 parachain_dir="$root_dir/parachain"
 ethereum_dir="$root_dir/ethereum"
 relay_dir="$root_dir/relayer"
+test_dir="$root_dir/test"
```
vgeddes, line 8: I think its fine for users to call the hardhat task manually to dump the addresses. So lets revert all these changes to the `start-services`.

ethereum/tasks/contractAddress.ts
sumitsnk: Yes, as suggested, I've adjusted the code.

test/scripts/start-services.sh
sumitsnk: Yes, as suggested, I've adjusted the code.

alistair-singh (approved):

vgeddes (approved):
LGTM, Lets merge it in.

PR #584, opened 2022-02-15
Inject ethereum config from shell environment

vgeddes:
For our deployment needs, we really want to configure hardhat from the shell environment rather than a static `.env` file loaded by `dotenv`. So we should use [direnv](https://direnv.net/) instead: 

This allows our deployment scripts in the snowbridge-infra repo to inject suitable config into hardhat.


claravanstaden (approved):

vgeddes:
I'm gonna hold off on merging this until https://github.com/Snowfork/snowbridge-infra/pull/17 is ready

PR #590, opened 2022-03-03
Fix nonces as part of recovering from a fork.

alistair-singh:
Resolves: SNO-218.

The script will now detect and fix nonce changes. Updating the nonce will allow double spend but this is a compromise to solve the longe range forks in our staging environment.

The way that `force_reset_to_fork` currently works is that it sets the `BestBlock` and rewinds the `FinalizedBlock` by 8 blocks(`DescendantsUntilFinalized`). This is not ideal as the relayer starts from the `FinalizedBlock` and will replay messages from there. This means there is a potential 8 block window in which transactions will be replayed. To change this we would need to change `force_reset_to_fork` to set the `FinalizedBlock` and walk forward 8 blocks to set the `BestBlock`.

### DIFF
--- a/test/scripts/detect-fork.ts
+++ b/test/scripts/detect-fork.ts
@@ -151,7 +151,7 @@ const fetchEthNonces = async (
     );
     // get all nonce changing events that happened after the new finalized
     pastEvents[nonce.name] = contract.getPastEvents(nonce.event, {
-      fromBlock: commonAnscestorBlockNumber + 1 - descendantsUntilFinalized,
+      fromBlock: commonAnscestorBlockNumber - descendantsUntilFinalized,
       toBlock: "latest",
     });
     nonces[nonce.name] = contract.methods.nonce().call();
@@ -323,7 +323,7 @@ const main = async () => {
     console.log(`Checking block number ${number}...`);
     const ethBlock = await ethApi.eth.getBlock(number, false);
     const paraBlock: any = await fetchImported(parachainApi, ethBlock.hash);
-    if (paraBlock === null && !paraBlock.finalized) continue;
+    if (paraBlock === null || !paraBlock.finalized) continue;
 
     assert(
       ethBlock.number === paraBlock.header.number,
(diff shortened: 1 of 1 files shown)


test/scripts/detect-fork.ts
```diff
@@ -73,6 +89,146 @@ const fetchImported = async (api: ApiPromise, hash: string) => {
...
+      "0x684b82bef882079feeabe54a5bd7b94a718368a0ace36e2b1b8b6dbd7f8093c0",
+  },
+  {
+    name: "IncentivizedInboundChannel",
+    event: "MessageDispatched",
+    storageKey:
+      "0xf0f4d0b91e760c07da58bc0498033acb718368a0ace36e2b1b8b6dbd7f8093c0",
+  },
+  {
+    name: "BasicOutboundChannel",
```
vgeddes, line 126: Can remove everything related to parachain outbound channels. The parachain outbound channels won't be affected in any way by a hard fork on ropsten. The same applies to the inbound channels on the ethereum side that receive messages from the parachain outbound channels.

test/scripts/detect-fork.ts
```diff
@@ -97,14 +258,25 @@ const main = async () => {
...
       default: null,
     },
-    "fix": {
+    fix: {
       type: "string",
       demandOption: false,
       describe: "Fix the fork with the following user. e.g. '//Alice'",
       default: null,
     },
+    "decendants-until-finalized": {
```
vgeddes, line 267: Nit: Spelling:

```suggestion
    "descendants-until-finalized": {
```

vgeddes (commented):
Looks great 🚀

test/scripts/detect-fork.ts
alistair-singh: I left the config for these nonces for displaying nonces in the script output. However I have changed the `generateUpdates` function to only generate updates for parachain inbound channels.

vgeddes (approved):
Looks good :rocket:

PR #591, opened 2022-03-09
Update substrate to 0.9.17

alistair-singh:
Resolves: SNO-198

### Require MaxEncodedLen per default
https://github.com/paritytech/substrate/pull/10662
* I have added the attribute `without_storage_info` to pallets which cannot implement `MaxEncodedLen` yet. This is a non trivial change to make as we have to change code to use `BoundedVec` and to check bounds.
* Added ticket to address this. Related: SNO-223

### Remove Default bound for AccountId
https://github.com/paritytech/substrate/pull/10403
* The `zero` account is no longer the default as it poses potential risks around external callers using the zero account to gain privileges based on `zero` account being used as default in the code base.
* I have switched to using `Optional<AccountId>` in pallets.
* XCM config uses `ParentIsPreset` as opposed to `ParentIsDefault`.
* Added `CheckNonZeroSender` as a signed extension to filter out `zero` accounts.

### Switch from StructOps to Clap for arg parsing
https://github.com/substrate-developer-hub/substrate-parachain-template/pull/101
* The substrate parachain node template now uses `clap` for args parsing. I have removed `StructOps` to keep some symmetry with the project to make future upgrades easier.

### Updated Workflow Toolchains
* Updated to stable-2022-01-20 (1.58.1)
* This is to support the edition 2021 required by the `snowbridge-test-collator`

### Allow unpaid execution from relay chain or relay chain executive.
https://github.com/substrate-developer-hub/substrate-parachain-template/pull/80
* Changed the XCM config to allow unpaid execution for the relay chain or it's executive. Previously this was set to `Unit`.

## E2E Tests
Currently e2e tests are failing from the parachain -> ethereum direction. This is because of an update to the BEEFY included in v0.9.17 version of polkadot. The beefy light client and relayer will need to be updated.

### DIFF
--- a/.github/workflows/parachain.yml
+++ b/.github/workflows/parachain.yml
@@ -35,13 +35,13 @@ jobs:
       - uses: actions-rs/toolchain@v1
         with:
           profile: minimal
-          toolchain: nightly-2021-07-01
+          toolchain: stable-2022-01-20
           target: wasm32-unknown-unknown
       - name: cargo check
         uses: actions-rs/cargo@v1
         with:
           command: check
-          toolchain: nightly-2021-07-01
+          toolchain: stable-2022-01-20
           args: >-
             --manifest-path parachain/Cargo.toml
             --workspace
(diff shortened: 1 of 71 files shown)


vgeddes (approved):
LGTM!

PR #594, opened 2022-03-11
Change descendants until finalized to 16 for snowblink

alistair-singh:
Resolves: SNO-221

### DIFF
--- a/parachain/runtime/snowblink/src/lib.rs
+++ b/parachain/runtime/snowblink/src/lib.rs
@@ -568,7 +568,7 @@ impl incentivized_channel_outbound::Config for Runtime {
 }
 
 parameter_types! {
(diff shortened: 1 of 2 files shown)


alistair-singh:
> Nice, LGTM.
> 
> Also make to update the relayer too, as it also has similar settings:
> 
> * https://github.com/Snowfork/snowbridge-infra/blob/056f1c4f5122a3fe007ca1ca60c92734b64a4045/ethereum-relay.nomad#L30
> * https://github.com/Snowfork/snowbridge-infra/blob/056f1c4f5122a3fe007ca1ca60c92734b64a4045/polkadot-relay.nomad#L26

I have opened a PR for this https://github.com/Snowfork/snowbridge-infra/pull/19.

PR #596, opened 2022-03-11
Updated libsecp256k1 dependency

alistair-singh:
Removed all references to libsecp256k1 not `>=0.5.0`.

https://rustsec.org/advisories/RUSTSEC-2021-0076

e2e tests pass.


musnit:
@alistair-singh do you know why github's bot didn't detect this and let us know? do we have it enabled? does it work well for Rust?

vgeddes:
Actually `libsecp256k1` is an unused dependency. We should remove it :)

vgeddes (approved):

PR #607, opened 2022-03-23
Support new Beefy commitment structure

vgeddes:
Companion PR: https://github.com/Snowfork/go-substrate-rpc-client/pull/38

Resolves: SNO-229


ethereum/contracts/BeefyLightClient.sol
```diff
@@ -287,11 +293,11 @@ contract BeefyLightClient {
         verifyCommitment(id, commitment, validatorProof);
         verifyNewestMMRLeaf(
             latestMMRLeaf,
-            commitment.payload,
+            bytes32(commitment.payload[0].data),
```
musnit, line 296: may wanna make 0 a constant VERSION_NUMBER

ethereum/contracts/BeefyLightClient.sol
```diff
@@ -543,23 +566,102 @@ contract BeefyLightClient {
...
-                abi.encodePacked(
-                    commitment.payload,
-                    commitment.blockNumber.encode32(),
-                    commitment.validatorSetId.encode64()
-                )
-            );
+        bytes memory payload;
+        if (commitment.payload.length == 1 && commitment.payload[0].data.length == 32) {
+            payload = encodePayloadFast(commitment.payload[0]);
+        } else {
```
musnit, line 577: hmm would an alternative solution like this work:
 - we submit Commitment in the same format as is currently, but also submit "commitmentPrefixBytes" and "commitmentSuffixBytes" and then encodeCommitment just does prefix+encodeCommitment+suffix ?

i guess it does mean that the validators could potentially sign multiple different/clashing commitments within the same signature, but it may be safe to assume that this will be handled/slashed on the polkadot side if so

ethereum/contracts/BeefyLightClient.sol
```diff
@@ -285,13 +295,15 @@ contract BeefyLightClient {
         SimplifiedMMRProof calldata proof
     ) public {
         verifyCommitment(id, commitment, validatorProof);
+
+        bytes32 mmrRootHash = getMmrRootHash(commitment.payload);
         verifyNewestMMRLeaf(
             latestMMRLeaf,
-            commitment.payload,
```
musnit, line 290: can we just keep using commitment.payload as is, based on my other suggestion below?

PR #614, opened 2022-04-08
BEEFY light client v2

vgeddes:
On-Chain changes:
* BEEFY light client
  * Both the current and next validator set are tracked in order to work with LEAN beefy
  * Added separate method for submitting new MMR leaves. in order to update the next validator set.
  * Various refactoring/cleanups
  * Renamed BeefyLightClient to BeefyClient
* Parachain light client
  * Is now a fully-fledged contract and acts as intermediary layer between channels and the BEEFY light client.
  * Parachain ID is configurable on contract construction
  * Renamed ParachainLightClient to ParachainClient
* SimplifiedMMRVerification has been downgraded to a library and renamed to MMRProofVerification
* Inbound channels now accept message "bundles" to simplify the protocol and nonce handling

Relayer:
* beefy relayer has mostly been rewritten to support LEAN beefy. Previously the relayer supported concurrent processing of commitments with the aid of a SQLite DB. This has been removed in favour of processing commitments sequentially. However the new logic allows some commitments to be skipped/ignored in order to keep up with the relay chain. So the SQLite DB is no longer needed and has been removed.
* The rewrite also improves on reliability by fixing various problems including https://linear.app/snowfork/issue/SNO-149/beefy-relay-sends-duplicate-final-signature-commitments and other issues.

Resolves: SNO-238, SNO-149, SNO-237, SNO-50, SNO-168


ethereum/contracts/BeefyLightClient.sol
```diff
@@ -140,10 +140,9 @@ contract BeefyLightClient {
 
     // We must ensure at least one block is processed every session,
     // so these constants are checked to enforce a maximum gap between commitments.
-    uint64 public constant NUMBER_OF_BLOCKS_PER_SESSION = 2400;
-    uint64 public constant ERROR_AND_SAFETY_BUFFER = 10;
-    uint64 public constant MAXIMUM_BLOCK_GAP =
-        NUMBER_OF_BLOCKS_PER_SESSION - ERROR_AND_SAFETY_BUFFER;
+    // TODO: verify SAFETY_BUFFER
```
musnit, line 143: don't forget to create issue flagged as security with this if you don't fix it before merge

ethereum/contracts/BeefyLightClient.sol
```diff
@@ -140,10 +140,9 @@ contract BeefyLightClient {
...
+    uint64 public numberOfBlocksPerSession;
```
musnit, line 144: should probably use immutable for this: https://solidity-by-example.org/immutable/

(immutable also i believe stores the variable within contract code, rather than in storage, so is way cheaper gas cost to access in future, much like a constant)

ethereum/contracts/BeefyLightClient.sol
```diff
@@ -153,16 +152,23 @@ contract BeefyLightClient {
...
         mmrVerification = _mmrVerification;
         currentId = 0;
         latestBeefyBlock = _startingBeefyBlock;
+        numberOfBlocksPerSession = _numberOfBlocksPerSession;
     }
 
     /* Public Functions */
 
+    // maximum block gap between commitments
+    function maximumCommitmentGap() internal view returns (uint64) {
```
musnit, line 168: maybe move this to an immutable var you set in constructor

ethereum/contracts/BeefyLightClient.sol
```diff
@@ -383,18 +368,30 @@ contract BeefyLightClient {
...
-                nextAuthoritySetRoot,
-                nextAuthoritySetLen,
-                nextAuthoritySetId
-            );
+
+        // Don't process older leafs
+        if (leaf.nextAuthoritySetId <= validatorRegistry.id()) {
+            return;
         }
+
```
musnit, line 381: so, beefy is changing to increase the id every time now right?

is there any cheap way we can check if the validator registry has actually changed or not so that we can skip the more expensive inclusion proof check if the validator set is all the same?

maybe if nextAuthoritySetRoot hasn't changed we could skip the verification, but i guess that opens up an attack where an incumbent validator set can frontrun a new validator set and get the bridge to fork.

we'd need beefy to provide us with a field that does not increment that allows us to skip this

ethereum/contracts/BeefyLightClient.sol
```diff
@@ -287,24 +293,16 @@ contract BeefyLightClient {
         uint256 id,
         Commitment calldata commitment,
         ValidatorProof calldata validatorProof,
-        BeefyMMRLeaf calldata latestMMRLeaf,
-        SimplifiedMMRProof calldata proof
+        BeefyMMRLeaf calldata leaf,
+        SimplifiedMMRProof calldata leafProof
     ) public {
         verifyCommitment(id, commitment, validatorProof);
-        verifyNewestMMRLeaf(
```
musnit, line 294: can we rather still do the verifyNewestMMRLeaf at the beginning, rather than within the  updateValidatorSet function? keeping up a check-effects-interactions pattern is still valuable for code readability

vgeddes: Yeah, I was thinking the same. The code path in the current draft just helped my thinking around the solution. Will see how I can reorganize :+1:

ethereum/contracts/BeefyLightClient.sol
vgeddes: The session length isn't immutable though. It can and probably will change through a runtime upgrade to Polkadot.

ethereum/contracts/BeefyLightClient.sol
vgeddes: > so, beefy is changing to increase the id every time now right?

Not really. Its just the following that is happening (according to Adrian):
1. The beefy validator set ID will increment on every new session, regardless of whether the validator root changes or not.
2. The session length on test networks is roughly 1 minutes. Not 4 hours like in production Polkadot

Change (1) is new and is part of Polkadot 0.9.17 / Lean Beefy. It's meant to ensure that each session has a unique validator set id, for high-level economic reasons (according to Adrian).

So the effect of (1) and (2) combined is that we see validator set ids incremented every minute (or roughly every 2 commitments on our E2E stack).

Don't think we need to change anything to improve performance. On production Polkadot and Kusama the IDs will only change every hour or more.



 On test polkadot networks such as our E2E stack its changing roughly 10 blocks to help tease out bugs. Is a new thing in Polkadot 0.9.17 / Lean BEEFY. However on production Polkadot the validator set id will only change every 2400 blocks as usual (regardless of the validator set root).



The validator set is still only changing every 2400 blocks on Polkadot. However

ethereum/contracts/BeefyLightClient.sol
musnit: ah cool yeh if the increment-every-session is now guaranteed by beefy, then this sounds like a safe assumption.

ethereum/.envrc-example
```diff
@@ -9,3 +9,6 @@ export BASIC_CHANNEL_PRINCIPAL=0x0000000000000000000000000000000000000042
 # Incentivized Channel
 # Default fee is 1 SnowDOT (12 decimal places)
 export INCENTIVIZED_CHANNEL_FEE=1000000000000000000
+
+# ParachainClient needs to be constructed with our parachain ID
+export PARACHAIN_ID=1000
```
alistair-singh, line 14: +1 this will get overridden from the Taskfile in snowbridge-infra.

vgeddes: Cool, sounds good :+1:

ethereum/contracts/BeefyClient.sol
```diff
@@ -0,0 +1,474 @@
...
+import "./utils/Bits.sol";
+import "./utils/Bitfield.sol";
+import "./utils/MMRProofVerification.sol";
+import "./ScaleCodec.sol";
+import "./utils/MerkleProof.sol";
+
+/**
+ * @title A entry contract for the BEEFY light client
+ */
+contract BeefyClient is AccessControl {
```
musnit, line 16: seems a bit of a waste to use AccessControl for just the initialization? can we just remove it and leave initialization open?

ethereum/contracts/BeefyClient.sol
```diff
@@ -0,0 +1,474 @@
...
+            "Sender address does not match original validation data"
+        );
+
+        // Verify that block wait period has passed
+        require(block.number >= data.blockNumber + BLOCK_WAIT_PERIOD, "Block wait period not over");
+
+        // Check that payload.leaf.block_number is > last_known_block_number;
+        require(commitment.blockNumber > latestBeefyBlock, "Commitment blocknumber is too old");
+
+        ValidatorSet memory vset = getValidatorSet(commitment.validatorSetId);
```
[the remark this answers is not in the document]
musnit: we could also have the BeefyClient inherit from a ValidatorSet contract instead of delegate to a ValidatorSet contract if the concern is related to multiple contract addresses/deployments rather than related to code style

ethereum/contracts/BeefyClient.sol
```diff
@@ -0,0 +1,474 @@
...
+        // Accept and save the commitment
+        validationData[nextID] = ValidationData(
+            msg.sender,
+            commitmentHash,
+            vset.id,
+            validatorClaimsBitfield,
+            block.number
+        );
+
+        emit CommitmentVerified(nextID, Phase.Initial, commitmentHash, msg.sender);
```
musnit, line 238: why choose to use the same event for both initial and final verification? seems a bit risky and may also not be ideal for consumers who only care about final commitments?

ethereum/contracts/BeefyClient.sol
```diff
@@ -0,0 +1,474 @@
...
+
+        bytes32 commitmentHash = verifyCommitment(data, commitment, proof);
+
+        latestMMRRoot = commitment.payload.mmrRootHash;
+        latestBeefyBlock = commitment.blockNumber;
+        emit NewMMRRoot(commitment.payload.mmrRootHash, commitment.blockNumber);
+
+        // Check if commitment signals an authority handover (new validator session)
+        if (commitment.validatorSetId == currentValidatorSet.id + 1) {
+            currentValidatorSet = nextValidatorSet;
```
musnit, line 289: the biggest change in the PR is of course this, ie, we no longer seem to need to check the MMR leaves to know what the next validator set ID/len/root should be, as this data is in the main commitment payload rather than in the MMR.

is this correct?

musnit: oh wait i see this has been moved to a separate function below?

musnit: will need to think about this flow further - especially in a scenario where there could be a malicious node running the initial commitment to overwrite validationData[id] at any time. maybe can discuss on a quick call

musnit: i suppose overall it may be a bit confusing with there being both nextID for validationData id as well as validatorset.id

i presume the former is used by the contracts only and the latter references the actual real validator set id?

ethereum/contracts/BeefyClient.sol
```diff
@@ -0,0 +1,474 @@
...
+        returns (bool)
+    {
+        return MMRProofVerification.verifyLeafProof(latestMMRRoot, leafHash, proof);
+    }
+
+    /**
+     * @notice Executed by the prover in order to begin the process of block
+     * acceptance by the light client
+     * @param commitmentHash contains the commitmentHash signed by the validator(s)
+     * @param validatorClaimsBitfield a bitfield containing a membership status of each
```
musnit, line 187: new param validatorSetID needs comment

ethereum/contracts/BeefyClient.sol
musnit: and so overall the flow looks like basically:

 - once complete is done, if there is a new validator set:
    - current becomes next, so next and current are both the same
    - it is now possible to only provide commitments for the new validatorset that do not have a handover
    - before any handover commitment, someone will need to call the new updateValidatorSet function to bring in the new next validator set
    - once that happens, new handover commitments are possible again and complete can be called for handover commitments again

musnit: this all seems fine and works, but thinking about if there's a way to make it less confusing.

one idea: maybe if we had separate functions like:
 - completeSignatureCommitment -> only used for commitments in current session, never does handover, reverts if  validatorSet is not in same session
 - goToNextSession -> only used for commitments in next session, always does handover in same tx

musnit: or maybe just use function overloading on completeSignatureCommitment, so each of the above 2 example seperate functions still have the same name, just different overload/implementation

ethereum/contracts/BeefyClient.sol
```diff
@@ -0,0 +1,474 @@
...
+     * validator who has claimed to have signed the commitmentHash
+     * @param validatorSignature the signature of one validator
+     * @param validatorPosition the position of the validator, index starting at 0
+     * @param validatorPublicKey the public key of the validator
+     * @param validatorPublicKeyMerkleProof proof required for validation of the public key in the validator merkle tree
+     */
+    function newSignatureCommitment(
+        bytes32 commitmentHash,
+        uint64 validatorSetID,
```
musnit, line 196: could this be just a boolean, like bool next true or false?

ethereum/contracts/BeefyClient.sol
musnit: the above suggestion i made to make the uint64 validatorSetID on the newSignatureCommitment function a boolean may also help simplify the confusion with the different ids?

ethereum/contracts/BeefyClient.sol
```diff
@@ -0,0 +1,474 @@
...
+        );
+
+        nextValidatorSet.id = leaf.nextAuthoritySetId;
+        nextValidatorSet.root = leaf.nextAuthoritySetRoot;
+        nextValidatorSet.length = leaf.nextAuthoritySetLen;
+    }
+
+    /* Private Functions */
+
+    function getValidatorSet(uint256 id) internal view returns (ValidatorSet memory) {
```
musnit, line 321: same suggestion as above - can we just use a boolean true/false for next vs current here?

ethereum/contracts/BeefyClient.sol
```diff
@@ -0,0 +1,474 @@
...
+        emit CommitmentVerified(id, Phase.Final, commitmentHash, msg.sender);
+    }
+
+    function updateValidatorSet(MMRLeaf calldata leaf, MMRProof calldata proof) public {
+        require(leaf.nextAuthoritySetId == nextValidatorSet.id + 1, "Leaf is invalid");
+
+        // Verify that the leaf suppied by the relayer is part of the MMR
+        bytes32 leafHash = keccak256(encodeMMRLeaf(leaf));
+        require(
+            MMRProofVerification.verifyLeafProof(latestMMRRoot, leafHash, proof),
```
musnit, line 310: you're calling MMRProofVerification.verifyLeafProof directly here, but using the verifyMMRLeafProof function elsewhere. should be consistent, either call directly everywhere or use the verifyMMRLeafProof function everywhere?

ethereum/contracts/utils/MMRProofVerification.sol
```diff
@@ -0,0 +1,48 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.5;
+
+struct MMRProof {
+    bytes32[] items;
+    uint64 order;
```
musnit, line 6: maybe add to comment to explain that order is a bitfield, since it's no longer in the variable name

ethereum/contracts/ParachainClient.sol
```diff
@@ -0,0 +1,99 @@
+// SPDX-License-Identifier: Apache-2.0
```
musnit, line 1: all looks great! nice cleanup :)

ethereum/contracts/ParachainClient.sol
```diff
@@ -0,0 +1,99 @@
...
+    constructor(BeefyClient _client, uint32 _parachainID) {
+        beefyClient = _client;
+        parachainID = _parachainID;
+        encodedParachainID = ScaleCodec.encode32(_parachainID);
+    }
+
+    function verifyCommitment(bytes32 commitment, Proof calldata proof)
+        external
+        view
+        returns (bool)
```
musnit, line 48: makes sense, just reminder to make sure we don't forget now that this no longer reverts on failure and that it is the responsibility of the consumer to check this.

maybe we should document this more strongly either in a comment or in README/solidity docs once we improve them

ethereum/contracts/BeefyClient.sol
vgeddes: > this all seems fine and works, but thinking about if there's a way to make it less confusing. not sure if i like the idea that the light client can exist in this "intermediary state" where currentValidatorSet === nextValidatorSet. may be better to ensure that all handovers happen atomically, without any intermediary state possible.

yeah, I agree. The handover needs some work :+1:

ethereum/contracts/BasicInboundChannel.sol
```diff
@@ -2,77 +2,62 @@
...
 contract BasicInboundChannel {
     uint256 public constant MAX_GAS_PER_MESSAGE = 100000;
     uint256 public constant GAS_BUFFER = 60000;
 
     uint64 public nonce;
 
-    BeefyLightClient public beefyLightClient;
+    ParachainClient public parachainClient;
+
+    struct MessageBundle {
```
musnit, line 15: Shifting to nonce per message bundle instead of nonce per message is a significant change (even though code-wise, looks pretty simple).

I think it makes sense and great that we've got that in now incidentally too, thanks!

Having said that, just want to think about/discuss the implications of the change a little more, as I haven't thought about it in a while - lets discuss it on our next call. Probably all good though.

Any other thoughts/comments from your side on this?

ethereum/contracts/BasicInboundChannel.sol
```diff
@@ -2,77 +2,62 @@
...
+        uint64 nonce;
+        Message[] messages;
+    }
 
     struct Message {
+        uint64 id;
```
musnit, line 21: These id's are not really used anywhere right, just for events/notification? I guess that's fine, just checking?

ethereum/contracts/BasicInboundChannel.sol
musnit: In terms of replay protection, our claim is now no longer "Your message will only be played once because each message nonce is increasing" but rather "Your message will only be played once because each bundle nonce is increasing AND we guarantee that your message will be included in EXACTLY one bundle, no more, no less"

Is that correct?

For the latter guarantee, it depends on new code changes to the parachain code. I have reviewed them and they look good, but can you also confirm that the new pallet code now guarantees that "your message will be included in EXACTLY one bundle, no more, no less", in all scenarios, irrespective of what interactions come into the channel pallets?

musnit: cc @alistair-singh for a second review on the pallet code to ensure the pallets provide this guarantee

ethereum/contracts/BeefyClient.sol
musnit: We should still have a step in the deployment runbook though to "manually make sure that nobody slipped a transaction in between yours" before using the bridge, though this shouldn't even matter as the bridge won't work anyway if someone slipped in a malicious transaction right, so will inevitably be detected before the bridge is used?

Is the above correct/would u agree?

ethereum/contracts/BasicInboundChannel.sol
vgeddes: yeah, its purely so that off-chain clients can track whether messages were processed successfully.

ethereum/contracts/BasicInboundChannel.sol
vgeddes: Will check :+1:

vgeddes:
@musnit pushed a new version of the Beefy Client contract. Its seems I was wrong in my previous understanding of LEAN beefy. Incoming commitments should always correspond to the current validator set recorded in the contract, so we don't need to optionally verify them against the next validator set.

This misunderstanding was due to a long-standing bug/misconception in our relay code, which was submitting MMR leafs which were too old. Essentially this line to convert block numbers to leaf index was incorrect: https://github.com/Snowfork/snowbridge/pull/614/commits/c4cbfc4857b8252935b4b2d5f906e07609b8473d#diff-b9769561fbce4da55796f378663d46d750def5fac20bdcc2f5a3d54c7151a24fR98

So by submitting the most recent leaf proven by the commitment, we can inform the light client of the next validator set change, satisfying commitment.validatorSetID = X, and leaf.NextValidatorSetId = X + 2. With that condition, we are able to do inform the light client of a session handover.

To understand better, notice the relationship between currentValidatorSetIDs and nextValidatorSetIDs in the following commitment stream:
```
Commitment { BlockNumber: 49 ValidatorSetID: 4}; Leaf { ParentNumber: 48, NextValidatorSetID: 5 }
Commitment { BlockNumber: 51 ValidatorSetID: 4}; Leaf { ParentNumber: 50, NextValidatorSetID: 6 }
Commitment { BlockNumber: 59 ValidatorSetID: 5}; Leaf { ParentNumber: 58, NextValidatorSetID: 6 }
Commitment { BlockNumber: 61 ValidatorSetID: 5}; Leaf { ParentNumber: 60, NextValidatorSetID: 7 }
Commitment { BlockNumber: 69 ValidatorSetID: 6}; Leaf { ParentNumber: 68, NextValidatorSetID: 7 }
Commitment { BlockNumber: 71 ValidatorSetID: 6}; Leaf { ParentNumber: 70, NextValidatorSetID: 8 }
Commitment { BlockNumber: 79 ValidatorSetID: 7}; Leaf { ParentNumber: 78, NextValidatorSetID: 8 }
```
Notice that Commitment.ValidatorSetIDs and nextValidatorSetIDS don't increment together on the same line. For commitment at block 61, its ValidatorSetID is 5, and the nextValidatorSet is 7. That's a commitment and leaf which we must submit to the light client to do a validator set handover.

ethereum/contracts/BasicInboundChannel.sol
vgeddes: Yeah is fine. If the app pallets (ETHApp, etc) submit a message to the outbound channel pallet, then its guaranteed to only exist in the next upcoming commitment. Upon creating a commitment, the channel's message queue buffer is converted into a message bundle, a commitment is inserted into the header, and then the message queue buffer is deleted.

ethereum/contracts/BeefyClient.sol
vgeddes: Changed a bit in my latest upload.

`BeefyClient.verifyMMRLeafProof` is basically a helper for external contracts (i.e ParachainClient) to verify a leaf proof against the latest MMR root.

However when submitting a commitment and leaf together, we want to verify the leaf against the mmr root in the bundled commitment, and so its easier to just call `MMRProofVerification.verifyLeafProof` directly, with some gas savings compared to calling the `BeefyClient.verifyMMRLeafProof` wrapper.

ethereum/contracts/BeefyClient.sol
vgeddes: obsolete

ethereum/contracts/BeefyClient.sol
vgeddes: obsolete.

ethereum/contracts/BeefyClient.sol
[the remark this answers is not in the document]
vgeddes: `getValidatorSet` no longer exists in the latest upload, so that simplifies things. Anyway I found it easier to reason about the validator set logic when its part of BeefyClient, rather than being scattered across multiple contracts.

Also simplifies deployment scripts.

ethereum/contracts/BasicInboundChannel.sol
musnit: thx!

ethereum/contracts/BeefyClient.sol
vgeddes: obsolete

ethereum/contracts/BeefyClient.sol
vgeddes: For now I've downgraded the initialization to just using `Ownable` for access control.

I think one foreseeable problem with leaving it open is that auditors will quickly pick it up, and we'll have to end up locking it down anyway.

Anyway, once we introduce governance/configuration we'll have to revisit access control again.

ethereum/contracts/BeefyClient.sol
vgeddes: Totally reworked in the latest code. Now a single `NewRequest` is emitted after initial verification, and nothing else is seems necessary.

ethereum/contracts/BeefyClient.sol
vgeddes: Updated all docstrings to be sync with latest code.

parachain/pallets/incentivized-channel/src/outbound/test.rs
```diff
@@ -168,10 +175,11 @@ fn test_submit() {
...
 
 		assert_ok!(IncentivizedOutboundChannel::submit(&who, target, &vec![0, 1, 2]));
-		assert_eq!(<Nonce<Test>>::get(), 1);
+		assert_eq!(<NextId<Test>>::get(), 1);
+		assert_eq!(<Nonce<Test>>::get(), 0);
 
-		assert_ok!(IncentivizedOutboundChannel::submit(&who, target, &vec![0, 1, 2]));
-		assert_eq!(<Nonce<Test>>::get(), 2);
+		run_to_block(2);
+		assert_eq!(<Nonce<Test>>::get(), 1);
```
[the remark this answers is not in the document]
vgeddes: NextId is only incremented when a new message is added, so the suggested change wouldn't apply.

ethereum/contracts/BeefyClient.sol
```diff
@@ -0,0 +1,478 @@
...
+     * @param bitfield a bitfield signalling which validators they claim have signed
+     * @param blockNumber the block number for this commitment
+     */
+    struct Request {
+        address sender;
+        uint64 validatorSetID;
+        bytes32 commitmentHash;
+        uint256[] bitfield;
+        uint256 blockNumber;
+        ValidatorSet vset;
```
musnit, line 103: looks unused?

ethereum/contracts/BeefyClient.sol
```diff
@@ -0,0 +1,478 @@
...
+    /**
+     * @dev A request is used to link initial and final submission of a commitment
+     * @param sender the sender of the initial transaction
+     * @param commitmentHash the hash of the commitment they are claiming has been signed
+     * @param bitfield a bitfield signalling which validators they claim have signed
+     * @param blockNumber the block number for this commitment
+     */
+    struct Request {
+        address sender;
+        uint64 validatorSetID;
```
musnit, line 99: also looks unused?

ethereum/contracts/BeefyClient.sol
musnit: is it? still looks like it could be a boolean rather?

ethereum/contracts/BeefyClient.sol
```diff
@@ -0,0 +1,478 @@
...
+
+    /**
+     * @dev Submit a commitment and leaf for final verification
+     * @param requestID identifier for the request generated by the initial submission
+     * @param commitment contains the full commitment that was used for the commitmentHash
+     * @param proof a struct containing the data needed to verify all validator signatures
+     * @param leaf an MMR leaf provable using the MMR root in the commitment payload
+     * @param leafProof an MMR leaf proof
+     */
+    function submitFinal(
```
musnit, line 264: should we use different names for these 2 functions? like complete and completeAndRollover ?

ethereum/contracts/BeefyClient.sol
vgeddes: removed in latest upload.

ethereum/contracts/BeefyClient.sol
vgeddes: its safer to use an ID instead of bool, so the verification code can discern between the following error cases and report properly:
1) The specified next validator set is no longer the currently-recorded next validator set.
2) The validator proof is incorrect.

musnit: solidity code should optimize for gas, not for error reporting :P

ethereum/contracts/BeefyClient.sol
musnit: yeh doesn't really matter

ethereum/contracts/BeefyClient.sol
musnit: right but this is only used offchain, so adds extra gas costs in production with no real on-chain use. maybe createFinalBitfield can take the vset or a boolean as an input instead of it being in Request?

ethereum/contracts/ParachainClient.sol
```diff
@@ -0,0 +1,99 @@
...
+import "./BeefyClient.sol";
+import "./utils/MerkleProof.sol";
+import "./ScaleCodec.sol";
+import "./utils/MMRProofVerification.sol";
+
+contract ParachainClient {
+    using ScaleCodec for uint32;
+
+    BeefyClient public immutable beefyClient;
+    uint32 public immutable parachainID;
```
alistair-singh, line 13: Having `parachainID` as immutable might be an issue if Parity decides to swap our para id. Maybe we should have a way to update this? Not really sure how that would effect validation process so maybe we need to stress that we need get a stable parachain id from them during the onboarding process?

musnit: i think governance upgrading to migrate the channels to a new parachain client should be fine to cover this risk

ethereum/contracts/BeefyClient.sol
vgeddes: Added https://linear.app/snowfork/issue/SNO-267/optimize-gas-cost-for-submitinitial to track gas optimizations.

PR #617, opened 2022-04-19
Remove ERC721 related code

alistair-singh:
Resolves: SNO-159


musnit:
@alistair-singh we should still keep the erc721 app - at least in a separate branch but ideally we move it to it's own separate repo and use it as an example app/tutorial for app developers.

alistair-singh:
> @alistair-singh we should still keep the erc721 app code atleast - maybe in a separate branch but ideally we move it to it's own separate repo and use it as an example app/tutorial for app developers.

I am currently working on moving the erc721 contracts, pallets and tests to another repo. The repo will reference `snowbridge` repo to get its dependencies so it should be minimal and suitable for a tutorial/example format.

vgeddes (approved):
LGTM! 🚀

PR #619, opened 2022-05-10
David/sno 257: Update the basic outbound channel to use BoundedVec

doubledup:
The parachain is compiling, but I haven't tested anything yet. Running the E2E tests now, will update if there are any failures.

### DIFF
--- a/parachain/pallets/basic-channel/src/outbound/mod.rs
+++ b/parachain/pallets/basic-channel/src/outbound/mod.rs
@@ -7,7 +7,6 @@ mod benchmarking;
 mod test;
 
 use codec::{Decode, Encode};
-// use codec::{Decode, Encode, Input, Compact};
 use ethabi::{self, Token};
 use frame_support::{
 	dispatch::DispatchResult,
@@ -35,14 +34,6 @@ pub struct Message {
 	/// Payload for target application.
 	payload: Vec<u8>,
 }
-
-// impl<T: Decode> Decode for BoundedVec<T, u64> {
-// 	fn decode<I: Input>(input: &mut I) -> Result<Self, Error> {
-// 		<Compact<u32>>::decode(input).and_then(move |Compact(len)| {
-// 			decode_vec_with_len(input, len as usize)
-// 		})
-// 	}
-// }
 
 pub use pallet::*;
 
--- a/parachain/pallets/basic-channel/src/outbound/mod.rs
+++ b/parachain/pallets/basic-channel/src/outbound/mod.rs
@@ -181,8 +181,8 @@ pub mod pallet {
 					nonce: *nonce,
 					payload: payload.to_vec(),
 				})
-				.map_err(|_unit| Error::<T>::QueueSizeLimitReached.into())
-				.map(|_unit| {
+				.map_err(|_| Error::<T>::QueueSizeLimitReached.into())
+				.map(|_| {
 					Self::deposit_event(Event::MessageAccepted(*nonce))
 				})
 			})
--- a/parachain/pallets/basic-channel/src/outbound/mod.rs
+++ b/parachain/pallets/basic-channel/src/outbound/mod.rs
@@ -203,7 +203,7 @@ pub mod pallet {
 					}
 
 					Err(()) => {
-						Err(DispatchError::Other("Outbound message channel full."))
+						Err(Error::<T>::QueueSizeLimitReached)
 					}
 				}
 			})
--- a/parachain/pallets/basic-channel/src/outbound/mod.rs
+++ b/parachain/pallets/basic-channel/src/outbound/mod.rs
@@ -192,13 +192,20 @@ pub mod pallet {
 					return Err(Error::<T>::Overflow.into())
 				}
 
-				<MessageQueue<T>>::append(Message {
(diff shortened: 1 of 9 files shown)


parachain/pallets/basic-channel/src/outbound/mod.rs
```diff
@@ -164,11 +164,6 @@ pub mod pallet {
 			let principal = Self::principal();
 			ensure!(principal.is_some(), Error::<T>::NotAuthorized,);
 			ensure!(*who == principal.unwrap(), Error::<T>::NotAuthorized,);
-			ensure!(
-				<MessageQueue<T>>::decode_len().unwrap_or(0) <
-					T::MaxMessagesPerCommit::get() as usize,
```
doubledup, line 169: With the queue being a BoundedVec, we don't need this validation anymore do we?

parachain/pallets/basic-channel/src/outbound/mod.rs
```diff
@@ -65,7 +65,7 @@ pub mod pallet {
 
 		/// Max number of messages per commitment
 		#[pallet::constant]
-		type MaxMessagesPerCommit: Get<u64>;
+		type MaxMessagesPerCommit: Get<u32>;
```
doubledup, line 68: ☝️ Because BoundedVec's [Decode implementation](https://crates.parity.io/frame_support/storage/bounded_vec/struct.BoundedVec.html#impl-Decode) requires the bound to be a u32.

parachain/pallets/basic-channel/src/outbound/mod.rs
```diff
@@ -181,18 +176,25 @@ pub mod pallet {
...
-				Self::deposit_event(Event::MessageAccepted(*nonce));
-				Ok(())
+				}) {
+					Ok(()) => {
+						Self::deposit_event(Event::MessageAccepted(*nonce));
+						Ok(())
+					}
+
+					Err(()) => {
+						Err(Error::<T>::QueueSizeLimitReached)
```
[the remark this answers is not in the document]
alistair-singh: Try:
```suggestion
						Err(Error::<T>::QueueSizeLimitReached.into())
```

parachain/pallets/basic-channel/src/outbound/mod.rs
```diff
@@ -181,18 +176,20 @@ pub mod pallet {
...
-				<MessageQueue<T>>::append(Message {
+				<MessageQueue<T>>::try_append(Message {
 					target,
 					nonce: *nonce,
 					payload: payload.to_vec(),
-				});
-				Self::deposit_event(Event::MessageAccepted(*nonce));
-				Ok(())
+				})
+				.map_err(|_unit| Err(Error::<T>::QueueSizeLimitReached))
```
doubledup, line 184: I used match initially, but this is shorter. Which style should we prefer when handling results?

parachain/pallets/basic-channel/src/outbound/mod.rs
alistair-singh: That is correct, `try_mutate` will only mutate if `Ok()` is returned. So it is safe to drop this `ensure!`.

https://docs.substrate.io/rustdocs/latest/src/frame_support/storage/mod.rs.html#178-179

parachain/pallets/basic-channel/src/outbound/mod.rs
alistair-singh: We do not have a hard and fast rule about `map` vs `match`. I think the `map` is more readable in this case.

parachain/pallets/basic-channel/src/outbound/mod.rs
```diff
@@ -181,18 +176,20 @@ pub mod pallet {
...
-				<MessageQueue<T>>::append(Message {
+				<MessageQueue<T>>::try_append(Message {
 					target,
 					nonce: *nonce,
 					payload: payload.to_vec(),
-				});
-				Self::deposit_event(Event::MessageAccepted(*nonce));
-				Ok(())
+				})
+				.map_err(|_unit| Error::<T>::QueueSizeLimitReached.into())
```
vgeddes, line 184: We use this convention for unused params:
```suggestion
				.map_err(|_| Error::<T>::QueueSizeLimitReached.into())
```

parachain/pallets/basic-channel/src/outbound/mod.rs
```diff
@@ -181,18 +176,20 @@ pub mod pallet {
...
+				.map(|_unit| {
```
doubledup, line 185:
```suggestion
				.map(|_| {
```

test/README.md
```diff
@@ -47,7 +47,7 @@ yarn install
 ### Polkadot
 
 * Clone the polkadot repository somewhere on your machine
-* Checkout commit `96d736ce75e8467862cd63037faa6d6e56122532`.
+* Checkout commit `f00a2772497aadddf75b8b4b475843ea0d910c48`.
```
vgeddes, line 50:
```suggestion
* Checkout branch `release-v0.9.19`.
```

alistair-singh (approved):

vgeddes (approved):
Looks great 🚀

PR #620, opened 2022-05-11
David/sno 260: Update the incentivized outbound channel to use BoundedVec

doubledup:


parachain/pallets/basic-channel/src/outbound/mod.rs
```diff
@@ -164,6 +165,11 @@ pub mod pallet {
 			let principal = Self::principal();
 			ensure!(principal.is_some(), Error::<T>::NotAuthorized,);
 			ensure!(*who == principal.unwrap(), Error::<T>::NotAuthorized,);
+			ensure!(
+				<MessageQueue<T>>::decode_len().unwrap_or(0) <
+					T::MaxMessagesPerCommit::get() as usize,
+				Error::<T>::QueueSizeLimitReached,
+			);
```
doubledup, line 172: Mentioned the need for this in my commit message, but effectively I'm adding this back as a preventative measure. If we don't check this and the queue is full, any mutations we add to the try_mutate closure before appending the message to the queue will occur, but the nonce and queue will be unaffected.

alistair-singh (approved):

PR #623, opened 2022-05-16
Beacon block sync

claravanstaden:
The beacon block contains an execution payload (the Eth1 header), and for this reason we'll send the whole beacon block to verify it, and then save the header in storage.


parachain/pallets/ethereum-beacon-client/src/merkleization.rs
```diff
@@ -0,0 +1,613 @@
...
+        },
+        attestation_2: SSZIndexedAttestation{
+            attesting_indices: attesting_indices2,
+            data: get_ssz_attestation_data(attester_slashing.attestation_2.data.clone())?,
+            signature: signature2,
+        },
+    })
+}
+
+pub fn get_ssz_checkpoint(checkpoint: Checkpoint) -> Result<SSZCheckpoint, MerkleizationError> {
```
vgeddes, line 203: Minor Nit: Its a convention to implement [TryInto](https://doc.rust-lang.org/std/convert/trait.TryInto.html) instead of creating ad hoc conversion functions.

But this is fine for now, just keep in mind for the future.

claravanstaden: Great, thanks for the tip! I'll do a refactor at some point.

parachain/pallets/ethereum-beacon-client/src/lib.rs
```diff
@@ -15,12 +17,8 @@ use sp_core::H256;
 use sp_io::hashing::sha2_256;
 use sp_runtime::RuntimeDebug;
 use sp_std::prelude::*;
-
-type Root = H256;
-type Domain = H256;
-type ValidatorIndex = u64;
-type ProofBranch = Vec<H256>;
-type ForkVersion = [u8; 4];
+use snowbridge_ethereum::Header as ExecutionHeader;
```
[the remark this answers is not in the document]
claravanstaden: Correct, I corrected this in a new branch where I added the new extrinsic, but let me fix it here too. 😄

PR #628, opened 2022-05-24
Updated logging

alistair-singh:
Resolves: SNO-275

### DIFF
--- a/relayer/relays/ethereum/parachain-writer.go
+++ b/relayer/relays/ethereum/parachain-writer.go
@@ -234,10 +234,10 @@ func (wr *ParachainWriter) queryImportedHeaderExists(hash types.H256) (bool, err
 	var headerOption types.OptionBytes
 	ok, err := wr.conn.API().RPC.State.GetStorageLatest(key, &headerOption)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


relayer/relays/ethereum/parachain-writer.go
```diff
@@ -228,16 +228,16 @@ func (wr *ParachainWriter) makeHeaderImportCall(header *chain.Header) (types.Cal
 func (wr *ParachainWriter) queryImportedHeaderExists(hash types.H256) (bool, error) {
 	key, err := types.CreateStorageKey(wr.conn.Metadata(), "EthereumLightClient", "Headers", hash[:], nil)
 	if err != nil {
-		return false, err
+		return false, fmt.Errorf("failed to create storeage key for hash %s with error: %s", hash.Hex(), err)
```
vgeddes, line 231: Just some Golang style guidelines for error messages:

1. Don't include redundant phrases such as "failed to" or "cannot".
2. Should generally use the `%w` verb when formatting error objects.

So an improved error message would look like this:
```golang
fmt.Errorf("create storage key for hash %s: %w", hash.Hex(), err)
```

Also, I recommend reading the "Wrapping errors with %w" section in https://go.dev/blog/go1.13-errors.

PR #629, opened 2022-05-25
Fix query header exists

alistair-singh:
Resolves: SNO-275

Changed the method which checks import status of headers to use `QueryStorageAtLatest`.
Changed error messages to be more standards compliant.

### DIFF
--- a/relayer/relays/ethereum/parachain-writer.go
+++ b/relayer/relays/ethereum/parachain-writer.go
@@ -231,14 +231,20 @@ func (wr *ParachainWriter) queryImportedHeaderExists(hash types.H256) (bool, err
 		return false, fmt.Errorf("create storeage key for hash %s: %w", hash.Hex(), err)
 	}
 
-	var headerOption types.OptionBytes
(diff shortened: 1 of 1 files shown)


relayer/relays/ethereum/parachain-writer.go
```diff
@@ -228,17 +234,23 @@ func (wr *ParachainWriter) makeHeaderImportCall(header *chain.Header) (types.Cal
 func (wr *ParachainWriter) queryImportedHeaderExists(hash types.H256) (bool, error) {
 	key, err := types.CreateStorageKey(wr.conn.Metadata(), "EthereumLightClient", "Headers", hash[:], nil)
 	if err != nil {
-		return false, fmt.Errorf("failed to create storeage key for hash %s with error: %s", hash.Hex(), err)
+		return false, fmt.Errorf("create storeage key for hash %s: %w", hash.Hex(), err)
```
vgeddes, line 237:
```suggestion
		return false, fmt.Errorf("create storage key for hash %s: %w", hash.Hex(), err)
```

vgeddes (approved):
Looks good!

PR #630, opened 2022-05-25
Rococo changes

alistair-singh:
Resolves: SNO-273

Added chainspecs
Changed currency from `DEV` to `SNO`


vgeddes:
Lets not commit the chainspec for staging, since that testnet is probably going to be deprecated.

PR #631, opened 2022-05-26
Improvements for channel protocols

vgeddes:
Several improvements to channel protocols, including:

1. Inbound channels on ethereum now receive opaque message proofs (SNO-272)
2. Incentivized outbound channel on parachain now rolls up per-message fees into a single field in the committed message bundle. This allows for greater flexibility in future when we introduce different fee-charging models, and also reduces gas usage on the ethereum side. Example of the new data structure for a committed bundle
  ```json
{
        "bundle": {
            "nonce": 1
            "fee": "10000000000000000",
            "messages": [
                {
                    "id": 0,
                    "target": "0xb8ea8cb425d85536b158d661da1ef0895bb92f1d"
                    "payload": "0x7e..0d"
                }
            ]
        }
}
```
4. Outbound channels on parachain no longer persist message bundles to offchain storage. This was always a dubious design decision, and an over-optimization. And since substrate RPCs to access offchain storage are marked unsafe (which we didn't know at the time), relayers can't query offchain storage from remote nodes. We'd have to provision our nodes to enable these unsafe RPCs, which is not very decentralized at all. So now the channels simply emit the bundles in pallet events.

One gotcha is that the relayer can't use GSRPC to read pallet events, since the library isn't equipped to do that. Instead I've developed a standalone rust binary `snowbridge-query-events` which queries for events. The relayer shells out to the binary to get the data it needs. This binary lives at `parachain/tools/query-events`.

### CODE REVIEW

File: parachain/pallets/incentivized-channel/src/outbound/mod.rs [hunk 1]
```diff
@@ -216,52 +222,49 @@ pub mod pallet {
 			let next_nonce = nonce.saturating_add(1);
 			<Nonce<T>>::put(next_nonce);
 
-			let bundle =
-				MessageBundle { nonce: next_nonce, messages: messages.clone().into_inner() };
+			let bundle: MessageBundleOf<T> = MessageBundle {
+				nonce: next_nonce,
+				fee: (messages.len() as u128).saturating_mul(Self::fee()),
```
musnit, line 227: hmm i'm just thinking about if anything has changed in terms of assumptions for how different messages depend on eachother. the goal of the design has always been to ensure that there is no dependency between messages, ie, if a message is broken or not dispatchable/callable or whatever, it does not affect other messages.

fees are fixed, and completely controlled and consumed by our channel logic with application-specific messages having no say/involvement in the fees, right? so this means we can roll them up and retain the no-dependency-between-messages goal? this all sounds good - will you just think about this yourself a bit more too and clarify/confirm 100% that this remains the case?

The other thing to keep in mind is that if we have different kind of channels in future that are not fixed-fee, and rather based on something like a fee market where applications can choose their fee, then this bundling interface won't be able to support that easily. Anyway, even though I mention that, it's not really an issue as would be dealt with in that later feature work anyway.

👍👍

vgeddes: Hey yeah, I did think about these considerations. Indeed one of primary motivators of this change was to enable support for more flexible fee models (fee markets, etc). So we can introduce fee models of varying complexity without having to upgrade our wire protocol every time. 

Firstly, when dealing with fees, we can take the analogy of stock market broker. The front office (equivalent to the outbound channel on the parachain), deals with fee charging according to market dynamics. The back office (equivalent to the inbound channel on ethereum) deals with settlement of rolled up fees to a bank account. The back office doesn't need to know that a specific message incurred a certain fee. In essence, back office only needs to know that a specific account needs to credited a certain amount. 

So in essence, I've decoupled the fee market from final fee settlement.

musnit: right though if we have a fee market, the ordering+inclusion of messages would be determined by the destination chain logic, not the source chain logic, so I don't think that would fit into this design

ethereum/contracts/BasicInboundChannel.sol
```diff
@@ -24,39 +27,34 @@ contract BasicInboundChannel {
...
+        require(bundle.sourceChannelID == sourceChannelID, "Invalid source channel");
+        require(bundle.nonce == nonce + 1, "Invalid nonce");
         require(
             gasleft() >= (bundle.messages.length * MAX_GAS_PER_MESSAGE) + GAS_BUFFER,
             "insufficient gas for delivery of all messages"
         );
 
-        processMessages(bundle);
+        dispatch(bundle);
+        nonce++;
```
alistair-singh, line 48: Is it not better practice to increment the nonce before calling other methods just in case of re-entrancy? `dispatch` calls `message.target` which could potentially be re-entrant and allow code malicious code to be executed before the nonce is incremented?

> ensure that all logic that changes state variables happen before ether is sent out of the contract

ethereum/contracts/IncentivizedInboundChannel.sol
```diff
@@ -28,64 +30,53 @@ contract IncentivizedInboundChannel is AccessControl {
...
+        require(bundle.nonce == nonce + 1, "Invalid nonce");
         require(
             gasleft() >= (bundle.messages.length * MAX_GAS_PER_MESSAGE) + GAS_BUFFER,
             "insufficient gas for delivery of all messages"
         );
 
-        processMessages(payable(msg.sender), bundle);
+        dispatch(bundle);
+        rewardController.handleReward(payable(msg.sender), bundle.fee);
+        nonce++;
```
alistair-singh, line 70: Ditto: re-entrancy comment from above.

parachain/pallets/incentivized-channel/src/outbound/mod.rs
```diff
@@ -26,24 +26,40 @@ use snowbridge_core::{types::AuxiliaryDigestItem, ChannelId};
...
 
 /// Wire-format for committed messages
-#[derive(Encode, Decode, Clone, PartialEq, RuntimeDebug, TypeInfo)]
-pub struct MessageBundle {
+#[derive(
+	Encode, Decode, CloneNoBound, PartialEqNoBound, RuntimeDebugNoBound, MaxEncodedLen, TypeInfo,
+)]
+#[scale_info(skip_type_params(M, N))]
+#[codec(mel_bound())]
+pub struct MessageBundle<M: Get<u32>, N: Get<u32>> {
```
alistair-singh, line 34: Why do we use both `skip_type_params` and `mel_bound`? My understanding of `mel_bound` is that it automatically skips all type params.

ethereum/contracts/IncentivizedInboundChannel.sol
vgeddes: Thanks! Fixed in latest upload.

parachain/pallets/incentivized-channel/src/outbound/mod.rs
vgeddes: Yeah its a bit complicated. They are used for different traits we are trying to derive. `#[scale_info(skip_type_params(M, N))]` to skip type bounds for `TypeInfo`, and `#[codec(mel_bound())]` to skip type bounds for `MaxEncodedLen`.

This answer it explains it some more: https://substrate.stackexchange.com/a/620/702

parachain/pallets/incentivized-channel/src/outbound/mod.rs
vgeddes: I know this discussion is getting a bit off-topic, but I'm keen to understand more about the details of the fee market you have in mind. :)

My understanding was that message inclusion & ordering (based on fee markets) would be determined by the source chain, since its in the best position to buffer messages, extract fees from users, and interact with Oracles if necessary. 

The destination chain could safely reorder the included messages I guess, but then again the source chain could do that too.

musnit: ordering on source chain is nice, but is always vulnerable to the risk that there is a gas price spike such that the fees banked are not enough to cover dispatch on the destination chain, and so messages won't be dispatched until the dispatch becomes affordable again with a gas price recovery.

this is acceptable and has graceful recovery in our current design for most real world scenarios, but a theoretical example in which gas price spikes up within a very short time window and never comes back down again would halt the bridge until some relayer accepts taking a loss to flush the channel.

a more theoretically pedantic design would need ordering&inclusion to occur on the destination chain to prevent the above edge case.

PR #632, opened 2022-05-26
Ethereum inbound channels should check identity of source channels

vgeddes:
Even though our channel pallets on the parachain are trusted, I can easily see an attack where a malicious relayer can submit a message bundle to the wrong channel, and it may actually end up being executed, and screw up the nonces in the process.

Right now this isn't possible as our current set of channels have different MessageBundle formats. But lets say we introduce another incentivised channel (call it channel C) which shares the same MessageBundle format as the existing incentivized channel (channel B). Then relayers are free to relay a message from B to C and the only thing preventing the bundle from being executed is the nonce check. But this nonce check won't always hold and so we should introduce an additional guard, at least for the sake of being explicit.

Since pallets don't have addresses, we can just assign numeric IDs to our current channel pallets.

Example of new message bundle format expected by the basic channel on ethereum:
```solidity
    struct MessageBundle {
        uint8 sourceChannelID; // <-- Must be checked before dispatching messages
        uint64 nonce;
        Message[] messages;
    }
```


ethereum/contracts/BasicInboundChannel.sol
```diff
@@ -7,11 +7,14 @@ contract BasicInboundChannel {
     uint256 public constant MAX_GAS_PER_MESSAGE = 100000;
     uint256 public constant GAS_BUFFER = 60000;
 
+    uint8 public constant sourceChannelID = 0;
```
musnit, line 10: maybe put in constructor and make immutable?

vgeddes: I thought about that but don't think its necessary. These channel IDs will never change and so leaving them as static constants should be fine.

vgeddes: sure, good idea :+1:

vgeddes:
Merging this into the parent branch #631

PR #635, opened 2022-05-31
Fix parsing of bitfields

vgeddes:
The beefy relayer calls out to the `BeefyClient.createFinalBitfield` contract method to retrieve a bitfield indicating the validators for which we need to include signatures and proofs. However the parsing of this bitfield had a bug at https://github.com/Snowfork/snowbridge/blob/1b5d8d5833a6ceb775d34f929404995088b6affe/relayer/relays/beefy/ethereum-writer.go#L274 where we truncate bits 65-256 of the bitfield, effectively ignoring those bits.

I've developed a new bitfield parser which replaces the old one.


relayer/relays/beefy/bitfield/bitfield.go
```diff
@@ -0,0 +1,42 @@
...
+func reverse(thing []byte) {
+	for i, j := 0, len(thing)-1; i < j; i, j = i+1, j-1 {
+		thing[i], thing[j] = thing[j], thing[i]
+	}
+}
+
+// New returns a Bitfield initialized from the Solidity representation of a Bitfield (an array of uint256). See below:
+// https://github.com/Snowfork/snowbridge/blob/18c6225b21782170156729d54a35404d876a2c7b/ethereum/contracts/utils/Bitfield.sol
+func New(input []*big.Int) Bitfield {
+	result := make(Bitfield, 256 * len(input))
```
alistair-singh, line 18: `uint256` is 256 bits wide (32 bytes). This line would allocate 256 bytes per `uint256`. Are we treating each byte as a bit which can only represent 1 or 0?

relayer/relays/beefy/bitfield/bitfield.go
```diff
@@ -0,0 +1,42 @@
...
+
+// Members returns the set bits in the bitfield
+func (b Bitfield) Members() []uint64 {
+	results := []uint64{}
+	for idx, bits := range b {
+		if bits == 0 {
+			continue
+		}
+		for i := 0; i < 8; i++ {
+			if bits & byte(1 << i) > 0 {
```
alistair-singh, line 36: This comment is linked to my comment above. This line means each byte does not represent a bit.

PR #637, opened 2022-06-02
SNO-304: Parachain -> Ethereum: bundle messages by account

doubledup:


parachain/pallets/basic-channel/src/outbound/mod.rs
```diff
@@ -30,14 +30,22 @@ pub use weights::WeightInfo;
...
 	#[codec(compact)]
 	nonce: u64,
-	messages: BoundedVec<Message<M>, N>,
+	messages: BoundedVec<EnqueuedMessage<AccountId, M>, N>,
 }
 
+#[derive(
+	Encode, Decode, CloneNoBound, PartialEqNoBound, RuntimeDebugNoBound, MaxEncodedLen, TypeInfo,
+)]
+#[scale_info(skip_type_params(AccountId, M))]
```
vgeddes, line 45: `<T as frame_system>::AccountId` likely already implements all those traits (Encode, Decode, etc), so you may not have to specify it in `skip_type_params`.

vgeddes: More info on those `mel_bound`, `skip_type_params` macros:
https://substrate.stackexchange.com/questions/1423/what-is-the-role-of-scale-infoskip-type-paramst
https://substrate.stackexchange.com/questions/1974/how-does-codecmel-bound-work/1977#1977

parachain/pallets/basic-channel/src/outbound/mod.rs
```diff
@@ -30,14 +30,22 @@ pub use weights::WeightInfo;
 )]
 #[scale_info(skip_type_params(M, N))]
 #[codec(mel_bound())]
-pub struct MessageBundle<M: Get<u32>, N: Get<u32>> {
+pub struct MessageBundle<M: Get<u32>, N: Get<u32>, AccountId> {
```
vgeddes, line 33: Can we have this order for the type parameters: AccountId, M, N

parachain/pallets/basic-channel/src/outbound/mod.rs
```diff
@@ -30,14 +30,22 @@ pub use weights::WeightInfo;
...
 	source_channel_id: u8,
+	account: AccountId,
 	/// Unique nonce for to prevent replaying bundles
 	#[codec(compact)]
 	nonce: u64,
-	messages: BoundedVec<Message<M>, N>,
+	messages: BoundedVec<EnqueuedMessage<AccountId, M>, N>,
```
vgeddes, line 39: Nah, I didn't intend for `EnqueuedMessage` to be used here. Should only be used for the `MessageQueue` storage item (which I see you've updated correctly), so that we can identify the owners of the messages in the queue.

parachain/pallets/basic-channel/src/outbound/mod.rs
```diff
@@ -30,14 +30,22 @@ pub use weights::WeightInfo;
...
 }
 
+#[derive(
+	Encode, Decode, CloneNoBound, PartialEqNoBound, RuntimeDebugNoBound, MaxEncodedLen, TypeInfo,
+)]
+#[scale_info(skip_type_params(AccountId, M))]
+#[codec(mel_bound())]
```
vgeddes, line 46: I think this may be needed to ensure that `AccountId` is substituted with something that implements `MaxEncodedLen`.
```suggestion
#[codec(mel_bound(AccountId: MaxEncodedLen))]
```

parachain/pallets/basic-channel/src/outbound/mod.rs
```diff
@@ -118,15 +126,16 @@ pub mod pallet {
...
 
 	/// Fee for accepting a message
 	#[pallet::storage]
 	#[pallet::getter(fn principal)]
 	pub type Principal<T: Config> = StorageValue<_, Option<T::AccountId>, ValueQuery>;
 
+	// Need a nonce for each account (message bundle) now
 	#[pallet::storage]
-	pub type Nonce<T: Config> = StorageValue<_, u64, ValueQuery>;
+	pub type Nonces<T: Config> = StorageMap<_, Identity, T::AccountId, u64, ValueQuery>;
```
vgeddes, line 138: Looks good 👍

parachain/pallets/basic-channel/src/outbound/mod.rs
```diff
@@ -118,15 +126,16 @@ pub mod pallet {
 	/// Messages waiting to be committed.
 	#[pallet::storage]
 	pub(super) type MessageQueue<T: Config> =
-		StorageValue<_, BoundedVec<MessageOf<T>, T::MaxMessagesPerCommit>, ValueQuery>;
+		StorageValue<_, BoundedVec<EnqueuedMessageOf<T>, T::MaxMessagesPerCommit>, ValueQuery>;
```
vgeddes, line 129: Looks good :+1:

parachain/pallets/basic-channel/src/outbound/mod.rs
```diff
@@ -202,52 +214,65 @@ pub mod pallet {
...
 			Ok(())
 		}
 
 		fn commit() -> Weight {
+			let who = Self::principal().unwrap();
+			let w2 = who.clone();
+
+			// for every account id in the message queues map, create an Eth ABI-encoded message bundle
+			// these encoded bundles will be the leaves of a merkle tree
 			let messages = <MessageQueue<T>>::take();
```
vgeddes, line 236: After reading the queued messages, you could convert it into an interim (BTreeMap)[https://crates.parity.io/sp_std/collections/btree_map/struct.BTreeMap.html], where the keys are the AccountId in the enqueued messages, and the values are all the collected messages for that that account.

Then that interim BTreeMap gets converted into a merkle tree.

parachain/pallets/basic-channel/src/outbound/mod.rs
```diff
@@ -263,8 +288,8 @@ pub mod pallet {
 			<T as Config>::Hashing::hash(&commitment)
 		}
 
-		fn average_payload_size(messages: &[MessageOf<T>]) -> u32 {
-			let sum: usize = messages.iter().fold(0, |acc, x| acc + x.payload.len());
+		fn average_payload_size(messages: &[EnqueuedMessage<T::AccountId, T::MaxMessagePayloadSize>]) -> u32 {
```
vgeddes, line 291: Just ignore or stub out `average_payload_size` for now. Its not on the critical path.

parachain/pallets/basic-channel/src/outbound/mod.rs
vgeddes: May potentially also need to add the following trait bounds for `AccountId`:

```
pub struct MessageBundle<AccountId, ...> where AccountId: Encode, Decode, CloneNoBound, PartialEqNoBound, RuntimeDebugNoBound, MaxEncodedLen, TypeInfo
```

doubledup: Just needed to add the Debug trait instead of RuntimeDebug

parachain/pallets/basic-channel/src/outbound/mod.rs
```diff
@@ -215,56 +269,121 @@ pub mod pallet {
...
-				messages: messages.clone(),
-			};
+			let message_bundles = Self::make_message_bundles(message_queue);
+			let eth_message_bundles: Vec<Vec<u8>> = message_bundles
+				.clone()
+				.into_iter()
+				.map(|bundle| ethabi::encode(&vec![bundle.into()]))
+				.collect();
+
+			// TODO: Maybe undo the O type param in merkle_root, since we can convert between H256
```
vgeddes, line 296: Yes please. Internally in `merkle_proof.rs` Its fine to hardcode `[u8; 32]` as the hash type and remove the `O` parameter. It will never change.

vgeddes (commented):
Looks very good! Nice progress so far 🚀

parachain/pallets/basic-channel/src/outbound/mod.rs
```diff
@@ -215,56 +271,128 @@ pub mod pallet {
...
+				hash: commitment_hash,
+				data: message_bundles.clone(),
+			});
+
+			// Persist ABI-encoded leaves to off-chain storage
+			let tokens = eth_message_bundles
+				.into_iter()
+				.map(|bundle| Token::Bytes(bundle))
+				.collect::<Vec<Token>>();
+			set(commitment_hash.as_bytes(), &ethabi::encode(&tokens));
```
vgeddes, line 322: While the individual leaves should be ABI-encoded, the container in which they are serialized should be SCALE-encoded for efficiency.

For example, you could persist this SCALE-encoded container to offchain storage. Each individual `Vec<u8>` still represents an ABI-encoded leaf.
```rust
#[derive(Encode, Decode)]
struct StoredLeaves(Vec<Vec<u8>>)
```

vgeddes, line 322: While the individual leaves should be ABI-encoded, the container in which they are serialized should be SCALE-encoded for efficiency.

For example, you could persist this SCALE-encoded container to offchain storage. Each individual `Vec<u8>` still represents an ABI-encoded leaf.
```rust
#[derive(Encode, Decode)]
struct StoredLeaves(Vec<Vec<u8>>)
```

parachain/pallets/basic-channel/src/outbound/mod.rs
```diff
@@ -215,56 +269,121 @@ pub mod pallet {
...
+			// 	} else {
+			// 		let mut messages = BoundedVec::default();
+			// 		messages.try_push(message);
+			// 		messages_per_account.insert(account, messages);
+			// 	}
+			// }
+
+			let message_bundles = account_message_map
+				.into_iter()
+				.map(|(account, messages)| {
```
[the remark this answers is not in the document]
doubledup: Perfect, also resolves the TODO above.

parachain/pallets/basic-channel/src/outbound/mod.rs
```diff
@@ -9,35 +9,80 @@ mod test;
...
+		])
+	}
+}
+
+impl<AccountId, M: Get<u32>, N: Get<u32>> AsRef<[u8]> for MessageBundle<AccountId, M, N>
+where
+	AccountId: Encode + Decode + Clone + PartialEq + Debug + MaxEncodedLen + TypeInfo,
+{
+	fn as_ref(&self) -> &[u8] {
+		&[0]
```
vgeddes, line 69: Looks like this isn't being used anywhere. Can it be removed?

parachain/pallets/basic-channel/src/outbound/mod.rs
```diff
@@ -215,56 +246,101 @@ pub mod pallet {
...
+		/// - Store the commitment hash on the parachain for the Ethereum light client to query.
+		/// - Emit an event with the commitment hash and SCALE-encoded message bundles for a
+		/// relayer to read.
+		/// - Persist the ethabi-encoded message bundles to off-chain storage.
 		fn commit() -> Weight {
-			let messages = <MessageQueue<T>>::take();
-			if messages.is_empty() {
+			// TODO: consider using mutate_exists here. If some part of emitting message bundles
+			// fails, we don't want the MessageQueue to be empty.
+			let message_queue = <MessageQueue<T>>::take();
```
doubledup, line 260: I'm still concerned about this, but the only failure case I can see is nonce overflow when creating the message bundles. Is this actually something we should be concerned about?

parachain/pallets/basic-channel/src/outbound/mod.rs
```diff
@@ -215,56 +246,101 @@ pub mod pallet {
...
+					messages_for_accounts
+				},
+			);
+
+			let mut message_bundles: Vec<MessageBundleOf<T>> = Vec::new();
+			for (account, messages) in account_message_map {
+				let next_nonce = <Nonces<T>>::mutate(&account, |nonce| {
+					// TODO: is this the right addition to use? u64 is large, but this will make us
+					// reuse nonces when we hit u64::MAX
+					*nonce = nonce.saturating_add(1);
```
doubledup, line 325: The only better option I'm seeing is checked_add and bubbling the error up to `commit`. Is that necessary here?

parachain/pallets/basic-channel/src/outbound/test.rs
```diff
@@ -97,38 +91,40 @@ pub fn new_tester() -> sp_io::TestExternalities {
 
 fn run_to_block(n: u64) {
 	while System::block_number() < n {
-			System::set_block_number(System::block_number() + 1);
-			BasicOutboundChannel::on_initialize(System::block_number());
+		System::set_block_number(System::block_number() + 1);
+		// TODO: is this necessary for the test? It's mentioned in the unit testing docs:
+		// https://docs.substrate.io/main-docs/test/unit-testing/
+		System::on_initialize(System::block_number());
```
doubledup, line 97: The tests run fine without this call, and it looks like this implementation in `System` doesn't do anything. Is it worth keeping it anyway?

parachain/pallets/basic-channel/merkle-proof/src/lib.rs
```diff
@@ -0,0 +1,792 @@
+// This file is part of Substrate.
+
+// Copyright (C) 2021-2022 Parity Technologies (UK) Ltd.
+// SPDX-License-Identifier: Apache-2.0
```
claravanstaden, line 4: Wondering if this should be here? Did we copy this file from Substrate?

Also asking because if it is, perhaps it's not necessary to review? 🤔

parachain/pallets/basic-channel/primitives/Cargo.toml
```diff
@@ -0,0 +1,18 @@
+[package]
+name = "snowbridge-basic-channel-primitives"
```
claravanstaden, line 2: I'm wondering if this shouldn't go in `parachain/primitives` to match the other code's structure?

parachain/pallets/basic-channel/rpc/Cargo.toml
```diff
@@ -0,0 +1,24 @@
+[package]
+name = "snowbridge-basic-channel-rpc"
```
claravanstaden, line 2: For my understanding, how can this crate work with `std`?

parachain/pallets/basic-channel/rpc/src/lib.rs
```diff
@@ -0,0 +1,44 @@
...
+pub trait BasicChannelApi {
+	#[rpc(name = "basicChannel_outbound_getMerkleProof")]
+	fn get_merkle_proof(&self, commitment_hash: H256, leaf_index: u64) -> Result<Vec<u8>>;
+}
+
+impl BasicChannelApi for BasicChannel {
+	fn get_merkle_proof(&self, commitment_hash: H256, leaf_index: u64) -> Result<Vec<u8>> {
+		let oci_mem = StorageValueRef::persistent(&commitment_hash.as_bytes());
+
+		if let Ok(Some(StoredLeaves(leaves))) = oci_mem.get::<StoredLeaves>() {
```
claravanstaden, line 28: I guess you could also replace this with a `match`?

parachain/pallets/basic-channel/src/outbound/test.rs
```diff
@@ -138,49 +134,47 @@ fn test_submit_exceeds_queue_limit() {
...
-	});
+		assert_eq!(<NextId<Test>>::get(), 1);
+		assert_eq!(<Nonces<Test>>::get(who), 1);
+		assert_eq!(<MessageQueue<Test>>::get().len(), 0);
+	})
 }
 
 #[test]
-fn test_set_principal() {
+fn test_commit_multi_user() {
```
claravanstaden, line 165: Cool test, this helps me understand what is the outcome of the work. 🥇

parachain/runtime/snowbase/Cargo.toml
```diff
@@ -72,8 +71,8 @@ runtime-primitives = { path = "../../primitives/runtime", default-features = fal
 
 snowbridge-xcm-support = { path = "../../primitives/xcm-support", default-features = false }
 snowbridge-asset-registry = { path = "../../pallets/asset-registry", default-features = false }
-basic-channel = { path = "../../pallets/basic-channel", package = "snowbridge-basic-channel", default-features = false }
-incentivized-channel = { path = "../../pallets/incentivized-channel", package = "snowbridge-incentivized-channel", default-features = false }
+snowbridge-basic-channel = { path = "../../pallets/basic-channel", default-features = false }
```
claravanstaden, line 74: Should we prefix the other pallets with `snowbridge` too? Checking what we want the convention to be.

claravanstaden (approved):
Very good! 👏

parachain/pallets/basic-channel/merkle-proof/src/lib.rs
doubledup: We did 😬 It's from the `beefy-merkle-tree` crate in FRAME's beefy-mmr pallet, but with beefy-mmr's `Hash` replaced with the equivalent `sp_core::H256` (also a 32 byte array) and the `Hasher` trait replaced with `sp_runtime::traits::Hash` (a more general form of `Hasher`).

Come to think of it, the `mmr` pallet looks like it _mostly_ does what we need here. A big difference is that the `mmr` pallet runs on every block, while our channel only runs every few blocks, determined by the `interval`.

@vgeddes What do you think of trying to use the `mmr` pallet here? It looks to me like it doesn't quite fit, but it might be close enough for us to contribute a change to the pallet.

claravanstaden: Thanks for explaining! 👍

parachain/pallets/basic-channel/rpc/src/lib.rs
doubledup: We could do, but we're only matching the one pattern, which is the use-case for `if let ... else`. If we had a second pattern (or more) we were interested in, we'd definitely prefer `match`.

parachain/runtime/snowbase/Cargo.toml
doubledup: I'd like to, but it's not a strong preference. It helps distinguish our pallets from others at a glance. I've made the change to the incentivized channel pallet already, but there are several others that don't have the prefix and only a few that do.

@vgeddes @alistair-singh Do you think it's worth having the `snowbridge-` prefix on our pallets? Is there an existing convention for when we use it?

vgeddes: I'd say we should keep the snowbridge prefix. I've been bitten before, when a local crate in the monorepo clashed with a public crate on https://crates.io/, causing compilation failures.

parachain/pallets/basic-channel/rpc/Cargo.toml
vgeddes: Indeed, yeah, thats correct :+1:

parachain/pallets/basic-channel/src/outbound/test.rs
vgeddes: yeah lets keep it.

alistair-singh (commented):
Looks good!

ethereum/contracts/BasicInboundChannelV2.sol
```diff
@@ -0,0 +1,66 @@
+// SPDX-License-Identifier: Apache-2.0
```
vgeddes, line 1: Should overwrite `BasicInboundChannel.sol` with this file, rather than naming it `..V2.sol`.

relayer/relays/parachain/beefy-listener.go
```diff
@@ -561,6 +563,17 @@ func (li *BeefyListener) scanForCommitments(
...
+		decodedAccount, err := hex.DecodeString(li.config.Account)
+		if err != nil {
+			return nil, fmt.Errorf("could not decode channel account id: %v", err)
+		} else if len(decodedAccount) != 32 {
+			// The conversion below will panic if decodedAccount has
+			// fewer than 32 bytes.
+			// We expect exactly 32 bytes.
+			return nil, fmt.Errorf("account id was not 32 bytes long: %v", decodedAccount)
+		}
+		accountAsArray := *(*[32]byte)(decodedAccount)
```
vgeddes, line 575: Ah, nice! A new feature I didn't know about. I see now that in Go 1.17 slices can be converted into arrays.

relayer/relays/parachain/beefy-listener.go
```diff
@@ -561,6 +563,17 @@ func (li *BeefyListener) scanForCommitments(
 			"blockNumber": currentBlockNumber,
 		}).Debug("Checking header")
 
+		decodedAccount, err := hex.DecodeString(li.config.Account)
+		if err != nil {
+			return nil, fmt.Errorf("could not decode channel account id: %v", err)
```
vgeddes, line 568: By convention, we don't include phrases such as "could not", "failed to", or "error while" in error messages. As such phrases end up being repeated redundantly in the final top-level wrapped error message".

Should generally also use the "%w" flag for formatting errors. See https://golangbyexample.com/wrapping-and-unwrapping-error-golang/

```suggestion
			return nil, fmt.Errorf("decode account id: %w", err)
```

relayer/relays/parachain/beefy-listener.go
```diff
@@ -602,18 +617,53 @@ func (li *BeefyListener) scanForCommitments(
...
+				err = api.Client.Call(&proofHex, "basicOutboundChannel_getMerkleProof", digestItem.AsCommitment.Hash.Hex(), bundleIndex)
+				if err != nil {
+					return nil, fmt.Errorf("call rpc: %w", err)
+				}
+
+				err = types.DecodeFromHexString(proofHex, &basicChannelBundleProof)
+				if err != nil {
+					return nil, fmt.Errorf("decode: %w", err)
+				}
+
```
vgeddes, line 654: Lets factor this logic into a new subroutine, as `scanForCommitments` is already quite complicated.

relayer/relays/parachain/beefy-listener.go
```diff
@@ -602,18 +617,53 @@ func (li *BeefyListener) scanForCommitments(
...
+				// Fetch Merkle proof for this bundle
+				api, err := gsrpc.NewSubstrateAPI(li.config.Parachain.Endpoint)
+				if err != nil {
+					return nil, fmt.Errorf("create client: %w", err)
+				}
+
+				var proofHex string
+				err = api.Client.Call(&proofHex, "basicOutboundChannel_getMerkleProof", digestItem.AsCommitment.Hash.Hex(), bundleIndex)
```
vgeddes, line 645: An initialized api already exists on the listener object:
```suggestion
				err = li.parachainConnection.API().Client.Call(&proofHex, "basicOutboundChannel_getMerkleProof", digestItem.AsCommitment.Hash.Hex(), bundleIndex)
```

relayer/relays/parachain/channels.go
```diff
@@ -19,13 +19,15 @@ func (b BasicOutboundChannelMessageBundle) IntoInboundMessageBundle() basic.Basi
...
 		SourceChannelID: b.SourceChannelID,
+		Account:         b.Account,
 		Nonce:           (*big.Int)(&b.Nonce).Uint64(),
 		Messages:        messages,
 	}
 }
 
 type BasicOutboundChannelMessageBundle struct {
 	SourceChannelID uint8
+	Account         [32]byte
```
vgeddes, line 30: Nice!

vgeddes (commented):
Implementation looks good so far! 🚀

relayer/relays/parachain/beefy-listener.go
```diff
@@ -561,6 +563,17 @@ func (li *BeefyListener) scanForCommitments(
 			"blockNumber": currentBlockNumber,
 		}).Debug("Checking header")
 
+		decodedAccount, err := hex.DecodeString(li.config.Account)
```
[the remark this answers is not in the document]
doubledup: I've moved the decoding itself to a getter method on the config type, but also moved the call to that method to `Start`, where the result is stored on the listener struct. Does it make more sense to remove the getter and put the decoding directly in start?

relayer/relays/parachain/ethereum-writer.go
```diff
@@ -283,6 +296,26 @@ func (wr *EthereumWriter) WriteBasicChannel(
...
+func generateHashSides(commitmentProof *MerkleProof) ([]bool, error) {
+	pos := commitmentProof.LeafIndex
+	width := commitmentProof.NumberOfLeaves
+
+	if pos < width {
+		return nil, fmt.Errorf("leaf position %v is too high in proof with %v leaves", pos, width)
+	}
+
+	// TODO: this float casting is lossy, find a base 2 log function that operates on uint64 or assert that the width doesn't exceed 2**63-1
+	numSides := (int)(math.Ceil(math.Log2(float64(width))))
```
alistair-singh, line 308: You could probably try something like `math/bits` `LeadingZeros64` function. This lots like you are trying to find the most significant set bit.

https://go.dev/play/p/WaW0jcHilXt

relayer/relays/parachain/beefy-listener.go
```diff
@@ -34,6 +35,7 @@ type BeefyListener struct {
 	paraID              uint32
 	tasks               chan<- *Task
 	eventQueryClient    QueryClient
+	AccountID           *[32]byte
```
vgeddes, line 38: Minor nit: Prefer we store the array here and not a pointer to the array:
```suggestion
	AccountID           [32]byte
```

relayer/relays/parachain/beefy-listener.go
```diff
@@ -674,6 +703,37 @@ func (li *BeefyListener) scanForCommitments(
...
+func fetchBundleProof(
+	basicChannelBundleProof *MerkleProof,
+	api *gsrpc.SubstrateAPI,
+	digestItem AuxiliaryDigestItem,
+	bundleIndex int,
+) error {
+	var proofHex string
+	err := api.Client.Call(&proofHex, "basicOutboundChannel_getMerkleProof", digestItem.AsCommitment.Hash.Hex(), bundleIndex)
+	if err != nil {
+		return fmt.Errorf("call rpc: %w", err)
```
vgeddes: Should also log the call params here:

```suggestion
		return fmt.Errorf("call rpc basicOutboundChannel_getMerkleProof(%v, %v): %w", digestItem.AsCommitment.Hash.Hex(), bundleIndex , err)
```

relayer/relays/parachain/beefy-listener.go
```diff
@@ -674,6 +703,37 @@ func (li *BeefyListener) scanForCommitments(
...
+	}
+
+	err = types.DecodeFromHexString(proofHex, basicChannelBundleProof)
+	if err != nil {
+		return fmt.Errorf("decode: %w", err)
```
vgeddes:
```suggestion
		return fmt.Errorf("decode merkle proof: %w", err)
```

ethereum/test/test_basic_inbound_channel.js
```diff
@@ -45,14 +46,16 @@ describe("BasicInboundChannel", function () {
     });
 
     it("should accept a valid commitment and dispatch messages", async function () {
-      const nonceBeforeSubmit = BigNumber(await this.channel.nonce());
+      const nonceBeforeSubmit = BigNumber(await this.channel.nonces(submitInput.params.bundle.account));
```
doubledup, line 49: Is it worth extending this test for multiple accounts? We're already using 2 accounts in the E2E test

doubledup: Leaving this to a later PR, the E2E test should be enough for now

relayer/relays/parachain/ethereum-writer.go
doubledup: That's clever! 😮 We could use the index of the most significant bit, assuming the least significant bit is at index 0. We'd need to add the bitwise-and of the smaller bits to account for trees that aren't full, like we do with `math.Ceil`. But then we sidestep casting.

doubledup: Looks like res1 in the example should be `res1 := bits.LeadingZeros64(n - 1)`, but otherwise perfect. I'll put this in, thanks! 🙏

test/test/bootstrap.js
```diff
@@ -29,25 +29,30 @@ describe('Bridge', function () {
       const ethAccount = ethClient.accounts[1];
 
       const beforeEthBalance = await ethClient.getDotBalance(ethAccount);
-      const beforeSubBalance = await subClient.queryAccountBalance(polkadotSenderSS58);
+      const beforeSubBalanceAlice = await subClient.queryAccountBalance(polkadotSenderSS58Alice);
+      const beforeSubBalanceBob = await subClient.queryAccountBalance(polkadotSenderSS58Bob);
 
       // lock DOT using basic channel
       await subClient.lockDOT(subClient.alice, ethAccount, amount.toFixed(), ChannelId.BASIC)
+      await subClient.lockDOT(subClient.bob, ethAccount, amount.toFixed(), ChannelId.BASIC)
```
vgeddes, line 37: A couple of things:

1. There's no guarantee the lockDOT extrinsics for alice and bob will be executed in the same block, and so there may still only be one account per commitment.

I would use the trick with sudo to guarantee we get at least 2 accounts per commitments: `Sudo.sudo(Utility.batch([utility.dispatchAs(Alice, dotApp.lock), ...]))`

Polkadot-JS api docs for utility pallet at https://polkadot.js.org/docs/substrate/extrinsics/#utility

2. I'd feel safer testing with more than 2 accounts. In order to exercise all the proof verification paths a bit more. Can we try 6 or more accounts? These builtin accounts would be: `Alice`, `Bob`, `Charlie`, `Dave`, `Eve`, and [so on](https://github.com/Snowfork/snowbridge/blob/f719545fe8d8cf9b5e0665b690c73077ac49258c/parachain/src/chain_spec/snowbase.rs#L38-L43).

3. Lets rather create a separate test in a new file. This one is purely meant for bootstrapping the bridge (and incidentally testing the basic channel)

test/scripts/start-services.sh
```diff
@@ -13,6 +13,8 @@ infura_endpoint_ws="${ETH_WS_ENDPOINT:-ws://localhost:8546}/${INFURA_PROJECT_ID:
 parachain_relay_eth_key="${PARACHAIN_RELAY_ETH_KEY:-0x8013383de6e5a891e7754ae1ef5a21e7661f1fe67cd47ca8ebf4acd6de66879a}"
 beefy_relay_eth_key="${BEEFY_RELAY_ETH_KEY:-0x935b65c833ced92c43ef9de6bff30703d941bd92a2637cb00cfad389f5862109}"
 
+account_ids="${ACCOUNT_IDS:-0xd43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d,0x8eaf04151687736326c9fea17e25fc5287613693c912909cb226aa4794f26a48}"
```
vgeddes, line 16: Please add a comment explaining what these accounts are for.

test/config/parachain-relay.json
```diff
@@ -14,7 +14,8 @@
             "BasicInboundChannel": null,
             "IncentivizedInboundChannel": null
         },
-        "beefy-activation-block": 0
+        "beefy-activation-block": 0,
+        "accountIDs": []
```
vgeddes, line 18:
```suggestion
        "accounts": []
```

parachain/pallets/basic-channel/merkle-proof/src/lib.rs
```diff
@@ -0,0 +1,791 @@
+// This file is part of Substrate.
+
+// Copyright (C) 2021-2022 Parity Technologies (UK) Ltd.
```
vgeddes, line 3: I think we need to attribute this code in a `NOTICE` file, as per https://opensource.stackexchange.com/questions/7872/how-to-properly-assemble-notice-file-for-new-software-under-apache-license-2-0. Can you check?

ethereum/test/test_basic_inbound_channel.js
vgeddes: Sounds fine. Though see my comments further down about the E2E tests.

.gitignore
```diff
@@ -8,4 +8,7 @@ workspace.code-workspace
 node_modules/
 
 #Intellij project file
-.idea
\ No newline at end of file
+.idea
```
vgeddes, line 11: Nit: https://sebastiandedeyne.com/setting-up-a-global-gitignore-file/

parachain/pallets/basic-channel/src/outbound/mod.rs
doubledup: I ended up SCALE-encoding the plain `Vec<Vec<u8>>`. Do we need the `StoredLeaves` wrapper?

parachain/pallets/basic-channel/src/outbound/mod.rs
```diff
@@ -6,38 +6,85 @@ mod benchmarking;
...
 
+impl<AccountId, M: Get<u32>, N: Get<u32>> Into<Token> for MessageBundle<AccountId, M, N>
+where
+	AccountId: Encode + Decode + Clone + PartialEq + Debug + MaxEncodedLen + TypeInfo,
+{
+	fn into(self) -> Token {
+		Token::Tuple(vec![
+			Token::Uint(self.source_channel_id.into()),
+			// TODO: check the AccountId encoding expectation in the Ethereum contract
+			Token::Uint(self.account.encode().as_slice().into()),
```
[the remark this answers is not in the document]
vgeddes: Ah yeah, oops, its actually `AsRef<[u8]>`. Using that sounds good :+1:

parachain/pallets/basic-channel/src/outbound/mod.rs
vgeddes: We can't really handle a nonce overflow since `commit` should never return errors. `commit` is called in an `on_initialize` hook which itself can never return errors either.

I think we need to just assume that nonce will never overflow. Indeed, for a user to encounter that case, they would need spend absolutely huge amounts on transaction fees to throw 2^64 messages through the bridge.

Still not an ideal situation for us though, but can't think of a better approach.

relayer/relays/parachain/config.go
```diff
@@ -13,7 +19,30 @@ type SourceConfig struct {
 	Ethereum  config.EthereumConfig  `mapstructure:"ethereum"`
 	Contracts SourceContractsConfig  `mapstructure:"contracts"`
 	// Block number when Beefy was activated
-	BeefyActivationBlock uint64 `mapstructure:"beefy-activation-block"`
+	BeefyActivationBlock uint64   `mapstructure:"beefy-activation-block"`
+	AccountIDs           []string `mapstructure:"accountIDs"`
```
vgeddes, line 23:
```suggestion
	AccountIDs           []string `mapstructure:"accounts"`
```

test/README.md
```diff
@@ -22,6 +22,7 @@ The E2E tests run against local deployments of the parachain, relayer, the ether
 * Development environment for the relay services. See setup [instructions](../relayer/README.md#development).
 * `jq` - https://stedolan.github.io/jq/download/
 * geth - https://geth.ethereum.org/docs/install-and-build/installing-geth
+* g++ (required for lodestar) is available in the `build-essential` package: `apt install build-essential`
```
claravanstaden, line 25: Interesting! I'm trying to remember if I had to install this. 🤔

claravanstaden (approved):
Very nice! 🚀

parachain/pallets/basic-channel/src/outbound/mod.rs
doubledup: If we made each nonce into something like a BigInt or even a [u64; n], we could resize the nonce as necessary. We'd need to set a max size to implement MaxEncodedLen for now, but we'd have an upgrade path.

Is it worth putting this in now? If not, do we have a place to record these limits?

test/test/bootstrap.js
doubledup: Cool, I'll set up a separate batch test with at least 6 accounts

doubledup: Will tackle this in [SNO-306](https://linear.app/snowfork/issue/SNO-306)

ethereum/test/test_basic_inbound_channel.js
doubledup: Will tackle this in [SNO-306](https://linear.app/snowfork/issue/SNO-306)

vgeddes (approved):
Looks good! Lets address remaining low-hanging fruit in a follow-up PR.

parachain/pallets/basic-channel/merkle-proof/src/lib.rs
doubledup: It sounds like a NOTICE file is only required for Apache Software Foundation projects:

> If your project is not an Apache Software Foundation project, but just happens to be Apache 2 licensed, you're not required to even keep a NOTICE file.

But the next sentence is:

> The NOTICE file is where you would mention if parts of the code have a different copyright...

This applies here, since our parachain is already under Apache but the copyright belongs to parity for this file.

I'll check in with Parity to see what the acceptable options are.

.gitignore
doubledup: I've got a global gitignore file, but tool-versions lets me pin the versions of things like Go in each project. Adding it to a global gitignore file prevents me from seeing changes to tool-versions in all projects.

Probably worth scrapping the tool-versions file in this case and just using the latest versions, since they're unlikely to be backwards-incompatible.

test/README.md
doubledup: It wasn't installed on my fresh instance, but you may well have installed it as part of the setup steps for something else.

parachain/pallets/basic-channel/merkle-proof/src/lib.rs
vgeddes: Cool, I guess we’re covered then. The copyright and license text in the header of the file should be sufficient

.gitignore
vgeddes: Ok, no worries, add whatever you need to the gitignore

doubledup:
Added an [issue](https://linear.app/snowfork/issue/SNO-308) to replace the copied Merkle proof implementation with a dependency on Substrate.

parachain/pallets/basic-channel/src/outbound/mod.rs
doubledup: Added an [issue](https://linear.app/snowfork/issue/SNO-309) for nonce overflow

parachain/pallets/basic-channel/src/outbound/mod.rs
doubledup: Created an [issue](https://linear.app/snowfork/issue/SNO-310) to track this

test/README.md
alistair-singh: Installing `build-essential` is part of the parachain setup documentation in `parachain/README.md`.
See https://docs.substrate.io/install/linux/

alistair-singh (approved):
🚀 Looks good! There are still some TODOs in the code that should tickets if they are still relevant.

PR #638, opened 2022-06-03
Execution Header processing and syncing

claravanstaden:
Includes:
- a new extrinsic called `import_execution_header` to store the execution header of beacon headers that are finalized.
- sends the beacon block from the relayer to the light client


relayer/relays/beacon/syncer/syncer.go
```diff
@@ -393,14 +411,14 @@ func IsInArray(values []uint64, toCheck uint64) bool {
 }
 
 func hexToBinaryString(rawHex string) string {
-	hex := strings.Replace(rawHex, "0x", "", -1)
```
alistair-singh, line 396: You can use `math/big` package to do this conversion for you. See https://go.dev/play/p/a3OICKkC9RN?v=goprev.go?download=true

Also instead of using `strings.Replace(rawHex, "0x", "", -1)` you could use `slices` with `rawHex[2:]`

parachain/pallets/ethereum-beacon-client/src/merkleization.rs
```diff
@@ -276,6 +258,22 @@ pub fn hash_tree_root<T: SimpleSerializeTrait>(mut object: T) -> Result<[u8; 32]
     }
 }
 
+pub fn get_sync_committee_bits(bits_hex: Vec<u8>) -> Result<Vec<u8>, MerkleizationError> {
+    let bitv = Bitvector::<{ config::SYNC_COMMITTEE_SIZE }>::deserialize(&bits_hex).map_err(|_| MerkleizationError::InvalidLength)?;
+
+    let mut result = Vec::new();
+
+    for bit in bitv.iter() {
+        if bit == true {
```
alistair-singh, line 267: Style: you can map over `bitv` and return a `vec` instead of using mutable `vec`.

claravanstaden: Addressed in [sync_committee_bits](https://github.com/Snowfork/snowbridge/pull/638/commits/8fdc1a156c72137a784def385d575baeb5483258).

relayer/relays/beacon/syncer/syncer.go
claravanstaden: Addressed in [sync_committee_bits](https://github.com/Snowfork/snowbridge/pull/638/commits/8fdc1a156c72137a784def385d575baeb5483258).

PR #639, opened 2022-06-07
Gas multiplier

alistair-singh:
Resolves: SNO-281

### DIFF
--- a/relayer/chain/ethereum/transact_opts_test.go
+++ b/relayer/chain/ethereum/transact_opts_test.go
@@ -1,0 +1,248 @@
+package ethereum_test
+
+import (
+	"math/big"
+	"testing"
+
+	"github.com/snowfork/snowbridge/relayer/chain/ethereum"
+	"github.com/snowfork/snowbridge/relayer/config"
+	"github.com/stretchr/testify/assert"
+)
+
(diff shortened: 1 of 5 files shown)


relayer/chain/ethereum/transact_opts_test.go
```diff
@@ -0,0 +1,248 @@
...
+		GasTipMultiplier: 0,
+	}
+
+	var suggestedFee *big.Int
+	var suggestedTip *big.Int
+
+	fee := ethereum.CalculateFee(config, suggestedFee, suggestedTip)
+
+	assert.Nil(t, fee.GasFeeCap)
+	assert.Nil(t, fee.GasTipCap)
```
claravanstaden, line 26: I guess you can convert these to table driven tests, but it is not really necessary.

claravanstaden (approved):
🏅 Very nice!

relayer/chain/ethereum/transact_opts_test.go
alistair-singh: Is this like a library? Or do I just create a table using an array and compress it all into one test?

vgeddes:
Hey, this fell of my radar. Will review next week.

vgeddes:
As discussed on the call, lets circle back to this once Rococo testnet is operational again.

PR #642, opened 2022-06-15
Ropsten start services script

claravanstaden:
All E2E tests now pass on local testnet. 8/13 tests pass on ropsten but are unreliable possibly due to an issue with the beefy relayer(more info below). The ethereum relayer continues to work even for ropsten because PoW verification is disabled on the parachain in the `snowbase` version of the runtime. Headers still import successfully even though they are PoS headers.

To use ropsten you need to have separate accounts for the relayers, an account for deployment and one for running the E2E stack.

Add the following to `ethereum/.envrc`:
```
export ROPSTEN_PRIVATE_KEY=0x…deployment.account
export INFURA_PROJECT_ID=…
```
Add the following to `test/.envrc`:
```
export INFURA_PROJECT_ID=…

export ETH_NETWORK=ropsten
export ETH_NETWORK_ID=3

export ETH_RPC_ENDPOINT=https://ropsten.infura.io/v3
export ETH_WS_ENDPOINT=wss://ropsten.infura.io/ws/v3

export ROPSTEN_PRIVATE_KEY=0x…deployment.account
export BEEFY_RELAY_ETH_KEY=0x…beefy.relayer.account
export PARACHAIN_RELAY_ETH_KEY=0x…parachain.relayer.account
export E2E_TEST_ETH_KEY=0x…test.account
```

The beefy relayer sometimes fails with the error below but does eventually recover and relay messages. I think this may be config related. I believe this is causing the timeouts as it means we miss a round of justifications.
```
{"@timestamp":"2022-06-22T00:23:49.246346738Z","error":"submit request: create validator bitfield: execution reverted: wait period not over","level":"fatal","message":"Unhandled error"}
```
Tests might take a bit longer to run so use a greater timeout e.g. `yarn test -t 1200000`

### Todo
- [x] Address beefy relayer issues. (Fixed by using different accounts to run the tests)
- [x] ERC20 Tests are currently failing for non timeout reasons. I am busy troubleshooting this.


claravanstaden:
@alistair-singh looks great! 👌 I can't approve because I created the PR, but I think it's good to go.

test/README.md
```diff
@@ -126,3 +126,29 @@ You should see the test pass, checking that message delivery works correctly and
 ## Generating/Updating new test fixtures
 
 Test fixtures are taken by running the service in full e2e test. The relayer should log the fixture data you need (code is in [the relayer here](../relayer/workers/beefyrelayer/fixture-data-logger.go), though may require a bit of manual copy/pasting to get perfectly it in the right format.
+
+## Running E2E tests on Ropsten
+
+To run the E2E tests on Ropsten you need to have separate accounts for the relayers, an account for deployment and one for running the E2E test stack. You will also require an [Infura](https://infura.io/) account and project.
+
+Add the following variables to `ethereum/.envrc`:
```
vgeddes, line 134: I don't the think the variables from `ethereum/.envrc` are used at all when starting the E2E stack? Can probably remove this part if thats the case.

test/test/bootstrap.js
```diff
@@ -43,7 +43,7 @@ describe('Bridge', function () {
     });
 
     it('should transfer ETH from Ethereum to Substrate (incentivized channel)', async function () {
-      const amount = BigNumber(Web3.utils.toWei('1', "ether"));
+      const amount = BigNumber(Web3.utils.toWei('0.1', "ether"));
```
vgeddes, line 46: `0.1` is still quite a lot for ropsten. May want to reduce it further to `0.001`.

test/README.md
```diff
@@ -126,3 +126,29 @@ You should see the test pass, checking that message delivery works correctly and
...
+```bash
+export ROPSTEN_PRIVATE_KEY=0x…                # Your deployment account private key
+export INFURA_PROJECT_ID=…                    # Your Infura project id
+```
+
+Add the following variables to `test/.envrc`:
+```bash
```
vgeddes, line 141: We already have an envrc template at `test/.envrc-example`, where these example vars should go (at least the mandatory ones). The optional vars could perhaps be commented out in the template file.

vgeddes (approved):
Looks good! Added some minor comments to address before merging.

PR #650, opened 2022-06-30
SNO-290, SNO-289 Header update verification fixes

claravanstaden:
- Changes relayer to work with block hashes rather than slots
- Fixes [SNO-289](https://linear.app/snowfork/issue/SNO-289/investigate-failed-block-verification-on-some-header-updates) where some execution header updates failed signature verification. [This was due to the relayer's conversion from hex string to byte array, where some of the leading zero bytes were truncated.](https://github.com/Snowfork/snowbridge/pull/650/files#diff-c6b15dc3d53ce6c2b26b27eeb75b67a60c66183d691df0bdfd0d497762e2d451R478) This caused the hash tree root of the beacon block to be different to what is expected, which in turned then failed BLS signature verification
- Fixes [SNO-290](https://linear.app/snowfork/issue/SNO-290/check-for-empty-fee-recipient). [The light client panicked when the fee recipient was empty.](https://github.com/Snowfork/snowbridge/pull/650/files#diff-a299082b2a90ace00c2385ea719fd1855bd9d0ff80ab426e4014ba893e96448aR471)


claravanstaden:
Screenshot 2022-06-30 at 13 21 04

No more errors. 🌟

claravanstaden:
This is one more issue that I am aware of. Now and then, the sync aggregate of a block is empty:
```
"sync_aggregate":
    {
        "sync_committee_bits": "0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
        "sync_committee_signature": "0xc00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"
    }
```
The previous block's signature is then invalid. I am not sure why the sync aggregate would be empty, I'm checking in on that. But our light client's behaviour is correct, it rejects the block with a `InvalidAggregatePublicKeys` error, because there are no pubkeys in this case.

vgeddes (approved):
LGTM!

PR #653, opened 2022-07-04
SNO-274 Message verification in Beacon Client

claravanstaden:
This PR mainly focuses on message verification:

- The beacon client `verify` implementation
- Reading events in the relayer
- Sending events from the relayer to the beacon client

This PR leaves the message verification implementation to the Ethereum light client. The message verification will be switched to the beacon client in a follow-up PR, where the E2E tests will also be updated.

To start the beacon relayer from the start-services script, add env variable `START_BEACON_SYNC=true`. It assumes that you have `geth` and `lodestar` installed on your testing machine, and that it has synced successfully.

[SNO-274](https://linear.app/snowfork/issue/SNO-274/send-event-with-proof-to-beacon-light-client)


parachain/pallets/ethereum-beacon-client/src/lib.rs
```diff
@@ -780,5 +771,102 @@ pub mod pallet {
...
+		fn verify(message: &Message) -> Result<Log, DispatchError> {
+			log::info!(
+				target: "ethereum-beacon-client",
+				"💫 Verifying message with block hash {}",
+				message.proof.block_hash,
+			);
+
+			let receipt_result = Self::verify_receipt_inclusion(&message.proof);
+			if let Err(receipt_result) = &receipt_result {
+				log::trace!(
```
claravanstaden, line 812: I added more trace logging for errors so that we can debug issues if any arise, especially because this is the newest part of the Eth2 work.

parachain/primitives/beacon/Cargo.toml
```diff
@@ -8,12 +8,16 @@ edition = "2021"
...
+rlp = { version = "0.5", default-features = false }
 
 frame-support = { git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.19", default-features = false }
 frame-system = { git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.19", default-features = false }
 sp-runtime = { git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.19", default-features = false }
 sp-std = { git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.19", default-features = false }
 sp-core = { git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.19", default-features = false }
+sp-io = { git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.19", default-features = false }
+
+snowbridge-ethereum = { path = "../../primitives/ethereum", default-features = false }
```
claravanstaden, line 20: I reuse what I can from the `snowbridge-ethereum` primitive crate.

parachain/primitives/beacon/src/lib.rs
```diff
@@ -193,3 +195,39 @@ pub struct BeaconBlock {
 	pub state_root: H256,
 	pub body: Body,
 }
+
+impl ExecutionHeader {
```
claravanstaden, line 199: `check_receipt_proof` and `apply_merkle_proof` are copied from the `snowbridge-ethereum` primitive crate to implement the new `ExecutionHeader`, but I reuse everything else.

test/scripts/start-services.sh
```diff
@@ -51,6 +51,28 @@ start_geth() {
         > "$output_dir/geth.log" 2>&1 &
 }
 
+start_geth_for_beacon_node() {
```
claravanstaden, line 54: This can be merged with `start_geth` in a follow-up PR. The reason why I added a new function for this is because the Lodestar-Geth combo requires a JWT secret, and I wasn't sure if that would break something if I added a `authrpc` parameter to our local geth setup.

parachain/pallets/ethereum-beacon-client/src/lib.rs
```diff
@@ -780,5 +768,102 @@ pub mod pallet {
...
+					Err(Error::<T>::InvalidProof.into())
+				},
+			}
+		}
+	}
+
+	impl<T: Config> Verifier for Pallet<T> {
+		/// Verify a message by verifying the existence of the corresponding
+		/// Ethereum log in a block. Returns the log if successful.
+		fn verify(message: &Message) -> Result<Log, DispatchError> {
```
claravanstaden, line 800: Copied from the Ethereum light client. I added more trace logs on error conditions so it is easier to debug if there is a verification failure, especially because this code is the newest in the beacon light client.

relayer/relays/beacon/main.go
```diff
@@ -269,13 +291,38 @@ func (r *Relay) SyncHeaders(ctx context.Context) error {
...
+	if err != nil {
+		return err
+	}
+
+	logrus.WithFields(logrus.Fields{
+		"start": secondLastBlockNumber,
+		"end":   lastBlockNumber - 1,
+	}).Info("Processing events for block numbers")
+
+	payload, err := r.listener.ProcessEvents(ctx, secondLastBlockNumber, lastBlockNumber-1)
```
vgeddes, line 311: Looking over the relayer code reminds me that we still need to handle certain recovery scenarios (in a later PR). Suppose the relayer dies somehow after importing an execution header into the parachain, but before importing the events for that execution header. Then when the relayer starts up it doesn't know that it still needs to import some old events, and moves onto the next header.

So we need to digest the problem and figure out a solution. In a later PR of course.

One idea is for the parachain inbound channels to keep track of the blockNumber of the last imported event. Then when the relayer startup it queries that block number, and begins relaying ethereum events from that block number. Furthermore, messages from ethereum are basically events like `Message(nonce, payload)`. So the relayer can inspect the message nonce to see if it is the next message expected by the inbound channel pallets in the Parachain.

I'm sure there other solutions too.

relayer/relays/beacon/syncer/syncer.go
```diff
@@ -362,6 +362,42 @@ func (s *Syncer) GetHeaderUpdate(blockRoot common.Hash) (HeaderUpdate, error) {
 	return headerUpdate, nil
 }
 
+func (s *Syncer) GetBlockRange(lastBlockHash, secondLastBlockHash common.Hash) (uint64, uint64, error) {
+	lastBlock, err := s.Client.GetBeaconBlock(lastBlockHash)
+	if err != nil {
+		logrus.WithError(err).Error("unable to fetch last block using hash")
+
+		return 0, 0, err
```
vgeddes, line 370: Earlier this year I refactored the relayer to use a new error reporting strategy, one that reduces the amount of logging boilerplate. The problem I saw was that each error handling block logs the error, and then returns the error. This ends up in the same error being logged repeatedly throughout the call chain.

So instead, most of the relayer codebase now uses [error wrapping](https://golangbyexample.com/wrapping-and-unwrapping-error-golang/) to provide context to errors:
```suggestion
		return 0, 0, fmt.Errorf("fetch last block: %w", err)
```

The other recommended thing was to avoid using the phrases "failed", "unable", "cannot" in error messages, as they end up being printed redundantly.

Then at the top of the call chain, the final wrapped error is logged once, printing a complete error message like `error: sync: determine block range: fetch last block: network connection timed out`

More docs about error wrapping:
* https://go.dev/blog/go1.13-errors

test/config/launch-config.json
```diff
@@ -50,6 +50,7 @@
           "port": 31200,
           "name": "alice",
           "flags": [
+            "-lethereum-light-client=trace",
```
vgeddes, line 53: This would be the beacon client surely?

parachain/pallets/ethereum-beacon-client/src/lib.rs
```diff
@@ -13,14 +13,12 @@ use codec::{Decode, Encode};
...
 use scale_info::TypeInfo;
-use snowbridge_beacon::{
-	BeaconBlock, BeaconHeader, Domain, ExecutionHeader, ForkData, PublicKey, Root, SigningData,
-	SyncAggregate, SyncCommittee,
-};
 use sp_core::H256;
 use sp_io::hashing::sha2_256;
 use sp_runtime::RuntimeDebug;
 use sp_std::prelude::*;
+use snowbridge_beacon::{SyncCommittee, BeaconHeader, SyncAggregate, ForkData, Root, Domain, PublicKey, SigningData, ExecutionHeader, BeaconBlock};
```
vgeddes, line 20: I would name the beacon primitives crate to this for clarity in what the crate provides (primitives):

```suggestion
use snowbridge_beacon_primitives::{SyncCommittee, BeaconHeader, SyncAggregate, ForkData, Root, Domain, PublicKey, SigningData, ExecutionHeader, BeaconBlock};
```

vgeddes (commented):
Looks great! 🚀

relayer/relays/beacon/main.go
claravanstaden: I logged Linear issue [SNO-292](https://linear.app/snowfork/issue/SNO-292/relayer-and-beacon-client-should-support-syncing-from-last-synced-slot) for this.

parachain/pallets/ethereum-beacon-client/src/lib.rs
```diff
@@ -780,5 +768,102 @@ pub mod pallet {
...
+					target: "ethereum-beacon-client",
+					"Verify receipt inclusion failed for block {}: {:?}",
+					message.proof.block_hash,
+					receipt_result
+				);
+
+				return Err(*receipt_result);
+			}
+
+			let receipt = receipt_result.unwrap();
```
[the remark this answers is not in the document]
claravanstaden: Addressed in https://github.com/Snowfork/snowbridge/pull/653/commits/45db3ef5727c1fa56e7c5b0f12b387bce7f0cacd.

parachain/pallets/ethereum-beacon-client/src/lib.rs
```diff
@@ -780,5 +768,102 @@ pub mod pallet {
...
+					target: "ethereum-beacon-client",
+					"RLP log decoded failed {}: {:?}",
+					message.proof.block_hash,
+					log_result
+				);
+
+				return Err(Error::<T>::DecodeFailed.into());
+			}
+
+			let log: Log = log_result.unwrap();
```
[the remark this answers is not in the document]
claravanstaden: Addressed in https://github.com/Snowfork/snowbridge/pull/653/commits/45db3ef5727c1fa56e7c5b0f12b387bce7f0cacd.

relayer/relays/beacon/syncer/syncer.go
claravanstaden: Thanks for reminding me, this is a nifty pattern! Addressed in https://github.com/Snowfork/snowbridge/pull/653/commits/45db3ef5727c1fa56e7c5b0f12b387bce7f0cacd.

test/config/launch-config.json
claravanstaden: I reverted this in https://github.com/Snowfork/snowbridge/pull/653/commits/45db3ef5727c1fa56e7c5b0f12b387bce7f0cacd, accidentally commit.

parachain/pallets/ethereum-beacon-client/src/lib.rs
claravanstaden: Addressed in https://github.com/Snowfork/snowbridge/pull/653/commits/45db3ef5727c1fa56e7c5b0f12b387bce7f0cacd

claravanstaden:
@vgeddes thanks for the review! I have addressed most of your comments, save for the resuming sync from the last synced block. I logged a Linear issue. 🙃

alistair-singh (approved):

vgeddes (approved):
LGTM!

PR #656, opened 2022-07-07
Updates scale metadata after adding beacon pallet.

claravanstaden:
The metadata.scale file should be updated after adding the beacon client.


vgeddes:
Could you also add a note to the parachain README about how to how to update the metadata.

I'm assuming you're using the steps at https://github.com/paritytech/subxt#downloading-metadata-from-a-substrate-node for fetching the new metadata?

claravanstaden:
@vgeddes I used these steps that @alistair-singh provided:

```
run start-services.sh to start up the parachain.
then run the following in the parachain sub folder
curl http://localhost:8081/ \
    -X POST \
    -H "Content-Type: application/json" \
    -d '{ "id": 1, "jsonrpc": "2.0", "method": "state_getMetadata", "params": [] }' | jq -r .result | xxd -r -p - > tools/query-events/metadata.scale
Do a git status and make sure the metadata.scale file has been updated.
Stop the parachain and re-launch with start-services.sh and then run tests.
```

I will try the subxt way too and update the readme. I'll do that in a separate PR because it may not be necessary to update the metadata. Alistair did a clean build with the metadata in main and it worked fine. I will try the same later, just want to finish off the E2E tests first since a clean build takes some time.

alistair-singh:
I think this could related to the type name issue we saw earlier. Where on some machines the type name has extra bytes. https://github.com/paritytech/subxt/issues/591

I closed the above issue because Parity opened up another issue to remove type name from the hash check. I believe this will fix our issues. https://github.com/paritytech/subxt/pull/595

But also in previous PRs we did change metadata. I have `diff`ed the generated rust code between this commit and `main` and the generated code has changed. I believe this is because the beacon types were moved to a `snowbridge-beacon-primitives` crate.

On my machine `main` builds with the old (checked in) metadata even with the latest changes to beacon. So I am at a little bit of a loss here as to whether we need to commit this or not. Maybe other members of the team need to try build `main` and see if this issue pops up and if the metadata needs to be updated.

claravanstaden:
> Maybe other members of the team need to try build main and see if this issue pops up and if the metadata needs to be updated.

I will try this this afternoon. 👍

vgeddes:
> I will try the subxt way too and update the readme. I'll do that in a separate PR because it may not be necessary to update the metadata. Alistair did a clean build with the metadata in main and it worked fine. I will try the same later, just want to finish off the E2E tests first since a clean build takes some time.

Ok, cool. sounds fine if that works better for you 👍

claravanstaden:
I did a clean build from main with no issues so the updated scale metadata isn't necessarily. I will still update the readme.

PR #658, opened 2022-07-11
GenesisConfig for beacon client

claravanstaden:
The `initial_sync` extrinsic should be removed - it should form part of the pallet GenesisConfig.

### DIFF
--- a/parachain/pallets/ethereum-beacon-client/src/lib.rs
+++ b/parachain/pallets/ethereum-beacon-client/src/lib.rs
@@ -335,8 +335,7 @@ pub mod pallet {
 			)?;
 
 			let current_period = Self::compute_current_sync_period(update.attested_header.slot);
-
-			let current_sync_committee = <SyncCommittees<T>>::get(current_period);
+			let current_sync_committee = Self::get_sync_committee_for_period(current_period)?;
 			let validators_root = <ValidatorsRoot<T>>::get();
 
 			Self::verify_signed_header(
--- a/parachain/pallets/ethereum-beacon-client/src/lib.rs
+++ b/parachain/pallets/ethereum-beacon-client/src/lib.rs
@@ -156,24 +156,33 @@ pub mod pallet {
 		StorageMap<_, Identity, u64, SyncCommittee, ValueQuery>;
 
 	#[pallet::storage]
+	pub(super) type LatestSyncCommitteePeriod<T: Config> = StorageValue<_, u64, ValueQuery>;
+
+	#[pallet::storage]
 	pub(super) type ValidatorsRoot<T: Config> = StorageValue<_, H256, ValueQuery>;
 
 	#[pallet::storage]
 	pub(super) type LatestFinalizedHeaderSlot<T: Config> = StorageValue<_, u64, ValueQuery>;
 
 	#[pallet::genesis_config]
-	pub struct GenesisConfig {}
+	pub struct GenesisConfig {
+		pub initial_header: BeaconHeader,
+	}
 
 	#[cfg(feature = "std")]
 	impl Default for GenesisConfig {
-		fn default() -> Self {
-			Self {}
-		}
+		fn default() -> Self { Self {
+			initial_header: Default::default()
+		}}
 	}
 
 	#[pallet::genesis_build]
 	impl<T: Config> GenesisBuild<T> for GenesisConfig {
-		fn build(&self) {}
+		fn build(&self) {
+			Pallet::<T>::initial_sync_test(
+				self.initial_header.clone(),
+			).unwrap();
+		}
 	}
 
 	#[pallet::call]
@@ -623,6 +632,17 @@ pub mod pallet {
 
 		fn store_sync_committee(period: u64, sync_committee: SyncCommittee) {
 			<SyncCommittees<T>>::insert(period, sync_committee);
+
+			let latest_committee_period = <LatestSyncCommitteePeriod<T>>::get();
+
+			if period > latest_committee_period {
+				log::trace!(
+					target: "ethereum-beacon-client",
+					"💫 Updated latest sync committee period stored to {}.",
+					period
+				);
+				<LatestSyncCommitteePeriod<T>>::set(period);
+			}
 		}
 
 		fn store_finalized_header(block_root: H256, header: BeaconHeader) {
@@ -791,6 +811,12 @@ pub mod pallet {
 					Err(Error::<T>::InvalidProof.into())
 				},
 			}
+		}
+
+		pub fn initial_sync_test(
+			_header: BeaconHeader,
+		) -> Result<(), &'static str> {
+			Ok(())
 		}
 	}
 
--- a/parachain/pallets/ethereum-beacon-client/src/lib.rs
+++ b/parachain/pallets/ethereum-beacon-client/src/lib.rs
@@ -613,7 +613,7 @@ pub mod pallet {
 			}
 		}
 
-		fn store_finalized_header(block_root: H256, header: BeaconHeader) {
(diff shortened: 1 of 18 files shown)


parachain/pallets/ethereum-beacon-client/src/lib.rs
```diff
@@ -712,7 +700,7 @@ pub mod pallet {
 			branch: Vec<H256>,
 			depth: u64,
 			index: u64,
-			root: Root,
+			root: H256,
```
claravanstaden, line 703: I should revert this change back to `Root`.

parachain/primitives/beacon/src/lib.rs
```diff
@@ -21,6 +29,47 @@ impl Default for PublicKey {
...
+}
+
+struct PublicKeyVisitor;
+
+#[cfg(feature = "std")]
+impl<'de> Visitor<'de> for PublicKeyVisitor {
+	type Value = PublicKey;
+
+	fn expecting(&self, formatter: &mut Formatter) -> StdResult {
+		formatter.write_str("an array of bytes")
```
claravanstaden, line 46:
```suggestion
		formatter.write_str("A hex string")
```

parachain/primitives/beacon/src/lib.rs
```diff
@@ -21,6 +29,47 @@ impl Default for PublicKey {
...
+			None => v,
+		};
+
+		let hex_bytes = match hex::decode(str_without_0x) {
+			Ok(bytes)=> bytes,
+			Err(e)=> return Err(Error::custom(e.to_string()))
+		};
+
+		let mut data = [0u8; 48];
+		data[0..48].copy_from_slice(&hex_bytes);
```
claravanstaden, line 61: Need to check if `hex_bytes` length is `48` else error.

parachain/pallets/ethereum-beacon-client/src/lib.rs
```diff
@@ -44,6 +47,7 @@ type ProofBranch = Vec<H256>;
 type ForkVersion = [u8; 4];
 
 #[derive(Clone, Default, Encode, Decode, PartialEq, RuntimeDebug, TypeInfo)]
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
```
vgeddes, line 50:
```suggestion
#[cfg_attr(feature = "std", derive(serde::Serialize, serde::Deserialize))]
```

Then we drop the serde imports on line 23.

parachain/primitives/beacon/src/lib.rs
```diff
@@ -21,6 +28,47 @@ impl Default for PublicKey {
...
+
+#[cfg(feature = "std")]
+impl<'de> Visitor<'de> for PublicKeyVisitor {
+	type Value = PublicKey;
+
+	fn expecting(&self, formatter: &mut Formatter) -> StdResult {
+		formatter.write_str("an array of bytes")
+	}
+
+	fn visit_str<E>(self, v: &str) -> Result<Self::Value, E>where E: Error,  {
```
vgeddes, line 48:
```suggestion
	fn visit_str<E>(self, v: &str) -> Result<Self::Value, E> where E: Error {	
```

parachain/primitives/beacon/src/lib.rs
```diff
@@ -21,6 +28,47 @@ impl Default for PublicKey {
...
+#[cfg(feature = "std")]
+impl<'de> Visitor<'de> for PublicKeyVisitor {
+	type Value = PublicKey;
+
+	fn expecting(&self, formatter: &mut Formatter) -> StdResult {
+		formatter.write_str("an array of bytes")
+	}
+
+	fn visit_str<E>(self, v: &str) -> Result<Self::Value, E>where E: Error,  {	
+        let	str_without_0x = match v.strip_prefix("0x") {
```
vgeddes, line 49:
```suggestion
        let str_without_0x = match v.strip_prefix("0x") {
```

relayer/relays/beacon/parachain-writer.go
```diff
@@ -120,3 +120,18 @@ func (wr *ParachainWriter) WriteToParachain(ctx context.Context, extrinsicName s
...
+func (wr *ParachainWriter) getLastSyncedSyncCommitteePeriod() (uint64, error) {
+	key, err := types.CreateStorageKey(wr.conn.Metadata(), "EthereumBeaconClient", "LatestSyncCommitteePeriod", nil, nil)
+	if err != nil {
+		return 0, fmt.Errorf("create storage key for last sync committee: %w", err)
+	}
+
+	var period types.U64
+	ok, err := wr.conn.API().RPC.State.GetStorageLatest(key, &period)
+	if err != nil || !ok {
+		panic(err)
```
vgeddes, line 133: Can't we rather return an error here instead of panicking?

vgeddes (approved):
Looks good! Added some misc comments

parachain/pallets/ethereum-beacon-client/src/lib.rs
```diff
@@ -792,6 +780,31 @@ pub mod pallet {
 				},
 			}
 		}
+
+		pub fn initial_sync(
```
alistair-singh, line 784: Does this need to be `pub` if it is only called from `GenesisBuild::build`?

test/.envrc-example
```diff
@@ -25,10 +25,11 @@ export PARACHAIN_ID=1000
...
+
 # export ROPSTEN_PRIVATE_KEY=         # Your deployment account private key
 # export BEEFY_RELAY_ETH_KEY=         # Your Beefy relayer account private key
 # export PARACHAIN_RELAY_ETH_KEY=     # Your Parachain relayer account private key
 # export E2E_TEST_ETH_KEY=            # Your E2E test account private key
 
-#export START_BEACON_SYNC=false       # Start the Beacon chain syncing (in development)
\ No newline at end of file
+# export START_BEACON_SYNC=false       # Start the Beacon chain syncing (in development)
+# export INITIAL_BEACON_BLOCK=0x088241fcf1cf63040b804498c945d3a6ae5484e65692483747fba5b8902c99c9 # Initial block to be synced from (this value is on ropsten)
```
alistair-singh, line 35: Can we get the initial beacon block using a script? We do this with `beefy`.

parachain/pallets/ethereum-beacon-client/src/lib.rs
claravanstaden: Good point! I changed it to `pub(super)` so that the tests can still access it.

test/.envrc-example
claravanstaden: We could just get the latest block, yeah. However, the light client protocol suggests always syncing from a manually validated block, so that you know you are syncing on the right chain. Is that not also an issue with Beefy?

alistair-singh: Sorry maybe beefy is the wrong example here. In the `ethereum-light-client` we know the block is trusted by getting the latest block from a trusted node (Infura) as opposed to setting an environment variable. Can we not do something similar for beacon? Fetch the latest sync committee from a trusted beacon node at time of building the genesis state?

claravanstaden: I can get the latest finalized block from a trusted node, yeah! Let me do that quick.

I think the sync committee fetching can stay in the relayer for now - I'll relook at it once I get to the [resuming syncing ticket](https://linear.app/snowfork/issue/SNO-292/relayer-and-beacon-client-should-support-syncing-from-last-synced-slot).

claravanstaden: Added in https://github.com/Snowfork/snowbridge/pull/658/commits/31bb446bed0110d837aaa3430b498ca1fccb08bb.

alistair-singh (approved):

PR #661, opened 2022-07-22
SNO-285 setup local beacon testnet for e2e tests

claravanstaden:
- Setup of local lodestar beacon net
- Minimal config for `snowbase` runtime
- Feature `minimal` in `ethereum-beacon-client` to toggle beacon net specs

PR #664, opened 2022-07-26
Update API endpoints and use public Lodestar Ropsten server for start-services

claravanstaden:
- Updates the Lodestar endpoints with the new light client spec version
- Uses the public Ropsten Lodestar server for start-services, instead of running own node

Resolves: [SNO-291](https://linear.app/snowfork/issue/SNO-291/update-endpoints-to-use-new-light-client-spec)


claravanstaden:
Changes included in https://github.com/Snowfork/snowbridge/pull/661.

PR #665, opened 2022-07-28
Swap verifier impl to the beacon light client

claravanstaden:
This switches from Ethereum PoW to PoS verification. 🥳 

Keeping in draft until Rococco deploy.


vgeddes (commented):
Looks good, you just need to make sure `start-services.sh` doesn't start the old deprecated ethereum-relay process.

alistair-singh (approved):
🚀

claravanstaden:
I ran the E2E tests against this branch again, all looks good! Only thing is I had to regenerate the metadata.scale file, @alistair-singh will check the diff to see if it needs to be checked in.

After that, we can merge. 🚢

PR #666, opened 2022-07-28
Remove TTD if beacon node is not used

claravanstaden:
When a TTD value is added to the geth genesis config and the beacon node doesn't run, contract deployment hangs.


test/scripts/start-services.sh
```diff
@@ -35,10 +35,15 @@ start_geth() {
 
     local data_dir="$output_dir/geth"
 
+    # removes TTD bomb for when beacon chain is not used
+    if [ "$start_beacon_sync" == "false" ]; then
+        jq 'del(.config.terminalTotalDifficulty)' config/genesis.json > "$output_dir/genesis.json"
```
alistair-singh, line 40: This will override the difficulty config from lines 27-34. I think you should use:
```
... "$output_dir/genesis.json" | sponge "$output_dir/genesis.json"
```

claravanstaden: Good point! Addressed in https://github.com/Snowfork/snowbridge/pull/666/commits/fea4e9963bfe684414a2fe6f7b22a08b422decd4.

Although interestingly `$output_dir/genesis.json` wasn't actually used before.

test/scripts/start-services.sh
```diff
@@ -35,10 +35,15 @@ start_geth() {
...
+    # removes TTD bomb for when beacon chain is not used
+    if [ "$start_beacon_sync" == "false" ]; then
+        jq 'del(.config.terminalTotalDifficulty)' "$output_dir/genesis.json" | sponge "$output_dir/genesis.json"
+    fi
+    
     if [ "$eth_network" == "localhost" ]; then
         echo "Starting geth local net"
 
-        geth init --datadir "$data_dir" config/genesis.json
+        geth init --datadir "$data_dir" "$output_dir/genesis.json"
```
claravanstaden, line 46: I changed it to use the modified file.

alistair-singh: Ah nice find! This is why removing TTD didnt work for me!

test/scripts/start-services.sh
alistair-singh: Ok tested and it does not hang on contract deploy anymore.

alistair-singh (approved):

PR #668, opened 2022-07-29
Relayer and beacon client should support syncing from last synced slot

claravanstaden:
- Stores last finalized header hash on beacon client
- Reads the above in relayer
- Syncs from the last finalized header

Resolves: [SNO-292](https://linear.app/snowfork/issue/SNO-292/relayer-and-beacon-client-should-support-syncing-from-last-synced-slot)

Note: Message verification backfill will be done in a separate PR since it requires tracking the basic channel and incentivized channel messages separately.


parachain/pallets/ethereum-beacon-client/src/lib.rs
```diff
@@ -851,6 +860,8 @@ pub mod pallet {
 				message.proof.block_hash,
 			);
 
+			<LatestVerifiedMessageBlock<T>>::set(block_number);
```
vgeddes, line 863: This approach is gonna be problematic, as it ignores that the basic and incentivized inbound channel pallets have their own logical message stream.

Depending on the message, the relayer will submit it either to the basic or incentivized inbound channel pallet.

So the problem that occurs in your approach is the following: Suppose the basic inbound pallet last processed a message at block 100.

A relayer submits incentivized messages to the incentivized channel, all the way up to block 10000. It ignores basic messages. Now `LatestVerifiedMessageBlock` points to block 1000, even though the basic inbound channel still needs to be backfed messages from 100-1000.

So actually we need to track LatestVerifiedMessageBlock separately for the basic and incentivized channels.

vgeddes: > So actually we need to track LatestVerifiedMessageBlock separately for the basic and incentivized channels.

And likely the relayer would have 2 separate goroutines that backfill messages to the basic channel or the incentivized channel.

claravanstaden: I reverted the message backfill, will open another PR then.

claravanstaden: @vgeddes so I've been thinking about this one... Both channel events are queried using the same start and end block range. So if the events are ordered by block number and then processed, I think the scenario you described above won't happen.

This does add requirements to the relayer (to sort the events by block number), which goes against the principle of keeping the relayer as dumb as possible. So I could still add tracking the last verified message separately per channel. Do you envision having more than 2 channels for the bridge in the future?

claravanstaden: One thing that I am also thinking about is that I don't know from which channel the message came in the beacon light client. Could I add that as a parameter to the `verify` method? I could perhaps send the whole message ID (channel and nonce)?

relayer/relays/beacon/main.go
```diff
@@ -110,15 +110,34 @@ func (r *Relay) Sync(ctx context.Context) error {
...
+	r.syncer.Cache.FinalizedHeaders = append(r.syncer.Cache.FinalizedHeaders, lastFinalizedHeader)
+
+	logrus.WithFields(logrus.Fields{
+		"hash": lastFinalizedHeader,
+		"slot": lastFinalizedSlot,
+	}).Info("set cache: last finalized header")
+
+	logrus.Info("starting to sync finalized headers")
+
+	err = r.SyncHeaders(ctx)
```
vgeddes, line 136: I see we end up calling SyncHeaders several times in this method. Here, and then in a loop further below on line 131.

Is this expected? Shouldn't we only call it in the loop?

relayer/relays/beacon/main.go
```diff
@@ -110,15 +110,34 @@ func (r *Relay) Sync(ctx context.Context) error {
...
+	if err != nil {
+		logrus.WithError(err).Error("error while syncing headers")
+	}
```
vgeddes, line 139: need to return the error if its not nil.

relayer/relays/beacon/main.go
```diff
@@ -110,15 +110,34 @@ func (r *Relay) Sync(ctx context.Context) error {
...
+		return err
+	}
 
-	_, _, err = r.SyncFinalizedHeader(ctx)
+	lastFinalizedSlot, err := r.writer.getLastStoredFinalizedHeaderSlot()
 	if err != nil {
+		logrus.WithError(err).Error("unable to get last finalized header slot")
+
 		return err
 	}
```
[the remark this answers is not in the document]
claravanstaden: Addressed in https://github.com/Snowfork/snowbridge/pull/668/commits/27119fdd6815bfb60e1684778a3347fa3d6a1582.

relayer/relays/beacon/main.go
```diff
@@ -110,15 +110,34 @@ func (r *Relay) Sync(ctx context.Context) error {
 		}
 	}
 
-	logrus.Info("done with sync committee updates ")
+	lastFinalizedHeader, err := r.writer.getLastStoredFinalizedHeader()
+	if err != nil {
+		logrus.WithError(err).Error("unable to get last finalized header")
```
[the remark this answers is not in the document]
claravanstaden: Addressed in https://github.com/Snowfork/snowbridge/pull/668/commits/27119fdd6815bfb60e1684778a3347fa3d6a1582.

relayer/relays/beacon/main.go
claravanstaden: I called `SyncHeaders` here too because the ticker started with a delay. I wanted to immediately call `SyncHeaders` (and not wait for the ticker delay) and then call it in the ticker increments. I found a way to call the function immediately, changes in https://github.com/Snowfork/snowbridge/pull/668/commits/27119fdd6815bfb60e1684778a3347fa3d6a1582.

Referenced: https://stackoverflow.com/questions/32705582/how-to-get-time-tick-to-tick-immediately

relayer/relays/beacon/main.go
claravanstaden: Addressed in https://github.com/Snowfork/snowbridge/pull/668/commits/27119fdd6815bfb60e1684778a3347fa3d6a1582.

vgeddes (approved):
LGMT 🚀

PR #671, opened 2022-08-17
Store last verified block number in inbound basic and incentivized channel

claravanstaden:
- stores last synced block number in the basic and incentivized channel
- relayer should backfills messages as finalized headers are synced
- if a last verified message nonce is set, use the nonce to resume message syncing from the last nonce and then continue syncing normally.

Also includes a refactor, header and message syncing are now in separate packages, called `header` and `message`.


relayer/chain/chain.go
```diff
@@ -7,8 +7,9 @@ type Message interface{}
 
 // Message from ethereum
 type EthereumOutboundMessage struct {
-	Call string
-	Args []interface{}
+	Call  string
+	Args  []interface{}
+	Nonce uint64
```
claravanstaden, line 12: I added the `Nonce` field here so that I can match it up with the stored nonce in the `incentivized` and `basic` inbound channels. This struct is used by the Ethereum layer too, but shouldn't have any effect since this new field isn't used in that relayer.

relayer/relays/beacon/header/header.go
```diff
@@ -0,0 +1,262 @@
+package header
```
claravanstaden, line 1: This file has just been moved from `main.go` to `header.go`, no major new changes here.

relayer/relays/beacon/header/header.go
```diff
@@ -0,0 +1,262 @@
...
+			switch {
+			case errors.Is(err, ErrFinalizedHeaderUnchanged):
+				log.WithField("finalized_header", h.cache.LastFinalizedHeader()).Info("finalized header unchanged")
+			case err != nil:
+				log.WithError(err).Error("error while syncing headers")
+
+				return
+			default:
+				basicChannel <- lastBlockNumber
+				incentivizedChannel <- lastBlockNumber
```
claravanstaden, line 97: I've made 2 separate channels for the basic and incentivized channel. I read up on patterns of 1 producer, 2 readers, and it looks like if you want to go that route you need to crate a publish and subscribe mechanism. That felt overkill for just the 2 channels so I left it like this.

relayer/relays/beacon/message/ethereum-listener.go
```diff
@@ -20,8 +21,13 @@ type ParachainPayload struct {
 	Messages []*chain.EthereumOutboundMessage
 }
 
+type EventContainer struct {
+	Event *etypes.Log
+	Nonce uint64
+}
```
claravanstaden, line 27: This is just a container struct to easily return the EthereumOutboundMessage nonce too.

relayer/relays/beacon/header/header.go
```diff
@@ -0,0 +1,262 @@
...
+	basicChannel := make(chan uint64)
+	incentivizedChannel := make(chan uint64)
+
+	go func() {
+		for {
+			lastBlockNumber, err := h.SyncHeaders(ctx)
+			switch {
+			case errors.Is(err, ErrFinalizedHeaderUnchanged):
+				log.WithField("finalized_header", h.cache.LastFinalizedHeader()).Info("finalized header unchanged")
+			case err != nil:
```
vgeddes, line 91: If we hit this error case, the goroutine will silently die, and then no more headers will be sync'ed until the process is restarted.

This has been a common issue in our relayer during its development, so we've moved over to spawning all goroutines using [errgroup](https://pkg.go.dev/golang.org/x/sync/errgroup). It allows the relayer to detect a failed goroutine and self-terminate itself (roughly equivalent to a  fail-fast error handling approach).

Can see an example of this approach in our old ethereum relay:
1. Spawn goroutine using errgroup: https://github.com/Snowfork/snowbridge/blob/f719545fe8d8cf9b5e0665b690c73077ac49258c/relayer/relays/ethereum/ethereum-listener.go#L125
2. If the goroutine is meant to run for ever, make sure it can self-terminate if the associated context.Context is done. (This is needed for orderly termination of all goroutines if one of them fails, or the process is signalled with SIGINT or something like that): https://github.com/Snowfork/snowbridge/blob/f719545fe8d8cf9b5e0665b690c73077ac49258c/relayer/relays/ethereum/ethereum-listener.go#L150
3. The process waits until all goroutines complete or one of them fails: https://github.com/Snowfork/snowbridge/blob/f719545fe8d8cf9b5e0665b690c73077ac49258c/relayer/cmd/run/ethereum/command.go#L97

relayer/relays/beacon/header/header.go
```diff
@@ -0,0 +1,262 @@
...
+
+func New(writer *writer.ParachainWriter, beaconEndpoint string, slotsInEpoch uint64, epochsPerSyncCommitteePeriod uint64) Header {
+	return Header{
+		cache:  cache.New(),
+		writer: writer,
+		syncer: syncer.New(beaconEndpoint, slotsInEpoch, epochsPerSyncCommitteePeriod),
+	}
+}
+
+func (h *Header) Sync(ctx context.Context) (<-chan uint64, <-chan uint64, error) {
```
vgeddes, line 34: Should also pass in the associated errgroup to spawn a goroutine (see my comment below).

parachain/pallets/basic-channel/src/inbound/mod.rs
```diff
@@ -93,7 +96,9 @@ pub mod pallet {
 		pub fn submit(origin: OriginFor<T>, message: Message) -> DispatchResult {
 			ensure_signed(origin)?;
 			// submit message to verifier for verification
-			let log = T::Verifier::verify(&message)?;
+			let result = T::Verifier::verify(&message)?;
```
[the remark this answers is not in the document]
claravanstaden: Cool tip, thanks! https://github.com/Snowfork/snowbridge/pull/671/commits/4dc4016fd75cb81bd6e383f2613ff9b042b07131

relayer/relays/beacon/header/header.go
claravanstaden: Good point, thanks for catching this! https://github.com/Snowfork/snowbridge/pull/671/commits/80b7345618449fb8643e4b8b90ed12486efbe073

relayer/relays/beacon/main.go
```diff
@@ -40,281 +31,53 @@ func NewRelay(
...
-}
-
-func (r *Relay) writeMessages(ctx context.Context, payload ParachainPayload) error {
-	for _, msg := range payload.Messages {
-		err := r.writer.WriteToParachain(ctx, msg.Call, msg.Args...)
-		if err != nil {
-			return err
-		}
-	}
+	go messages.SyncBasic(ctx, basicChannel)
```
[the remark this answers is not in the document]
claravanstaden: Addressed in https://github.com/Snowfork/snowbridge/pull/671/commits/80b7345618449fb8643e4b8b90ed12486efbe073.

relayer/relays/beacon/header/header.go
```diff
@@ -0,0 +1,266 @@
...
+
+				return err
+			default:
+				basicChannel <- lastBlockNumber
+				incentivizedChannel <- lastBlockNumber
+			}
+
+			select {
+			case <-ctx.Done():
+				return ctx.Err()
```
vgeddes, line 106: Should rather return `nil`. This ensures that upon an error, the subsequent `errgroup.Wait()` call doesn't return this `ctx.Err()` but rather the actual error returned by some other gouroutine that failed. 

```suggestion
				return nil
```

claravanstaden: Cool! Committed 😄

alistair-singh (approved):

PR #673, opened 2022-08-25
Test Georli network

claravanstaden:
- [x] Add Goerli network to Hardhat config
- [x] Test `start-services.sh`
- [x] Test `Bootstrap` tests

Resolves [SNO-303](https://linear.app/snowfork/issue/SNO-303)


claravanstaden:
To test against Goerli, I had to upgrade Lodestar to 0.41, and as I was already upgrading Lodestar, I figured we should upgrade to the latest version, while we are making changes. I ran Lodestar 1.0.0, but ran into memory issues using Geth 1.10.18, so I upgraded Geth to the latest version as well (1.10.23). The latest Geth introduced a change that required us to specify the syncmode for a local testnet, see related issue https://github.com/ethereum/go-ethereum/issues/25633#issuecomment-1232976006.

The Bootstrap tests against Goerli works passes, both using https://lodestar-goerli.chainsafe.io and my own Goerli node. There is one related issue that I need to address: https://linear.app/snowfork/issue/SNO-321 However this shouldn't block this PR to be merged. I saw the BEEFY relayer restart when running Goerli tests, but I ran it a few times again this morning and haven't come across it again.

alistair-singh (approved):
🚢

PR #676, opened 2022-08-26
Remove PoW light client from runtimes

vgeddes:


claravanstaden (approved):
Nice!

I'm just wondering if `start-services.sh` won't give an error because the genesis config for the ethereum light client is still set [here](https://github.com/Snowfork/snowbridge/blob/main/test/scripts/start-services.sh#L165)? Maybe that will be ignored in the chain spec but I guess we can remove it as well?

vgeddes:
> Nice!
> 
> I'm just wondering if `start-services.sh` won't give an error because the genesis config for the ethereum light client is still set [here](https://github.com/Snowfork/snowbridge/blob/main/test/scripts/start-services.sh#L165)? Maybe that will be ignored in the chain spec but I guess we can remove it as well?

ah, good catch! let me update

PR #678, opened 2022-08-26
Misc fixes for query-event-tool

vgeddes:
Subxt 0.23.0 seems to be buggy. David and I get `IncompatibleMetadataError` errors when `snowbridge-query-events` is built with with version. But subxt master seems to work.

Also added docs about fetching new metadata


alistair-singh (approved):
🚀

PR #679, opened 2022-08-30
SNO-306: parachain -> ethereum E2E test

doubledup:
Figured it out: generateHashSides only worked for leaves in the last level of the Merkle tree, not leaves in the second-last level. So with 6 bundles, the first 4 were working because they're in the last level. The last 2 are now being relayed as well 🙌 

[SNO-306](https://linear.app/snowfork/issue/SNO-306)

### DIFF
--- a/relayer/relays/parachain/beefy-listener.go
+++ b/relayer/relays/parachain/beefy-listener.go
@@ -632,7 +632,8 @@ func (li *BeefyListener) scanForCommitments(
 					digestItemHash,
 					basicChannelAccountNonces,
 					basicChannelScanAccounts,
-					events.Basic.Bundles)
+					events.Basic.Bundles,
+				)
 				if err != nil {
 					return nil, err
 				}
--- a/relayer/relays/parachain/beefy-listener.go
+++ b/relayer/relays/parachain/beefy-listener.go
@@ -13,7 +13,6 @@ import (
 	"golang.org/x/sync/errgroup"
 
 	gsrpc "github.com/snowfork/go-substrate-rpc-client/v4"
-	"github.com/snowfork/go-substrate-rpc-client/v4/rpc/offchain"
 	"github.com/snowfork/go-substrate-rpc-client/v4/types"
 	"github.com/snowfork/snowbridge/relayer/chain/ethereum"
 	"github.com/snowfork/snowbridge/relayer/chain/parachain"
@@ -797,27 +796,4 @@ func fetchBundleProof(
 type OffchainStorageValue struct {
 	Nonce      uint64
 	Commitment []byte
-}
-
-func (li *BeefyListener) fetchOffchainData(digestItem AuxiliaryDigestItem) (*OffchainStorageValue, error) {
-	storageKey, err := makeStorageKey(digestItem.AsCommitment.ChannelID, digestItem.AsCommitment.Hash)
-	if err != nil {
-		return nil, err
-	}
-
-	data, err := li.parachainConnection.API().RPC.Offchain.LocalStorageGet(offchain.Persistent, storageKey)
-	if err != nil {
-		return nil, fmt.Errorf("read commitment from offchain storage: %w", err)
-	}
-	if data == nil {
-		return nil, fmt.Errorf("offchain storage item not found")
-	}
-
-	var offchainStorageValue OffchainStorageValue
-	err = types.DecodeFromBytes(*data, &offchainStorageValue)
-	if err != nil {
-		return nil, fmt.Errorf("decode offchain storage value: %w", err)
-	}
-
-	return &offchainStorageValue, nil
 }
--- a/relayer/relays/parachain/beefy-listener.go
+++ b/relayer/relays/parachain/beefy-listener.go
@@ -727,6 +727,10 @@ func (li *BeefyListener) scanForCommitments(
 			}
 		}
 
+		for _, proof := range basicChannelProofs {
+			log.Debugf("proof root: %v", proof.Root)
+		}
+
 		if len(basicChannelProofs) > 0 || incentivizedChannelCommitment != nil {
 			task := Task{
 				ParaID:                        li.paraID,
@@ -767,11 +771,13 @@ func fetchBundleProof(
 	}
 
 	err = types.DecodeFromHexString(proofHex, &rawProof)
+	log.Debugf("raw proof: %+v", rawProof)
 	if err != nil {
 		return proof, fmt.Errorf("decode merkle proof: %w", err)
 	}
 
 	proof, err = NewMerkleProof(rawProof, bundle)
+	log.Debugf("proof: %+v", proof)
 	if err != nil {
 		return proof, fmt.Errorf("decode merkle proof: %w", err)
 	}
--- a/relayer/relays/parachain/beefy-listener.go
+++ b/relayer/relays/parachain/beefy-listener.go
@@ -670,10 +670,6 @@ func (li *BeefyListener) scanForCommitments(
 			}
 		}
 
-		for _, proof := range basicChannelProofs {
-			log.Debugf("proof root: %v", proof.Root)
-		}
-
 		if len(basicChannelProofs) > 0 || incentivizedChannelCommitment != nil {
 			task := Task{
 				ParaID:                        li.paraID,
@@ -785,13 +781,11 @@ func fetchBundleProof(
 	}
 
 	err = types.DecodeFromHexString(proofHex, &rawProof)
-	log.Debugf("raw proof: %+v", rawProof)
 	if err != nil {
 		return proof, fmt.Errorf("decode merkle proof: %w", err)
 	}
 
 	proof, err = NewMerkleProof(rawProof, bundle)
-	log.Debugf("proof: %+v", proof)
 	if err != nil {
 		return proof, fmt.Errorf("decode merkle proof: %w", err)
 	}
--- a/relayer/relays/parachain/beefy-listener.go
+++ b/relayer/relays/parachain/beefy-listener.go
@@ -698,12 +698,10 @@ func (li *BeefyListener) scanForCommitments(
 					continue
 				} else if bundleNonce > incentivizedNonceToFind {
 					// Collect these commitments
-					commitment := NewIncentivizedChannelCommitment(digestItem.AsCommitment.Hash, bundle)
-					incentivizedChannelCommitment = &commitment
+					incentivizedChannelCommitment = NewIncentivizedChannelCommitment(digestItem.AsCommitment.Hash, bundle)
 				} else if bundleNonce == incentivizedNonceToFind {
 					// Collect this commitment and terminate scan
-					commitment := NewIncentivizedChannelCommitment(digestItem.AsCommitment.Hash, bundle)
-					incentivizedChannelCommitment = &commitment
+					incentivizedChannelCommitment = NewIncentivizedChannelCommitment(digestItem.AsCommitment.Hash, bundle)
 					scanIncentivizedChannelDone = true
 				}
 			}
--- a/relayer/relays/parachain/beefy-listener.go
+++ b/relayer/relays/parachain/beefy-listener.go
@@ -306,11 +306,6 @@ type AccountNonces struct {
 	paraBasicNonce, ethBasicNonce uint64
 }
 
-type AccountNonceToFind struct {
-	account     [32]byte
-	nonceToFind uint64
-}
-
 // discoverCatchupTasks finds all the commitments which need to be relayed
 func (li *BeefyListener) discoverCatchupTasks(
 	ctx context.Context,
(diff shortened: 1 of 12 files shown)


test/test/basic-channel.js
```diff
@@ -0,0 +1,96 @@
...
+      const afterEthBalance = await ethClient.getDotBalance(ethAccount);
+      const afterSubBalanceAlice = await subClient.queryAccountBalance(polkadotSenderSS58Alice);
+      const afterSubBalanceBob = await subClient.queryAccountBalance(polkadotSenderSS58Bob);
+      const afterSubBalanceCharlie = await subClient.queryAccountBalance(polkadotSenderSS58Charlie);
+      const afterSubBalanceDave = await subClient.queryAccountBalance(polkadotSenderSS58Dave);
+      const afterSubBalanceEve = await subClient.queryAccountBalance(polkadotSenderSS58Eve);
+      const afterSubBalanceFerdie = await subClient.queryAccountBalance(polkadotSenderSS58Ferdie);
+
+      expect(afterEthBalance.minus(beforeEthBalance)).to.be.bignumber.equal(6 * amountWrapped);
+      expect(beforeSubBalanceAlice.minus(afterSubBalanceAlice)).to.be.bignumber.gte(amount);
```
doubledup, line 88: This was a bit painful: turns out that chai has greaterThanOrEqual, but that failed here because it expected a number or a date. The [chai-bignumber plugin](https://www.chaijs.com/plugins/chai-bignumber/) we're using here has *almost* all comparison assertions, except lessThanOrEqual and greaterThanOrEqual. Those are still the base chai assertions, which fail when used on BigNumbers.

test/test/basic-channel.js
```diff
@@ -0,0 +1,96 @@
...
+      const beforeSubBalanceCharlie = await subClient.queryAccountBalance(polkadotSenderSS58Charlie);
+      const beforeSubBalanceDave = await subClient.queryAccountBalance(polkadotSenderSS58Dave);
+      const beforeSubBalanceEve = await subClient.queryAccountBalance(polkadotSenderSS58Eve);
+      const beforeSubBalanceFerdie = await subClient.queryAccountBalance(polkadotSenderSS58Ferdie);
+
+      // lock DOT using basic channel
+      await subClient.api.tx.sudo.sudo(
+        subClient.api.tx.utility.batchAll([
+          subClient.api.tx.utility.dispatchAs(
+            {system: {signed: subClient.alice.address}},
```
doubledup, line 52: Luckily found an example of this in an [Acala repo](https://github.com/AcalaNetwork/bodhi.js/blob/930cb66af4b5f94978c683dd390fb18778f720ac/scripts/virtual-tx-test.ts#L59)

relayer/relays/parachain/types.go
```diff
@@ -1,43 +1,131 @@
...
+	hashSides, err := generateHashSides(rawProof)
+	if err != nil {
+		return proof, err
+	}
+
+	proof = MerkleProof{
+		Root:      rawProof.Root,
+		Proof:     byteArrayProof,
+		HashSides: hashSides,
+		Leaf:      bundle,
```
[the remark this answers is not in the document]
doubledup: I'm not sure how much we gain from this extra validation. The bundle in the event is the same data as in the proof's leaf. Both are emitted from the parachain, only difference is that the event is SCALE-encoded while the proof's leaf is ethabi-encoded.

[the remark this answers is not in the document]
doubledup: I'm leaning towards removing the leaf field (the message bundle) from the proof that the RPC returns.
That way, the only source of message bundles for relayers is the events. The events and proofs come from the parachain anyway, so we shouldn't need to validate them until we get to the inbound channel contract. If a bundle is paired with the wrong proof, the contract will reject it.

relayer/relays/parachain/beefy-listener.go
```diff
@@ -696,26 +696,32 @@ func (li *BeefyListener) scanForCommitments(
 				if bundleNonce < incentivizedNonceToFind {
 					scanIncentivizedChannelDone = true
 					continue
-					// Collect these commitments
 				} else if bundleNonce > incentivizedNonceToFind {
-					commitments[channelID] = append(commitments[channelID], NewCommitment(digestItem.AsCommitment.Hash, bundle))
-					// collect this commitment and terminate scan
+					// Collect these commitments
+					commitment := NewIncentivizedChannelCommitment(digestItem.AsCommitment.Hash, bundle)
+					incentivizedChannelCommitment = &commitment
```
claravanstaden, line 702: Perhaps `NewIncentivizedChannelCommitment` can return a reference to `IncentivizedChannelCommitment`, since it seems like it is only used here and in line `705`? That would remove the need for the extra line on `702` and `706`.

relayer/relays/parachain/beefy-listener.go
```diff
@@ -634,45 +628,51 @@ func (li *BeefyListener) scanForCommitments(
...
-					if bundleNonce < accountAndNonce.nonceToFind {
+					if bundleNonce < nonceToFind {
 						log.Debugf(
 							"Halting scan for account '%v'. Messages not committed yet on basic channel",
-							types.HexEncodeToString(accountAndNonce.account[:]),
+							types.HexEncodeToString(bundle.Account[:]),
 						)
-						delete(basicAccountsToCheck, accountAndNonce.account)
-						scanBasicChannelDone = len(basicAccountsToCheck) == 0
+						delete(scanBasicChannelAccounts, bundle.Account)
```
claravanstaden, line 649: Not directly related to this bit of code, but this method has become deeply nested and a little unwieldy (no doubt from all the new functionality we've introduced! 😄 ). It could benefit from splitting chunks of logic like these into smaller methods to make it more digestible.

relayer/relays/parachain/beefy-listener.go
```diff
@@ -735,20 +741,30 @@ func fetchBundleProof(
...
 	if err != nil {
-		return basicChannelBundleProof, fmt.Errorf("decode merkle proof: %w", err)
+		return proof, fmt.Errorf("decode merkle proof: %w", err)
 	}
 
-	return basicChannelBundleProof, nil
+	return proof, nil
 }
 
 func bundleIndexForAccount(bundles []BasicOutboundChannelMessageBundle, account *[32]byte) int {
```
claravanstaden, line 770: This method doesn't seem to be used anymore and could probably be removed.

claravanstaden: `fetchOffchainData` also doesn't seem to be used anymore.

relayer/relays/parachain/channels.go
```diff
@@ -27,7 +27,7 @@ func (b BasicOutboundChannelMessageBundle) IntoInboundMessageBundle() basic.Basi
 
 type BasicOutboundChannelMessageBundle struct {
 	SourceChannelID uint8
-	Account         [32]byte
+	Account         types.AccountID
```
claravanstaden, line 30: Nice! ✨

relayer/relays/parachain/types.go
```diff
@@ -1,45 +1,131 @@
...
+type IncentivizedChannelCommitment struct {
 	Hash types.H256
-	Data interface{}
+	Data IncentivizedOutboundChannelMessageBundle
 }
 
-func NewCommitment(hash types.H256, data interface{}) Commitment {
-	return Commitment{
+func NewIncentivizedChannelCommitment(hash types.H256, data IncentivizedOutboundChannelMessageBundle) IncentivizedChannelCommitment {
+	return IncentivizedChannelCommitment{
```
claravanstaden, line 32: As noted in comments for the `beefy-listener`, you can probably just return `&IncentivizedChannelCommitment` here.

claravanstaden (commented):
Nice work!! 👏 What a tricky thing to find. I've made a few minor comments but looks great!

relayer/relays/parachain/beefy-listener.go
```diff
@@ -702,26 +716,30 @@ func (li *BeefyListener) scanForCommitments(
...
+					// Collect this commitment and terminate scan
+					*incentivizedChannelCommitment = NewIncentivizedChannelCommitment(digestItem.AsCommitment.Hash, bundle)
 					scanIncentivizedChannelDone = true
 				}
 			}
 		}
 
-		if len(commitments) > 0 {
+		for _, proof := range basicChannelProofs {
+			log.Debugf("proof root: %v", proof.Root)
```
vgeddes, line 713: Looks like this was for debugging and can be removed?

relayer/relays/parachain/beefy-listener.go
```diff
@@ -741,20 +759,30 @@ func fetchBundleProof(
...
+	var proof MerkleProof
+
 	err := api.Client.Call(&proofHex, "basicOutboundChannel_getMerkleProof", digestItem.AsCommitment.Hash.Hex(), bundleIndex)
 	if err != nil {
-		return basicChannelBundleProof, fmt.Errorf("call rpc basicOutboundChannel_getMerkleProof(%v, %v): %w", digestItem.AsCommitment.Hash.Hex(), bundleIndex, err)
+		return proof, fmt.Errorf("call rpc basicOutboundChannel_getMerkleProof(%v, %v): %w", digestItem.AsCommitment.Hash.Hex(), bundleIndex, err)
+	}
+
+	err = types.DecodeFromHexString(proofHex, &rawProof)
+	log.Debugf("raw proof: %+v", rawProof)
```
vgeddes, line 756: can these debug logs be removed?

relayer/relays/parachain/beefy-listener.go
```diff
@@ -741,20 +759,30 @@ func fetchBundleProof(
...
+	if err != nil {
+		return proof, fmt.Errorf("decode merkle proof: %w", err)
 	}
 
-	err = types.DecodeFromHexString(proofHex, &basicChannelBundleProof)
+	proof, err = NewMerkleProof(rawProof, bundle)
+	log.Debugf("proof: %+v", proof)
```
vgeddes, line 762: ditto

relayer/relays/parachain/types.go
```diff
@@ -1,43 +1,131 @@
...
+		Root:      rawProof.Root,
+		Proof:     byteArrayProof,
+		HashSides: hashSides,
+		Leaf:      bundle,
+	}
+
+	return proof, nil
+}
+
+func generateHashSides(commitmentProof RawMerkleProof) ([]bool, error) {
```
vgeddes, line 76: Can we give this function a more generic signature. Would make it easier to test.

```suggestion
func generateHashSides(nodePosition int, breadth int) ([]bool, error) {
```

vgeddes: I'd also add a few test cases, mostly to guard against future regressions.

test/scripts/start-services.sh
```diff
@@ -15,7 +15,7 @@ beefy_relay_eth_key="${BEEFY_RELAY_ETH_KEY:-0x935b65c833ced92c43ef9de6bff30703d9
 
 # Accounts for which the relayer will relay messages over the basic channel.
```
vgeddes, line 16: Should also comment that these are addresses for Alice, Bob, etc...

test/test/basic-channel.js
vgeddes: Nice! yeah I also search for code all the time on github to help with undocumented APIs

relayer/relays/parachain/types.go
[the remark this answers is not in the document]
vgeddes: I'm fine with that. I did intend for events querying to be the primary API for retrieving for bundles. :+1:

Though I would leave the leaf field in the RawMerkleProof RPC result. It may be useful for troubleshooting further down the line.

vgeddes (commented):
Looks good! Added some minor comments.

relayer/relays/parachain/beefy-listener.go
doubledup: Will do. I was trying to stick to values, especially for a constructor function like this. But this is the only place it's used and we need a pointer for `Task` anyway

relayer/relays/parachain/beefy-listener.go
doubledup: Alistair mentioned this too, I'll find a way to make this clearer 👌

relayer/relays/parachain/beefy-listener.go
doubledup: Removed `bundleIndexForAccount`, but I can't see `fetchOffchainData` used in [the commit where it was added](https://github.com/Snowfork/snowbridge/commit/18c6225b21782170156729d54a35404d876a2c7b). @vgeddes are there any plans to use `fetchOffchainData`?

relayer/relays/parachain/types.go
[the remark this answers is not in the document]
doubledup: I've left the leaf field in the raw proof, but removed it from `MerkleProof`. There's a `BundleProof` struct now that pairs a `MerkleProof` with its message bundle.

test/test/basic-channel.js
doubledup: I'm not sure I understand here. This was more about incompleteness in the chai-bignumber plugin than the way we're testing. What pain points have you seen with using dotApp & co to test?

alistair-singh: Sorry for not being clear, it is not directly related. I have run into this issue before where I have to convert down to `Number` or `String` to assert using chai.

My comment was more general. Channels are a low level concept in the bridge. Where the various apps (e.g. dotApp) are higher level. And we are using the high level apps in this test case to test features of the lower level channel. And it would be nice to have a way to test channels directly without high level apps. Maybe instead of asserting balance updates to test the permission-less channel feature we could assert against the existence of events that the channels emit.

Also our tests which use balances for assertions are side effecting and running them can affect the outcome of other tests (e.g Depleting balances). This is fine for the rest of our tests because they are testing the high level apps directly as well as the underlying channel. I think this is our first test which is isolated to testing a feature of a channel.

But this is non blocking and is more of an observation.

vgeddes (approved):
:+1:

test/test/basic-channel.js
doubledup: Made a PR to add the aliases [here](https://github.com/asmarques/chai-bignumber/pull/15).

Chatted a bit offline, and testing the channel separately would be ideal but we'd have to have an extrinsic to send messages across the channel. We don't want to expose `submit` directly, so a `TestChannelApp` would be best. But that's a fair bit of work for limited gain, given how simple `DOTApp` is.

@vgeddes @claravanstaden Do you think it's worth isolating this test to the channel? Using `DOTApp` seems good enough to me, but I might be missing something.

PR #680, opened 2022-08-31
Filter out parathreads

vgeddes:
See my comments at https://linear.app/snowfork/issue/SNO-320#comment-2afe34e2


alistair-singh (approved):
Nice find!

PR #682, opened 2022-09-05
Improve event query tool

vgeddes:
The event query tool is used by the relayer to query the parachain for events emitted by the outbound channels.

The tool is hardcoded to use APIs and types from the `snowbase` runtime. To make it more production suitable, we should make it support the `snowblink` runtime too.

* To compile for snowbase:
```
cargo build --features parachain-snowbase ...
```

* To compile for snowblink:
```
cargo build --features parachain-snowblink ...
```


claravanstaden (approved):
Nice! 🙌 

How does the snowbridge runtime fit into this?

alistair-singh (approved):

vgeddes:
> 
> How does the snowbridge runtime fit into this?

The snowbridge runtime is meant to be used for Kusama and production Polkadot networks. Since we're not on those networks yet, the runtime is basically unused and dormant.

PR #686, opened 2022-09-16
Implements Benchmarking and Weight for Beacon Client

claravanstaden:
Resolves [SNO-314](https://linear.app/snowfork/issue/SNO-314)

- Note: I had to regenerate the test data since the initial sync and sync committee updates didn't contain adjacent sync periods, which I required for the benchmarks.


parachain/Cargo.toml
```diff
@@ -95,12 +95,14 @@ substrate-build-script-utils = { git = "https://github.com/paritytech/substrate.
...
-    "snowbase-runtime/runtime-benchmarks",
-    "polkadot-service/runtime-benchmarks"
-]
+
+snowbridge-runtime-benchmarks = [ "snowbridge-runtime/runtime-benchmarks" ]
+snowblink-runtime-benchmarks = [ "snowblink-runtime/runtime-benchmarks" ]
+snowbase-runtime-benchmarks = [ "snowbase-runtime/runtime-benchmarks" ]
+polkadot-runtime-benchmarks = [ "polkadot-service/runtime-benchmarks" ]
+
+runtime-benchmarks = []
```
claravanstaden, line 104: This change is to prevent all the runtimes from being included at once, and doing so enabling the `minimal` feature for all runtimes, which then breaks the `snowblink` and `snowbridge` benchmarks for the beacon pallet.

The downside is that if we want to run benchmarks for a runtime, we need to include 2 features:

`features = "snowblink-runtime-benchmarks,runtime-benchmarks"`

We cannot just have the runtime benchmark feature (i.e. `snowblink-runtime-benchmarks`) because Substrate checks if the pallet was compiled with feature `runtime-benchmarks`, so we need both.

relayer/relays/beacon/header/syncer/api.go
```diff
@@ -416,59 +414,6 @@ func (b *BeaconClient) GetSyncCommitteePeriodUpdate(from uint64) (SyncCommitteeP
...
-
-type SyncCommitteeIndexes struct {
-	Indexes []uint64
-}
-
-func (b *BeaconClient) GetSyncCommittee(epoch uint64) (SyncCommitteeIndexes, error) {
-	req, err := http.NewRequest(http.MethodGet, fmt.Sprintf("%s/eth/v1/beacon/states/finalized/sync_committees?epoch=%v", b.endpoint, epoch), nil)
-	if err != nil {
-		return SyncCommitteeIndexes{}, fmt.Errorf("%s: %w", ConstructRequestErrorMessage, err)
-	}
```
claravanstaden, line 433: These methods weren't used.

alistair-singh (approved):

parachain/Cargo.toml
vgeddes: Ok, yeah I see what you mean. Not ideal but I don't see another way around it.

parachain/pallets/ethereum-beacon-client/src/weights.rs
```diff
@@ -0,0 +1,48 @@
...
+use frame_support::{traits::Get, weights::{Weight, constants::RocksDbWeight}};
+use sp_std::marker::PhantomData;
+
+/// Weight functions needed for ethereum_beacon_client.
+pub trait WeightInfo {	fn sync_committee_period_update() -> Weight;	fn import_finalized_header() -> Weight;	fn import_execution_header() -> Weight;}
+
+/// Weights for ethereum_beacon_client using the Snowbridge node and recommended hardware.
+pub struct SnowbridgeWeight<T>(PhantomData<T>);
+impl<T: frame_system::Config> WeightInfo for SnowbridgeWeight<T> {	fn sync_committee_period_update() -> Weight {
+		(170_683_416_000 as Weight)			.saturating_add(T::DbWeight::get().reads(4 as Weight))			.saturating_add(T::DbWeight::get().writes(2 as Weight))	}	fn import_finalized_header() -> Weight {
```
vgeddes, line 40: Are these weights generated using full mainnet config? Should make sure that we default to the most expensive weights, even if it isn't accurate for snowbase runtime

vgeddes (approved):

parachain/pallets/ethereum-beacon-client/src/weights.rs
claravanstaden: Yes, they are generated using Snowblink using the mainnet config. I figured the same. 😄

PR #687, opened 2022-09-16
SNO-325: Enable multiple accounts in inbound channel pallet

doubledup:
Sorted out the new test data & added a section in the parachain README to help the next time we need to do it.

[SNO-325](https://linear.app/snowfork/issue/SNO-325)


parachain/primitives/core/src/types.rs
```diff
@@ -3,20 +3,14 @@
...
-	pub channel_id: ChannelId,
-	pub nonce: u64,
-}
-
-impl MessageId {
-	pub fn new(channel_id: ChannelId, nonce: u64) -> Self {
-		Self { channel_id, nonce }
-	}
+pub enum MessageId {
+	Basic(ChannelId, H160, u64),
```
doubledup, line 12: @vgeddes I'm not sure about adding the origin address to `MessageId` here, vs adding an extra id field to the message event from the outbound contract like we discussed.
Looks like the event's id field will just be the concatenation of the origin address and the nonce. And we need the separate origin address and nonce in the contract's event for the `Nonces` storage map anyway. Does it make more sense to do this concatenation only on the parachain, or to add a separate id field to the `Message` event in the contract?

parachain/primitives/core/src/types.rs
```diff
@@ -3,20 +3,14 @@
...
+	Incentivized(ChannelId, u64),
```
doubledup, line 13: @vgeddes I'm also not happy with the use of `enum` for `MessageId` yet. The `ChannelId` field is redundant because we already have that in the enum variant. But to remove `ChannelId` we'd need custom `Encode` and `Decode` implementations that include the channel id in the serialized `MessageId`. Is it worth going to that effort? As-is, incorrectly use looks like `MessageId::Basic(ChannelId::Incentivized...`, which seems easy to spot.

parachain/pallets/basic-channel/src/inbound/mod.rs
```diff
@@ -117,7 +118,7 @@ pub mod pallet {
 				}
 			})?;
 
-			let message_id = MessageId::new(ChannelId::Basic, envelope.nonce);
+			let message_id = MessageId::Basic(ChannelId::Basic, envelope.origin, envelope.nonce);
 			T::MessageDispatch::dispatch(envelope.source, message_id, &envelope.payload);
```
doubledup, line 122: @vgeddes We call the source address `origin` in the [dispatch call](https://github.com/Snowfork/snowbridge/blob/29a05937b5775be5b644f04888a75051be809484/parachain/pallets/dispatch/src/lib.rs#L121-L122). Is this a change in the meaning of "origin" in the dispatch pallet vs the outbound contract?

ie. in the contract "source" is an app and "origin" is a user, while the dispatch pallet uses "origin" to mean the app that sent the message.

parachain/primitives/core/src/types.rs
alistair-singh: I didnt see this as a redundancy. The enum represents the channel type where as channel id represents an instance of a channel. Unless we are saying that we will always have one instance of a channel per channel type. i.e. We will never have two different incentivized channels that we use for different purposes?

parachain/primitives/core/src/types.rs
vgeddes: Yeah I would not add a redundant ID field to the `Message`, since we can generate the full ID on the parachain using the various fields in Message.

To add more background info, this `MessageId` type isn't really used in the core bridging protocol.  When a message is dispatched on the parachain, a `MessageDispatched` event is emitted [here](https://github.com/Snowfork/snowbridge/blob/820ec5d8d8e1a9973790e05ccc9f7d3ca88fac53/parachain/pallets/dispatch/src/lib.rs#L125), with the `MessageId` as a member. Offchain clients query these events to get progress updates on bridge transfers, and so the ID is necessary to correlate events happening on either chain.

parachain/primitives/core/src/types.rs
vgeddes: @doubledup moving over to enum's here is the right approach. And you can drop the `ChannelId` field, isn't needed at all since client's can inspect the enum variant using match statements to figure that out.

Also  see my comment further above for details on `MessageId`. The type actually isn't that important.

parachain/pallets/basic-channel/src/inbound/envelope.rs
```diff
@@ -20,6 +21,8 @@ static EVENT_ABI: &Event = &Event {
 pub struct Envelope {
 	/// The address of the outbound channel on Ethereum that forwarded this message.
 	pub channel: H160,
+	/// The user on Ethereum that authorized the source to send the message.
+	pub origin: H160,
```
vgeddes, line 25: In this context, lets rather name this `user`.

Also, we can switch the order of `source` and `user`, so that user is grouped with nonce?

parachain/pallets/basic-channel/src/inbound/mod.rs
vgeddes: > We call the source address origin in the [dispatch call](https://github.com/Snowfork/snowbridge/blob/29a05937b5775be5b644f04888a75051be809484/parachain/pallets/dispatch/src/lib.rs#L121-L122).

Yeah that's purely a Substrate thing, where an [origin](https://docs.substrate.io/build/origins/) is something that invoked a pallet's dispatchable function. For example, a user could submit a transaction that ends up invoking a pallet dispatchable. Then that user's account is the origin.

In the code you've linked to, we are actually doing something very custom and not seen on many other Substrate chains. The dispatch pallet implements a custom origin which is an ethereum account ID (H160). When the dispatch pallet is asked to dispatch a message, it calls the dispatchable function (encoded in the message), and passes the the contract address of the Ethereum app as the origin.

This is important part of our security model. Our app pallets such as EthApp, DotApp, etc, need to ensure that their pallet dispatchables are being called only by their peer apps on the Ethereum side.

For example, in the code at https://github.com/Snowfork/snowbridge/blob/820ec5d8d8e1a9973790e05ccc9f7d3ca88fac53/parachain/pallets/eth-app/src/lib.rs#L151, the Ethapp::mint() dispatchable checks that the origin is a trusted peer app (EthApp.sol).

parachain/primitives/core/src/types.rs
vgeddes: > Do we have any plans for multiple channels of a given type?

Nope, not yet. Though I do have vague ideas about supporting multiple "lanes" within a channel. And each lane would have different delivery properties (slow, fast, etc). But this is all in the future.

parachain/pallets/basic-channel/src/inbound/mod.rs
```diff
@@ -66,7 +66,8 @@ pub mod pallet {
 	pub type SourceChannel<T: Config> = StorageValue<_, H160, ValueQuery>;
 
 	#[pallet::storage]
-	pub type Nonce<T: Config> = StorageValue<_, u64, ValueQuery>;
+	// TODO: if eth addresses are user-generated, we need to use the blake2_128_concat hasher
```
vgeddes, line 69: Actually, according to the doc below (see section "Real Accounts"), the eth addresses in this context wouldn't really be entirely user generated. Users would need to possess a valid private key that corresponds to user generated address in order to submit ethereum transactions which use the bridge.

https://www.shawntabrizi.com/substrate/transparent-keys-in-substrate/

So the suggestion would be to use the `Twox64Concat` hasher.

parachain/primitives/core/src/types.rs
```diff
@@ -3,20 +3,14 @@
...
-	pub channel_id: ChannelId,
-	pub nonce: u64,
-}
-
-impl MessageId {
-	pub fn new(channel_id: ChannelId, nonce: u64) -> Self {
-		Self { channel_id, nonce }
-	}
+pub enum MessageId {
+	Basic(H160, u64),
```
vgeddes, line 12: Lets use named fields here for improved descriptiveness. like `Basic { user: H160, nonce: u64 }`

parachain/pallets/basic-channel/src/inbound/mod.rs
doubledup: @vgeddes Thanks! I've updated the hasher. If I'm understanding correctly, the eth address creation process is to create a key pair and use the last 20 bytes of `keccak(pubkey)` as the address. So we're relying on that section of a keccak hash being cryptographically secure. An attacker might be able to choose the pubkey and get hashes with a shared prefix from Twox64, but they still can't create storage keys with a shared prefix any easier than brute force because of keccak.

The pipeline I'm imagining for our storage key prefix is this:

pubkey -> keccak256 -> trunc to last 20 bytes -> twox64

Is keccak still secure wrt shared prefixes after being truncated? Haven't found an answer to this online, but seems like it would be a useful thing for Ethereum as a whole.

parachain/README.md
```diff
@@ -65,6 +65,32 @@ For an example configuration, consult the [setup script](https://github.com/Snow
 
 To run the parachain tests locally, use `cargo test --release`. For the full suite of tests, use `cargo test --release --features runtime-benchmarks`.
 
+### Updating test data for inbound channel unit tests
```
alistair-singh, line 68: This feels like it belongs in the `ethereum` sub folder's README?

parachain/pallets/basic-channel/src/inbound/envelope.rs
```diff
@@ -22,6 +23,8 @@ pub struct Envelope {
 	pub channel: H160,
 	/// The application on Ethereum where the message originated from.
 	pub source: H160,
+	/// The user on Ethereum that authorized the source to send the message.
+	pub user: H160,
```
alistair-singh, line 27: This maps to `origin` on the [BasicOutboundChannel](https://github.com/Snowfork/snowbridge/pull/687/files#diff-0c18c33e86533e5721bc116be59e03b14f120dc10c751f4136a99547fdc56cb4R20) end. Why did we change the change name on the parachain side to `user`?

parachain/pallets/basic-channel/src/inbound/envelope.rs
```diff
@@ -63,36 +71,34 @@ mod tests {
 	use super::*;
 	use hex_literal::hex;
 
-	const LOG: [u8; 284] = hex!(
+	const LOG: [u8; 251] = hex!(
```
claravanstaden, line 74: Has this changed just because the test data is now shorter?

parachain/pallets/basic-channel/src/inbound/envelope.rs
```diff
@@ -63,36 +71,34 @@ mod tests {
...
+		0000000000000000000000000000000000000000000000000000206172626974
+		726172792d7061796c6f6164000000000000000000000000000000
 	"
 	);
 
 	#[test]
 	fn test_try_from_log() {
 		let log: Log = rlp::decode(&LOG).unwrap();
 		let envelope = Envelope::try_from(log).unwrap();
+		println!("payload: {:?}",  envelope.payload);
```
claravanstaden, line 91: Perhaps this should be removed?

claravanstaden (approved):
Looks good! 💃

parachain/README.md
doubledup: The ethereum README has a section about updating test data for its unit tests (using E2E tests), and this is the equivalent for updating the parachain unit test data (using ethereum tests). It probably feels like it should move because it's using tests from `ethereum` to generate fixture data, but its purpose is to help update the parachain unit test data. Using the ethereum tests is a detail: if we started getting the data elsewhere (eg. E2E tests), we'd keep this in the parachain README.

parachain/pallets/basic-channel/src/inbound/envelope.rs
doubledup: This came from a [previous comment](https://github.com/Snowfork/snowbridge/pull/687#discussion_r972933413). My understanding is that `origin` has a slightly different meaning [in Substrate](https://docs.substrate.io/reference/glossary/#origin): "... the source of a dispatched function call...", which corresponds to the `source` field here in the envelope or the `operator` in contract authorization. The `user` field here is the address that authorized the message but not necessarily the one that sent it, which is the `origin` in contract authorization.

As I understand it:
- Substrate origin == Parachain envelope source == Contract operator
- Parachain envelope user == Contract origin

Our dispatch pallet already [converts the source to an origin](https://github.com/Snowfork/snowbridge/blob/main/parachain/pallets/dispatch/src/lib.rs#L121), so avoiding using `origin` in the envelope (for a Contract origin) avoids confusing it with a Substrate origin.

cc @vgeddes please keep me honest here.

parachain/pallets/basic-channel/src/inbound/envelope.rs
doubledup: Exactly! I didn't check what made the old log longer 🤔

parachain/pallets/basic-channel/src/inbound/envelope.rs
doubledup: Yup, it's a rogue debug statement 🙈

alistair-singh (approved):

PR #688, opened 2022-09-16
Rate limit backfilling execution headers

claravanstaden:
If the relayer was offline for a long time since the last finalized block, the execution headers that need to be backfilled can pile up quite a bit. This PR reuses outgoing_extrinisc.go in the parachain package and rate limits 10 execution headers synced at a time.

Finalized headers are watched synchronously, as it used to, since the update infrequently.

Message updates can potentially also be rate-limited, but it is not necessary right now (for Rococco).

### DIFF
--- a/relayer/relays/ethereum/parachain-writer.go
+++ b/relayer/relays/ethereum/parachain-writer.go
@@ -17,6 +17,8 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
+const MaxWatchedExtrinsics = 10
+
 type ParachainPayload struct {
 	Header   *chain.Header
 	Messages []*chain.EthereumOutboundMessage
@@ -53,7 +55,7 @@ func (wr *ParachainWriter) Start(ctx context.Context, eg *errgroup.Group) error 
 	}
 	wr.genesisHash = genesisHash
 
-	wr.pool = parachain.NewExtrinsicPool(eg, wr.conn)
+	wr.pool = parachain.NewExtrinsicPool(eg, wr.conn, 10)
 
 	eg.Go(func() error {
 		err := wr.writeLoop(ctx)
--- a/relayer/relays/ethereum/parachain-writer.go
+++ b/relayer/relays/ethereum/parachain-writer.go
@@ -17,28 +17,29 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
-const MaxWatchedExtrinsics = 10
-
 type ParachainPayload struct {
 	Header   *chain.Header
 	Messages []*chain.EthereumOutboundMessage
 }
 
 type ParachainWriter struct {
-	conn        *parachain.Connection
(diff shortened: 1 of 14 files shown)


claravanstaden:
I just realized this might not work with the messages, the nonces might get mixed up. I need to test, will put back in review after.

claravanstaden:
@vgeddes, @alistair-singh, @doubledup  I have tested this, works well! I specified the beacon max watched extrinsics to be `32`. On my EC2 this looks manageable for the parachain, and speeds up execution header importing a bit.

@alistair-singh do you think it would be worthwhile to add the max watched extrinsics as beacon relay config? So we can change it easily without a redeploy?

relayer/relays/ethereum/parachain-writer.go
```diff
@@ -53,7 +55,7 @@ func (wr *ParachainWriter) Start(ctx context.Context, eg *errgroup.Group) error
 	}
 	wr.genesisHash = genesisHash
 
-	wr.pool = parachain.NewExtrinsicPool(eg, wr.conn)
+	wr.pool = parachain.NewExtrinsicPool(eg, wr.conn, 10)
```
alistair-singh, line 58: Does this need to use the constant `MaxWatchedExtrinsics`? And can we set to a larger value such as 32? I think by default the ethereum relayer submits 12 headers at a time
```suggestion
	wr.pool = parachain.NewExtrinsicPool(eg, wr.conn, MaxWatchedExtrinsics)
```

vgeddes:
> I specified the beacon max watched extrinsics to be `32` (so that an epoch can be synced at a time). On my EC2 instance this looks manageable for the parachain, and speeds up execution header importing a bit.

Cool, that's quite some good news. Means that the eth2 light client can verify headers more efficiently than the eth1 light client, and not use up too much of the allowed block weight.

alistair-singh:
> @alistair-singh do you think it would be worthwhile to add the max watched extrinsics as beacon relay config? So we can change it easily without a redeploy?

I think since the code is new and we do not know how catch up is going to behave in prod this might be very useful to add.

vgeddes (approved):
+1

relayer/relays/ethereum/parachain-writer.go
claravanstaden: @alistair-singh I'll add the config! 😄

Whoops, you're right, it should be using the constant.

I left the old Ethereum relayer with MaxWatchedExtrinsics = `10` (as it was), but the Beacon relayer [submits `32` header at a time](https://github.com/Snowfork/snowbridge/pull/688/files#diff-8788c1408defd89dcb3fb8a2b398665badf8ef0fe2abe6451a7b0bad9aafec6aR14).

alistair-singh (approved):

alistair-singh (approved):

PR #690, opened 2022-09-20
Fix first execution header with incorrect sync aggregate

claravanstaden:


relayer/relays/beacon/header/header.go
```diff
@@ -265,7 +265,10 @@ func (h *Header) SyncHeaders(ctx context.Context) error {
 	}).Info("starting to back-fill headers")
 
 	blockRoot := common.HexToHash(finalizedHeader.FinalizedHeader.ParentRoot.Hex())
-	prevSyncAggregate := finalizedHeader.SyncAggregate
```
claravanstaden, line 268: This sync aggregate is actually the `attested_header` sync witness, not the `finalized_header`'s sync witness.

alistair-singh (approved):
🚀

vgeddes (approved):
+1

PR #697, opened 2022-10-12
Vault contract for ETH app.

alistair-singh:
Added a vault to store ether on behalf of EthApp.

* The vault is secured with `Ownable` openzeppelin contract.
* Only one version of EthApp can hold ownership of the vault at a time.
* EthApp controls transfer of vault ownership via messages from the inbound channel.
* Ownership transfer is assumed to be part of a consistent upgrade procedure. i.e. No new transfers across the bridge can be initiated and all pending transfers are complete.

### DIFF
--- a/ethereum/contracts/ETHApp.sol
+++ b/ethereum/contracts/ETHApp.sol
@@ -36,8 +36,6 @@ contract ETHApp is RewardController, AccessControl {
         Channel incentivized
     );
 
-    event VaultTransferred(address newOwner);
-
     bytes2 constant MINT_CALL = 0x4101;
 
     bytes32 public constant REWARD_ROLE = keccak256("REWARD_ROLE");
@@ -199,6 +197,5 @@ contract ETHApp is RewardController, AccessControl {
         address newOwner
     ) external onlyRole(CHANNEL_UPGRADE_ROLE) {
         vault.transferOwnership(newOwner);
-        emit VaultTransferred(newOwner);
     }
 }
--- a/ethereum/contracts/ETHApp.sol
+++ b/ethereum/contracts/ETHApp.sol
@@ -193,4 +193,10 @@ contract ETHApp is RewardController, AccessControl {
         grantRole(INBOUND_CHANNEL_ROLE, _incentivized.inbound);
         emit Upgraded(msg.sender, c1, c2);
     }
+
+     function transferVaultOwnership(
+        address newOwner
+    ) external onlyRole(CHANNEL_UPGRADE_ROLE) {
+        vault.transferOwnership(newOwner);
+    }
 }
--- a/ethereum/contracts/ETHApp.sol
+++ b/ethereum/contracts/ETHApp.sol
@@ -36,6 +36,8 @@ contract ETHApp is RewardController, AccessControl {
         Channel incentivized
     );
 
+    event VaultTransferred(address newOwner);
+
     bytes2 constant MINT_CALL = 0x4101;
 
     bytes32 public constant REWARD_ROLE = keccak256("REWARD_ROLE");
@@ -197,5 +199,6 @@ contract ETHApp is RewardController, AccessControl {
         address newOwner
     ) external onlyRole(CHANNEL_UPGRADE_ROLE) {
         vault.transferOwnership(newOwner);
+        emit VaultTransferred(newOwner);
     }
 }
--- a/ethereum/contracts/ETHApp.sol
+++ b/ethereum/contracts/ETHApp.sol
@@ -92,7 +92,7 @@ contract ETHApp is RewardController, AccessControl {
 
         // revert in case of overflow.
         uint128 value = (msg.value).toUint128();
-        vault.deposit{value: msg.value}();
+        vault.deposit{value: msg.value}(msg.sender);
 
         emit Locked(msg.sender, _recipient, value, _paraId, _fee);
 
--- a/ethereum/contracts/ETHApp.sol
+++ b/ethereum/contracts/ETHApp.sol
@@ -51,7 +51,6 @@ contract ETHApp is RewardController, AccessControl {
     bytes32 public constant CHANNEL_UPGRADE_ROLE =
         keccak256("CHANNEL_UPGRADE_ROLE");
 
-        
     ETHVault public vault;
 
     constructor(
--- a/ethereum/contracts/ETHApp.sol
+++ b/ethereum/contracts/ETHApp.sol
@@ -6,6 +6,7 @@ import "@openzeppelin/contracts/utils/math/SafeCast.sol";
 import "./RewardController.sol";
 import "./ScaleCodec.sol";
 import "./OutboundChannel.sol";
+import "./ETHVault.sol";
 
 enum ChannelId {
     Basic,
@@ -50,11 +51,16 @@ contract ETHApp is RewardController, AccessControl {
     bytes32 public constant CHANNEL_UPGRADE_ROLE =
         keccak256("CHANNEL_UPGRADE_ROLE");
 
+        
+    ETHVault public vault;
+
     constructor(
         address rewarder,
+        ETHVault _vault,
         Channel memory _basic,
         Channel memory _incentivized
     ) {
+        vault = _vault;
 
         Channel storage c1 = channels[ChannelId.Basic];
         c1.inbound = _basic.inbound;
@@ -72,6 +78,10 @@ contract ETHApp is RewardController, AccessControl {
         _setupRole(INBOUND_CHANNEL_ROLE, _incentivized.inbound);
     }
 
+    receive() external payable {
+        require(msg.sender == address(vault), "Only receive funds from vault");
+    }
+
     function lock(
         bytes32 _recipient,
         ChannelId _channelId,
@@ -87,6 +97,7 @@ contract ETHApp is RewardController, AccessControl {
 
         // revert in case of overflow.
         uint128 value = (msg.value).toUint128();
+        vault.lock{value: msg.value}();
 
         emit Locked(msg.sender, _recipient, value, _paraId, _fee);
 
@@ -110,6 +121,7 @@ contract ETHApp is RewardController, AccessControl {
     ) public onlyRole(INBOUND_CHANNEL_ROLE) {
         require(_amount > 0, "Must unlock a positive amount");
 
+        vault.unlock(_amount);
         (bool success, ) = _recipient.call{value: _amount}("");
         require(success, "Unable to send Ether");
         emit Unlocked(_sender, _recipient, _amount);
--- a/ethereum/contracts/ETHApp.sol
+++ b/ethereum/contracts/ETHApp.sol
@@ -77,10 +77,6 @@ contract ETHApp is RewardController, AccessControl {
         _setupRole(INBOUND_CHANNEL_ROLE, _incentivized.inbound);
     }
 
-    receive() external payable {
-        require(msg.sender == address(vault), "Only receive funds from vault");
-    }
-
     function lock(
         bytes32 _recipient,
         ChannelId _channelId,
@@ -96,7 +92,7 @@ contract ETHApp is RewardController, AccessControl {
 
         // revert in case of overflow.
         uint128 value = (msg.value).toUint128();
-        
+
         (bool success, ) = address(vault).call{value: msg.value}("");
         require(success, "Vault must accept funds");
 
@@ -120,11 +116,7 @@ contract ETHApp is RewardController, AccessControl {
         address payable _recipient,
         uint128 _amount
     ) public onlyRole(INBOUND_CHANNEL_ROLE) {
-        require(_amount > 0, "Must unlock a positive amount");
-
-        vault.unlock(_amount);
-        (bool success, ) = _recipient.call{value: _amount}("");
-        require(success, "Unable to send Ether");
+        vault.withdraw(_recipient, _amount);
         emit Unlocked(_sender, _recipient, _amount);
     }
 
--- a/ethereum/contracts/ETHApp.sol
+++ b/ethereum/contracts/ETHApp.sol
@@ -92,7 +92,7 @@ contract ETHApp is RewardController, AccessControl {
 
         // revert in case of overflow.
         uint128 value = (msg.value).toUint128();
-        vault.deposit{value: msg.value}(tx.origin);
+        vault.deposit{value: msg.value}();
 
         emit Locked(msg.sender, _recipient, value, _paraId, _fee);
 
--- a/ethereum/contracts/ETHApp.sol
+++ b/ethereum/contracts/ETHApp.sol
@@ -96,7 +96,9 @@ contract ETHApp is RewardController, AccessControl {
 
         // revert in case of overflow.
         uint128 value = (msg.value).toUint128();
-        vault.lock{value: msg.value}();
+        
+        (bool success, ) = address(vault).call{value: msg.value}("");
+        require(success, "Vault must accept funds");
 
         emit Locked(msg.sender, _recipient, value, _paraId, _fee);
 
--- a/ethereum/contracts/ETHApp.sol
+++ b/ethereum/contracts/ETHApp.sol
@@ -51,7 +51,7 @@ contract ETHApp is RewardController, AccessControl {
     bytes32 public constant CHANNEL_UPGRADE_ROLE =
         keccak256("CHANNEL_UPGRADE_ROLE");
 
-    ETHVault public vault;
+    ETHVault public immutable vault;
 
     constructor(
         address rewarder,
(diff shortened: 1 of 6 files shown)


ethereum/contracts/ETHApp.sol
```diff
@@ -72,6 +78,10 @@ contract ETHApp is RewardController, AccessControl {
         _setupRole(INBOUND_CHANNEL_ROLE, _incentivized.inbound);
     }
 
+    receive() external payable {
```
alistair-singh, line 81: This will allow the vault to transfer funds when calling `unlock`.

ethereum/contracts/ETHApp.sol
```diff
@@ -110,6 +120,7 @@ contract ETHApp is RewardController, AccessControl {
     ) public onlyRole(INBOUND_CHANNEL_ROLE) {
         require(_amount > 0, "Must unlock a positive amount");
 
+        vault.unlock(_amount);
         (bool success, ) = _recipient.call{value: _amount}("");
```
vgeddes, line 124: Double withdrawal here 😄

ethereum/contracts/ETHApp.sol
```diff
@@ -181,4 +192,10 @@ contract ETHApp is RewardController, AccessControl {
         grantRole(INBOUND_CHANNEL_ROLE, _incentivized.inbound);
         emit Upgraded(msg.sender, c1, c2);
     }
+
+     function transferVaultOwnership(
+        address newOwner
+    ) external onlyRole(CHANNEL_UPGRADE_ROLE) {
+        vault.transferOwnership(newOwner);
```
vgeddes, line 199: Should also emit an event, is useful for debugging.

ethereum/contracts/ETHVault.sol
```diff
@@ -0,0 +1,26 @@
...
+    function lock() 
+        public
+        payable
+        onlyOwner
+    {
+        require(msg.value > 0, "Value of transaction must be positive");
+    }
+
+    // Returns ETH to the caller.
+    function unlock(uint128 _amount)
```
vgeddes, line 18: I think at this scope, at the level of vaults, we want to use the verbs deposit/withdraw instead of lock/unlock. Kind like of a real bank.

ethereum/contracts/ETHVault.sol
```diff
@@ -0,0 +1,26 @@
...
+        public
```
vgeddes, line 19: Should use `external` instead of `public`, since we know that `unlock` won't ever be called internally (in the contract its defined). Using `external` saves a small amount of gas.

See details here: https://ethereum.stackexchange.com/questions/19380/external-vs-public-best-practices

ethereum/contracts/ETHVault.sol
```diff
@@ -0,0 +1,26 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity ^0.8.9;
+
+import "@openzeppelin/contracts/access/Ownable.sol";
+
+// Holds ETH on behalf of ETHApp
+contract ETHVault is Ownable {
+    // Accepts ETH from the caller.
+    function lock()
```
vgeddes, line 9: So the ether asset is a special case and we should ideally implement the `receive` function instead: https://docs.soliditylang.org/en/latest/contracts.html#receive-ether-function

ethereum/contracts/ETHVault.sol
```diff
@@ -0,0 +1,26 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity ^0.8.9;
+
+import "@openzeppelin/contracts/access/Ownable.sol";
+
+// Holds ETH on behalf of ETHApp
+contract ETHVault is Ownable {
```
vgeddes, line 7: I'd like us to use `Ether` instead of `ETH` in names. Ether is the proper name of the asset and it reads a bit better than the all-caps ETH. I've been wanting to update our contracts/pallets but its very low-hanging fruit.

Can just rename this contract and filename for now.

ethereum/contracts/ETHVault.sol
vgeddes: Also I think its fine for the anyone untrusted to deposit ether. In fact we need to allow that if there ever comes a time when we need to recover from a hack and add funds to compensate from a hack.

ethereum/contracts/ETHApp.sol
```diff
@@ -50,11 +51,15 @@ contract ETHApp is RewardController, AccessControl {
     bytes32 public constant CHANNEL_UPGRADE_ROLE =
         keccak256("CHANNEL_UPGRADE_ROLE");
 
+    ETHVault public vault;
```
vgeddes, line 54: This storage variable is only ever set once in the constructor so can make it an immutable constant.

Especially since we're designing EthApp to be immutable. Don't think we'll ever need to swap new vaults into an app that has already been deployed.

ethereum/contracts/ETHApp.sol
vgeddes: Hmm, I think its simpler to make the vault give the funds directly to the user, rather than passing the funds upwards to the app and to the user.

ethereum/contracts/ETHVault.sol
```diff
@@ -0,0 +1,26 @@
...
+    }
+
+    // Returns ETH to the caller.
+    function unlock(uint128 _amount)
+        public
+        onlyOwner
+    {
+        require(_amount > 0, "Must unlock a positive amount");
+        (bool success, ) = msg.sender.call{ value: _amount }("");
+        require(success, "Unable to send Ether");
```
vgeddes, line 24: Lets also emit an event `Withdrawn(user, amount)` as debugging aid.

vgeddes (commented):
Good start! Added some suggestions 🚀

ethereum/contracts/ETHApp.sol
vgeddes: Oops, not a double withdrawal. but see my comment above that relates to this.

ethereum/contracts/ETHVault.sol
vgeddes: and we should emit a `Deposit(account, amount)` too, where `account` is the address of the app.

claravanstaden (commented):
I don't know Solidity that well yet, but learnt a ton from this PR and Vincent's comments. 👍

ethereum/contracts/ETHApp.sol
alistair-singh: The `Ownable` contract emits `OwnershipTransferred` when the EtherVault is upgraded. Do we still need to add our own event to this contract?

ethereum/contracts/ETHVault.sol
alistair-singh: Will the `user` here be the `app` which did the withdraw or the recipient?

alistair-singh: I have added both to the withdraw event.

ethereum/contracts/ETHApp.sol
```diff
@@ -88,6 +93,9 @@ contract ETHApp is RewardController, AccessControl {
         // revert in case of overflow.
         uint128 value = (msg.value).toUint128();
 
+        (bool success, ) = address(vault).call{value: msg.value}("");
```
musnit, line 96: what's the intention here? i guess to check that the vault is working? but value will probably be 0 in the constructor tx so seems kinda weird?

ethereum/contracts/EtherVault.sol
```diff
@@ -0,0 +1,26 @@
...
+
+    // Returns ETH to the caller.
+    function withdraw(address payable _recipient, uint256 _amount)
+        external
+        onlyOwner
+    {
+        require(_amount > 0, "Must unlock a positive amount");
+        (bool success, ) = _recipient.call{ value: _amount }("");
+        require(success, "Unable to send Ether");
+        emit Withdraw(msg.sender, _recipient, _amount);
```
musnit, line 24: msg.sender here will always be the ethapp? so kinda a waste to emit?

musnit: also may be a bit redundant to have both Withdraw and Unlocked events?

ethereum/contracts/EtherVault.sol
```diff
@@ -0,0 +1,26 @@
...
+
+import "@openzeppelin/contracts/access/Ownable.sol";
+
+// Holds ETH on behalf of ETHApp
+contract EtherVault is Ownable {
+    event Deposit(address account, uint256 amount);
+    event Withdraw(address account, address recipient, uint256 amount);
+
+    // Accepts ETH from the caller.
+    receive() external payable {
```
musnit, line 12: There is nothing that prevents someone from sending eth directly to the vault, rather than going through the channel. This is fine from a security perspective, but I think bad due to the following:
 - UX issue -> people will likely mistakenly send eth to the vault expecting something to happen when they need to go via channel
 - Accounting issue -> there will be a mismatch between eth in the vault and wrapped eth on polkadot, so annoying/messy from an accounting perspective

We should probably make this onlyOwner too?

ethereum/contracts/EtherVault.sol
```diff
@@ -0,0 +1,26 @@
...
+        emit Deposit(msg.sender, msg.value);
```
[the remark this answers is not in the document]
vgeddes: If we want to scan for events for long historical periods, then the events are not redundant. Vaults can be transferred between multiple apps, and so we lose deposit information (the Lock events) every time ownership is transferred.

For auditing and analysis while in production, I'm very keen for vaults to log everything in detail.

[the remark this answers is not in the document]
musnit: That's fair but keep in mind that every user will need to pay extra for this on every single tx they make, and it could be a significant extra cost. We can't decide on whether this trade off makes sense or not without evaluating the actual cost difference, so we should do that.

ethereum/contracts/EtherVault.sol
vgeddes: > There is nothing that prevents someone from sending eth directly to the vault, rather than going through the channel. This is fine from a security perspective, but I think bad due to the following:
> 
> * UX issue -> people will likely mistakenly send eth to the vault expecting something to happen when they need to go via channel

This is already an issue in all our app contracts. Someone sending Ether to our EtherApp will lose it forever. Same with ERC20App, and DotApp.

> * Accounting issue -> there will be a mismatch between eth in the vault and wrapped eth on polkadot, so annoying/messy from an accounting perspective
> 

Not really possible to prevent this. There is no way to reliable way to prevent a contract from receiving Ether: https://consensys.github.io/smart-contract-best-practices/development-recommendations/general/force-feeding/

> We should probably make this onlyOwner too?

I'd rather not, because it would imply we're trying to prevent unauthorized deposits, when in fact that's not technically possible.

musnit: On accounting issue - yeah fair, not possible to prevent in theory but in practice you can prevent everything except self destruct, so i do think that's still worthwhile.

On UX issue - yeah then we should ideally be fixing this issue across all our contracts, not just this one. It is very, very likely that if we don't fix this, we'll end up with random people messing up and giving us operational tickets to get their funds back due to it. I've seen this happen on almost every other project that accepts tokens/eth directly.

musnit: > I'd rather not, because it would imply we're trying to prevent unauthorized deposits, when in fact that's not technically possible.

We can always just add a comment that acknowledges this and explains that it's just to prevent transfers for UX reasons, not for security purposes and that self-destructs are still fine.

ethereum/contracts/EtherVault.sol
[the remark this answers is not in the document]
vgeddes: My goal for the governance/upgrade project is to optimize for clarity, security, and auditability. Especially as most of our team aren't very experienced with Solidity and the tradeoffs involved with EVM programming.

If our bridge has huge gas issues, it will likely not be due to low-hanging fruit such as event emission. Rather it would be the high level architecture of our contracts and upgradability design. And that's where my responsibility comes in to make sure we have the right design in place that balances those tradeoffs.

We can always remove the events later on if that ends up being our last resort. That's more efficient than spending time now analysing gas costs in detail for event submission.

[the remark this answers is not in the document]
alistair-singh: We could aggregate multiple events to infer this data. Such as the `OwnershipTransferred` event that the vault would emit.

ethereum/contracts/EtherVault.sol
vgeddes: Sounds good :+1:

@alistair-singh yeah, so put a `revert` in the `receive` function, and then add an explicit payable `deposit` function only callable by the app. Would be good to emit a `Deposit` event with the app address and the actual owner of the funds.

ethereum/contracts/EtherVault.sol
```diff
@@ -0,0 +1,34 @@
...
+        revert("Must use deposit function");
+    }
+
+    // Accepts ETH from the caller.
+    function deposit() 
+        external
+        payable
+        onlyOwner
+    {
+        emit Deposit(msg.sender, tx.origin, msg.value);
```
vgeddes, line 21: Use of `tx.origin` isn't recommended: https://consensys.github.io/smart-contract-best-practices/development-recommendations/solidity-specific/tx-origin/

Rather pass the depositee address as a parameter to `deposit`.

Then the EthApp can call `vault.deposit(msg.sender)`, because in that context `msg.sender` will be the account which interacted with EthApp

alistair-singh:
I have added deployment and support scripts for the `EtherVault`. e2e tests pass. This is ready for final review.

ethereum/contracts/EtherVault.sol
```diff
@@ -0,0 +1,34 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity ^0.8.9;
+
+import "@openzeppelin/contracts/access/Ownable.sol";
+
+// Holds ETH on behalf of ETHApp
+contract EtherVault is Ownable {
+    event Deposit(address account, address origin, uint256 amount);
```
doubledup, line 8: Similar to the [Message event](https://github.com/Snowfork/snowbridge/pull/699#discussion_r995592247) in the outbound channel contract, should we avoid the name `origin` to avoid the `tx.origin` association? We could use `sender` to match Withdraw's `recipient`
cc @vgeddes

ethereum/contracts/EtherVault.sol
```diff
@@ -0,0 +1,34 @@
...
+        emit Deposit(msg.sender, _sender, msg.value);
+    }
+
+    // Returns ETH to the caller.
+    function withdraw(address payable _recipient, uint256 _amount)
+        external
+        onlyOwner
+    {
+        require(_amount > 0, "Must unlock a positive amount");
+        (bool success, ) = _recipient.call{ value: _amount }("");
```
doubledup, line 30: Similar to [Ownable's `transferOwnership`](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/access/Ownable.sol#L70), should we require the recipient to be non-zero?

My understanding is that the zero address is the burn address: nobody has the keys to it, so the funds can't be retrieved. It's unlikely that a user would want to transfer eth across the bridge only to immediately burn it. They can always burn the eth later themselves if that's really what they want.

If this is correct, we should also require a non-zero address in the [EthApp pallet's burn call](https://github.com/Snowfork/snowbridge/blob/main/parachain/pallets/eth-app/src/lib.rs#L122), to catch this before it gets across the bridge.

ethereum/test/test_eth_app.js
```diff
@@ -156,29 +172,88 @@ describe("ETHApp", function () {
...
       var iface = new ethers.utils.Interface(ETHApp.abi);
-      let event = iface.decodeEventLog(
+      let unlockEvent = iface.decodeEventLog(
         'Unlocked(bytes32,address,uint128)',
+        receipt.rawLogs[1].data,
+        receipt.rawLogs[1].topics
+      );
+
+      unlockEvent.recipient.should.be.equal(recipient);
+      unlockEvent.amount.eq(ethers.BigNumber.from(amount)).should.be.true;
```
doubledup, line 182:
```suggestion
      unlockEvent.sender.should.be.equal(this.app.address);
      unlockEvent.recipient.should.be.equal(recipient);
      unlockEvent.amount.eq(ethers.BigNumber.from(amount)).should.be.true;
```

doubledup (approved):

ethereum/test/test_eth_app.js
alistair-singh: Addressed in https://github.com/Snowfork/snowbridge/pull/697/commits/9350f767fbc05395172d31650b60bbffa5591c42

ethereum/contracts/EtherVault.sol
alistair-singh: None of our apps originally do this check. It may be useful but I think it is out of scope for this PR.

PR #696, opened 2022-10-06
Clara/sno 329 better checking for header and message sync errors

claravanstaden:
- [x] Keep execution header state in storage in beacon client
- [x] Reverse execution header syncing so that headers are synced sequentially instead of backwards
- [x] Resume syncing execution headers on relay startup 
- [x] Check sync committee updates succeed
- [x] Check message verification succeeded after sending to the inbound pallets

PR #699, opened 2022-10-13
SNO-305: Ethereum -> Parachain: bundle messages by address

doubledup:
Opening this now because some tests are passing in both directions, while the others are timing out. The Basic Channel, Bootstrap and DOT App tests are passing. I haven't seen all of the tests pass yet, but I've been getting timeouts & closed connections with no errors in the logs. The ERC20 App tests in particular look like they're not terminating - running the tests on main now to compare with this branch.

Bumped the default timeout in the E2E test command to 25 mins, up from 6m 40s. This seems to be enough for the tests that have terminated. For the run on main, I'm trying a timeout of 2 hours to ensure that the tests have enough time.


vgeddes:
Thanks David, looks good! Will review properly tomorrow or early next week. In any case lets hold off on merging this until after our demo on Monday.

I think some of Clara's fixes in https://github.com/Snowfork/snowbridge/pull/696 may help with reliability. If not we should do a peer session together and see why those ERC20 tests are so troublesome.

ethereum/contracts/BasicOutboundChannel.sol
```diff
@@ -11,14 +11,15 @@ contract BasicOutboundChannel is OutboundChannel, ChannelAccess, AccessControl {
     // Governance contracts will administer using this role.
     bytes32 public constant CONFIG_UPDATE_ROLE = keccak256("CONFIG_UPDATE_ROLE");
 
-    uint64 public nonce;
+    mapping(address => uint64) public nonces;
```
vgeddes, line 14: I think I prefer the singular `nonce`. It reads better than `nonces`.

ethereum/contracts/BasicOutboundChannel.sol
```diff
@@ -11,14 +11,15 @@ contract BasicOutboundChannel is OutboundChannel, ChannelAccess, AccessControl {
...
 
     // Only messages originating from this account will
     // be allowed through the channel.
     address public principal;
 
     event Message(
         address source,
+        address origin,
```
vgeddes, line 22: I think we should drop the `origin` name, as it could be misleading for reviewers and auditors. It makes it sound like we're relying on [tx.origin](https://hackernoon.com/hacking-solidity-contracts-using-txorigin-for-authorization-are-vulnerable-to-phishing) which has flaws.

I think we should perhaps just call it account:
```suggestion
        address account,
```

ethereum/contracts/BasicOutboundChannel.sol
```diff
@@ -11,14 +11,15 @@ contract BasicOutboundChannel is OutboundChannel, ChannelAccess, AccessControl {
     // Governance contracts will administer using this role.
     bytes32 public constant CONFIG_UPDATE_ROLE = keccak256("CONFIG_UPDATE_ROLE");
 
-    uint64 public nonce;
+    mapping(address => uint64) public nonces;
 
     // Only messages originating from this account will
     // be allowed through the channel.
     address public principal;
```
vgeddes, line 18: Should remove

ethereum/contracts/BasicOutboundChannel.sol
```diff
@@ -55,29 +54,15 @@ contract BasicOutboundChannel is OutboundChannel, ChannelAccess, AccessControl {
...
      * Submission is a privileged action involving two parties: The operator and the origin.
      * Apps (aka operators) need to be authorized by the `origin` to submit messages via this channel.
-     *
-     * Furthermore, this channel restricts the origin to a single account, that of the principal.
-     * In essence this ensures that only the principal account can send messages via this channel.
-     *
-     * For pre-production testing, the restriction to the principal account can be bypassed by using
-     * `setPrincipal` to set the principal to the address 0x0000000000000000000000000000000000000042.
      */
     function submit(address _origin, bytes calldata _payload) external override {
```
vgeddes, line 63:
```suggestion
    function submit(address _account, bytes calldata _payload) external override {
```

parachain/pallets/basic-channel/src/inbound/envelope.rs
```diff
@@ -22,6 +23,8 @@ pub struct Envelope {
 	pub channel: H160,
 	/// The application on Ethereum where the message originated from.
 	pub source: H160,
+	/// The user on Ethereum that authorized the source to send the message.
+	pub user: H160,
```
vgeddes, line 27: In line with suggested changes in Solidity.

```suggestion
	pub account: H160,
```

parachain/pallets/basic-channel/src/inbound/mod.rs
```diff
@@ -66,7 +66,7 @@ pub mod pallet {
 	pub type SourceChannel<T: Config> = StorageValue<_, H160, ValueQuery>;
 
 	#[pallet::storage]
-	pub type Nonce<T: Config> = StorageValue<_, u64, ValueQuery>;
+	pub type Nonces<T: Config> = StorageMap<_, Twox64Concat, H160, u64, ValueQuery>;
```
vgeddes, line 69: I know this is subjective, but I prefer the singular form. If the data type were an actual or array/vector then nonces would make more sense. 

```suggestion
	pub type Nonce<T: Config> = StorageMap<_, Twox64Concat, H160, u64, ValueQuery>;
```

parachain/primitives/core/src/types.rs
```diff
@@ -3,20 +3,14 @@
...
-	pub channel_id: ChannelId,
-	pub nonce: u64,
-}
-
-impl MessageId {
-	pub fn new(channel_id: ChannelId, nonce: u64) -> Self {
-		Self { channel_id, nonce }
-	}
+pub enum MessageId {
+	Basic { user: H160, nonce: u64 },
```
vgeddes, line 12:
```suggestion
	Basic { account: H160, nonce: u64 },
```

vgeddes (commented):
Looks great! Just needs some naming changes.

parachain/pallets/basic-channel/src/inbound/mod.rs
doubledup: I'm happy to rename it - should we also rename the mappings in the Parachain -> Ethereum direction?

doubledup: Made this change in #704 for consistency

PR #700, opened 2022-10-15
Implement minimumRequiredSignatures

vgeddes:
The previous version was not very accurate, see discussion in SNO-150

Fixes: SNO-150


alistair-singh (commented):
👍 looks good!

PR #701, opened 2022-10-20
Adds minimal config for snowbase.

claravanstaden:
Adds minimal spec weights for the `snowbase` runtime, so that tests can complete quicker.


vgeddes (commented):
Are these changes strictly necessary right now? I'm assuming they make the E2E tests pass again. If so, I think they are fine as a temporary band-aid to get the tests working again.

A better solution: If each runtime needs different weights, then we should rather put the specialized weights alongside the runtime code. For example `runtime/snowblink/weights/ethereum_beacon_client.rs`.

The minimal weight config could be left as the default in `parachain/pallets/ethereum-beacon-client/src/weights.rs`, to be used by runtimes if necessary.

This is what Acala and a few other parachains do: https://github.com/AcalaNetwork/Acala/tree/master/runtime/karura/src/weights

The other issue is that these weights seem to be hard-coded, which complicates the process of generating new weights.

claravanstaden:
@vgeddes great point! Let me do that rather. 

The E2E tests failing is caused by something else (sync committees not being backfilled if block production is faster than header syncing - this occurs specifically on our local net because block production is faster than on mainnet. A fix is included in this PR: https://github.com/Snowfork/snowbridge/pull/696).

The snowbase weights in this PR speeds the E2E tests up 2x times (from 10mins to 5mins) for Eth->Parachain.

alistair-singh (approved):

doubledup (approved):

parachain/runtime/snowbase/src/weights/ethereum_beacon_client.rs
```diff
@@ -0,0 +1,39 @@
...
+#![allow(unused_parens)]
+#![allow(unused_imports)]
+
+use frame_support::{traits::Get, weights::Weight};
+use sp_std::marker::PhantomData;
+
+/// Weight functions for ethereum_beacon_client.
+pub struct SnowbridgeWeight<T>(PhantomData<T>);
+impl<T: frame_system::Config> ethereum_beacon_client::WeightInfo for SnowbridgeWeight<T> {	fn sync_committee_period_update() -> Weight {
+		(48_498_137_000 as Weight)			.saturating_add(T::DbWeight::get().reads(4 as Weight))			.saturating_add(T::DbWeight::get().writes(3 as Weight))	}	fn import_finalized_header() -> Weight {
```
vgeddes, line 37: Formatting seems to be a bit off.  Is missing some newlines or something like that.

parachain/runtime/snowbase/src/weights/mod.rs
```diff
@@ -0,0 +1 @@
+pub mod ethereum_beacon_client;
```
vgeddes, line 1: Missing a trailing newline

parachain/templates/runtime-weight-template.hbs
```diff
@@ -0,0 +1,47 @@
...
+			{{~/each}}
+			{{~#if (ne benchmark.base_writes "0")}}
+			.saturating_add(T::DbWeight::get().writes({{benchmark.base_writes}} as Weight))
+			{{~/if}}
+			{{~#each benchmark.component_writes as |cw|}}
+			.saturating_add(T::DbWeight::get().writes(({{cw.slope}} as Weight).saturating_mul({{cw.name}} as Weight)))
+			{{~/each}}
+	}
+	{{~/each}}
+}
```
vgeddes, line 47: Missing a trailing newline

vgeddes (commented):
Looks much better! Though I think the templates aren't inserting newlines when appropriate, causing the generated weight files to not be formatted properly

parachain/runtime/snowbase/src/weights/ethereum_beacon_client.rs
claravanstaden: Yeah, that's from the tilde (`~#each`) in the Handlebars template (it strips whitespace), like [here](https://github.com/Snowfork/snowbridge/blob/main/parachain/module-weight-template.hbs#L40). I noticed our template `module-weight-template.hbs` used it and so I also used it for the [runtime-weight-template.hbs](https://github.com/Snowfork/snowbridge/pull/701/files#diff-2b8a58c029afebd7be9c95411a9f5e49b8b5915babee07021a7de98677a277cc) because I thought there might be reason for it (but it did look very weird). 

I'll go ahead and remove those from both templates and regenerate the weights.

claravanstaden:
@vgeddes I think I fixed the missing newlines, but they don't show in the files diffs, strange. If you check the code out, it's there though.

vgeddes (approved):
Cool, looks all good now.

PR #704, opened 2022-10-21
SNO-352: Rename nonces map to nonce in parachain -> eth basic channel

doubledup:
Follow-on from a [previous comment](https://github.com/Snowfork/snowbridge/pull/699#discussion_r995597978) about renaming the nonce map.

This is another breaking change for the parachain & contracts, so might be a good opportunity for an upgradability exercise.


alistair-singh (approved):

vgeddes (approved):
+1

PR #706, opened 2022-10-21
Add erc20vault

alistair-singh:
Adds a vault for ERC20 tokens.

E2E tests pass.

### TODO
- [x] better unit testing for the new contract

### DIFF
--- a/ethereum/contracts/ERC20Vault.sol
+++ b/ethereum/contracts/ERC20Vault.sol
@@ -5,20 +5,32 @@ import "@openzeppelin/contracts/access/Ownable.sol";
 import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
 
-// Holds ERC20 Tokens on behalf of ERC20App
+/// @title ERC20 Vault
+/// @notice Holds ERC20 Tokens on behalf of ERC20App.
 contract ERC20Vault is Ownable {
     using SafeERC20 for IERC20;
 
-    event Deposit(address account, address sender, uint256 amount);
-    event Withdraw(address account, address recipient, uint256 amount);
+    /// @dev Emitted when funds are deposited.
+    /// @param account The address of the ERC20App contract.
+    /// @param sender The address of the sender.
+    /// @param token The address of the ERC20 token.
+    /// @param amount The amount being deposited.
+    event Deposit(address account, address sender, address token, uint256 amount);
 
+    /// @dev Emitted when funds are withdrawn.
+    /// @param account The address of the ERC20App contract.
+    /// @param recipient The address of the sender.
+    /// @param token The address of the ERC20 token.
+    /// @param amount The amount being withdrawn.
+    event Withdraw(address account, address recipient, address token, uint256 amount);
+
+    /* State */
     mapping(address => uint256) public balances;
 
-    receive() external payable {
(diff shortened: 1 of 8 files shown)


ethereum/contracts/ERC20Vault.sol
```diff
@@ -0,0 +1,49 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity ^0.8.9;
+
+import "@openzeppelin/contracts/access/Ownable.sol";
+import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+
+// Holds ETH on behalf of ETHApp
```
vgeddes, line 8: Comment needs to be updated.

ethereum/contracts/ERC20Vault.sol
```diff
@@ -0,0 +1,49 @@
...
+    event Deposit(address account, address sender, uint256 amount);
+    event Withdraw(address account, address recipient, uint256 amount);
+
+    mapping(address => uint256) public balances;
+
+    receive() external payable {
+        revert("Must use deposit function");
+    }
+
+    // Accepts ETH from the caller.
```
vgeddes, line 21: Please use the [NATSPEC](https://docs.soliditylang.org/en/v0.8.17/natspec-format.html) style comments. See https://github.com/Snowfork/snowbridge/blob/main/ethereum/contracts/BeefyClient.sol for an example.

ethereum/contracts/ERC20Vault.sol
```diff
@@ -0,0 +1,49 @@
...
+// Holds ETH on behalf of ETHApp
+contract ERC20Vault is Ownable {
+    using SafeERC20 for IERC20;
+
+    event Deposit(address account, address sender, uint256 amount);
+    event Withdraw(address account, address recipient, uint256 amount);
+
+    mapping(address => uint256) public balances;
+
+    receive() external payable {
```
vgeddes, line 17: Doesn't make sense in this context since `receive` is only for receiving Ether.

ethereum/contracts/ERC20Vault.sol
```diff
@@ -0,0 +1,49 @@
...
+        balances[_token] = balances[_token] + _amount;
+        require(
+            IERC20(_token).transferFrom(_sender, address(this), _amount),
+            "Contract token allowances insufficient to complete this lock request"
+        );
+        emit Deposit(msg.sender, _sender, _amount);
+    }
+
+    // Returns ETH to the caller.
+    function withdraw(address _recipient, address _token, uint256 _amount)
```
vgeddes, line 35: Like the parent app, the the vault should also use `uint128` for amounts. `uint256` isn't really supported effectively on the Polkadot side, so we want to prevent users from locking up amounts that can't be represented using 128-bits.

alistair-singh: Addressed in https://github.com/Snowfork/snowbridge/pull/706/commits/c379a00d63327d29527ed236bf969a0c347572cb.

ethereum/contracts/ERC20Vault.sol
alistair-singh: Addressed in https://github.com/Snowfork/snowbridge/pull/706/commits/4f7524253c72c6ccf7f518245e67b4d91caff153.

vgeddes (approved):
+1

PR #709, opened 2022-10-26
SNO-358: Make apps declare weights of messages

doubledup:


vgeddes:
Looks good so far 👍

ethereum/contracts/BasicOutboundChannel.sol
```diff
@@ -17,7 +17,8 @@ contract BasicOutboundChannel is OutboundChannel, ChannelAccess, AccessControl {
         address source,
         address account,
         uint64 nonce,
-        bytes payload
+        bytes payload,
+        uint64 weight
```
vgeddes, line 21: I would swap the order of these fields so that the payload is last.

parachain/primitives/core/src/types.rs
```diff
@@ -26,8 +26,10 @@ pub enum ChannelId {
 pub struct Message {
 	/// The raw RLP-encoded message data.
 	pub data: Vec<u8>,
-	/// Input to the message verifier
+	/// Input to the message verifier.
 	pub proof: Proof,
+	/// Untrusted weight declared by relayer.
+	pub weight: Weight,
```
vgeddes, line 32: Let's rename this field to `dispatch_weight` in this struct, so that readers know its referring to the weight of the dispatchable call.

Also, lets expand on the comment a bit: The declared weight of the dispatchable call, extracted from the message payload by relayers. This is untrusted, so the message payload will be rejected if the actual dispatchable weight is higher than the declared weight.

ethereum/contracts/BasicOutboundChannel.sol
doubledup: Swapped in eff0bb0d

parachain/primitives/core/src/types.rs
doubledup: Renamed in 8ccbef4c & comment updated in c9afd97e

parachain/pallets/basic-channel/src/inbound/mod.rs
```diff
@@ -92,7 +92,7 @@ pub mod pallet {
 
 	#[pallet::call]
 	impl<T: Config> Pallet<T> {
-		#[pallet::weight(100_000_000)]
+		#[pallet::weight(T::WeightInfo::submit() + message.dispatch_weight)]
```
vgeddes, line 95: Now that you've got the protocol changes figured out, we need to implement benchmarks for `T::WeightInfo::submit()`.

I think that's going to involve some restructuring of this submit() dispatchable, to factor out the message verification logic into a separate function that can be benchmarked.

After reading https://docs.substrate.io/test/benchmark/, I'd recommend chatting to @claravanstaden who has experience writing benchmarks.

parachain/pallets/basic-channel/src/inbound/mod.rs
```diff
@@ -118,7 +118,7 @@ pub mod pallet {
 			})?;
 
 			let message_id = MessageId::Basic { account: envelope.account, nonce: envelope.nonce };
-			T::MessageDispatch::dispatch(envelope.source, message_id, &envelope.payload);
+			T::MessageDispatch::dispatch(envelope.source, message_id, &envelope.payload, Some(message.dispatch_weight));
 
 			<LatestVerifiedBlockNumber<T>>::set(block_number);
```
vgeddes, line 123: Should move this before message dispatch, to make implementing benchmarks easier.

doubledup:
Not sure why `parachain/check` is failing. I've run `cargo +stable-2022-07-19 check --manifest-path parachain/Cargo.toml --workspace --exclude snowbridge --features runtime-benchmarks` from the root directory passes.

ethereum/contracts/ETHApp.sol
```diff
@@ -37,6 +37,9 @@ contract ETHApp is RewardController, AccessControl {
     // Value of transaction must be positive
     error MinimumAmount();
 
+    // TODO: replace with weight generated by parachain
+    uint64 constant WEIGHT = 42;
```
vgeddes, line 41: obsolete, since the weights are now recorded in ETHAppPallet.sol

vgeddes (commented):
Looks good! lets go ahead and do the incentivized channel now.

parachain/README.md
```diff
@@ -104,13 +120,50 @@ cargo install subxt-cli
...
+
+```bash
+runtime=snowbase
+cargo build \
+    --release \
+    --no-default-features \
+    --features "${runtime}-native,rococo-native,runtime-benchmarks,${runtime}-runtime-benchmarks" \
+    --bin snowbridge
+./target/release/snowbridge build-spec --chain $runtime --disable-default-bootnode > /tmp/snowbridge/spec.json
+cat "/tmp/snowbridge/spec.json" | node ../test/scripts/helpers/mutateSpec.js "/tmp/snowbridge/contracts.json" "/tmp/snowbridge/initialBeaconSync.json" | sponge "/tmp/snowbridge/spec.json"
```
vgeddes, line 146: @claravanstaden, just checking, surely benchmarks can be run using the default builtin chainspec? ie.
```
/target/release/snowbridge build-spec --chain $runtime --disable-default-bootnode
```

claravanstaden: I ran the benchmarks with the chainspec generated from the `start-services.sh` script. I must confess I didn't think of using the built-in chainspec.

vgeddes: ahh, ok. So I tried to run the benchmarks using the default spec, and they failed with:

```
vincent@lemur:~/code/snowbridge/parachain (main)$ target/release/snowbridge benchmark pallet --chain snowbase --execution wasm --wasm-execution compiled --pallet snowbridge_ethereum_beacoclient --extrinsic '*' --r
epeat 20 --steps 50 --template templates/module-weight-template.hbs
2022-11-08 11:27:06 💫 Sync committee size is: 32
2022-11-08 11:27:06 💫 Received initial sync, starting processing.
2022-11-08 11:27:06 Merkle proof branch length doesn't match depth.
2022-11-08 11:27:06 Initial sync failed with error Module(ModuleError { index: 18, error: [7, 0, 0, 0], message: Some("InvalidSyncCommitteeMerkleProof") })
```

The problem is that when the benchmarks tool is started with the default spec, it generates the genesis config using the default spec, and that causes this line to error out: https://github.com/Snowfork/snowbridge/blob/81348245bfeaa66b9e02a09c09aa696167bfa1d1/parachain/pallets/ethereum-beacon-client/src/lib.rs#L176. Because the default spec contains an empty `InitialSync`

I think the fix is to bypass processing of the InitialSync if its the default value. This would allow benchmarks to run normally using the default spec which is highly desirable.

vgeddes: Added a ticket here: https://linear.app/snowfork/issue/SNO-369

doubledup: This is generating that same chainspec ☝️ Just without starting an E2E env. The benchmark fails if we don't run mutateSpec on the spec first: `Error: Input("Error parsing spec file: invalid type: sequence, expected a hex string at line 122 column 32")`. Looks like `.genesis.runtime.ethereumBeaconClient.initialSync.current_sync_committee.aggregate_pubkey` is an array of zeroes where the benchmark subcommand expects a hash string like `0x8fe11476a05750c52618deb79918e2e674f56dfbf12dbce55ae4386d108e8a1e83c6326f5957e2ef19137582ce270dc6`

doubledup:
Switching to using XCM for passing messages over the bridge.

PR #712, opened 2022-10-28
Remove usage of ERC-777

vgeddes:
Wrapped DOT should be ERC20, not ERC-777. The latter standard is too complicated and no one really uses it. Lots of security concerns too.

Fixes: SNO-362


ethereum/contracts/WrappedToken.sol
```diff
@@ -1,35 +1,19 @@
...
 pragma solidity ^0.8.9;
 
-import "@openzeppelin/contracts/token/ERC777/ERC777.sol";
+import "@openzeppelin/contracts/token/ERC20/ERC20.sol";
 import "@openzeppelin/contracts/access/Ownable.sol";
+import "@openzeppelin/contracts/token/ERC20/extensions/draft-ERC20Permit.sol";
 
-contract WrappedToken is ERC777, Ownable {
+/// @custom:security-contact contact@snowfork.com
+contract WrappedToken is ERC20, Ownable, ERC20Permit {
```
alistair-singh, line 9: WrappedToken is just a Vault for dot-app and my assumption is that it will be changed to a vault in the upgradability branch?

ethereum/contracts/DOTApp.sol
```diff
@@ -72,26 +73,29 @@ contract DOTApp is FeeController, AccessControl {
...
-        token.burn(msg.sender, _amount, abi.encodePacked(_recipient));
+        token.burn(msg.sender, _amount);
 
         OutboundChannel channel =
             OutboundChannel(channels[_channelId].outbound);
 
         bytes memory call = encodeCall(msg.sender, _recipient, _amount);
         channel.submit(msg.sender, call);
+
+        emit Burned(msg.sender, _recipient, _amount);
```
alistair-singh, line 84: We should move the `emit Burned(...)` to right below the call to `token.burn(...)`. This will make the order in which the events  are logged the same as the order when using ERC777.

ethereum/contracts/DOTApp.sol
```diff
@@ -72,26 +73,29 @@ contract DOTApp is FeeController, AccessControl {
...
     ) external onlyRole(INBOUND_CHANNEL_ROLE) {
-        token.mint(_recipient, _amount, abi.encodePacked(_sender));
+        token.mint(_recipient, _amount);
+        emit Minted(_sender, _recipient, _amount);
     }
 
     // Incentivized channel calls this to charge (burn) fees
     function handleFee(address feePayer, uint256 _amount) external override onlyRole(FEE_BURNER_ROLE) {
-        token.burn(feePayer, _amount, "");
+        token.burn(feePayer, _amount);
```
alistair-singh, line 98: The ERC777 code would emit a `Burned` event here. Do we need to also emit a burned event?

ethereum/contracts/DOTApp.sol
```diff
@@ -118,7 +122,7 @@ contract DOTApp is FeeController, AccessControl {
         // revoke old channel
         revokeRole(INBOUND_CHANNEL_ROLE, c1.inbound);
         revokeRole(INBOUND_CHANNEL_ROLE, c2.inbound);
-        // set new channel
+       // set new channel
```
alistair-singh, line 125: Nit: whitespace

ethereum/contracts/DOTApp.sol
vgeddes: I don't think it matters that much, and the convention in Substrate and Solidity seems to be that events are emitted last in each dispatchable.

ethereum/contracts/DOTApp.sol
vgeddes: Nope. Nothing is using that event.

ethereum/contracts/DOTApp.sol
vgeddes: Thanks, though this code is all going to be removed by my other open prs.

ethereum/contracts/WrappedToken.sol
vgeddes: WrappedToken already acts like a vault. So in production we can transfer ownership of it to a new DOTApp.

PR #714, opened 2022-10-31
Introduce channel registry contract

vgeddes:
All app contracts now lookup channel info from the new `ChannelRegistry` contract. This eliminates all the redundant channel management and upgrade code which existed in each app.

Governance should call `ChannelRegistry.updateChannel` to add new versions of channel contracts.

Also updated all the Solidity unit tests to use ethers and various hardhat test helpers instead of the archaic legacy truffle/web3 libraries. Much easier to write unit tests now.


musnit:
> hardhat

hardhat it getting out of fashion now too lol ppl moving to https://github.com/foundry-rs/foundry but probably not worth us investing time into that atm yet

ethereum/contracts/BasicOutboundChannel.sol
```diff
@@ -56,9 +56,9 @@ contract BasicOutboundChannel is OutboundChannel, ChannelAccess, AccessControl {
      * Submission is a privileged action involving two parties: The operator and the origin (called account here).
      * Apps (aka operators) need to be authorized by the `account` to submit messages via this channel.
      */
-    function submit(address _account, bytes calldata _payload) external override {
-        require(isOperatorFor(msg.sender, _account), "Caller is unauthorized");
-        nonce[_account] = nonce[_account] + 1;
-        emit Message(msg.sender, _account, nonce[_account], _payload);
+    function submit(address account, bytes calldata payload, uint64) external override {
```
alistair-singh, line 59: Are we not using the `_variable` parameter naming convention anymore?

ethereum/contracts/ChannelRegistry.sol
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    function outboundChannelForID(uint32 id) external view returns (address) {
+        return channels[id].outbound;
+    }
+
+    function updateChannel(uint32 id, address inbound, address outbound) external onlyOwner {
+        delete validInboundChannels[channels[id].inbound];
+        channels[id] = Channel(inbound, outbound);
+        validInboundChannels[inbound] = true;
```
alistair-singh, line 32: We should emit an event when a channel is updated.

ethereum/contracts/DOTApp.sol
```diff
@@ -6,77 +6,44 @@ import "./WrappedToken.sol";
...
-    }
+    error UnknownChannel(uint32 id);
+    error Unauthorized();
 
     constructor(
         WrappedToken _token,
         address feeBurner,
-        Channel memory _basic,
-        Channel memory _incentivized
+        address channelRegistry
```
alistair-singh, line 26: Use contract type instead of address.
```suggestion
        ChannelRegistry channelRegistry
```

ethereum/contracts/ERC20App.sol
```diff
@@ -46,116 +53,69 @@ contract ERC20App is AccessControl {
...
-                _channelId == ChannelId.Incentivized,
-            "Invalid channel ID"
-        );
+        address channel = registry.outboundChannelForID(_channelID);
+        if (channel == address(0)) {
+            revert UnknownChannel(_channelID);
+        }
 
+        if (!IERC20(_token).transferFrom(msg.sender, address(this), _amount)) {
+            revert InsufficientAllowance();
```
alistair-singh, line 76: `transferFrom` can fail with both an insufficient allowance but also insufficient balance. This error might hide the real error. 
https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/ERC20.sol#L237

ethereum/contracts/BasicOutboundChannel.sol
vgeddes: yeah, not using the `_variable` convention anymore. It kills readability. I don't think the convention is useful in our case, because at an individual contract level, there are not many storage fields which can shadowed by parameters.

ethereum/contracts/DOTApp.sol
vgeddes: Good catch!

PR #720, opened 2022-11-09
Small changes from work on weights

doubledup:
Minor tweaks from the weights work and using the new project structure. Added a section in the parachain's README about generating benchmarks and updated the instructions for generating test data.

### DIFF
--- a/core/packages/test/package.json
+++ b/core/packages/test/package.json
@@ -17,7 +17,6 @@
     "@polkadot/types": "^9.6.2",
     "@polkadot/util": "^10.1.11",
     "@polkadot/util-crypto": "^10.1.11",
-    "@snowfork/snowbridge-types": "^0.2.7",
     "@types/keccak": "^3.0.1",
     "@types/node": "^16.4.2",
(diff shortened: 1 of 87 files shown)


core/packages/test/package.json
```diff
@@ -14,8 +14,10 @@
   "devDependencies": {
     "@chainsafe/lodestar": "1.1.0",
     "@polkadot/api": "^9.6.2",
+    "@polkadot/types": "^9.6.2",
     "@polkadot/util": "^10.1.11",
     "@polkadot/util-crypto": "^10.1.11",
+    "@snowfork/snowbridge-types": "^0.2.7",
```
vgeddes, line 20: This package is obsolete, we shouldn't rely on it anymore. I see you're adding it because you included `make-xcm-transfer.ts` in the compilation. But that script is obsolete too, so we should probably exclude it from compilation until its replaced.

vgeddes (approved):

alistair-singh (approved):
🚢

PR #721, opened 2022-11-14
Split Beacon Relayer

claravanstaden:
Split beacon relayer into two separate relayer processes for syncing:
- beacon headers
- execution header messages

Resolves: #https://linear.app/snowfork/issue/SNO-372


relayer/chain/ethereum/private_key.go
```diff
@@ -0,0 +1,33 @@
...
+
+import (
+	"fmt"
+	"io/ioutil"
+	"strings"
+
+	"github.com/snowfork/snowbridge/relayer/crypto/secp256k1"
+)
+
+func ResolvePrivateKey(privateKey, privateKeyFile string) (*secp256k1.Keypair, error) {
```
claravanstaden, line 11: Refactored out from the relays packages.

relayer/relays/execution/message/message.go
```diff
@@ -0,0 +1,320 @@
...
+
+	log.WithFields(log.Fields{
+		"block_number": lastVerifiedBlockNumber,
+		"nonce":        nonce,
+	}).Info("checking last synced incentivized channel messages on startup")
+
+	// If the last nonce is set, there could be messages that have not been processed in the same block.
+	// Messages that have already been verified will not be reprocessed because they will be filtered out
+	// in filterMessagesByLastNonce.
+	// Messages after the lastVerifiedBlockNumber will be processed normally in the go routine below.
```
doubledup, line 289:
```suggestion
	// Messages after the lastVerifiedBlockNumber will be processed separately in the Sync method.
```

relayer/relays/execution/message/message.go
```diff
@@ -0,0 +1,320 @@
...
+	if nonce == 0 {
+		return 0, nil
```
doubledup, line 291: When the nonce is zero, there aren't any messages to relay up to and including `lastVerifiedBlockNumber`, so we should be able to return that here.

```suggestion
	if nonce == 0 {
		return lastVerifiedBlockNumber, nil
```

relayer/relays/execution/message/message.go
```diff
@@ -0,0 +1,320 @@
...
+			"nonce":        nonce,
+		}).Info("last basic channel")
+
+		if nonce != 0 {
+			addressNonzeroNonceMap[address] = nonce
+		}
+	}
+
+	if len(addressNonzeroNonceMap) == 0 {
+		return 0, nil
```
doubledup, line 248: Similar to the incentivized channel method, if all addresses' nonces are 0, we should be able to ignore all blocks up to `lastVerifiedBlockNumber`.

```suggestion
	if len(addressNonzeroNonceMap) == 0 {
		return lastVerifiedBlockNumber, nil
```

relayer/relays/execution/message/message.go
```diff
@@ -0,0 +1,320 @@
...
+		return fmt.Errorf("sync unprocessed basic messages: %w", err)
+	}
+
+	ticker := time.NewTicker(time.Second * 20)
+
+	var secondLastSyncedBlockNumber uint64
+	if lastSyncedIncentivizedBlock > lastSyncedBasicBlock {
+		secondLastSyncedBlockNumber = lastSyncedIncentivizedBlock
+	} else {
+		secondLastSyncedBlockNumber = lastSyncedBasicBlock
```
doubledup, line 59: Why is `secondLastSyncedBlockNumber` the max of the last synced blocks for each channel? If syncing fails for one channel and not the other for one block, we could ignore some blocks for the channel that was failing. If that block contains messages, a skipped nonce could prevent the channel from receiving future messages.

eg.
```
lastSyncedIncentivizedBlock = 11
lastSyncedBasicBlock = 10
secondLastSyncedBlockNumber = 11
lastSyncedBlockNumber = 12
```

Then we'll try to sync block 12 for both channels, but if block 11 contains a basic channel message, the channel will reject it because the nonce isn't incrementing by 1.

Given that syncing can fail independently (basic channel sync could fail after incentivized channel sync succeeds), could we use `lastSyncedIncentivizedBlock` and `lastSyncedBasicBlock` separately? `secondLastSyncedBlockNumber` would still be useful in the loop to detect unchanged execution blocks, but we'd need to start with the min of `lastSyncedIncentivizedBlock` and `lastSyncedBasicBlock` here.

relayer/relays/execution/message/message.go
```diff
@@ -0,0 +1,320 @@
...
+			return err
+		}
+
+		lastNonce, err := m.writer.GetLastIncentivizedChannelNonce()
+		if err != nil {
+			return err
+		}
+
+		if lastNonce != msg.Nonce {
+			return fmt.Errorf("last incentivized message verification failed (nonce: %d)", lastNonce)
```
doubledup, line 188: Let's include the expected nonce, like with the basic channel verification:

```suggestion
			return fmt.Errorf("last incentivized message verification failed (expected nonce: %d, actual nonce: %d)", msg.Nonce, lastNonce)
```

doubledup (commented):
Nice! The independent relayers are much easier to work with. It looks like the basic channel relayer could effectively halt if we get an error in `syncBasic` - let's double-check this.

alistair-singh (approved):

core/packages/test/scripts/start-services.sh
```diff
@@ -277,21 +277,27 @@ start_relayer()
 
     # Configure beacon relay
     jq \
-        --arg k1 "$(address_for BasicOutboundChannel)" \
-        --arg k2 "$(address_for IncentivizedOutboundChannel)" \
         --arg infura_endpoint_ws $infura_endpoint_ws \
         --arg beacon_endpoint_http $beacon_endpoint_http \
         --arg active_spec $active_spec \
+    '
+     .source.ethereum.endpoint = $infura_endpoint_ws
```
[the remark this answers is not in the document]
claravanstaden: Good catch! Addressed in https://github.com/Snowfork/snowbridge/pull/721/commits/e4d75222b1055b2ab63933708f8ae1a27b4a9856.

relayer/cmd/run/execution/command.go
```diff
@@ -0,0 +1,104 @@
...
+	"context"
+	"log"
+	"os"
+	"os/signal"
+	"syscall"
+
+	"github.com/sirupsen/logrus"
+	"github.com/snowfork/snowbridge/relayer/chain/parachain"
+	"github.com/snowfork/snowbridge/relayer/relays/execution"
+	"github.com/snowfork/snowbridge/relayer/relays/execution/config"
```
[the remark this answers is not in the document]
claravanstaden: I agree the package structure should be consistent. However, I added the config package to avoid a cyclical import in `message/message.go` (using `config.go` in the parent package produces a cyclical import). This is kind of related to the [question here](https://github.com/Snowfork/snowbridge/pull/721#discussion_r1022712303). See comment there too.

I think using consistent packages for the parachain and beefy relayers would be great, but not in the scope of this PR. I'll make a follow-up.

relayer/relays/execution/main.go
```diff
@@ -0,0 +1,70 @@
...
+	if err != nil {
+		return err
+	}
+
+	messages, err := message.New(
+		ctx,
+		eg,
+		writer,
+		&r.config.Source,
+		ethconn,
```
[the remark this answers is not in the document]
claravanstaden: This is a great point! It's actually a better practice to inject the dependency instances rather than creating it in the package, because it makes testing easier (can inject mock interfaces instead of the real connection to Ethereum). I've pulled out the dependencies and inject them now. https://github.com/Snowfork/snowbridge/pull/721/commits/4a2a0d6015a4b431aa494924cead6ce3aaf7fc1e

relayer/relays/execution/message/message.go
```diff
@@ -0,0 +1,320 @@
...
+
+	incentivizedPayload.Messages = filterMessagesByLastNonce(incentivizedPayload.Messages, nonce)
+	// Reset the nonce so that the next block processing range will exclude the block that was synced,
+	// and start syncing from the next block instead
+
+	incentivizedPayload.Messages = filterMessagesByLastNonce(incentivizedPayload.Messages, nonce)
+	// Reset the nonce so that the next block processing range will exclude the block that was synced,
+	// and start syncing from the next block instead
+
+	nonce = 0
```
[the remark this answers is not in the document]
claravanstaden: Ah nice catch! I think I wanted to something here and forgot to remove it. 😬

relayer/relays/execution/message/message.go
claravanstaden: Good point!

relayer/relays/execution/message/message.go
claravanstaden: Good point! I thought it would be OK because blocks are always processed sequentially, but realized that is not actually the case if one channel gets processed before the other and the other channel has an earlier block with a message. Fixed in https://github.com/Snowfork/snowbridge/pull/721/commits/dc0ef6af10d04299d197a92a6fb564d5f086448b. The 2 3 variables assigned to `lastSyncedBlockNumber` isn't the best, I'll see if I can refactor that a bit.

claravanstaden: I ended up removing `secondLastSyncedBlockNumber ` here: https://github.com/Snowfork/snowbridge/pull/721/commits/e4b51ee9109baff6ffcfaddaaf5de0f7d83a76ec

relayer/relays/execution/main.go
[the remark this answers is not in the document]
doubledup: oh right, makes a lot of sense for testing 👌

relayer/relays/execution/message/message.go
```diff
@@ -72,22 +50,25 @@ func (m *Message) Sync(ctx context.Context, eg *errgroup.Group) error {
...
 				}).Info("last synced execution block changed, fetching messages")
 
-				err = m.syncIncentivized(ctx, eg, secondLastSyncedBlockNumber+1, lastSyncedBlockNumber)
+				err = m.syncIncentivized(ctx, eg, lastSyncedIncentivizedBlock+1, lastSyncedBlockNumber)
 				if err != nil {
 					return fmt.Errorf("sync incentivized messages: %w", err)
 				}
 
-				err = m.syncBasic(ctx, eg, secondLastSyncedBlockNumber+1, lastSyncedBlockNumber)
+				err = m.syncBasic(ctx, eg, lastSyncedBasicBlock+1, lastSyncedBlockNumber)
```
doubledup, line 65: When only one channel is behind, the up-to-date channel will have its sync method called with a start block that's 1 greater than the end block. Do the `FilterMessage` implementations work when the start block is greater than the end block? Just double-checking, I expect they do

doubledup (approved):

alistair-singh (approved):

relayer/relays/execution/message/message.go
```diff
@@ -20,22 +19,7 @@ type Message struct {
...
-		configSource,
-		ethconn,
-	)
-
-	err = listener.Start(ctx, eg)
-	if err != nil {
-		return &Message{}, err
-	}
-
+func New(ctx context.Context, eg *errgroup.Group, writer *parachain.ParachainWriter, listener *EthereumListener, ethconn *ethereum.Connection, addresses []common.Address) (*Message, error) {
```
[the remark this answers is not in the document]
claravanstaden: Ah nice catch! I've removed it. https://github.com/Snowfork/snowbridge/pull/721/commits/6b732cc193a24114dd1f9891b20e280f9c1cb895

relayer/relays/execution/message/message.go
claravanstaden: Good question! I've added a check to make sure the start block if smaller than the end block. https://github.com/Snowfork/snowbridge/pull/721/commits/6b732cc193a24114dd1f9891b20e280f9c1cb895

doubledup (approved):

PR #728, opened 2022-11-21
Adds ssz tests

claravanstaden:
- Changes ssz-rs crate to use latest version of https://github.com/Snowfork/ssz_rs. This will be updated to the upstream crate once https://github.com/ralexstokes/ssz-rs/pull/25 is merged.
- Adds tests for deserializing sync committee bits, with too many bits and too few bits.
- The other usages of the ssz crate is for hash_tree_root, which is applied on a struct with size limits, so overflow cases are already handled on that level.

### DIFF
--- a/parachain/Cargo.lock
+++ b/parachain/Cargo.lock
@@ -11667,7 +11667,7 @@ dependencies = [
 [[package]]
 name = "ssz-rs"
 version = "0.8.0"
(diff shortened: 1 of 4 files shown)


alistair-singh (approved):
🚢

vgeddes (approved):
+1

doubledup (approved):

PR #729, opened 2022-11-22
SNO-373: Optimize Bitfield.randomNbitsWithPriorCheck

doubledup:
E2E tests have passed. There's a small decrease in gas cost for `submitFinalWithLeaf` and a moderate one for deploying `Bitfield`, with other minor decreases.

___

## Hardhat gas reporter 2 (optimiser enabled)

Running `REPORT_GAS=1 pnpm test` with the optimizer enabled showed these differences with main (da5bf982ec38d90e714e48cb9f357a782c40f59c):

### Methods

BeefyClientPublic.submitFinalWithLeaf: 380950 -> 354014 (~7.07% decrease)
BeefyClientPublic.submitInitial: 176430 -> 175367 (~0.60% decrease)

### Deployments

BeefyClientPublic: 2421120 -> 2361685 (~2.45% decrease; 8.1% -> 7.9% of limit)
Bitfield: 549365 -> 418530 (~23.82% decrease; 1.8% -> 1.4% of limit)

___

## Hardhat gas reporter 1 (optimiser disabled)

Running `REPORT_GAS=1 pnpm test` with the optimizer disabled showed these differences with main (da5bf982ec38d90e714e48cb9f357a782c40f59c):

### Methods

BeefyClientPublic.submitFinalWithLeaf: 567156 -> 500862 (~11.69% decrease)
BeefyClientPublic.submitInitial: 196877 -> 193570 (~1.68% decrease)

### Deployments

BeefyClientPublic: 4206703 -> 4085950 (~2.87% decrease; 14% -> 13.6% of limit)
Bitfield: 793710 -> 610278 (~23.11% decrease; 2.6% -> 2% of limit)
IncentivizedInboundChannel: 2013596 -> 2013608 (~0.00% increase; 6.7% -> 6.7% of limit)


core/packages/contracts/contracts/utils/Bitfield.sol
```diff
@@ -48,87 +41,103 @@ library Bitfield {
...
+            return count;
+        }
+    }
+
+    struct Location {
+        uint256 element;
+        uint8 within;
+    }
+
+    function toLocation(uint256 index) internal pure returns (Location memory location) {
```
vgeddes, line 121: This function is a good start, but its still inefficient because it allocates memory on the "heap" rather than the stack. This happens time a variable is declared with "memory" qualifier. 

Currently, Solidity doesn't allow structs to be stored on the stack, but I don't think we need a struct here, so its not a big loss.

So rather allocate the return value on the stack roughly like this:

```solidity
function toLocation(uint256 index) internal pure returns (uint265, uint256) {
    unchecked {
        uint256 element = index / 256;
        uint256 within = uint8(index % 256);
        return (element, within)
    }
}
```

core/packages/contracts/contracts/utils/Bitfield.sol
```diff
@@ -48,87 +41,103 @@ library Bitfield {
...
+                unchecked { ++i; }
                 continue;
             }
 
-            set(bitfield, index);
+            set(bitfield, loc);
 
-            found++;
+            unchecked { ++found; }
+            unchecked { ++i; }
```
vgeddes, line 68: Rather use a single `unchecked` statement for these mutations

core/packages/contracts/contracts/utils/Bitfield.sol
```diff
@@ -48,87 +41,103 @@ library Bitfield {
...
+            if (!isSet(prior, loc)) {
+                unchecked { ++i; }
                 continue;
             }
 
             // require a not yet set (new) bit to be set
-            if (isSet(bitfield, index)) {
+            if (isSet(bitfield, loc)) {
+                unchecked { ++i; }
                 continue;
```
vgeddes, line 62: I have a feeling we can combine these two checks into a single operation with some fancy bit twiddling. 

```solidity
if (IsNotSetInAandIsSetInB(a, b, loc)) {
  unchecked { i++; }
  continue;
}
```

core/packages/contracts/contracts/utils/Bitfield.sol
```diff
@@ -48,87 +41,103 @@ library Bitfield {
...
 
-            found++;
+            unchecked { ++found; }
+            unchecked { ++i; }
         }
 
         return bitfield;
     }
 
     function createBitfield(uint256[] calldata bitsToSet, uint256 length)
```
vgeddes, line 74: This function does not need to be optimized at all since it is never called from on any on-chain code. So I wouldn't bother using `unchecked` inside it.

core/packages/contracts/contracts/ScaleCodec.sol
```diff
@@ -82,27 +82,27 @@ library ScaleCodec {
         v = (v >> 8) | (v << 8);
     }
 
-    function encodeU256(uint256 input) public pure returns (bytes32) {
+    function encodeU256(uint256 input) internal pure returns (bytes32) {
```
vgeddes, line 85: I've already optimized ScaleCodec in my https://github.com/snowfork/snowbridge/tree/vincent/optimize-mmr branch which I'm going to make into PR soon.

core/packages/contracts/contracts/utils/Bitfield.sol
```diff
@@ -48,87 +41,103 @@ library Bitfield {
         bitfield = new uint256[](prior.length);
         uint256 found = 0;
 
-        for (uint256 i = 0; found < n; i++) {
+        for (uint256 i = 0; found < n;) {
             bytes32 randomness = keccak256(abi.encodePacked(seed, i));
```
vgeddes, line 45: You can still optimize this hashing operation by using this assembly here: https://github.com/Snowfork/snowbridge/blob/a17e18c6425c4389ef2bcbc27113663c3eac95b3/core/packages/contracts/contracts/BeefyClient.sol#L371

Also this guide may help add context to what that assembly is doing: https://docs.soliditylang.org/en/v0.8.15/internals/layout_in_memory.html

core/packages/contracts/contracts/utils/Bitfield.sol
```diff
@@ -48,87 +41,103 @@ library Bitfield {
...
-            uint256 index = uint256(randomness) % length;
+            uint256 index;
+            unchecked {
+                index = uint256(randomness) % length;
+            }
+
+            Location memory loc = toLocation(index);
```
vgeddes, line 51: See my comment above about storing the return values on the stack rather than memory

core/packages/contracts/contracts/utils/Bitfield.sol
```diff
@@ -48,87 +41,103 @@ library Bitfield {
...
-            x = (x & M1) + ((x >> 1) & M1); //put count of each  2 bits into those  2 bits
-            x = (x & M2) + ((x >> 2) & M2); //put count of each  4 bits into those  4 bits
-            x = (x & M4) + ((x >> 4) & M4); //put count of each  8 bits into those  8 bits
-            x = (x & M8) + ((x >> 8) & M8); //put count of each 16 bits into those 16 bits
-            x = (x & M16) + ((x >> 16) & M16); //put count of each 32 bits into those 32 bits
-            x = (x & M32) + ((x >> 32) & M32); //put count of each 64 bits into those 64 bits
-            x = (x & M64) + ((x >> 64) & M64); //put count of each 128 bits into those 128 bits
-            x = (x & M128) + ((x >> 128) & M128); //put count of each 256 bits into those 256 bits
-            count += x;
+    function countSetBits(uint256[] memory self) internal pure returns (uint256) {
```
vgeddes, line 99: Hey these changes look good, didn't know further optimizations were possible. However, I'd prefer if we make them in a separate PR, especially as we'll need to add a bunch of tests.

core/packages/contracts/contracts/utils/Bitfield.sol
```diff
@@ -37,7 +30,7 @@ library Bitfield {
         uint256[] memory prior,
         uint256 n,
         uint256 length
-    ) public pure returns (uint256[] memory bitfield) {
+    ) internal pure returns (uint256[] memory bitfield) {
 
         // `n` must be <= number of set bits in `prior`
         require(
```
vgeddes, line 36: So it turns out that this `require` check is redundant, as we already verify the number of set bits in the `prior` bitfield in `submitInitial`. So we can remove it.

Am still keen to incorporate your optimizations to countSetBits, in a separate PR though.

vgeddes (commented):
This is a good start. I'd suggest the following changes:

1. The changes here are very broad, which make it harder to understand the impact of just optimizing `randomNBitsWithPriorCheck` in isolation. As such, I would request you:
  a. Revert the changes to `i++`, `ScaleCodec`.
  b. Move the changes to `countSetBits` to a separate PR which we can review later
2. Provide new before/after gas benchmarks of `submitFinalWithLeaf `. Its seems the benchmarks you put into the PR description are more about evaluating `countSetBits`.

core/packages/contracts/contracts/BasicInboundChannel.sol
```diff
@@ -50,12 +50,12 @@ contract BasicInboundChannel {
             gasleft() >= (bundle.messages.length * MAX_GAS_PER_MESSAGE) + GAS_BUFFER,
             "insufficient gas"
         );
-        nonce[bundle.account]++;
```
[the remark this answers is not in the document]
doubledup: There's a small gas reduction after the change, even with the optimizer enabled. This comes from `i++` storing & returning the previous value of `i`, which we don't use. I'm happy to remove this change here, but it's a small optimization (~900 gas for `submitFinalWithLeaf` with the optimizer enabled) without a downside, so I'd like to make the change in a separate PR.

core/packages/contracts/contracts/utils/Bitfield.sol
doubledup: Should we move these test helper functions to their own library that doesn't get deployed?

core/packages/contracts/contracts/utils/Bitfield.sol
doubledup: Cool, will do! I was following the Wikipedia article mentioned in the comment

core/packages/contracts/contracts/BasicInboundChannel.sol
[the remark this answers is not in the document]
vgeddes: Sure, sounds good :+1:

core/packages/contracts/contracts/utils/Bitfield.sol
vgeddes: Actually it is used in production by the relayer, which calls it via this function on BeefyClient: https://github.com/Snowfork/snowbridge/blob/da5bf982ec38d90e714e48cb9f357a782c40f59c/core/packages/contracts/contracts/BeefyClient.sol#L511

I'm open to removing `createBitfield` if we re-implement it in Golang and Typescript.

core/packages/contracts/contracts/utils/Bitfield.sol
doubledup: Right, because n is the signature threshold which is always <= the 2/3+1 set bits minimum

core/packages/contracts/contracts/utils/Bitfield.sol
doubledup: Right, it's used before submitInitial. Not sure what the savings are, but I'm happy to move it out of Solidity 👍

vgeddes (commented):
Looks much better! Can you run the benchmarks again with optimization enabled please?

alistair-singh (approved):

core/packages/contracts/test/tsconfig.json
```diff
@@ -0,0 +1,30 @@
+{
```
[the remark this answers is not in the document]
doubledup: ja, I was getting weird errors in test files (eg. when loading JSON test data) about not having settings that were in the base tsconfig. Figured that came from the test files not being included in the base tsconfig, but I've tweaked it to include them.

Please let me know if this has worked for you, otherwise let's handle it later.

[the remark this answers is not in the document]
vgeddes: We can't include the test files in the compilation though. Only the [auto-generated typescript bindings](https://github.com/dethcrypto/TypeChain) for our contracts should be included (the files under `src`), as they will be packaged and uploaded to NPM eventually.

This is why I suggest having two different tsconfigs, one for editors, and one for build & packaging. But this is something we can handle in a separate PR.

core/packages/contracts/contracts/utils/Bitfield.sol
```diff
@@ -110,25 +110,45 @@ library Bitfield {
...
+    function clear(uint256[] memory self, uint256 element, uint8 within) internal pure {
+        self[element] = Bits.clearBit(self[element], within);
+    }
+
+    function makeIndex(uint256 seed, uint256 iteration, uint256 length) internal pure returns (uint256 index) {
+        assembly {
+            mstore(0x00, seed)
+            mstore(0x20, iteration)
+            let randomness := keccak256(0x00, 0x40)
+            index := mod(randomness, length)
```
vgeddes, line 151:
```suggestion
            index := mod(keccak256(0x00, 0x40), length)
```

vgeddes (approved):
🚀

PR #731, opened 2022-12-03
refactor: use weightToFee for xcm Trader

yrong:


yrong:
https://github.com/Snowfork/snowbridge/pull/752

PR #732, opened 2022-12-03
style: format code

yrong:


core/packages/contracts/.prettierrc
```diff
@@ -7,7 +7,8 @@
           "tabWidth": 4,
           "singleQuote": false,
           "bracketSpacing": true,
-          "explicitTypes": "always"
+          "explicitTypes": "always",
+          "plugins": ["prettier-plugin-solidity"]
```
vgeddes, line 11: nice!

vgeddes (commented):
Thanks! Though you need to refresh the PR since https://github.com/Snowfork/snowbridge/pull/735 introduced some conflicts.

core/packages/contracts/.prettierrc
yrong: @vgeddes  conflict resolved, no logic but only format change in this pr and also add a git hook for it when commit

.husky/pre-commit
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env sh
+. "$(dirname -- "$0")/_/husky.sh"
+
+cd core && pnpm format
```
vgeddes, line 4: Nice! Can we also make the hook fail if `pnpm lint` fails?

vgeddes (approved):
🚀

vgeddes (approved):
+1

alistair-singh (approved):

PR #734, opened 2022-12-06
feat: introduce foundry test

yrong:
seems low coverage for some core contracts/libraries, so introduce foundry for unit test which is 5x faster than hardhat and has some advanced features like fuzz and ffi which maybe useful in some cases


vgeddes:
Hey lets chat more on Discord about this.

While I very much like Foundry, I don't want the project to use both Foundry and hardhat. And we can already fuzz test contracts with [echidna](https://github.com/crytic/echidna), without using Foundry.

So this is not a decision to take lightly, as our use of hardhat is everywhere (deployments, scripts, etc).

PR #739, opened 2022-12-09
Adds Rustfmt to Github Actions workflow and format files

claravanstaden:
- Adds cargo fmt job to Github Actions
- Formats files using `cargo fmt` and `rustfmt.toml` definition that was already in project


.github/workflows/parachain.yml
```diff
@@ -15,6 +15,32 @@ on:
       - '!parachain/LICENSE'
 
 jobs:
+  fmt:
```
claravanstaden, line 18: I created a new job here because `rustfmt` requires the `nightly` toolchain, while the check toolchain uses the `stable` toolchain.

.github/workflows/parachain.yml
```diff
@@ -15,6 +15,32 @@ on:
...
+          restore-keys: |
+            ${{ runner.os }}-cargo-
+      - uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: nightly
+          target: wasm32-unknown-unknown
+          components: rustfmt
+      - name: cargo fmt
+        run: cd parachain && cargo +nightly fmt -- --config-path rustfmt.toml && cd -
```
claravanstaden, line 42: I used the run option here instead of the cargo runner, because the `actions-rs/cargo@v1` fmt command doesn't work when the rust project is in a subdirectory. I tested using the `manifest-path` flag with `cargo fmt`, doesn't seem to work.

claravanstaden:
@vgeddes @alistair-singh this PR does not conflict with https://github.com/Snowfork/snowbridge/pull/732, since that PR only touches the `core` directory.

alistair-singh (approved):

vgeddes (approved):

vgeddes (changes requested):
shouldn't you be running cargo fmt in `--check` mode in github actions?

claravanstaden:
> shouldn't you be running cargo fmt in `--check` mode in github actions?

Good point, @vgeddes! Added in https://github.com/Snowfork/snowbridge/pull/739/commits/578be11bfb3b1681fa6cc9a9190c99ccd07c31e9 and also tested with [run](https://github.com/Snowfork/snowbridge/actions/runs/3683689706/jobs/6232555891).

vgeddes (approved):
+1

PR #740, opened 2022-12-15
fix: relayer test

yrong:
fix 2 test cases fail during  `mage test`
```
2022/12/15 16:31:22 Connecting to ws://127.0.0.1:11144/...
time="2022-12-15T16:31:22+08:00" level=info msg="Connected to chain" endpoint="ws://127.0.0.1:11144/" metaVersion=14
--- FAIL: TestWrite (0.11s)
    parachain-writer_test.go:67: module EthereumLightClient not found in metadata for call EthereumLightClient.import_header
FAIL
FAIL    github.com/snowfork/snowbridge/relayer/relays/ethereum  0.135s
ok      github.com/snowfork/snowbridge/relayer/relays/ethereum/syncer   (cached)
?       github.com/snowfork/snowbridge/relayer/relays/execution [no test files]
?       github.com/snowfork/snowbridge/relayer/relays/execution/config  [no test files]
?       github.com/snowfork/snowbridge/relayer/relays/execution/message [no test files]
--- FAIL: TestQueryEvents (0.00s)
    query_events_test.go:49: type *parachain.BasicOutboundChannelMessageBundle does not support Decodeable interface and could not be decoded field by field, error: type *parachain.BasicOutboundChannelMessage does not support Decodeable interface and could not be decoded field by field, error: expected more bytes, but could not decode any more
FAIL
FAIL    github.com/snowfork/snowbridge/relayer/relays/parachain 0.021s
?       github.com/snowfork/snowbridge/relayer/substrate        [no test files]
FAIL
Error: running "go test ./..." failed with exit code 1
```

seems first is because we deprecated `EthereumLightClient` pallet and replace with `EthereumBeaconClient` after the merge, so would assume removing the unused test should be fine.

 second is because the previous mocked event data outdated, just add some logs to output it so that later we can copy from logs and change mock data as required.


yrong:
@vgeddes the second commit is for adding lint&format for relayer codes, ideally should be in another PR but since this PR is only for some test fix and format change so do it together

relayer/relays/parachain/query_events.go
```diff
@@ -73,6 +73,10 @@ func (q *QueryClient) QueryEvents(ctx context.Context, api string, blockHash typ
 		return nil, err
 	}
 
+	log.WithFields(log.Fields{
+		"inputItems": items,
+	}).Info("parachain.QueryEvents")
```
alistair-singh, line 78: Maybe log this out as debug information.
```suggestion
	}).Debug("parachain.QueryEvents")
```

relayer/relays/parachain/query_events.go
```diff
@@ -107,6 +111,9 @@ func (q *QueryClient) QueryEvents(ctx context.Context, api string, blockHash typ
 			return nil, fmt.Errorf("unknown channel")
 		}
 	}
+	log.WithFields(log.Fields{
+		"events": events,
+	}).Info("parachain.QueryEvents")
```
alistair-singh, line 116: Same as above
```suggestion
	}).Debug("parachain.QueryEvents")
```

alistair-singh (approved):

vgeddes (approved):

PR #742, opened 2022-12-19
Clara/sno 367 check for extended gap between header imports

claravanstaden:
Adds weak subjectivity check to prevent long range attacks. Set to 27 hours, as it is the [minimum recommended weak](https://notes.ethereum.org/@adiasg/weak-subjectvity-eth2#:~:text=The%20maximum%20safe%20weak%20subjectivity%20period%20is%20always%20at%20least%2027%20hours.) subjectivity period.

When the weak subjectivity threshold period is exceeded, the light client panics and will have to be reinitialized.

### DIFF
--- a/parachain/pallets/ethereum-beacon-client/src/lib.rs
+++ b/parachain/pallets/ethereum-beacon-client/src/lib.rs
@@ -17,7 +17,7 @@ mod benchmarking;
 pub use weights::WeightInfo;
 
 use crate::merkleization::get_sync_committee_bits;
-use frame_support::{dispatch::DispatchResult, log, transactional, traits::UnixTime};
+use frame_support::{dispatch::DispatchResult, log, traits::UnixTime, transactional};
 use frame_system::ensure_signed;
 use snowbridge_beacon_primitives::{
 	BeaconHeader, BlockUpdate, Domain, ExecutionHeader, ExecutionHeaderState,
@@ -178,8 +178,7 @@ pub mod pallet {
 	pub(super) type LatestFinalizedHeaderSlot<T: Config> = StorageValue<_, u64, ValueQuery>;
 
 	#[pallet::storage]
-	pub(super) type LatestFinalizedHeaderImportTime<T: Config> =
-		StorageValue<_, u64, ValueQuery>;
+	pub(super) type LatestFinalizedHeaderImportTime<T: Config> = StorageValue<_, u64, ValueQuery>;
 
 	#[pallet::storage]
 	pub(super) type LatestExecutionHeaderState<T: Config> =
@@ -333,9 +332,7 @@ pub mod pallet {
 
 		#[pallet::weight(1000)]
 		#[transactional]
-		pub fn unblock_bridge(
-			origin: OriginFor<T>,
-		) -> DispatchResult {
+		pub fn unblock_bridge(origin: OriginFor<T>) -> DispatchResult {
 			let _sender = ensure_root(origin)?;
 
 			log::info!(target: "ethereum-beacon-client","💫 Unblocking bridge.");
@@ -431,7 +428,7 @@ pub mod pallet {
 			if time > weak_subjectivity_period_check {
 				log::info!(target: "ethereum-beacon-client","💫 Weak subjectivity period exceeded, blocking bridge.",);
 				<Blocked<T>>::set(true);
-				return Err(Error::<T>::BridgeBlocked.into());
+				return Err(Error::<T>::BridgeBlocked.into())
 			}
 
 			let sync_committee_bits =
--- a/parachain/pallets/ethereum-beacon-client/src/lib.rs
+++ b/parachain/pallets/ethereum-beacon-client/src/lib.rs
@@ -176,7 +176,8 @@ pub mod pallet {
 	pub(super) type LatestFinalizedHeaderSlot<T: Config> = StorageValue<_, u64, ValueQuery>;
 
 	#[pallet::storage]
-	pub(super) type LatestFinalizedHeaderImportBlockNumber<T: Config> = StorageValue<_, T::BlockNumber, ValueQuery>;
+	pub(super) type LatestFinalizedHeaderImportBlockNumber<T: Config> =
+		StorageValue<_, T::BlockNumber, ValueQuery>;
 
 	#[pallet::storage]
 	pub(super) type LatestExecutionHeaderState<T: Config> =
@@ -393,7 +394,8 @@ pub mod pallet {
 		fn process_finalized_header(update: FinalizedHeaderUpdateOf<T>) -> DispatchResult {
 			let current_block_number = <frame_system::Pallet<T>>::block_number();
 			let last_finalized_block_number = <LatestFinalizedHeaderImportBlockNumber<T>>::get();
-			let weak_subjectivity_period_check = last_finalized_block_number + T::WeakSubjectivityPeriod::get().into();
+			let weak_subjectivity_period_check =
+				last_finalized_block_number + T::WeakSubjectivityPeriod::get().into();
 
 			log::info!(
 				target: "ethereum-beacon-client",
--- a/parachain/pallets/ethereum-beacon-client/src/lib.rs
+++ b/parachain/pallets/ethereum-beacon-client/src/lib.rs
@@ -17,7 +17,7 @@ mod benchmarking;
 pub use weights::WeightInfo;
 
 use crate::merkleization::get_sync_committee_bits;
-use frame_support::{dispatch::DispatchResult, log, transactional};
+use frame_support::{dispatch::DispatchResult, log, transactional, traits::UnixTime};
 use frame_system::ensure_signed;
 use snowbridge_beacon_primitives::{
 	BeaconHeader, BlockUpdate, Domain, ExecutionHeader, ExecutionHeaderState,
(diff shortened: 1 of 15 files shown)


parachain/runtime/snowbase/src/lib.rs
```diff
@@ -628,6 +628,7 @@ parameter_types! {
 	pub const MaxVoluntaryExitSize: u32 = 16;
 	pub const MaxAttestationSize: u32 = 128;
 	pub const MaxValidatorsPerCommittee: u32 = 2048;
+	pub const WeakSubjectivityPeriod: u32 = 27 * HOURS;
```
vgeddes, line 631: Can't we go with 2 weeks? There are currently [489599](https://beaconscan.com/validators#active) validators, much more than the minimum of 262144 validators needed to sustain a 2-week period.

Also I suspect there are going to be many times when the light client won't be updated for a day or 2, due to operational reasons (like upgrades, etc). So don't want to trigger the WS check unnecessary.

vgeddes: Also this line needs a comment with a brief explanation of what this period is.

parachain/pallets/ethereum-beacon-client/src/lib.rs
```diff
@@ -388,6 +393,22 @@ pub mod pallet {
...
+
+			log::info!(
+				target: "ethereum-beacon-client",
+				"💫 Checking weak subjectivity period. Current block number :{:?} Weak subjectvitity period check: {:?}.",
+				current_block_number,
+				weak_subjectivity_period_check
+			);
+
+			if current_block_number > weak_subjectivity_period_check.into() {
+				panic!("weak subjectvity period expired, reinitialize bridge")
```
vgeddes, line 409: Panicking is a big no-no. It will essentially brick the parachain and require a runtime upgrade to fix. Remember that we are going to be deploying our pallets onto the common-good BridgeHub parachain, which is shared with other bridges. We don't want to affect our neighbours.

vgeddes (commented):
Is a good start. However, the changes need to include a pallet dispatchable that Governance can call to re-activate header imports. I wonder if that would entail applying a new InitialSync?

parachain/pallets/ethereum-beacon-client/src/lib.rs
claravanstaden: Whoops! 😄 I'll need to rethink this then. Perhaps set a flag in storage that is checked, so no further processing happens in the light client until this is resolved. 

I am unsure of how the light client would get the go-ahead from governance to sync from an accepted checkpoint? Should there be an extrinsic with a specific origin? Or how could that work?

parachain/runtime/snowbase/src/lib.rs
claravanstaden: I am worried about a case where there is a major attack on the network and the vast majority of validators get slashed, and then the weak subjectivity threshold lowers to less than 2 weeks before we can update the threshold. It's unlikely, but not impossible.

parachain/pallets/ethereum-beacon-client/src/lib.rs
vgeddes: yeah a pallet dispatchable that can only be called by governance.

An example here: https://github.com/Snowfork/snowbridge/blob/317917cca47d62740ede01ceebc5d9e7d32ce24b/parachain/pallets/incentivized-channel/src/inbound/mod.rs#L164

claravanstaden:
I need to think about this a bit more, for example:

- Will the update be a light client initial sync? Need to think about sync committee updates that could have been skipped
- Need to potentially sync backwards in this case, not forwards, to make sure we are still on the right chain

parachain/pallets/ethereum-beacon-client/src/lib.rs
yrong: yeah, then maybe we just refactor the current dispatchable `sync_committee_period_update` a little bit for some [origin check](https://github.com/Snowfork/snowbridge/blob/317917cca47d62740ede01ceebc5d9e7d32ce24b/parachain/pallets/ethereum-beacon-client/src/lib.rs#L220) which we ignore now.

@vgeddes @claravanstaden and could we also allow to config these [beacon envs](https://github.com/Snowfork/snowbridge/blob/main/parachain/pallets/ethereum-beacon-client/src/config.rs#L13-L20) without hardcode?

claravanstaden: I think the complication here is that if we process a `sync_committee_period_update` as sent by governance, the last synced committee will be, for example, for period `50`. Say the light client went offline at period `40`. So now the relayers won't know to sync the committee periods `41` - `49`.

Besides syncing the sync committees, syncing the execution headers in the offline period should ideally be backwards as well, syncing from the provided governance checkpoint backwards to the last imported header. If we don't sync backwards and sync forwards (like the current relayer does), we could accidentally sync the forked chain and then never reach the provided checkpoint.

claravanstaden: > @vgeddes @claravanstaden and could we also allow to config these [beacon envs](https://github.com/Snowfork/snowbridge/blob/main/parachain/pallets/ethereum-beacon-client/src/config.rs#L13-L20) without hardcode?

Sure, in another PR though. 😄

claravanstaden:
Screenshot 2022-12-20 at 14 21 46

After chatting to @alistair-singh, he had a really good idea of solving the complexity of verifying the last synced checkpoint up to the provided weak subjectivity checkpoint. (The complexity comes in where the light client essentially needs to sync backwards to make sure we are syncing the right chain).

**Option 1:** Is to use the governance provided checkpoint and check the chain backwards (or sync forwards but only imports headers to storage once confirmed that the chain is syncing the checkpoint chain). The disadvantage to this approach is that it requires a specialized approach to syncing the gap in headers.

**Option 2:** Governance provides the forked header, so that when the light client verifies headers and comes across this fork header, it can reject it. This denylist of headers will allow normal forward processing of headers, as usual. The con is that if governance doesn't provide all the forked headers, it could still sync a forked chain. The likelihood of such a fork is quite slim, and multiple forks even less likely, it needs to be said.

vgeddes:
> **Option 2:** Governance provides the forked header, so that when the light client verifies headers and comes across this fork header, it can reject it. This denylist of headers will allow normal forward processing of headers, as usual. The con is that if governance doesn't provide all the forked headers, it could still sync a forked chain. The likelihood of such a fork is quite slim, and multiple forks even less likely, it needs to be said.

The assumption here is that governance knows upfront about the forked chain. I rather think that malicious header relays will be hiding their forked chain up until the moment they try to submit it to the pallet.

claravanstaden:
> The assumption here is that governance knows upfront about the maliciously forked chain. I rather think that malicious header relayers will be hiding their forked chain up until the moment they try to submit it to the pallet, when its too late for governance to intervene.

@vgeddes that is a good point... I was thinking it would work because the fork would be known on the Ethereum network, but I guess there's nothing stopping malicious relayers of starting their own fork (especially when paired with a eclipse attack).

We can chat about it in our meeting later, but in any case I'll continue working on the governance provided checkpoint sync then.

claravanstaden:
@vgeddes @alistair-singh @doubledup @yrong I've slimmed down this PR just to do the 27 hour header gap check and the blocked bridge flag. I will deal with the linear ancestry proofs in a separate PR.

parachain/pallets/ethereum-beacon-client/src/lib.rs
```diff
@@ -175,13 +178,19 @@ pub mod pallet {
 	#[pallet::storage]
 	pub(super) type LatestFinalizedHeaderSlot<T: Config> = StorageValue<_, u64, ValueQuery>;
 
+	#[pallet::storage]
+	pub(super) type LatestFinalizedHeaderImportTime<T: Config> = StorageValue<_, u64, ValueQuery>;
+
```
yrong, line 183: would it be better we move all LatestFinalizedHeader*** to a structure and make sure updated atomiclly?

parachain/pallets/ethereum-beacon-client/src/lib.rs
```diff
@@ -388,6 +415,23 @@ pub mod pallet {
 		}
 
 		fn process_finalized_header(update: FinalizedHeaderUpdateOf<T>) -> DispatchResult {
+			let import_time = <LatestFinalizedHeaderImportTime<T>>::get();
+			let weak_subjectivity_period_check = import_time + 97200;
```
yrong, line 419: seems better to move to a constant instead of hardcode

parachain/pallets/ethereum-beacon-client/src/lib.rs
claravanstaden: @yrong good point! I do this for the execution header: https://github.com/Snowfork/snowbridge/blob/main/parachain/pallets/ethereum-beacon-client/src/lib.rs#L178

I considered doing the same for the finalized header, but back then I kept it separate to simplify the forkless upgrade to Rococco, but since that's no longer a consideration, I agree, let me put them in one storage item.

parachain/pallets/ethereum-beacon-client/src/lib.rs
claravanstaden: Addressed in https://github.com/Snowfork/snowbridge/pull/742/commits/fb7483fd3fefee6127f2e5e5a100c93a7ec9b3a4.

yrong (approved):
LGTM!

PR #743, opened 2022-12-19
Xcm V3 support: XcmApp contracts

alistair-singh:


core/packages/contracts/contracts/XcmApp.sol
```diff
@@ -0,0 +1,145 @@
...
+
+import "./ChannelRegistry.sol";
+import "@openzeppelin/contracts/access/Ownable.sol";
+
+// TODO: put encoding functions here
+//import "./XCMAppPallet.sol";
+
+/// @title Proxy
+/// @notice A simple pass through proxy.
+contract Proxy is Ownable {
```
musnit, line 12: should we call this SovereignAccount instead of Proxy?

vgeddes: Nope, I think Proxies will eventually be used for all kinds of accounts on Polkadot. Specifically, `DescendOrigin` will result in another layering of proxying for simple accounts on parachains.

core/packages/contracts/contracts/XcmApp.sol
```diff
@@ -0,0 +1,148 @@
...
+/// @notice Implements XCM for the EVM.
+contract XcmApp {
+    /// @dev Channels which are allowed to call this app.
+    ChannelRegistry public immutable registry;
+
+    /// @dev A mapping or 32 byte hashed origins to proxy accounts.
+    mapping(bytes32 => Proxy) public proxies;
+
+    /// @dev A list of approved xcm executors.
+    mapping(bytes8 => address) public executors;
```
vgeddes, line 37: Lets drop support for multiple executors (and adding and removing them), so we can focus on the most important bits first. Can re-introduce this later on

core/packages/contracts/contracts/XcmApp.sol
```diff
@@ -0,0 +1,148 @@
...
+
+    /// @param _registry The channel registry which is the source of messages.
+    constructor(ChannelRegistry _registry) {
+        registry = _registry;
+    }
+
+    /// @dev Approves an executor version.
+    /// @param _version An identifier for the version.
+    /// @param _executor The executor to approve.
+    function approveExecutor(bytes8 _version, address _executor) external {
```
vgeddes, line 63: ditto, not important right now.

core/packages/contracts/contracts/XcmApp.sol
```diff
@@ -0,0 +1,148 @@
...
+        }
+        if (executors[_version] != address(0)) {
+            revert ExecutorExists();
+        }
+        executors[_version] = _executor;
+    }
+
+    /// @dev Revokes an executor version.
+    /// @param _version An identifier for the version.
+    function revokeExecutor(bytes8 _version) external {
```
vgeddes, line 76: ditto, not important right now.

core/packages/contracts/contracts/XcmApp.sol
musnit: Maybe simple accounts are still SovereignAccounts though? We should confirm if there is even a distinction between the two

core/packages/contracts/contracts/XcmApp.sol
```diff
@@ -0,0 +1,148 @@
...
+}
+
+/// @dev Executes Xcm instructions.
+contract XcmExecutor {
+    /// @dev The entry point for an payload.
+    function execute(bytes[] calldata instructions) external {
+        // TODO: registers like origin, holding, etc...
+        for (uint i = 0; i < instructions.length; i++) {
+            bytes1 instruction = instructions[i][0];
+            if (instruction == 0x00) {
```
vgeddes, line 123: Lets use [Enums](https://solidity-by-example.org/enum/) to identify instructions

vgeddes (commented):
Looks great! 🚀 

Aidan and I have been reviewing this and chatting about next steps. We can go over them tomorrow. But for now the important highlights:
* We probably won't represent asset transfer using XCM Transact
* Proxies may end up being stateful
* The proxy account for Statemint will likely need to own the ERC20 wrapped token contracts for tokens which are native to Statemint. Otherwise there is no way for the proxy account (or its xcm executor delegate) to mint new wrapped tokens.

musnit:
Based on discussion with Vincent, it seems like the best design should look something like the example below:

Example: Vincent wants to send 3 DOT to Ethereum

1) Vincent goes to UI and presses button, signs on wallet
2) UI constructs extrinsic:
 - ReserveAssetTransfer(DestinationOrigin, LocalAssetId, Amount, TargetAccount)
 - ReserveAssetTransfer("ethereum", "DOT", 3, "0x1234")
3) Extrinsic is submitted to Statemint
4) Statemint crafts an XCMv3 program/message that will eventually result in:
 a) The DOT is withdrawn from Vincent's account
 b) The DOT is deposited into the sovereign account of Ethereum
 c) An ExportMessage, with an origin from Statemint that contains a message to report the DepositAsset message to Ethereum
     DepositAsset(assets: [3 DOT], beneficiary: "0x1234")
5) The ExportMessage is routed via the bridge, through channels to Ethereum and picked up by the global XCMManager on Ethereum
6) The XCMManager processes the ExportedMessage as follows:
   i) It checks if the origin Statemint has a sovereign account contract on Ethereum or not
   ii) (optional) It deploys a sovereign account contract for Statemint if needed, as well as a DerivativeAssetManager contract for Statemint.
   iii) It submits the contents of ExportMessage to Statemint's Sovereign account on Ethereum
7) Statemint's Sovereign account on Ethereum gets the message(s) and executes the DepositAsset via submitting a delegatecall to the XCMExecution smart contract.
8) The XCMExecution code executes the DepositAsset message as follows:
   i) It gets the address of it's DerivativeAssetManager contract from the XCMManager
   ii) it executed a call to the DerivativeAssetManager
9)  DerivativeAssetManager runs as follows:
   i) Ensure the call comes from Statemint's sovereign account contract
   i) (optional) Creates the wDOT token if it does not exist yet and storing a mapping of Polkadot AssetId to corresponding ERC20 Smart Contract
   iii) Calls the mint function on the corresponding ERC20 Smart contract for each deposited asset.
10) The wDOT smart contract mints the token
   - i) It ensures the sender of the message is Statemint's Sovereign Contract Account
   - ii) It mints the tokens to the TargetAccount as requested

**Note on this design:**

We need to store the mapping between an AssetID and a ERC20 Smart Contract address for derivative assets.

For this, we have some requirements:
   - It should be possible for new mappings to be created dynamically to allow for new parachains to send their assets over the bridge automatically, without needing to set up anything beforehand
   - It should be possible for new assets to be added to the mappings dynamically, without asset token contracts automatically deployed as new assets are received
   - Different parachains should not share the same mappings or have shared access to each others mappings, since they may not trust each other
   - Sovereign Account contracts should be super minimal and ideally stateless, since there will be one for each parachain (and potentially even one for each user), so that we don't have to risk needing to upgrade thousands of them in the future

This brings a challenge - we need to store state on a per-parachain basis, but we want sovereign account contracts to be stateless, so there is a question of where to store this state. The above design opts to store this state on the XCMManager, via a DerivativeAssetManager contract created for every sovereign account that is deployed alongside that sovereign account's contract to manage asset mappings for derivative assets.

There are some other options that could be used here, that should be explored before finalizing the design.

Other potential options:
 a - Just store the asset mapping per sovereign account on the XCMManager directly, rather than in a DerivativeAssetManager contract
 b - Create a SovereignAccountState contract, with an address stored directly on the Sovereign Account proxy contract, giving up full statelessness to store just a single contract address in state.
 c - Use an alternative proxy implementation from https://docs.openzeppelin.com/contracts/4.x/api/proxy

core/packages/contracts/contracts/XcmApp.sol
vgeddes: I don't think so. My reading of the xcm spec leads me to believe that only _consensus systems_ can have sovereign accounts, where a consensus system is a relaychain, parachain, pallet, or a contract within a pallet. Simple accounts are not consensus systems because they don't have specialized consensus code. They are more like data structures.

Indeed the xcm spec refers to simple accounts as _Locations_:

>Location: A Consensus System, or an addressable account or data structure that exists therein. Examples include the Treasury account on the Polkadot Relay-chain

core/packages/contracts/contracts/XcmApp.sol
alistair-singh: Used enums and structures.

core/packages/contracts/contracts/XcmApp.sol
musnit: ok. though in the ethereum context there is no difference really

core/packages/contracts/contracts/XcmApp.sol
```diff
@@ -0,0 +1,77 @@
...
+
+import "./ChannelRegistry.sol";
+import "./XcmProxy.sol";
+import "./XcmAssetManager.sol";
+
+/// @title Xcm App
+/// @notice Implements XCM for the EVM.
+contract XcmApp {
+    /// @dev Channels which are allowed to call this app.
+    ChannelRegistry public immutable registry;
```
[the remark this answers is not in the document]
alistair-singh: Ok for now I will remove all references to channels from the prototype and will make XcmApp ownable. It will be owned by the basic channel.

core/packages/contracts/contracts/XcmApp.sol
```diff
@@ -0,0 +1,77 @@
...
+
+        XcmProxy proxy = proxies[_origin];
+        // JIT create proxy if it does not exist.
+        if (proxy == XcmProxy(address(0))) {
+            proxy = new XcmProxy();
+            proxies[_origin] = proxy;
+        }
+
+        // encode a call to the xcm executor
+        bytes memory encodedCall = bytes.concat(
```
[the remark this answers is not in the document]
alistair-singh: The reason I use `bytes.concat` instead of `abi.encodeCall` is because to `XcmApp` the xcm payload is opaque(i.e. `bytes` and not `Instruction[]`). `abi.encodeCall` looks at the the type of parameters and creates the wrong encoding. 

`XcmApp` sees opaque `bytes`. The `XcmExecutor` converts this to an `Instruction[]`. The reason for the opaqueness is because we want to eventually switch out the implementation of XcmExecutor and the data layout of instructions when there are new versions of Xcm.

core/packages/contracts/contracts/XcmExecutor.sol
```diff
@@ -0,0 +1,103 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity ^0.8.9;
+
+import "./XcmAssetLookup.sol";
+
+/// @dev Executes Xcm instructions.
+contract XcmExecutor {
+    /// @dev Represents the type of instruction.
+    enum InstructionKind {
```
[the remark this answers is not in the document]
alistair-singh: Yeah, this has deviated from what we discussed during our early design talks. During implementation, I realized that a lot of the Xcm instructions map well to the ERC20 spec.

`TransferAsset`
`ERC20.transferFrom(source, destination, amount)` - the proxy never controls the asset. The source needs to approve the proxy as a spender using ERC20 allowance.

`WithdrawAsset`
`ERC20.transferFrom(source, proxy, amount)` - pull asset from the source into the proxy. The source needs to approve the proxy as a spender using ERC20 allowance.

`DepositAsset`
`ERC20.transfer(destination, amount)` - move asset owned by the proxy to some destination.

`ReserveAssetDeposited`
`ERC20.mint(proxy, amount)` - Create/mint a new asset owned by the proxy.

We probably won't need the following however for completeness:

`BurnAsset`
`ERC20.burn(proxy, amount)` - Burn asset owned by the proxy. 

`BuyExecution`
`ERC20.transfer(relayer, amount)` - BuyExecution can even be a transfer from the proxy to the relayer account.

`Transact`
An arbitrary low-level `call` with the proxy as `msg.sender`.

So basically I decided to build the `mintPolkadotAssets` in terms of `(ReserveAssetsDeposited, DepositAsset)` and then we could discuss it here. It is more complex than a batch instruction but just this subset of Xcm instructions will allow a user to manage ERC20 tokens using Xcm and the bridge.

The design also becomes simpler if we think of the XcmApp as a low-level, "dumb", component that allows a sovereign account to use ERC20 tokens via a proxy and Xcm instructions. The XcmApp should not be aware of how these tokens are created, how they are secured, or how they map to assets on the other side of the bridge. I feel like that should be handled at a higher level of abstraction. I go into a little more detail about asset management specifically in this notion doc.
https://www.notion.so/snowfork/XCM-v3-The-four-major-transfer-cases-f0aad787455444d4a25f28829637ee45

core/packages/contracts/contracts/XcmExecutor.sol
```diff
@@ -0,0 +1,103 @@
...
+
+    /// @dev an asset that is held in reserved was deposited. This equates to an ERC20 mint.
+    /// @param lookup lookup ERC20 tokens for an asset.
+    /// @param data the instruction data.
+    function reserveAssetDeposited(
+        XcmAssetLookup lookup,
+        ReserveAssetDepositedData memory data
+    ) internal {
+        require(data.amount > 0, "must reserve a positive amount");
+        XcmFungibleAsset asset = lookup.lookupOrCreate(data.assetHash);
```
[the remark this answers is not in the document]
alistair-singh: Yeah so I have done a lot of thinking around asset management. And I am basically inclined to do asset management on the Substrate side. More in this doc:
https://www.notion.so/snowfork/XCM-v3-The-four-major-transfer-cases-f0aad787455444d4a25f28829637ee45

TLDR; 
1. AssetManager lives on the substrate side.
2. All MultiAssets are resolved to H160 contract addresses during Xcm lowering.
3. XcmExecutor is “dump” and just invokes ERC20 methods on ERC20 token contracts.
4. Assets are added to the AssetManager manually through governance.
5. We need escrow(vaults) accounts. Proxies, specifically Statemint’s Proxy, is an escrow(vault) for Ethereum native assets.

core/packages/contracts/contracts/XcmExecutor.sol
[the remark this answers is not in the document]
vgeddes: Hmm, I think the equivalence of those XCM instructions to ERC20 ops is fairly weak. For example, `WithdrawAsset` reduces an account balance and transfers the amount to an ephemeral holding register. There's no equivalence to that in ERC20.

And we may have to support certain XCM instructions such as `ClearOrigin` for which there is no equivalence in ERC20. Though for our limited executor, `ClearOrigin` should just be a no-op.
 
I think for reasons of performance, security, and simplicity, we should batch instructions. We need only support these two batches: 

* Transfer Polkadot-native assets to Ethereum:
  * (ReserveAssetDeposited, ClearOrigin, DepositAsset)
* Transfer Ethereum-native assets to Ethereum:
  * (WithdrawAsset, ClearOrigin, DepositAsset)

core/packages/contracts/test/apps/xcm_app.test.ts
```diff
@@ -0,0 +1,206 @@
+import { defaultAbiCoder } from "@ethersproject/abi"
+import { expect, loadFixture } from "../setup"
+import { xcmAppFixture } from "./fixtures"
+import { XcmFungibleAsset__factory } from "../../src"
+
+let PARA_2001_ORIGIN = "0x2001000000000000000000000000000000000000000000000000000000000000"
+let PARA_2002_ORIGIN = "0x2002000000000000000000000000000000000000000000000000000000000000"
+
+describe("XCMApp", function () {
```
vgeddes, line 9: We gonna migrate to Foundry, so I'd pause writing hardhat tests until then (and when our design is more concrete)

PR #744, opened 2022-12-20
Refactor type conversions using TryFrom impls

claravanstaden:
Refactor of SSZ conversion types to use `TryFrom` impls instead of manual conversions.

Also adds error type conversions so manual `map_err` isn't necessary anymore.


alistair-singh (approved):
Nice! Code is much neater.

vgeddes (approved):

PR #746, opened 2023-01-05
Undo Changes for Mac dev

claravanstaden:
Turns out I didn't need to change anything in our project to work on Mac after @yrong's changes. 😄 

This turned out to be the culprit of all my problems:

/etc/hosts
```
127.0.0.1       localhost
255.255.255.255 broadcasthost
#::1             localhost <---- comment this out
```

### DIFF
--- a/core/packages/test/scripts/deploy-ethereum.sh
+++ b/core/packages/test/scripts/deploy-ethereum.sh
@@ -14,8 +14,6 @@ start_geth() {
             --http --http.api debug,personal,eth,net,web3,txpool,engine,miner --ws --ws.api debug,eth,net,web3 \
             --rpc.allow-unprotected-txs --mine --miner.threads=1 \
             --miner.etherbase=0x0000000000000000000000000000000000000000 \
-            --authrpc.addr="127.0.0.1" \
-            --http.addr="127.0.0.1" \
             --allow-insecure-unlock \
             --authrpc.jwtsecret config/jwtsecret \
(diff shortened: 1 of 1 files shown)


vgeddes (commented):
Hey your initial change was good and we don't need to revert it. I also have the `::1  localhost` line in my config, and so presumably other macbooks will also have it.

alistair-singh:
> Hey your initial change was good and we don't need to revert it. I also have the `::1 localhost` line in my config, and so presumably other macbooks will also have it.

Yeah I thinks its there by default. Mine has it too.

claravanstaden:
Unfortunately, the work is half done if these changes are left in. To get the setup to work with `::1 localhost`, there needs to be lodestar config changes too. I know what to add for it to work for lodestar, but unfortunately the parachain/relaychain setup doesn't work with `::1 localhost` and I haven't figured out how to fix it yet. Presumably also some replacement of `localhost` with `127.0.0.1` somewhere in the config.

So I was hesitant to leave something half-baked in. But perhaps I should just at least add the lodestar config changes then too? It's also interesting that @yrong didn't come across this, please weigh in if you did. 😄

yrong:
> So I was hesitant to leave something half-baked in. But perhaps I should just at least add the lodestar config changes then too? It's also interesting that @yrong didn't come across this, please weigh in if you did. 😄

yes, I did not met the issue with or without this config in my Mac-Air-M1.  But I would assume the script is only for dev local setup so I'm ok if we add it or not.

claravanstaden:
Closing for now, will reopen if other team members have issues with this.

PR #749, opened 2023-01-09
Test/e2e speed up

yrong:


yrong:
- [x] split bootstrap test to two directions as to be parallelized
- [x] hack lodestar param

with this hack we can decrease execution time of bootstrap test by at least 50%

claravanstaden (approved):
🤩 Incredible!! 👏 This will save us so much time, thank you @yrong! I tested locally and it works perfectly. I am surprised that the Bootstrap tests can be run in parallel because I know the Substrate -> Eth one needed to be run first to set up SnowETH to transfer in the first place, but looks like it works fine. 🤷‍♀️ 

I made a small PR to your PR to print out the time start-services took and also to remove an unnecessary check for the beacon relayer that speeds up the script a little. https://github.com/Snowfork/snowbridge/pull/750

claravanstaden:
PS. The Lodestar `SLOTS_PER_EPOCH` is brilliant! 😂 Works like a charm.

core/packages/test/src/subclient/index.js
```diff
@@ -13,6 +13,23 @@ class SubClient {
     const provider = new WsProvider(this.endpoint);
     this.api = await ApiPromise.create({
       provider,
+      rpc: {
+        basicOutboundChannel: {
+          getMerkleProof: {
```
doubledup, line 18: Thanks for this! That warning was annoying

doubledup (approved):
This looks great! 🚀

vgeddes (approved):
+1

PR #754, opened 2023-01-11
test: zombienet launch

yrong:
replace `polkadot-launch` with [zombienet](https://github.com/paritytech/zombienet)

### DIFF
--- a/core/packages/test/README.md
+++ b/core/packages/test/README.md
@@ -18,6 +18,7 @@ The E2E tests run against local deployments of the parachain, relayer, the ether
 
 * direnv - https://direnv.net/
 
+
 ## Setup
 
 ### Install NPM dependencies
@@ -27,6 +28,10 @@ Make sure to install dependencies for all packages (contracts, api, test):
 ```bash
 cd ../.. && pnpm install
 ```
+
+### Install Zombienet
+
+Zombienet is a cli tool to easily spawn ephemeral Polkadot/Substrate networks and perform tests. Just download [latest binary release](https://github.com/paritytech/zombienet/releases/) and copy it to $PATH which can be executed directly from command line
 
 ### Configure testnet
 
--- a/core/packages/test/config/launch-config.toml
+++ b/core/packages/test/config/launch-config.toml
@@ -1,0 +1,95 @@
+[settings]
+node_spawn_timeout = 120
+provider = "native"
+timeout = 600
+
+## Rococo
+[relaychain]
+chain = "rococo-local"
+default_command = "/tmp/snowbridge/bin/polkadot"
+default_args = ["-lparachain=debug","--beefy","--enable-offchain-indexing=true","--pruning","archive"]
+
+[relaychain.genesis.runtime.runtime_genesis_config.configuration.config]
+max_validators_per_core = 2
+needed_approvals = 1
+validation_upgrade_cooldown = 2
+validation_upgrade_delay = 2
+
+[[relaychain.nodes]]
+name = "rococo-alice"
+rpc_port = 30444
+validator = true
+ws_port = 9944
+
+[[relaychain.nodes]]
+name = "rococo-bob"
+validator = true
+
+[[relaychain.nodes]]
+name = "rococo-charlie"
+validator = true
+
+[[relaychain.nodes]]
+name = "rococo-dave"
+validator = true
+
+# Snowbridge
+[[parachains]]
+add_to_genesis = true
+chain_spec_path = "/tmp/snowbridge/spec.json"
+cumulus_based = true
+id = 1000
+
+[[parachains.collators]]
+command = "/tmp/snowbridge/bin/snowbridge"
+name = "snowbrige-alice"                              
+rpc_port = 31200
+validator = true
+ws_port = 11144
+args = ["--enable-offchain-indexing=true","--pruning=archive"]
+
+[[parachains.collators.env]]
+name = "RUST_LOG"
+value = "runtime=debug,parachain=trace,cumulus-collator=trace,aura=trace,xcm=trace"
+
+[[parachains.collators]]
+command = "/tmp/snowbridge/bin/snowbridge"
+name = "snowbrige-bob"
+validator = true
+args = ["--enable-offchain-indexing=true","--pruning=archive"]
+
+# Test Parachain
+[[parachains]]
+add_to_genesis = true
+chain_spec_path = "/tmp/snowbridge/test_spec.json"
+cumulus_based = true
+id = 1001
+
+[[parachains.collators]]
+command = "/tmp/snowbridge/bin/snowbridge-test-node"
+name = "test-parachain-alice"                              
+rpc_port = 31202
+validator = true
+ws_port = 13144
+
+[[parachains.collators.env]]
+name = "RUST_LOG"
+value = "runtime=debug,parachain=trace,cumulus-collator=trace,aura=trace,xcm=trace"
+
+[[parachains.collators]]
+command = "/tmp/snowbridge/bin/snowbridge-test-node"
+name = "test-parachain-bob"
+validator = true
+
+# XCM Channels
+[[hrmp_channels]]
+max_capacity = 8
+max_message_size = 512
+recipient = 1000
+sender = 1001
+
+[[hrmp_channels]]
+max_capacity = 8
+max_message_size = 512
+recipient = 1001
+sender = 1000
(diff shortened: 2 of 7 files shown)


alistair-singh (approved):

core/packages/test/README.md
```diff
@@ -28,6 +29,10 @@ Make sure to install dependencies for all packages (contracts, api, test):
 cd ../.. && pnpm install
 ```
 
+### Install Zombienet
```
vgeddes, line 32: Should perhaps add a link to this section from the Requirements section.

core/packages/test/config/launch-config.toml
```diff
@@ -0,0 +1,95 @@
...
+[relaychain]
+chain = "rococo-local"
+default_command = "/tmp/snowbridge/bin/polkadot"
+default_args = ["-lparachain=debug","--beefy","--enable-offchain-indexing=true","--pruning","archive"]
+
+[relaychain.genesis.runtime.runtime_genesis_config.configuration.config]
+max_validators_per_core = 2
+needed_approvals = 1
+validation_upgrade_cooldown = 2
+validation_upgrade_delay = 2
```
vgeddes, line 16: Think we should also port over the following from the old config:
```
    "genesis": {
      "runtime": {
        "session_length_in_blocks": 10
      }
    }
```

For testing BEEFY, we may want to alter the session length.

core/packages/test/config/launch-config.toml
```diff
@@ -0,0 +1,95 @@
...
+[[parachains]]
+add_to_genesis = true
+chain_spec_path = "/tmp/snowbridge/spec.json"
+cumulus_based = true
+id = 1000
+
+[[parachains.collators]]
+command = "/tmp/snowbridge/bin/snowbridge"
+name = "snowbrige-alice"                              
+rpc_port = 31200
```
vgeddes, line 46: The rpc port for alice is actually 8081, according to the old config. And the p2p port is 31200.

core/packages/test/scripts/deploy-polkadot.sh
```diff
@@ -51,7 +50,15 @@ start_polkadot_launch()
...
         ' \
         "$output_dir/test_spec.json" | sponge "$output_dir/test_spec.json"
+}
+
+wait_start() {
+    scripts/wait-for-it.sh -t 120 localhost:11144
+    scripts/wait-for-it.sh -t 120 localhost:13144
+}
 
+polkadot_launch() {
```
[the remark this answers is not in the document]
yrong: would suggest for keep it for some cooling period, and remove it later in next pr

core/packages/test/config/launch-config.toml
yrong: already defined in rococo chain spec so should be duplicated
https://github.com/paritytech/polkadot/blob/a779892f8ee2a072ac9676d7949ac4539f630ecd/node/service/src/chain_spec.rs#L1925

claravanstaden (approved):
🧟 Very nice!!

core/packages/test/config/launch-config.toml
```diff
@@ -0,0 +1,95 @@
...
+
+[[parachains.collators.env]]
+name = "RUST_LOG"
+value = "runtime=debug,parachain=trace,cumulus-collator=trace,aura=trace,xcm=trace"
+
+[[parachains.collators]]
+command = "{{output_bin_dir}}/snowbridge"
+name = "snowbrige-bob"
+validator = true
+args = ["--enable-offchain-indexing=true","--pruning=archive"]
```
doubledup, line 59: Should we add the rpc_port and ws_port values here too?
```suggestion
[[parachains.collators]]
command = "{{output_bin_dir}}/snowbridge"
name = "snowbrige-bob"
rpc_port = 8082
validator = true
ws_port = 11155
args = ["--enable-offchain-indexing=true","--pruning=archive"]
```

core/packages/test/config/launch-config.toml
```diff
@@ -0,0 +1,95 @@
...
+ws_port = 13144
+
+[[parachains.collators.env]]
+name = "RUST_LOG"
+value = "runtime=debug,parachain=trace,cumulus-collator=trace,aura=trace,xcm=trace"
+
+[[parachains.collators]]
+command = "{{output_bin_dir}}/snowbridge-test-node"
+name = "test-parachain-bob"
+validator = true
```
doubledup, line 82: Similar to the change on the Snowbridge chain, don't we need these port numbers?
```suggestion
[[parachains.collators]]
command = "{{output_bin_dir}}/snowbridge-test-node"
name = "test-parachain-bob"
rpc_port = 8084
validator = true
ws_port = 13155
```

core/packages/test/config/launch-config.toml
```diff
@@ -0,0 +1,95 @@
...
+name = "rococo-bob"
+validator = true
+
+[[relaychain.nodes]]
+name = "rococo-charlie"
+validator = true
+
+[[relaychain.nodes]]
+name = "rococo-dave"
+validator = true
```
doubledup, line 34: Should we include the RPC and websocket ports here, like with the rococo-alice node?
```suggestion
[[relaychain.nodes]]
name = "rococo-bob"
rpc_port = 30555
validator = true
ws_port = 9955

[[relaychain.nodes]]
name = "rococo-charlie"
rpc_port = 30666
validator = true
ws_port = 9966

[[relaychain.nodes]]
name = "rococo-dave"
rpc_port = 30777
validator = true
ws_port = 9977
```

core/packages/test/scripts/set-env.sh
```diff
@@ -11,9 +11,9 @@ lodestar_version="${LODESTAR_VER:-1.2.2}"
 contract_dir="$core_dir/packages/contracts"
 relay_dir="$root_dir/relayer"
 relay_bin="$relay_dir/build/snowbridge-relay"
-output_dir=/tmp/snowbridge
-output_bin_dir="$output_dir/bin"
-ethereum_data_dir="$output_dir/geth"
+export output_dir=/tmp/snowbridge
+export output_bin_dir="$output_dir/bin"
+export ethereum_data_dir="$output_dir/geth"
```
doubledup, line 16: Why do we need to export `ethereum_data_dir`? I see we're using `output_dir` and `output_bin_dir` in the zombienet config, but can't find a use of `ethereum_data_dir`

doubledup (approved):
Looks good! There are some missing port numbers, not sure if we need them

core/packages/test/scripts/set-env.sh
alistair-singh: It is used to launch a local eth node. 
https://github.com/Snowfork/snowbridge/pull/754/files#diff-d6b5dd8f20f1dcdd2b05128403e3822304831554c07c815c8e94206a9f1cb567R11-R12

alistair-singh (approved):

core/packages/test/config/launch-config.toml
doubledup: Can we define `session_length_in_blocks` here and use it in the chain spec?

vgeddes: I'm fine with duplicating it in the config. Its there for documentation purposes, so that we know what the current session length is, and how to change it.

yrong:
> Looks good! There are some missing port numbers, not sure if we need them

Yes, we can explicit declaring so to connect with `polkadot-app` to check info. But I would assume explicit one node for each chain should be enough to make config a bit clean

core/packages/test/config/launch-config.toml
yrong: https://github.com/Snowfork/snowbridge/pull/754#issuecomment-1384709684

core/packages/test/config/launch-config.toml
yrong: https://github.com/Snowfork/snowbridge/pull/754#issuecomment-1384709684

core/packages/test/config/launch-config.toml
yrong: https://github.com/Snowfork/snowbridge/pull/754#issuecomment-1384709684

yrong: > Should we add the rpc_port and ws_port values here too?

Yes, we can explicit declaring all, but I would assume one node for each chain should be enough, normally we only connect to one node to check runtime info and others are all the same. And if you really want to connect to other node, you can just check the output of zoombienet which you can find the port to connect with.

core/packages/test/scripts/set-env.sh
yrong: yes, though we use it we do need to export this param. But seems not a big issue will not hurt? Anyway just change it.

vgeddes (approved):
+1

alistair-singh (approved):

PR #755, opened 2023-01-12
fix benchark params

yrong:
then we need to re-benchark based on it to update the weight required
https://github.com/Snowfork/snowbridge/blob/09404845e0fb34783576a9513b59043b8bff6349/parachain/pallets/basic-channel/src/outbound/weights.rs#L49-L55


yrong:
closed as we may re-design-implement the current batching model

PR #753, opened 2023-01-10
Remove apps & incentivized channel from parachain

doubledup:
Builds pass for all runtimes, the query-events tool and the test parachain. Not sure I've caught everything that can be removed in this pass. Notably the assets and treasury pallets can probably be removed with the move to Statemint and BridgeHub.

This is merging into a feature branch to keep the PRs (relatively) small. I'll make a PR for the E2E tests once the relayer & contracts are updated.

PR #756, opened 2023-01-12
Remove incentivized channel  from relayer

doubledup:
Will regenerate the bindings in `contracts/` in the Solidity PR. `mage build` passes. Will run E2E tests once Solidity code is removed.

The main changes here are follow-ons from #753:
- Remove the incentivized channel.
- Remove channelID.

Also removed now-unused code & simplified code that now only has a single path. eg. `Events` struct is now just `BasicChannelEvent` and we don't need a mapping from address to contract method.


relayer/relays/ethereum/ethereum-listener.go
```diff
@@ -20,25 +22,24 @@ import (
...
-	headerSyncer                *syncer.Syncer
-	initBlockHeight             uint64
-	descendantsUntilFinal       uint64
+	ethashDataDir        string
+	ethashCacheDir       string
+	config               *SourceConfig
+	conn                 *ethereum.Connection
+	basicOutboundChannel *basic.BasicOutboundChannel
+	// TODO: remove mapping, it only maps the basic channel address to the string "BasicInboundChannel.submit"
+	mapping               map[common.Address]string
```
vgeddes, line 38: Prefer to remove/refactor in this PR.

doubledup: Definitely. I had done it, just hadn't pushed until now.

yrong:
LGTM, Simplifies the codebase a lot, Great! @doubledup

claravanstaden (approved):
Very nice! ✨ 

Can we remove these files too? https://github.com/Snowfork/snowbridge/tree/main/relayer/contracts/incentivized

doubledup:
@claravanstaden The Go contract bindings are removed in #757 where we're removing the contracts. I'd actually forgotten to check that the incentivized channel bindings are removed there, so good catch! 🙏

vgeddes (approved):
+1

PR #758, opened 2023-01-13
Move commit to on_idle with weight check

yrong:
- [x] move commit to `on_idle`


vgeddes:
> * [ ]   re-run benchmark in collator node

I would not worry about benchmarks in this PR. I would just use dummy weights if necessary. Only after we implement https://www.notion.so/snowfork/Basic-Channel-V3-0e46926b19d644f0be4f0d9be18d18f1 do we need proper benchmarks

alistair-singh (commented):
This change looks good! Reading up on the `on_idle` hook, it will not run if there is not enough weight left before block finalization. During testing in Rococo last year our beacon light client consumed quite a bit of block weight. We also will be sharing block space with other bridge pallets when we run on the BridgeHub.

So I am wondering if a scenario where the bridge is never "idle" enough to process the commit could happen. And how would we recover from something like that?

https://paritytech.github.io/substrate/master/frame_support/traits/trait.Hooks.html#method.on_idle

vgeddes (approved):
Looks good!

Can consider moving to OCW workers later.

parachain/pallets/basic-channel/src/outbound/test.rs
```diff
@@ -135,7 +135,8 @@ fn test_submit_exceeds_payload_limit() {
 		let who: &AccountId = &Keyring::Bob.into();
 
 		let max_payload_bytes = MaxMessagePayloadSize::get();
-		let payload: Vec<u8> = (0..).take(max_payload_bytes as usize + 1).collect();
+
+		let payload: Vec<u8> = vec![1u8; (max_payload_bytes as usize + 1)].try_into().unwrap();
```
doubledup, line 139: What's the reason for this change? Are we just replacing `0, 1, 2, ...` with `1, 1, 1, ...`?

doubledup (approved):

alistair-singh (approved):

parachain/pallets/basic-channel/src/outbound/test.rs
yrong: because we change `MaxMessagePayloadSize` to 256 which does not apply to u8 if we still sample data like this. But I did  a little tweak to work as previously and also make test passed

PR #759, opened 2023-01-16
Remove incentivized channel and apps from test & cleanup 

doubledup:
Removes all of the E2E tests, as these all depended on the apps. This also removes some left over references like channelID, along with the assets-registry pallet. There are more things we could remove, but I want to solicit feedback here first:

- `core/packages/test/src/` has clients, helpers & fixtures that aren't used now that the tests aren't around. We could remove these, especially with the move to Foundry I'm not sure we can get much value out of them.
- `core/packages/api/` is an example of using the apps, so we can probably remove it now that we don't have apps.
- The assets pallet is [used by the XCM pallet](https://github.com/Snowfork/snowbridge/blob/main/parachain/runtime/snowbridge/src/lib.rs#L286) (Assets -> FungiblesTransactor -> AssetTransactors -> XcmConfig). Do we want to remove the XCM pallet as well, or leave the assets pallet in? Do we need our own AssetTransactors when using Statemint?


core/packages/test/scripts/detect-fork.ts
```diff
@@ -118,24 +118,12 @@ const NONCES = [
     storageKey:
       "0x684b82bef882079feeabe54a5bd7b94a718368a0ace36e2b1b8b6dbd7f8093c0",
   },
-  {
```
alistair-singh, line 121: This file can probably deleted completely. It was made to handle hardforks in the ethereum 1 light client which we do not use anymore.

alistair-singh (approved):
Nice!

claravanstaden (approved):
🗡️ Nice!!

vgeddes:
> * `core/packages/test/src/` has clients, helpers & fixtures that aren't used now that the tests aren't around. We could remove these, especially with the move to Foundry I'm not sure we can get much value out of them.

yes, lets remove.

> * `core/packages/api/` is an example of using the apps, so we can probably remove it now that we don't have apps.

I want to keep the `api` package even if its just a stub. We'll re-implement the package once the XCMv3 work is done.

> * The assets pallet is [used by the XCM pallet](https://github.com/Snowfork/snowbridge/blob/main/parachain/runtime/snowbridge/src/lib.rs#L286) (Assets -> FungiblesTransactor -> AssetTransactors -> XcmConfig). Do we want to remove the XCM pallet as well, or leave the assets pallet in? Do we need our own AssetTransactors when using Statemint?

The assets pallet won't be needed. Instead, we'll likely want to run Statemint as a second parachain in our testnet. But that can come later.

PR #760, opened 2023-01-17
Add nix env

vgeddes:
The idea is that we can have a reproducible dev environment across Linux and Mac machines. Seems to be working on my MacOs machine so far.

Steps:

1. [Install](https://nixos.org/download.html) nix
2. run `nix-shell` in the root of our monorepo

Then our developer environment should mostly be ready

TODO
- [x] Install utility packages needed for E2E stack.
- [x] Install Rust stable and nightly using the `rustup` nix package. see https://nixos.wiki/wiki/Rust.
- [x] Compatibility with code editors, e,g VSCode


doubledup:
This is awesome! 😄 There's also `direnv` integration, which removes the need for an explicit `nix-shell`. Just go to the project and `nix-shell` loads automatically! Unfortunately there are at least 3 options for this:

- [plain `direnv`](https://determinate.systems/posts/nix-direnv): simple enough, not sure about editor integration.
- [`lorri`](https://github.com/nix-community/lorri): allegedly faster than `direnv` and aims to support editor integration.
- [`nix-direnv`](https://github.com/nix-community/nix-direnv): allegedly simpler (no external daemon) & faster than `lorri`, with flake support.

I'm keen to start with `direnv` alone and see how much friction we have. All the other options rely on `direnv` anyway.

doubledup:
[`niv`](https://github.com/nmattia/niv/) could be handy for managing dependencies, but also something for later.

doubledup (approved):
This is a great start! Looking forward to it 😄

doubledup:
It's probably a good idea to use [flakes](https://www.tweag.io/blog/2020-05-25-flakes/) instead of plain Nix channels.

alistair-singh (approved):
Looks good! Will test this out on my mac.

shell.nix
```diff
@@ -0,0 +1,23 @@
...
+        pkgs.go
+        pkgs.rustup
+
+        pkgs.go-ethereum
+        pkgs.gcc
+        pkgs.jq
+        pkgs.moreutils
+
+        pkgs.cowsay
+    ];
```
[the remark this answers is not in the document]
vgeddes: Nice! added this.

shell.nix
```diff
@@ -0,0 +1,23 @@
+{pkgs ? import <nixpkgs> {}}:
```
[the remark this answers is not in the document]
vgeddes: Added

vgeddes:
> Looks good! Will test this out on my mac.

Got it working on my mac with the latest changes

.gitignore
```diff
@@ -15,3 +15,9 @@ node_modules/
 
 relaychain/
 .vscode
+
+.rustup
+.cargo
+
+.DS_Store
+
```
doubledup, line 23: (nit) Extra whitespace
```suggestion
```

shell.nix
```diff
@@ -0,0 +1,44 @@
...
+    shellHook = ''
+        export CARGO_HOME=$PWD/.cargo
+        export RUSTUP_HOME=$PWD/.rustup
+        export PATH=$CARGO_HOME/bin:$PATH
+
+        rustup install 1.66.1
+        rustup install nightly-2022-11-15
+        rustup target add wasm32-unknown-unknown --toolchain nightly-2022-11-15
+        rustup default 1.66.1
+
```
doubledup, line 41: We also need to ensure that the direnv hook runs:
```suggestion

        eval "$(direnv hook bash)"

```

doubledup (approved):
LGTM. Save for the pkill issue blocking pure shells on MacOS, so we can't be sure that this is reproducible yet (unless you've run it with a pure shell on Linux already?). But at least we have a single place for all external dependencies now 🙌

core/packages/test/scripts/set-env.sh
```diff
@@ -37,42 +37,9 @@ address_for()
...
-}
-
-kill_relayer() {
-    echo "Killing relayer"
-    kill_trap
-    sleep 3
-    pkill -9 snowbridge-relay
-    sleep 1
+    trap - SIGTERM
+    kill 0
```
vgeddes, line 42: @yrong I changed the kill behavior here to be a bit more simpler. `kill 0` will kill the whole process group of the current process. I verified that this works on MacOS, and should also work on Linux too (will test).

shell.nix
```diff
@@ -0,0 +1,23 @@
...
+    buildInputs = [
+        pkgs.nodejs-18_x
+        pkgs.nodePackages.pnpm
+        pkgs.go
+        pkgs.rustup
+
+        pkgs.go-ethereum
+        pkgs.gcc
+        pkgs.jq
+        pkgs.moreutils
```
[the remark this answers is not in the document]
vgeddes: Got a pure shell working.

core/packages/test/scripts/start-ethereum.sh
```diff
@@ -12,7 +12,7 @@ start_chains()
 
 if [ -z "${from_start_services:-}" ]; then
     echo "start ethereum only!"
-    trap kill_ethereum SIGINT SIGTERM EXIT
+    trap kill_all SIGINT SIGTERM EXIT
```
yrong, line 15: IMHO starting and killing chains separately could be helpful in dev stage when we only change some code/logic in parachain, then no need to restart ethereum and deploy contracts again

core/packages/test/scripts/start-polkadot.sh
```diff
@@ -19,7 +19,7 @@ start_chains()
 
 if [ -z "${from_start_services:-}" ]; then
     echo "start polkadot only!"
-    trap kill_polkadot SIGINT SIGTERM EXIT
+    trap kill_all SIGINT SIGTERM EXIT
```
yrong, line 22: https://github.com/Snowfork/snowbridge/pull/760/files#r1090111239

core/packages/test/scripts/set-env.sh
```diff
@@ -37,42 +37,9 @@ address_for()
...
-    pkill -9 -f polkadot-launch
-    pkill -9 zombienet
-}
-
-kill_relayer() {
-    echo "Killing relayer"
-    kill_trap
-    sleep 3
-    pkill -9 snowbridge-relay
-    sleep 1
```
yrong, line 75: IMHO starting and killing chains separately could be helpful in dev stage, 
e.g: when we only change some code/logic in parachain, then no need to restart ethereum and deploy contracts again

doubledup:
A pure shell is hanging for me: I reach `Deploying contracts` but not `Exported contract artifacts`. Trying `git clean` with a fresh build.

vgeddes:
> A pure shell is hanging for me: I reach `Deploying contracts` but not `Exported contract artifacts`. Trying `git clean` with a fresh build.

Ah yeah, so that's actually an incompatibility Geth/Lodestar has with MacOS, where it binds ::1 to localhost in /etc/hosts.

I had exactly the same issue on MacOS

Comment out that line, create a new terminal shell, and things should be fine

doubledup:
Hmm, that line's already commented out for me (has been for at least a week), and I've even restarted since.

doubledup:
Looks like [nix-shell is intended for making packages](https://discourse.nixos.org/t/nix-shell-does-not-use-my-users-shell-zsh/5588/11) rather than consuming them. [`lorri` addresses specific pain points](https://www.tweag.io/blog/2019-03-28-introducing-lorri/) and [devenv](https://devenv.sh/) looks like an alternative. Let's just bear this in mind and see how far we get though.

yrong (approved):
Looks good!

alistair-singh:
I have seen this issue on Ubuntu 22.04
```
error: failed to run custom build command for `librocksdb-sys v0.6.1+6.28.2`

Caused by:
  process didn't exit successfully: `/home/alistair/snowbridge/relaychain/target/release/build/librocksdb-sys-e159a757d1e18cc7/build-script-build` (exit status: 101)
  --- stderr
  thread 'main' panicked at 'Unable to find libclang: "the `libclang` shared library at /usr/lib/x86_64-linux-gnu/libclang-14.so.14.0.0 could not be opened: libLLVM-14.so.1: cannot open shared object file: No such file or directory"', /home/alistair/snowbridge/.cargo/registry/src/github.com-1ecc6299db9ec823/bindgen-0.59.2/src/lib.rs:2144:31
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
warning: build failed, waiting for other jobs to finish...
Terminated
```

I have tried the following packages in the nix config to remedy this with no luck.
* clang
* llvm
* libclang
* libllvm
* clang_14
* llvm_14

Has anyone seen this on Ubuntu?

doubledup:
@alistair-singh Haven't tried Linux, I'll give that a look. Does the file `/usr/lib/x86_64-linux-gnu/libLLVM-14.so.1` exist, or does running `sudo find / -name libLLVM-14.so.1` return any results? (with all of the mentioned packages installed)

alistair-singh:
> @alistair-singh Haven't tried Linux, I'll give that a look. Does the file `/usr/lib/x86_64-linux-gnu/libLLVM-14.so.1` exist, or does running `sudo find / -name libLLVM-14.so.1` return any results? (with all of the mentioned packages installed)

Yeah so it does exist and the build works from outside a nix shell. Its almost like nix is ignoring it because it's not explicitly listed in the config. I think the root cause is that librocksdb is c package wrapped in a rust crate and it is not picking up nix's linker flags.

Looking at `echo $NIX_LDFLAGS` llvm is installed and present.
```
 -L/nix/store/x5sgaacc0y94wix2yl0nkim56qm946lq-llvm-14.0.6-lib/lib
```

core/packages/test/scripts/set-env.sh
vgeddes: Sure, can revert my change.

doubledup:
Got it building by hard-coding the path to the clang shared object in the derivation. Finding a way to reference this now.

doubledup:
@alistair-singh Give it a go now. I've set `LIBCLANG_PATH` so that rocksdb should be able to find the clang shared object. (it can on my remote machine)

alistair-singh:
Tested successfully on Ubuntu 22.04 using flake.
* Testnet launches.
* Block production on eth and substrate sides.
* No errors in relayer logs.
* Header imports from beacon relayer.

core/packages/test/scripts/deploy-ethereum.sh
```diff
@@ -38,18 +38,15 @@ start_lodestar() {
...
-            timestamp=$(date -d'+10second' +%s)
-        fi
+        timestamp=$(date -d'+10second' +%s)
 
         npx lodestar dev \
             --genesisValidators 8 \
             --genesisTime $timestamp \
             --startValidators "0..7" \
-            --enr.ip "127.0.0.1" \
+            --enr.ip6 "::1" \
```
vgeddes, line 47: What's the reasoning behind making both geth and lodestar bind to ipv6 interfaces? Usually ipv4 is the best default, since its more widely supported. For example, not sure if this will work in Docker or Github Actions runners, etc

I vaguely recall geth preferentially binding to ipv6 interfaces due to config in `/etc/hosts`. If that's the case though, maybe we need to sort that out instead.

flake.nix
```diff
@@ -0,0 +1,82 @@
...
+                    #
+                    export HOME=~
+
+                    eval "$(direnv hook bash)"
+
+                    # rocksdb requires a clang.so
+                    export LIBCLANG_PATH="$(readlink -f ${pkgs.clang}/resource-root/include | xargs dirname | xargs dirname | xargs dirname)"
+
+                    pushd parachain
+                    cargo install --git https://github.com/paritytech/polkadot --tag v0.9.30 polkadot --locked --root .cargo
```
vgeddes, line 72: Won't this time-consuming process cause the shell to hang every time its activated?

vgeddes: Also its fine for the installation of polkadot to remain the responsibility of start-services.

Nice trick with the `cargo install` approach though. Much simpler than cloning the polkadot repo, so am keen to retain that part.

core/packages/test/scripts/deploy-ethereum.sh
doubledup: ipv6 was a shortcut to finding more differences between system configs, though maybe commenting out `/etc/hosts` would be simpler. Happy to switch to ipv4, just want to pick one to reduce confusion.

flake.nix
doubledup: If the binary exists, `cargo install` skips it. This only happens on the first `nix develop` invocation, and whenever the binary is removed, eg. accidentally running `git clean` without filtering it out. You can also copy another version of a binary into `parachain/.cargo/bin/polkadot` if you need to test other versions.

Happy to move this into `start-services`, it is a bit annoying when the binary doesn't exist.

core/packages/test/scripts/set-env.sh
```diff
@@ -1,10 +1,12 @@
...
+# exports enable use in core/packages/test/config/launch-config.toml
 root_dir="$(realpath ../../..)"
-relaychain_dir="${POLKADOT_DIR:-$root_dir/relaychain}"
-relaychain_version="${POLKADOT_VER:-v0.9.30}"
-relaychain_bin="${POLKADOT_BIN:-$relaychain_dir/target/release/polkadot}"
 parachain_dir="$root_dir/parachain"
 parachain_runtime="${PARACHAIN_RUNTIME:-snowbase}"
 parachain_bin="$parachain_dir/target/release/snowbridge"
+relaychain_version="${POLKADOT_VER:-v0.9.30}"
+relaychain_bin="${POLKADOT_BIN:-$parachain_dir/.cargo/bin/polkadot.d/$relaychain_version/polkadot}"
```
doubledup, line 9: Including `$relaychain_version` here allows us to swap out polkadot versions by changing `POLKADOT_VER`. The way `relaychain_bin` gets checked, we won't rebuild the same polkadot version.

core/packages/test/scripts/deploy-ethereum.sh
yrong: @doubledup so what is the error thrown when not explicted add `127.0.0.1` in the setup?  I met the [issue](https://github.com/Snowfork/snowbridge/pull/771#issuecomment-1417907080) in one of my physical machine 
and resolved after Clara updates `lodestar` to `1.4.2`, not sure if will also resolve the issue here.

doubledup: Hardhat and lodestar had trouble talking to geth, I can't remember the exact error.

I believe `localhost` gets mapped to an address in `/etc/hosts`, but we don't want to assume anything about system-level config like `/etc/hosts`. eg. using ipv6 & localhost had issues because on our Linux boxes localhost is only mapped to `127.0.0.1`, while the Macs also map it to `::1`.

flake.nix
vgeddes: Ah ok, cool. Well if cargo install skips it then it's fine to put the install in the shellhook

doubledup:
After merging main, we need to merge #774 before we can do the last round of testing.

core/packages/test/config/launch-config.toml
```diff
@@ -43,23 +48,23 @@ chain_spec_path = "{{output_dir}}/spec.json"
...
+    validator = true
+    ws_port = 11144
+    args = ["--enable-offchain-indexing=true","--pruning=archive"]
 
-[[parachains.collators.env]]
-name = "RUST_LOG"
-value = "runtime=debug,parachain=trace,cumulus-collator=trace,aura=trace,xcm=trace"
+        [[parachains.collators.env]]
+        name = "RUST_LOG"
+        value = "runtime=debug,parachain=trace,cumulus-collator=trace,aura=trace,xcm=trace"
```
yrong, line 61: better to add `ethereum-beacon-client` to log

shell.nix
```diff
@@ -0,0 +1,67 @@
+{ system ? builtins.currentSystem }:
```
vgeddes, line 1: This file is obsolete and can be deleted right?

core/packages/test/config/launch-config.toml
doubledup: I've gone with trace logs here, since we have some trace logs in this pallet and no debug logs

core/packages/test/README.md
```diff
@@ -4,24 +4,19 @@ The E2E tests run against local deployments of the parachain, relayer, the ether
...
+## Setup
 
-* zombienet - Just download binary from [latest release](https://github.com/paritytech/zombienet/releases/), rename to `zombienet` and copy to $PATH
+### Activate nix developer shell
 
+In the root directory of our repository, where `shell.nix` is located, activate a developer shell:
 
-## Setup
+```
+nix-shell
```
vgeddes, line 18: These steps are a bit obsolete too.

doubledup:
@vgeddes @yrong @claravanstaden This is ready for general testing. Follow the main README and let me know if you run into any problems.

I've got an ongoing issue with `@chainsafe/blst` not finding a dylib on both Mac & Linux, but Alistair hasn't seen this on Linux. There's another outstanding issue on x86 Mac, but we're aiming for arm Mac & x86 Linux for now.

yrong:
@doubledup Awesome! Thanks for the work and will also test on my mac setup.

vgeddes (commented):
Works for me on my M1 Macbook. 👍

alistair-singh (approved):
Works on Ubuntu 22.04! LGTM!

doubledup:
Found the problem with my envs: needed to clear `$PNPM_HOME` (set to `~/Library/pnpm` on my machine). All fine after that 👌

PR #762, opened 2023-01-18
Remove incentivized channel and apps

doubledup:
This builds & runs locally. The logs in `core/packages/test/*.log` show that the parachain & relaychain are producing blocks (alice, bob, charlie, 11144, etc.). The logs for the beacon, execution, beefy & parachain relayers all show that they're running, though nothing's being relayed.

Not sure how to check geth & the contracts, though I've reached the `Testnet has been initialized` message locally. From the `start-services` logs, the contracts have been built & deployed successfully.

I haven't been able to connect polkadot.js to the [relaychain or parachain mentioned in the README](https://github.com/Snowfork/snowbridge/blob/main/core/packages/test/README.md?plain=1#L56-L59), but the logs at least show blocks being produced. This is probably a local environment issue.


doubledup:
Is it worth removing the "basic" from "basic channel", while we're at it?

parachain/primitives/core/src/types.rs
```diff
@@ -10,17 +9,10 @@ use sp_std::vec::Vec;
 #[derive(Encode, Decode, Copy, Clone, PartialEq, Eq, RuntimeDebug, TypeInfo)]
 pub enum MessageId {
```
alistair-singh, line 10: This is non-blocking but hould we collapse this enum into a struct?

alistair-singh (approved):

parachain/primitives/core/src/types.rs
doubledup: Makes sense. It's easy enough, I'll add it here.

PR #763, opened 2023-01-18
docs: correct parachain destination

evilrobot-01:


alistair-singh:
Thanks for the contribution, however, we are currently removing the incentivized channel so related documentation will also be removed.

Please see this related PR:
https://github.com/Snowfork/snowbridge/pull/762

PR #764, opened 2023-01-19
Set node & pnpm versions in package.json

doubledup:
Also bumps the lockfile, as that's happening with this version of `pnpm` anyway.


claravanstaden (approved):
Thanks! 👌

PR #765, opened 2023-01-23
Ancestry proofs

claravanstaden:
Adds ancestry proofs to the beacon client.

[ ] Downloads the beacon state using lodestar
[ ] Send block roots, block root hash and block roots proof to beacon client
[ ] Verify block root proof
[ ] Store block roots
[ ] Check block proofs in execution header imports
[ ] Figure out how to unmarshal minimal config Beacon state


parachain/pallets/ethereum-beacon-client/src/lib.rs
```diff
@@ -466,11 +475,58 @@ pub mod pallet {
...
 
+		fn process_and_store_block_roots(
+			block_roots: BoundedVec<H256, <T>::MaxSlotsPerHistoricalRoot>,
+			slot: u64,
+			finalized_header_block_root: H256,
+		) {
+			let max_slots = T::MaxSlotsPerHistoricalRoot::get() as u64;
+			let start = slot - max_slots;
+			let end = slot;
+			for n in start..end {
```
vgeddes, line 521: Performing multiple storage writes in a loop may be problematic from a performance/weight perspective. How many loop iterations are you expecting?

claravanstaden: Max `8192` at worst case. 😬 Is there a better way to do a bulk insert? I'll check anyway. This is a the first basic implementation of ancestry proofs, where I send the entire `block_roots` array to the beacon client because there isn't a way to get a proof for just the single `block_root` you want to prove. I figured it would be best to implement a stop-gap for the vulnerability and then see if I can get the single block root proof generation to work using a Prysm work.

To summarize:
This PR: sends all the `block_roots`, does the proof and then stores the expected block root per slot in storage.
Final solution: only send the blocks proofs I'm interested in, with a multi-proof.

claravanstaden: OK cool! I'll use most of what I've done so far, I'll just work on the multiproof bit then. There are two alternatives here, will explain on Monday.

yrong: is `MMR` also an option here to verify the ancestry proof? something like what we do for Beefy in relaychain?

relayer/relays/beacon/header/syncer/syncer.go
```diff
@@ -163,12 +188,104 @@ func (s *Syncer) GetSyncCommitteePeriodUpdate(from uint64) (SyncCommitteePeriodU
...
+	state, err := cf.UnmarshalBeaconState(data)
+	if err != nil {
+		return types.H256{}, nil, fmt.Errorf("unmarshal beacon state using prysm: %w", err)
+	}
+
+	currentBlockHash, err := state.LatestBlockHeader().HashTreeRoot()
+
+	log.WithField("currentFinalizedHeader", s.currentFinalizedHeader).Info("setting current finalized header")
+
+	blockRootsHash, err := stateutil.ArraysRoot(state.BlockRoots(), uint64(len(state.BlockRoots())))
```
yrong, line 245: @claravanstaden  notice that we did some [customization](https://github.com/Snowfork/prysm/commit/4aec9d4e72a1f946dcd7088ce8018328f47518b1) in `prysm` to add proof for blockRoots which we will use later for the ancestry execution block verification, right? 
  But the original code I checked [here](https://github.com/Snowfork/prysm/blob/4aec9d4e72a1f946dcd7088ce8018328f47518b1/beacon-chain/state/state-native/hasher.go#L72) to calculate the leaf is based on `SLOTS_PER_HISTORICAL_ROOT` while we use the current length here.
  so is that correct or I misunderstand something?

claravanstaden: Hey @yrong! Nice catch! You are right, it should be `SLOTS_PER_HISTORICAL_ROOT`. However, I am going to likely going remove the Prysm dependency and rather use https://github.com/ferranbt/fastssz. I am busy fixing this issue at the moment: https://github.com/ferranbt/fastssz/issues/119.

relayer/relays/beacon/header/syncer/syncer.go
```diff
@@ -246,13 +411,73 @@ func (s *Syncer) GetHeaderUpdate(blockRoot common.Hash) (HeaderUpdate, error) {
...
+	tree, err := blockRootsContainer.GetTree() // TODO check memory cost to save this in memory instead of block root array
+	if err != nil {
+		return nil, fmt.Errorf("convert block roots to tree: %w", err)
+	}
+
+	log.WithField("tree root", common.BytesToHash(tree.Hash())).Info("tree root")
+
+	treeDepth := math.Floor(math.Log2(float64(s.MaxSlotsPerHistoricalRoot)))
+	leavesStartIndex := int(math.Pow(2, treeDepth))
+	leafIndex := leavesStartIndex + indexInArray
```
yrong, line 446: just curious why we add an extra `leavesStartIndex` for the leafIndex?

parachain/pallets/ethereum-beacon-client/src/lib.rs
```diff
@@ -497,7 +523,30 @@ pub mod pallet {
...
+					.into();
+
+			let finalized_block_root_hash =
+				<FinalizedBeaconHeadersBlockRoot<T>>::get(last_finalized_header.beacon_block_root);
+
+			let max_slots_per_historical_root = T::MaxSlotsPerHistoricalRoot::get();
+			let depth = max_slots_per_historical_root.ilog2();
+			let index_in_array = block_slot % max_slots_per_historical_root as u64;
+			let leaves_start_index = 2u32.pow(depth as u32) as u64;
+			let leaf_index: u64 = (leaves_start_index + index_in_array) as u64;
```
yrong, line 535: same as https://github.com/Snowfork/snowbridge/pull/765/files#r1105271929

and also not sure we can use exp/log functions in `no_std` runtime

parachain/pallets/ethereum-beacon-client/src/lib.rs
```diff
@@ -497,7 +523,30 @@ pub mod pallet {
 			let beacon_block_root: H256 =
 				merkleization::hash_tree_root_beacon_header(header.clone())
 					.map_err(|_| Error::<T>::HeaderHashTreeRootFailed)?
-					.into(); // TODO check denylist headers
+					.into();
+
+			let finalized_block_root_hash =
```
yrong, line 528: IMHO `finalized_block_root_hash` a little confusing here, would it better to rename as `block_roots_hash` or  some other to repensent not the root of the beacon block but the root of the blockRoots in the beacon state?

parachain/pallets/ethereum-beacon-client/src/lib.rs
```diff
@@ -497,7 +523,30 @@ pub mod pallet {
...
+				<FinalizedBeaconHeadersBlockRoot<T>>::get(last_finalized_header.beacon_block_root);
```
yrong, line 529: IMHO `finalized_block_root_hash` a little confusing here, would it better to rename as `block_roots_root_hash` or  some other to repensent not the root of the beacon block but the root of the blockRoots in the beacon state?

yrong: Also seems we change to always use the `last_finalized_header` for the verification, does that mean the original `FinalizedBeaconHeaders` storage can be safely removed now?

since it is related to the impl of https://linear.app/snowfork/issue/SNO-328 so I may pause that one until we finalize this PR

relayer/relays/beacon/header/syncer/syncer.go
claravanstaden: The tree root leaves are the block_root values. For example, using this diagram: https://github.com/protolambda/eth2-docs#ssz-hash-tree-root-and-merkleization

`block_root[0]` would be located at tree index `8` in the example tree. So if you want to prove `block_root` at index `2` (`block_root[2]`), you would have to calculate `8` + `2` = tree index `10`. Hope this makes sense. If you have a better way to calculate it I'd be happy to change it.

parachain/pallets/ethereum-beacon-client/src/lib.rs
claravanstaden: Same answer as here: https://github.com/Snowfork/snowbridge/pull/765#discussion_r1105337368

I'll check the no-std usage, just wanted to get it working for now. 😄

parachain/pallets/ethereum-beacon-client/src/lib.rs
claravanstaden: > Also seems we change to always use the last_finalized_header for the verification, does that mean the original FinalizedBeaconHeaders storage can be safely removed now?

I might still change this, because if we need to import older headers we may have to use an older finalized header for the block proof. Working on this at the moment.

relayer/relays/beacon/header/syncer/syncer.go
yrong: I see, the base index of `block_root` in original state tree, right?

I would assume `leavesStartIndex` is a const value for mainet spec or mimal spec. so maybe we can precaculate as a const in runtime without the exp/log functions?

claravanstaden:
The only thing still outstanding is to update the test and benchmark data fixtures. I would like to do so after merging https://github.com/Snowfork/snowbridge/pull/773.

core/package.json
```diff
@@ -7,7 +7,7 @@
   },
   "scripts": {
     "preinstall": "npx only-allow pnpm",
-    "postinstall": "cd .. && husky install && cd core/packages/test && ./scripts/hack-ethereum.sh",
+    "postinstall": "cd .. && husky install && cd core/packages/test",
```
claravanstaden, line 10: I am removing the `hack-ethereum.sh` script, because deviating from the official [Ethereum minimal spec](https://github.com/ethereum/consensus-specs/blob/ffacbb442b8d4469e857a79dee6932fb2420cf89/configs/minimal.yaml) gave me some issues with the ancestry proofs and I believe in this case to stick to the standard (at the cost of slightly slower tests), make sense.

parachain/pallets/ethereum-beacon-client/src/benchmarking/data_mainnet.rs
```diff
@@ -1,3 +1,5 @@
+// Generated, do not edit!
+// See README.md for instructions to generate
```
claravanstaden, line 2: These files are now generated by go command `generate-beacon-data`.

parachain/pallets/ethereum-beacon-client/src/benchmarking/mod.rs
```diff
@@ -86,11 +85,23 @@ benchmarks! {
...
 			), initial_sync_data.current_sync_committee);
 
+		let finalized_update: FinalizedHeaderUpdate<T::MaxSignatureSize, T::MaxProofBranchSize, T::MaxSyncCommitteeSize> = finalized_header_update();
+
+		let finalized_slot = finalized_update.finalized_header.slot;
+		let finalized_block_root: H256 =
+			merkleization::hash_tree_root_beacon_header(finalized_update.finalized_header)
+				.unwrap()
+				.into();
+
```
claravanstaden, line 95: The execution header import is now dependent on a child finalized header after it for the ancestry proof.

parachain/pallets/ethereum-beacon-client/src/tests_mainnet.rs
```diff
@@ -0,0 +1,222 @@
+#[cfg(not(feature = "minimal"))]
```
claravanstaden, line 1: I moved the `minimal` and `mainnet` specific tests to their own test files.

parachain/pallets/ethereum-beacon-client/tests/fixtures/goerli_block_update.json
```diff
@@ -1,2224 +0,0 @@
-{
```
claravanstaden, line 1: I renamed these test files to `mainnet_` instead of `goerli_`, to rather use the config spec than the actual testnet name.

parachain/pallets/ethereum-beacon-client/tests/fixtures/mainnet_finalized_header_update.json
```diff
@@ -0,0 +1,37 @@
+{
```
claravanstaden, line 1: These files are now generated with the `generate-beacon-data` Go command.

relayer/relays/beacon/header/syncer/json/beacon_json.go
```diff
@@ -0,0 +1,315 @@
+package json
```
claravanstaden, line 1: This package contains all the beacon structs necessary to generate the test data and output to the test data fixture JSON files, as well as the benchmark data file.

relayer/relays/beacon/header/syncer/syncer.go
claravanstaden: That's a good point. I added the config in the parachain but not in the relayer yet, will do so.

claravanstaden (commented):
@vgeddes @alistair-singh @doubledup @yrong This is a big one, apologies. 😬 I've tried adding some comments to explain some things, hope it helps. I will explain how the ancestry proofs are now implemented in the Notion doc. I'll post the link with the updated details here in a bit.

core/package.json
vgeddes: I agree, it complicates testing.

parachain/README.md
```diff
@@ -151,3 +151,50 @@ target/release/snowbridge benchmark pallet \
...
+
+Wait for output `Testnet has been initialized`.
+
+In a separate terminal, from the `snowbridge` directory, run:
+
+```bash
+mage -d relayer build && relayer/build/snowbridge-relay generate-beacon-data --spec "minimal" && cd parachain && cargo +nightly fmt -- --config-path rustfmt.toml && cd -
+```
+
+### Mainnet Spec
```
vgeddes, line 174: nice! very comprehensive docs.

parachain/pallets/ethereum-beacon-client/src/lib.rs
```diff
@@ -595,6 +658,71 @@ pub mod pallet {
...
+
+			if finalized_block_root_hash.is_zero() {
+				log::error!(
+					target: "ethereum-beacon-client",
+					"💫 Finalized block root {} slot {} for ancestry proof not found.", beacon_block_root, block_slot
+				);
+				return Err(Error::<T>::ExpectedFinalizedHeaderNotStored.into())
+			}
+
+			let max_slots_per_historical_root = T::MaxSlotsPerHistoricalRoot::get() as u64;
```
vgeddes, line 703: If we are type casting like this, shouldn't T::MaxSlotsPerHistoricalRoot be `u64` in the first place?

parachain/pallets/ethereum-beacon-client/src/lib.rs
```diff
@@ -781,6 +910,20 @@ pub mod pallet {
 			});
 
 			Self::deposit_event(Event::BeaconHeaderImported { block_hash: block_root, slot });
+
+			Ok(())
+		}
+
+		fn add_finalized_header_slot(slot: u64) -> DispatchResult {
+			<FinalizedBeaconHeaderSlots<T>>::try_mutate(|b_vec| {
+				if b_vec.len() as u32 == T::MaxFinalizedHeaderSlotArray::get() {
```
vgeddes, line 919: perhaps `T::MaxFinalizedHeaderSlotArray` should be u64 too? If in doubt.

I think anywhere we are casting from u32 to u64, perhaps the original value should be u64 to being with.

core/packages/test/scripts/helpers/mutateSpec.js
```diff
@@ -20,6 +24,7 @@ function run() {
     let initialSync = JSON.parse(fs.readFileSync(process.argv[3]));
 
     data['genesis']['runtime']['ethereumBeaconClient']['initialSync'] = initialSync;
+    data['genesis']['runtime']['ethereumBeaconClient']['initialSync']['import_time'] = getTimestampInSeconds()
```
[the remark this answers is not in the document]
claravanstaden: That's a good point. Although if we take longer than 27 hours to deploy and future date the import time, it opens up the risk of a long range attack because the initial sync checkpoint header is older than 27 hours. But maybe that's OK, even if it blocks the bridge on deployment, and then we provide another checkpoint through governance. Thanks for picking this up, I'll remember to take this into account with the new bridge hub architecture. 🤞

parachain/pallets/ethereum-beacon-client/src/lib.rs
```diff
@@ -781,6 +910,20 @@ pub mod pallet {
...
+					b_vec.remove(0);
+				}
+				b_vec.try_push(slot)
+			})
+			.map_err(|_| <Error<T>>::FinalizedBeaconHeaderSlotsExceeded)?;
+
```
yrong, line 925: seems we'are doing some prune work for [FinalizedBeaconHeaderSlots](https://github.com/Snowfork/snowbridge/blob/36c99689aab88a82529931212d72a2157dd8954b/parachain/pallets/ethereum-beacon-client/src/lib.rs#L183), is it also nessessary to prune the related [FinalizedBeaconHeadersBlockRoot](https://github.com/Snowfork/snowbridge/blob/36c99689aab88a82529931212d72a2157dd8954b/parachain/pallets/ethereum-beacon-client/src/lib.rs#L187) storage here?

parachain/runtime/snowbridge/src/lib.rs
```diff
@@ -585,6 +585,8 @@ parameter_types! {
 	pub const MaxVoluntaryExitSize: u32 = 16;
 	pub const MaxAttestationSize: u32 = 128;
 	pub const MaxValidatorsPerCommittee: u32 = 2048;
+	pub const MaxSlotsPerHistoricalRoot: u64 = 8192;
+	pub const MaxFinalizedHeaderSlotArray: u32 = 1000;
```
yrong, line 589: Just curios how to estimate the threshhold here? 
seems we still need to cache enough finalized checkpoints make sure we can always find the merkle root of the block roots [here](https://github.com/Snowfork/snowbridge/blob/36c99689aab88a82529931212d72a2157dd8954b/parachain/pallets/ethereum-beacon-client/src/lib.rs#L692-L701), right?

parachain/pallets/ethereum-beacon-client/src/lib.rs
claravanstaden: The main reason why I added pruning for `FinalizedBeaconHeaderSlots` is because the beacon relayer queries the whole array, so it needs to be kept in a queryable size. I think we can prune `FinalizedBeaconHeadersBlockRoot` as part of the other pruning ticket: https://linear.app/snowfork/issue/SNO-328

parachain/runtime/snowbridge/src/lib.rs
claravanstaden: For now, it's just an arbitrary number which can still be finetuned. That array is used for the beacon relayer, so it knows which finalized headers are already available on the parachain so we can send a block root proof for a much older block.

parachain/pallets/ethereum-beacon-client/src/lib.rs
claravanstaden: I've added these as parachain config: https://github.com/Snowfork/snowbridge/pull/765/files#diff-20b75e7903673f591cfb5dd3c153d51797bd4069c9b71dfb6b11e06f2c8c675eR7

relayer/relays/beacon/header/syncer/syncer.go
claravanstaden: I've cleaned this up: https://github.com/Snowfork/snowbridge/pull/765/files#diff-c589b232c311e67a8690c53fe1886c463a185178fde8ed4aa9dd2def499603b3R497

relayer/relays/beacon/header/syncer/syncer.go
claravanstaden: I've removed this since using Prysm is no longer necessary.

relayer/relays/beacon/cache/cache.go
```diff
@@ -35,6 +68,89 @@ func (b *BeaconCache) SetLastSyncedSyncCommitteePeriod(period uint64) {
...
+	for _, slot := range slots {
+		b.addSlot(slot)
+	}
+}
+
+func (b *BeaconCache) GetClosestCheckpoint(slot uint64) (Proof, error) {
+	checkpointSlot, err := b.calculateClosestCheckpointSlot(slot)
+	if err != nil {
+		return Proof{}, err
+	}
```
yrong, line 92: So what we will do to handle here? is that the long range attack you mentioned before and we need to manuallay or from governance call to add some ancestry checkpoints into beacon light client?

yrong, line 92: So what we will do to handle the error here? @claravanstaden is that the long range attack you mentioned before and we need to manually add some ancestry checkpoints into beacon light client?

parachain/runtime/snowbridge/src/lib.rs
yrong: That is ok. Just a little confusing about threshold in relayer with some very different value
https://github.com/Snowfork/snowbridge/blob/36c99689aab88a82529931212d72a2157dd8954b/relayer/relays/beacon/cache/cache.go#L13 

Any way, I understand these threshhold values are only for pruning the old checkpoint data which is not the main scope in this PR and we can handle it in https://linear.app/snowfork/issue/SNO-328

claravanstaden: Ah I see! That is a different kind of value. Let me explain the difference.

`MaxFinalizedHeaderSlotArray` is the max number of items in `FinalizedBeaconHeaderSlots`. `FinalizedBeaconHeaderSlots` holds the slots of the finalized headers we have verified and imported into the light client. The array is used for the relayer to know which finalized headers the parachain has already imported and can be used for `block_root` proofs. For instance, if the relayer goes down and starts up, the relayer won't know which finalized headers are available for proofs, so the relayer queries the parachain to get this information.

In the relayer, `FinalizedCheckpointsLimit` is the number of finalized header's block root proof tree should be kept in memory. I need to double check if I do this, but this limit can be smaller since if we clear the cache, we can download the beacon state to get this value again. The limit needs to be smaller because it is a larger structure to keep in memory, while `FinalizedBeaconHeaderSlots` is just an array of ints.

relayer/relays/beacon/cache/cache.go
claravanstaden: Exactly, I will handle that in https://linear.app/snowfork/issue/SNO-405, since if we reach this condition where there isn't a checkpoint for the sync committee period, we would have reached the condition where the bridge would be blocked.

parachain/runtime/snowbridge/src/lib.rs
yrong: Thanks for the detailed explaination. Yes, that make sense.

claravanstaden:
Updated docs including a explanation of how the ancestry proofs work: https://www.notion.so/snowfork/Beacon-Client-074ca039288247bba2f96bd6cb8674bf#13645ee8badc4cf183547cb51151d15a

relayer/relays/beacon/cache/cache.go
```diff
@@ -35,6 +68,89 @@ func (b *BeaconCache) SetLastSyncedSyncCommitteePeriod(period uint64) {
...
+		}
+
+		if i == slot { // if the slot is at a finalized checkpoint, we don't need to do the ancestry proof
+			return i, nil
+		}
+
+		checkpointSlot := int(i) // convert to int since it can be negative
+		if checkpointSlot-blockRootThreshold < int(slot) {
+			return i, nil
+		}
```
yrong, line 152: is it better replace to `if i < slot+blockRootThreshold` to avoid the int cast?

yrong: then maybe we can also add another check in light client following  https://github.com/Snowfork/snowbridge/blob/36c99689aab88a82529931212d72a2157dd8954b/parachain/pallets/ethereum-beacon-client/src/lib.rs#L575

```
ensure!(
				block_slot + T::MaxSlotsPerHistoricalRoot::get() >= latest_finalized_header_slot,
				Error::<T>::ExpectedFinalizedHeaderNotStored
			);
```

yrong (approved):
Awesome!

alistair-singh (approved):

PR #766, opened 2023-01-23
Basic channel v3: blockchain updates

doubledup:
Parachain and contracts both build. More context in [Remove message bundles](https://www.notion.so/snowfork/Remove-message-bundles-d46d2f002d444979a39ce9a208ce23cc) and [Basic Channel V3](https://www.notion.so/snowfork/Basic-Channel-V3-0e46926b19d644f0be4f0d9be18d18f1).


core/packages/contracts/contracts/BasicInboundChannel.sol
```diff
@@ -5,61 +5,63 @@ import "./ParachainClient.sol";
...
+        require(message.nonce == nonce[message.sourceId] + 1, "Invalid nonce");
+        // TODO: should we check the remaining gas?
+        nonce[message.sourceId]++;
+        dispatch(message);
+        // TODO: should we emit a MessageDispatched event?
+    }
 
-        require(parachainClient.verifyCommitment(commitment, proof), "Invalid proof");
-        require(bundle.nonce == nonce[bundle.account] + 1, "Invalid nonce");
+    function submitBatch(
```
vgeddes, line 40: Hey, I'd rather implement `submitBatch` in a future PR. I just described it in Notion to show that using multiproofs was possible

parachain/pallets/basic-channel/src/outbound/mod.rs
```diff
@@ -30,79 +30,33 @@ pub use weights::WeightInfo;
...
-	Encode, Decode, CloneNoBound, PartialEqNoBound, RuntimeDebugNoBound, MaxEncodedLen, TypeInfo,
-)]
-#[scale_info(skip_type_params(M))]
-#[codec(mel_bound())]
-pub struct Message<M: Get<u32>> {
-	/// Unique message ID
+	/// ID of source parachain
+	source_id: SourceId,
+	/// Unique nonce to prevent replaying messages
 	#[codec(compact)]
```
vgeddes, line 41: The nonce shouldnt be compact encoded, as that has as that it is quite expensive to parse on the Solidity side.

doubledup: Agreed, though I don't think we send any SCALE-encoded values to the Solidity side. We use SCALE for the event emitted, but everything related to the proof is ethabi-encoded. We also didn't have any special parsing for it on the Solidity side before, just `bundle.nonce` from calldata

vgeddes: oh right, my mistake, that's very true

parachain/pallets/basic-channel/src/outbound/mod.rs
```diff
@@ -121,6 +75,10 @@ pub mod pallet {
 
 		type Hashing: Hash<Output = H256>;
 
+		/// ID of message source
+		/// + TypeInfo,
+		type SourceId: Parameter + Member + MaxEncodedLen;
```
vgeddes, line 80: Should also be bounded by `PartialEq`. I know that's already defined by Member, but it is good to repeat it here for documentation purposes

parachain/pallets/basic-channel/src/outbound/mod.rs
```diff
@@ -203,12 +158,12 @@ pub mod pallet {
...
-	where
-		T::AccountId: AsRef<[u8]>,
-	{
+	impl<T: Config> Pallet<T> {
 		/// Submit message on the outbound channel
-		pub fn submit(who: &T::AccountId, target: H160, payload: &[u8]) -> DispatchResult {
+		// TODO: Remove target once incentivized channel & app removal are done
+		pub fn submit(who: &T::SourceId, _target: H160, payload: &[u8]) -> DispatchResult {
+			// TODO: Do we need to convert the error values passed to ensure! to
+			// sp_runtime::DispatchError? Does the ensure! macro hide a possible type error here?
```
vgeddes, line 166: Pallet errors such as `Error::<T>::QueueSizeLimitReached` implement the necessary traits to convert to DispatchError. So not needed.

vgeddes: The `ensure!` macro would not compile if the passed error was not convertible to DispatchError

parachain/pallets/basic-channel/src/outbound/mod.rs
```diff
@@ -219,34 +171,27 @@ pub mod pallet {
 				Error::<T>::PayloadTooLarge,
 			);
 
-			let next_id = <NextId<T>>::get();
-			if next_id.checked_add(1).is_none() {
-				return Err(Error::<T>::Overflow.into())
-			}
+			let nonce = <Nonce<T>>::get(who);
+			let next_nonce = nonce.saturating_add(1);
+			ensure!(next_nonce != nonce, Error::<T>::Overflow);
```
yrong, line 176: seems the `ensure` logic is not correct here, `next_nonce` will be zero when overflow, use this instead
`nonce.checked_add(1).ok_or(Error::<T>::Overflow)?`

parachain/pallets/basic-channel/src/outbound/mod.rs
```diff
@@ -219,34 +171,27 @@ pub mod pallet {
...
+				source_id: who.clone(),
+				nonce,
+				payload: payload.to_vec().try_into().map_err(|_| Error::<T>::PayloadTooLarge)?,
 			})
 			.map_err(|_| Error::<T>::QueueSizeLimitReached)?;
-			Self::deposit_event(Event::MessageAccepted(next_id));
+			Self::deposit_event(Event::MessageAccepted(nonce));
 
-			<NextId<T>>::put(next_id + 1);
+			<Nonce<T>>::set(who, next_nonce);
```
yrong, line 186: `who` seems a little ambiguous, better to rename `source_id' ?

parachain/pallets/basic-channel/src/outbound/test.rs
```diff
@@ -69,6 +69,7 @@ parameter_types! {
 }
 
 impl basic_outbound_channel::Config for Test {
+	type SourceId = AccountId;
```
yrong, line 72: Is `AccountId` temporary  used for compatible with legacy codes and finally change to `ParaId` ? 
If so is it nessessary to also add `assetId` into `Message` structure to support MultiAsset from same parachain?

parachain/pallets/basic-channel/src/outbound/mod.rs
doubledup: Saturating_add doesn't give 0 on an overflow:
```
let x = 255u8.saturating_add(1);
println!("{:?}", x); // prints "255"
```

That said, we don't actually have a use for the saturating behaviour. We're returning an error anyway, so `checked_add` is a better choice.

parachain/pallets/basic-channel/src/outbound/test.rs
doubledup: Yes, using `AccountId` here purely to keep the tests passing. If we need to use another type for `SourceId`, we can update the runtimes.

I believe `assetId` will be part of the `Message.payload` field. The payload will be an encoded set of XCM messages that's opaque to the channels. Then if we want to send messages that don't need an `assetId`, we change the XCM messages and not the channel's `Message`.

vgeddes (approved):

PR #769, opened 2023-01-26
Basic channel v3: relayer updates

doubledup:
Relayer builds and the beefy & parachain relayer logs look good on startup.


vgeddes (approved):
+1

PR #770, opened 2023-01-27
Basic channel v3

doubledup:
Builds and runs locally, though there are no E2E tests yet.


core/packages/contracts/contracts/BasicInboundChannel.sol
```diff
@@ -5,61 +5,45 @@ import "./ParachainClient.sol";
...
+        require(message.nonce == nonce[message.sourceId] + 1, "Invalid nonce");
+        // TODO: should we check the remaining gas?
+        nonce[message.sourceId]++;
+        dispatch(message);
+        // TODO: should we emit a MessageDispatched event?
+    }
+
+    // solhint-disable no-empty-blocks
+    function dispatch(Message calldata messages) internal {
+        // TODO: dispatch to XCM interpreter
```
vgeddes, line 42: lets emit an event like `MessageDispatched`. Useful for writing unit tests.

core/packages/contracts/contracts/BasicInboundChannel.sol
```diff
@@ -5,61 +5,45 @@ import "./ParachainClient.sol";
...
-        dispatch(bundle);
+        require(message.nonce == nonce[message.sourceId] + 1, "Invalid nonce");
+        // TODO: should we check the remaining gas?
+        nonce[message.sourceId]++;
+        dispatch(message);
+        // TODO: should we emit a MessageDispatched event?
+    }
+
+    // solhint-disable no-empty-blocks
+    function dispatch(Message calldata messages) internal {
```
vgeddes, line 41:
```suggestion
    function dispatch(Message calldata message) internal {
```

core/packages/contracts/contracts/BasicInboundChannel.sol
```diff
@@ -5,61 +5,45 @@ import "./ParachainClient.sol";
...
         require(
-            gasleft() >= (bundle.messages.length * MAX_GAS_PER_MESSAGE) + GAS_BUFFER,
-            "insufficient gas"
+            parachainClient.verifyCommitment(commitment, parachainHeaderProof),
+            "Invalid proof"
         );
-        nonce[bundle.account]++;
-        dispatch(bundle);
+        require(message.nonce == nonce[message.sourceId] + 1, "Invalid nonce");
+        // TODO: should we check the remaining gas?
```
vgeddes, line 34: Yeah, we'll want retain some form of gas checking with those `MAX_GAS_PER_MESSAGE` `GAS_BUFFER` constants.

Its important that if the nonce is updated, and the message is dispatched, that the target app has enough gas to execute the message.

core/packages/contracts/contracts/BasicInboundChannel.sol
```diff
@@ -5,61 +5,45 @@ import "./ParachainClient.sol";
...
-    function dispatch(MessageBundle calldata bundle) internal {
-        for (uint256 i = 0; i < bundle.messages.length; i++) {
-            Message calldata message = bundle.messages[i];
-            (bool success, ) = message.target.call{ value: 0, gas: MAX_GAS_PER_MESSAGE }(
-                message.payload
-            );
-            emit MessageDispatched(message.id, success);
-        }
+    // solhint-disable no-empty-blocks
+    function dispatchBatch(Message[] calldata messages) internal {
```
vgeddes, line 46: Not necessary for now, can remove.

core/packages/contracts/contracts/BasicInboundChannel.sol
```diff
@@ -5,61 +5,45 @@ import "./ParachainClient.sol";
...
-    struct MessageBundle {
-        bytes32 account;
-        uint64 nonce;
-        Message[] messages;
-    }
-
     struct Message {
-        uint64 id;
-        address target;
+        bytes32 sourceId;
```
vgeddes, line 13: We use this style in Solidity: https://docs.soliditylang.org/en/v0.8.17/style-guide.html#naming-styles

```suggestion
        bytes32 sourceID;
```

doubledup: Got it. Updated the tests as well for consistency, even though they're in Typescript.

parachain/pallets/basic-channel/src/outbound/mod.rs
```diff
@@ -30,79 +30,33 @@ pub use weights::WeightInfo;
...
-	AccountId: AsRef<[u8]> + Encode + Decode + Clone + PartialEq + Debug + MaxEncodedLen + TypeInfo,
+	SourceId: Decode + Parameter + Member + MaxEncodedLen, //+ TypeInfo,
 {
 	fn into(self) -> Token {
 		Token::Tuple(vec![
-			Token::FixedBytes(self.account.as_ref().into()),
+			Token::Bytes(self.source_id.encode()),
 			Token::Uint(self.nonce.into()),
-			Token::Array(self.messages.into_iter().map(|message| message.into()).collect()),
+			Token::Bytes(self.payload.to_vec()),
```
yrong, line 55: is this `Message->Token` Convert still in use now? did not find referenced somewhere

parachain/pallets/basic-channel/src/outbound/mod.rs
```diff
@@ -228,34 +183,26 @@ pub mod pallet {
...
-					target,
-					payload: payload
-						.to_vec()
-						.try_into()
-						.map_err(|_| Error::<T>::PayloadTooLarge)?,
-				},
+			<MessageQueue<T>>::try_append(Message {
+				source_id: source_id.clone(),
+				nonce,
+				payload: payload.to_vec().try_into().map_err(|_| Error::<T>::PayloadTooLarge)?,
```
yrong, line 192: seems `PayloadTooLarge` is already checked in previous step, so `try_into` not nessessary here?

parachain/pallets/basic-channel/src/outbound/mod.rs
```diff
@@ -228,34 +183,26 @@ pub mod pallet {
...
-				account: who.clone(),
-				message: Message {
-					id: next_id,
-					target,
-					payload: payload
-						.to_vec()
-						.try_into()
-						.map_err(|_| Error::<T>::PayloadTooLarge)?,
-				},
+			<MessageQueue<T>>::try_append(Message {
```
yrong, line 189: seems  `append` directly should be fine since already checked in previous step?

parachain/pallets/basic-channel/src/outbound/mod.rs
```diff
@@ -228,34 +183,26 @@ pub mod pallet {
 				Error::<T>::PayloadTooLarge,
 			);
 
-			let next_id = <NextId<T>>::get();
-			if next_id.checked_add(1).is_none() {
-				return Err(Error::<T>::Overflow.into())
-			}
+			let nonce = <Nonce<T>>::get(source_id);
+			let next_nonce = nonce.checked_add(1).ok_or(Error::<T>::Overflow)?;
```
yrong, line 187: though not possible in near future, just curious how to recover if `nonce` overflow here, is it better not use  `checked_add` and just let it to be zero?

parachain/pallets/basic-channel/src/outbound/mod.rs
doubledup: It's [used with `ethabi::encode`](https://github.com/Snowfork/snowbridge/pull/770/files/a343f4ae848411b053ea2b87f61abb92a19116f4#diff-3100992310c8525bbe27b87da7cc75b5a6a091d313a4fed5001406f24993bed9R225) to serialize the messages. `ethabi::encode` operates on Tokens, so we need to convert the message first.

parachain/pallets/basic-channel/src/outbound/mod.rs
doubledup: Yes, you're right. I'll move this conversion into the `ensure!` above.

parachain/pallets/basic-channel/src/outbound/mod.rs
doubledup: The length validation above isn't encoded in the types, so we have to treat `MessageQueue` as though it's unvalidated. `StorageValue<T>` only has `append` when `T` implements `StorageAppend`, but `BoundedVec` implements `StorageTryAppend` instead.

We could use a phantom type around `BoundedVec` to record the validation in the type, but that won't give us access to `append`. Even if we convert to a mutable `Vec` after validation, that removes the bounds from `BoundedVec`.

We could make our own type like `VecAppendableOnce`, but I'm not sure that's worth it to remove the single `map_err` line below.

Maybe we can remove the `ensure!` above? `try_append` performs the same validation. When the queue is full, we would do some redundant work constructing the `Message` instead of failing early.

parachain/pallets/basic-channel/src/outbound/mod.rs
doubledup: Allowing overflow to nonces exposes us to replay attacks. When the nonce resets to zero, a malicious attacker could resubmit an old message with its valid proofs.

In practice this attack takes a lot of messages over the bridge (2**64 ~= 18.4*10**18), but a known bricking condition we need to upgrade to fix is better than a silently drained bridge.

A lesser concern is that we also lose the ability to sequence all messages for a source, which could make auditing harder. When there are two messages with nonce 0, we can't know which one came first just by looking at the message. We need external information like a timestamp.

doubledup: A permanent fix could be making `Nonce` map to a resizable uint type, instead of `u64`. A wrapper around `&[u64]` might work if there aren't existing libraries.

doubledup: A simpler workaround in the current design is to have sources that approach 2**64 messages switch to a different `sourceId`. Might have knock-on effects though, like the need for a new proxy account on Ethereum. Or others from breaking the assumption of a unique `sourceId` per source.

yrong: yes, totally make sense.  And 2**64 is big enough the concern here may be unnessessary

parachain/pallets/basic-channel/src/outbound/mod.rs
yrong: Gotcha,  not worry it is not a big deal and just keep it is also fine

yrong:
LGTM!

yrong (approved):
LGTM!

PR #774, opened 2023-02-10
Fix e2e setup with foundry

yrong:


yrong:
Since we use foundry for e2e deployment, the previous `configure-beefy.ts` which depends on hardhat does not work any more and breaks the e2e setup,  this PR aims to fix it.

core/packages/contracts/scripts/configure-beefy.ts
```diff
@@ -9,17 +7,13 @@ import type {
     BeefyNextAuthoritySet,
     BeefyId,
 } from "@polkadot/types/interfaces/beefy/types"
+import fs from "fs"
 
 let endpoint = process.env.RELAYCHAIN_ENDPOINT || "ws://localhost:9944"
 const beefyStartBlock = process.env.BEEFY_START_BLOCK ? parseInt(process.env.BEEFY_START_BLOCK) : 15
+const BeefyStateFile = process.env.BEEFY_STATE_FILE || "/tmp/snowbridge/beefy-state.json"
```
doubledup, line 14: Let's use `OUTPUT_DIR` here too, in case we change it:
```suggestion
const BeefyStateFile = process.env.BEEFY_STATE_FILE || path.join(process.env.OUTPUT_DIR, "beefy-state.json")
```

core/packages/contracts/package.json
```diff
@@ -35,7 +35,10 @@
     "@nomiclabs/hardhat-ethers": "^2.0.2",
     "@nomiclabs/hardhat-etherscan": "^2.1.4",
     "@openzeppelin/contracts": "^4.1.0",
-    "@polkadot/api": "^9.6.2",
+    "@polkadot/api": "9.11.1",
+    "@polkadot/types": "9.11.1",
+    "@polkadot/util": "10.2.3",
+    "@polkadot/util-crypto": "10.2.3",
```
doubledup, line 41: Do we need types, util & util-crypto? Not seeing any uses in this PR.

doubledup (approved):
LGTM! It's cleaner with contract config separate from starting the chains.

alistair-singh (approved):
LGTM!

core/packages/contracts/package.json
yrong: yes, I think `types` is required here and I remove the others which is not used https://github.com/Snowfork/snowbridge/blob/b6546280cf2ad67da4cf69702971be50c3013789/core/packages/contracts/scripts/configure-beefy.ts#L5-L9

core/packages/test/scripts/deploy-ethereum.sh
```diff
@@ -33,7 +33,7 @@ start_geth() {
 start_lodestar() {
     if [ "$eth_network" == "localhost" ]; then
         echo "Starting lodestar local node"
-        genesisHash=$(curl http://localhost:8545 \
+        genesisHash=$(curl $infura_endpoint_http \
```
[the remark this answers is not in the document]
yrong: yes, actually the env here applies also for the local setup. So the `infura` prefix seems 
a little confusing here and just renamed to  eth_*** as David suggested

doubledup (approved):

core/README.md
```diff
@@ -20,6 +21,12 @@ corepack enable
 corepack prepare pnpm@7.14.2 --activate
 ```
 
+Install [foundry](https://book.getfoundry.sh/getting-started/installation):
+
+```bash
+curl -L https://foundry.paradigm.xyz | bash
+```
+
```
claravanstaden, line 29: Nice! ✨

core/packages/test/scripts/start-ethereum.sh
```diff
@@ -8,6 +8,7 @@ start_chains()
 {   #1 start ethereum and deploy contracts
     echo "start ethereum and deploy bridge contrancts"
     deploy_ethereum
+    echo "Chain started!"
```
claravanstaden, line 11: Nit: Shouldn't it maybe say:
```suggestion
    echo "Ethereum started"
```

claravanstaden (approved):
Nice! ✨

core/README.md
```diff
@@ -20,6 +21,12 @@ corepack enable
 corepack prepare pnpm@7.14.2 --activate
 ```
 
+Install [foundry](https://book.getfoundry.sh/getting-started/installation):
+
+```bash
+curl -L https://foundry.paradigm.xyz | bash
```
doubledup, line 27: Just remembered: we also need to run `foundryup` to install forge etc:
```suggestion
curl -L https://foundry.paradigm.xyz | bash
foundryup
```

yrong: Nice catch!

vgeddes (approved):
+1

core/README.md
doubledup: Thanks! @claravanstaden's comment in the README jogged my memory 😅

PR #776, opened 2023-02-17
SNO-399: Implement SovereignTreasury contract

doubledup:
Fixes: SNO-399

### DIFF
--- a/core/packages/contracts/contracts/SovereignTreasury.sol
+++ b/core/packages/contracts/contracts/SovereignTreasury.sol
@@ -6,8 +6,6 @@ import "./Vault.sol";
 contract SovereignTreasury {
     Vault public vault;
 
-    // TODO: emit events?
-
     constructor(Vault _vault) {
         vault = _vault;
     }
--- a/core/packages/contracts/contracts/SovereignTreasury.sol
+++ b/core/packages/contracts/contracts/SovereignTreasury.sol
@@ -12,13 +12,13 @@ contract SovereignTreasury is Ownable {
     }
 
     // Handle a message from the bridge.
-    function handle(bytes32 sovereignID, bytes calldata message) external onlyOwner {
+    function handle(bytes32 origin, bytes calldata message) external onlyOwner {
         Message memory message = abi.decode(message, (Message));
 
         if (message.action == Action.Withdraw) {
             WithdrawPayload memory payload = abi.decode(message.payload, (WithdrawPayload));
 
-            transfer(sovereignID, payload.recipient, payload.amount);
+            transfer(origin, payload.recipient, payload.amount);
         }
     }
 
--- a/core/packages/contracts/contracts/SovereignTreasury.sol
+++ b/core/packages/contracts/contracts/SovereignTreasury.sol
@@ -20,11 +20,6 @@ contract SovereignTreasury {
 
             transfer(_sovereignID, payload.recipient, payload.amount);
         }
-
-        // TODO: refund relayer
-        // TODO: reward relayer
-        /* uint256 rewardAmount = 1; */
-        /* transfer(sovereignID, msg.sender, rewardAmount); */
     }
 
     // Deposit ETH into a sovereign account. Permissionless.
--- a/core/packages/contracts/contracts/SovereignTreasury.sol
+++ b/core/packages/contracts/contracts/SovereignTreasury.sol
@@ -20,7 +20,7 @@ contract SovereignTreasury {
             // TODO: how do we know that the recipient address is payable?
             WithdrawPayload memory payload = abi.decode(message.payload, (WithdrawPayload));
 
-            vault.withdraw(_sovereignID, payload.recipient, payload.amount);
+            transfer(_sovereignID, payload.recipient, payload.amount);
         }
 
         // TODO: refund relayer
@@ -32,6 +32,10 @@ contract SovereignTreasury {
     // Deposit ETH into a sovereign account. Permissionless.
     function deposit(bytes32 _sovereignID) external payable {
         vault.deposit{ value: msg.value }(_sovereignID);
+    }
+
+    function transfer(bytes32 sovereignID, address recipient, uint256 amount) private {
+        vault.withdraw(sovereignID, recipient, amount);
     }
 }
 
--- a/core/packages/contracts/contracts/SovereignTreasury.sol
+++ b/core/packages/contracts/contracts/SovereignTreasury.sol
@@ -1,0 +1,64 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity ^0.8.9;
+
+import "./Vault.sol";
+
+contract SovereignTreasury {
+    Vault public vault;
+
+    // TODO: emit events?
+
+    constructor(Vault _vault) {
+        vault = _vault;
+    }
+
+    // Handle a message from the bridge.
+    function handle(bytes32 sovereignID, bytes calldata _message) external {
+        Message memory message = abi.decode(_message, (Message));
+
+        if (message.action == Action.Withdraw) {
+            // TODO: how do we know that the recipient address is payable?
+            WithdrawPayload memory payload = abi.decode(message.payload, (WithdrawPayload));
+
+            transfer(sovereignID, payload.recipient, payload.amount);
+        }
+
+        // TODO: refund relayer
+        // TODO: reward relayer
+        /* uint256 rewardAmount = 1; */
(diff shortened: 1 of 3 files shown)


core/packages/contracts/contracts/SovereignTreasury.sol
```diff
@@ -0,0 +1,50 @@
...
+}
+
+enum Action {
+    // Withdraw from sovereign account and transfer to recipient.
+    // Parachain teams will occasionally send this message to retrieve collected fees.
+    Withdraw
+}
+
+struct WithdrawPayload {
+    address payable recipient;
```
doubledup, line 48: `address payable` is [not part of the ABI](https://docs.soliditylang.org/en/latest/abi-spec.html#mapping-solidity-to-abi-types), so here we're assuming that the recipient is payable.

core/packages/contracts/contracts/Vault.sol
```diff
@@ -0,0 +1,45 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity ^0.8.9;
+
+import "@openzeppelin/contracts/utils/Address.sol";
+import "@openzeppelin/contracts/access/Ownable.sol";
+
+// TODO: transfer ownership from deployer to SovereignTreasury
+// This contract actually holds Ether balances for each sovereignID.
+contract Vault is Ownable {
```
doubledup, line 9: Based on OpenZeppelin's [Escrow](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/escrow/Escrow.sol), but stores money per depositer. The vault owner can withdraw the money to any payable address.

doubledup, line 9: Another option here is a [PullPayment](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/security/PullPayment.sol). This would also prevent malicious recipients from blocking execution: a potential attack is using up all remaining gas on a withdrawal, blocking all future messages for a given sourceID/parachain.

This isn't much of a concern though, given that sources/parachains will control the `recipient` of Withdraw actions and the impact is isolated to their ability to use the bridge.

It might be a consideration for dev experience if the recipient account can be a contract controlled by a source/parachain, but for now we're assuming it's an EOA.

core/packages/contracts/contracts/SovereignTreasury.sol
```diff
@@ -0,0 +1,49 @@
...
+// TODO: transfer ownership from deployer to inbound channel
+contract SovereignTreasury is Ownable {
+    Vault public vault;
+
+    constructor(Vault _vault) {
+        vault = _vault;
+    }
+
+    // Handle a message from the bridge.
+    function handle(bytes32 _sovereignID, bytes calldata _message) external onlyOwner {
```
alistair-singh, line 15: Non Blocking: I do not think we want to use the `_` underscore convention anymore.
```suggestion
    function handle(bytes32 sovereignID, bytes calldata message) external onlyOwner {
```

core/packages/contracts/contracts/SovereignTreasury.sol
```diff
@@ -0,0 +1,49 @@
...
+        Message memory message = abi.decode(_message, (Message));
+
+        if (message.action == Action.Withdraw) {
+            WithdrawPayload memory payload = abi.decode(message.payload, (WithdrawPayload));
+
+            transfer(_sovereignID, payload.recipient, payload.amount);
+        }
```
alistair-singh, line 22: We should revert if the message is not an action we know about.

core/packages/contracts/contracts/Vault.sol
```diff
@@ -0,0 +1,45 @@
...
+    using Address for address payable;
+
+    event Deposited(bytes32 indexed sovereignID, uint256 amount);
```
alistair-singh, line 12: In my vault implementation, I store `msg.sender` in Deposited/Withdraw events as opposed to sovereignId. The implementation emits Lock/Unlock events from NativeTokens contract which contains the sovereignId or origin of the request. This allows us to see which version of NativeTokens contract initiated the withdraw/deposit. After an upgrade of `SovereignTreasury` contract the `msg.sender` should change to reflect the new version of `SovereignTreasury` in these events.

core/packages/contracts/contracts/SovereignTreasury.sol
vgeddes: Agreed!

vgeddes: Also `sovereignID` should be `origin` in this context. Its the parachain from whench the message originated

core/packages/contracts/contracts/SovereignTreasury.sol
vgeddes: Agree.

Should raise a custom error like `UnknownMessage`: https://blog.soliditylang.org/2021/04/21/custom-errors/

core/packages/contracts/contracts/Vault.sol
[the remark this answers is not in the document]
vgeddes: Agree

core/packages/contracts/contracts/Vault.sol
```diff
@@ -0,0 +1,45 @@
...
+    event Withdrawn(bytes32 indexed sovereignID, address recipient, uint256 amount);
+
+    // Mapping of sovereignID to balance
+    mapping(bytes32 => uint256) private _balances;
```
vgeddes, line 16: No underscore prefixes please. That's really for larger, more complicated contracts where local variables can shadow global variables.

```suggestion
    mapping(bytes32 => uint256) private balances;
```

core/packages/contracts/contracts/SovereignTreasury.sol
doubledup: ah ok. We talked about them being the same [in the doc](https://www.notion.so/snowfork/SovereignTreasury-ed3e8bf757be426b800c8918e77aa4ea?d=3c1d360cd770409c941e8d03ecc422e6#d05a7d66f68c41ec98a4f915e63082ce), so I didn't see the distinction. Is it that it's called `origin` for messages and `sovereignID` for transactions?

core/packages/contracts/contracts/SovereignTreasury.sol
doubledup: Won't `abi.decode` have to revert if the action isn't recognised? Else it has to choose an action for `message` to be defined.

Makes sense to have a better error message for it 👌

core/packages/contracts/contracts/SovereignTreasury.sol
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    // Deposit ETH into a sovereign account. Permissionless.
+    function deposit(bytes32 _sovereignID) external payable {
+        vault.deposit{ value: msg.value }(_sovereignID);
+    }
+
+    function transfer(bytes32 sovereignID, address payable recipient, uint256 amount) private {
```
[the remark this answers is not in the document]
doubledup: iirc, passing around reference types in `memory` like this is more expensive than passing value types, but the abstraction makes more sense & we can optimise later.

core/packages/contracts/contracts/Vault.sol
doubledup: Nice, adding `msg.sender` sounds like a good idea for audits.

I still think we should keep the `sovereignID` though. The deposit happens on behalf of a `sovereignID`, and is recorded in the `balances` mapping.

core/packages/contracts/contracts/SovereignTreasury.sol
alistair-singh: I have tested in Remix VM and foundry VM and enums are validated when using `abi.decode`. So the code that reverts with that custom error will be unreachable.

core/packages/contracts/contracts/SovereignTreasury.sol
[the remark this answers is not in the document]
alistair-singh: I believe it's okay when passing them to private/internal functions which are optimized to jumps and data is not copied. It might just be the same when optimized.

core/packages/contracts/contracts/Vault.sol
alistair-singh: `sovereignId` can be logged in events emitted from `SovereignTreasury.sol`.

vgeddes (approved):
Looks good! 

Gonna merge this into my `vincent/xcmv3` working branch for some further fine-tuning.

PR #777, opened 2023-02-17
Fix: e2e setup for goerli network & Refactor: move initialization to nix hook

yrong:


flake.nix
```diff
@@ -72,7 +72,10 @@
                     # LIBCLANG_PATH points rocksdb to a clang.so on Linux
                     export LIBCLANG_PATH="$(readlink -f ${pkgs.clang}/resource-root/include | xargs dirname | xargs dirname | xargs dirname)"
 
-                    cowsay "Snowbridge Dev Environment"
+                    echo "Initializing Dev Environment..."
+                    ./init.sh
```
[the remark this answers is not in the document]
yrong: Could you elaborate some more abou benefit of inline script here?  IMHO seperate to another shell script is more maintainable and yes the script here applies also for none nix env

core/package.json
```diff
@@ -2,12 +2,11 @@
   "name": "@snowbridge/core",
   "private": true,
   "engines": {
-    "node": "v18.13.0",
+    "node": "v18.12.1",
```
yrong, line 5: @doubledup  saw this warning message when initializing nix env seems node in nix is v18.12 by default? so downgrade version here.
```
WARN: Unsupported engine: wanted: {"node":"v18.13.0"} (current: {"node":"v18.12.1","pnpm":"7.14.2"})
```

core/packages/contracts/scripts/generateContractInfo.js
```diff
@@ -1,6 +1,7 @@
 const fs = require("fs");
 const path = require("path");
-const DeployInfoFile = "./broadcast/Deploy.sol/15/run-latest.json";
+const NetworkId = process.env.ETH_NETWORK_ID || 15;
+const DeployInfoFile = `./broadcast/Deploy.sol/${NetworkId}/run-latest.json`;
```
claravanstaden, line 4: Nice, thank you!

claravanstaden (approved):

core/package.json
doubledup: Whoops, that's right. Should be 18.12.1: https://search.nixos.org/packages?channel=22.11&show=nodejs-18_x&from=0&size=50&sort=relevance&type=packages&query=nodejs-18

flake.nix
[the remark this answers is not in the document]
doubledup: `init.sh` looks like it's only `forge install` and `pnpm install`, which are dev env setup commands that aren't directly related to Nix. We could run them outside Nix, but we're not planning to afaik.

I can't see us gaining much from a separate script here. We could separate Nix-related commands in `shellHook` vs non-Nix-related commands in `init.sh`, but that only saves us a small amount of time if we decide to stop using Nix.

core/package.json
```diff
@@ -2,12 +2,11 @@
...
     "pnpm": "7.14.2"
   },
   "scripts": {
     "preinstall": "npx only-allow pnpm",
-    "postinstall": "cd .. && husky install && cd core/packages/test && ./scripts/hack-ethereum.sh",
```
doubledup, line 10: `husky` and `hack-ethereum` are tightly coupled to Node, so we might as well keep this script.

alistair-singh (approved):

core/packages/test/scripts/configure-contracts.sh
```diff
@@ -25,7 +25,7 @@ configure_contracts()
     	"initialize(uint64,(uint128,uint128,bytes32),(uint128,uint128,bytes32))" \
     	--rpc-url $eth_endpoint_http \
     	--private-key $PRIVATE_KEY \
-      $beefy_start_block \($current_id,$current_length,$current_root\) \($next_id,$next_length,$next_root\)
+      $beefy_start_block \($current_id,$current_length,$current_root\) \($next_id,$next_length,$next_root\) || true
```
claravanstaden, line 28:
```suggestion
      $beefy_start_block \($current_id,$current_length,$current_root\) \($next_id,$next_length,$next_root\) || true
      # "|| true" can be removed once https://github.com/foundry-rs/foundry/pull/4010 has been released
```

core/packages/test/scripts/deploy-polkadot.sh
```diff
@@ -11,7 +11,7 @@ generate_chain_spec() {
     while [ -z "$initial_beacon_block" ] || [ "$initial_beacon_block" == "0x0000000000000000000000000000000000000000000000000000000000000000" ]
     do
         echo "Waiting for beacon chain to finalize to get initial block..."
-        initial_beacon_block=$(curl -s "$beacon_endpoint_http/eth/v1/beacon/states/head/finality_checkpoints" \
+        initial_beacon_block=$(curl -s "$beacon_endpoint_http/eth/v1/beacon/states/justified/finality_checkpoints" \
```
yrong, line 14: @claravanstaden not sure this is nessessary, but with `head` sometimes failed in my setup

core/packages/test/config/beefy-relay.json
```diff
@@ -8,8 +8,7 @@
   },
   "sink": {
     "ethereum": {
-      "endpoint": "ws://127.0.0.1:8546",
-      "gas-limit": 5000000
```
yrong, line 12: would like to know is this required for local setup, remove it since does not work for goerli

yrong: from tests seems still required for local setup, so set as variable here

core/packages/test/.envrc-example
```diff
@@ -1,4 +1,4 @@
-source_up_if_exists
+# source_up_if_exists
```
claravanstaden, line 1: What is this for? Since it is commented out, perhaps we can remove it?

core/packages/test/scripts/deploy-polkadot.sh
claravanstaden: Hmmm, I wonder if we shouldn't rather retry getting the head finalized checkpoint and bootstrap if it fails?

claravanstaden (commented):
Looks great! Thanks especially for getting the Goerli deployment working again. I made a few small comments but almost ready to merge. 👍

flake.nix
```diff
@@ -72,7 +72,10 @@
                     # LIBCLANG_PATH points rocksdb to a clang.so on Linux
                     export LIBCLANG_PATH="$(readlink -f ${pkgs.clang}/resource-root/include | xargs dirname | xargs dirname | xargs dirname)"
 
-                    cowsay "Snowbridge Dev Environment"
+                    echo "Initializing Snowbridge Dev Environment..."
+                    cd core && pnpm install && cd -
```
vgeddes, line 76: or more succinctly:
```suggestion
                    (cd core && pnpm install)
```

alistair-singh (approved):

core/packages/test/scripts/deploy-polkadot.sh
yrong: I would assume both as check point will be fine, but anyway we can still use `head` and retry util finished so I'll just reverted.

core/packages/test/.envrc-example
yrong: I do not know what it is for and guess it is just some hint to source the env. Since we've introduced `direnv` and that will be done automaticlly so would assume we do not need it.
So if no other comment here I will just remove it.

core/packages/test/scripts/deploy-polkadot.sh
claravanstaden: `justified` checkpoints can still be reorged (as I understand it) and so `finalized` should be used instead for safety.

claravanstaden (approved):

core/packages/contracts/.envrc-example
```diff
@@ -1,6 +1,4 @@
-source_up_if_exists
-
-export PRIVATE_KEY=0x4e9444a6efd6d42725a250b650a781da2737ea308c839eaccb0f7f3dbd2fea77
-export RANDAO_COMMIT_DELAY=3
-export RANDAO_COMMIT_EXP=8
-export PARAID=1000
+// export PRIVATE_KEY=0x4e9444a6efd6d42725a250b650a781da2737ea308c839eaccb0f7f3dbd2fea77
+// export RANDAO_COMMIT_DELAY=3
+// export RANDAO_COMMIT_EXP=8
+// export PARAID=1000
```
claravanstaden, line 4: What is the `//` for? If it is comments, shouldn't it be `#`?

flake.nix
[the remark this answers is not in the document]
doubledup: Agreed about `husky` 👌 It's possible, though unlikely, that we'll want to run the non-`husky` commands in start-services. This would allow start-services to work in the same Nix shell after a "clean" operation (eg. deleting `node_modules/` and the submodules in `lib/`).

Assuming we use `git clean` for the clean action, this won't delete the Foundry submodules. So adding `forge install` to start-services is less of a concern.

This leaves the `pnpm install` and `hack-ethereum` commands, which are necessary after removing node_modules and before starting the test environment. A hack for this is restarting the Nix shell: with `direnv` set up, `cd ~ && cd -` will rerun the shellHook, including `pnpm install` and the postinstall script.

If we go with this approach, let's add a note to the README about restarting the Nix shell after a `git clean`.

core/packages/test/.envrc-example
doubledup: This runs the root `.envrc`, if it exists. It's for the [nix-direnv integration](https://github.com/Snowfork/snowbridge/blob/main/.envrc.example). The end of the [Development section of the root README](https://github.com/Snowfork/snowbridge/blob/main/README.md#development) shows how to set that up.

When the direnv integration is set up in the root and there's an `.envrc` in the current directory without this line, we'll leave the Nix shell. This removes our dev dependencies, which definitely isn't what we want in the test directory.

[direnv docs](https://direnv.net/man/direnv-stdlib.1.html#codesourceupifexists-ltfilenamegtcode)

core/packages/test/config/parachain-relay.json
```diff
@@ -1,28 +1,27 @@
...
+    "contracts": {
+      "BeefyClient": null,
+      "BasicInboundChannel": null
+    },
+    "beefy-activation-block": 0,
+    "basicChannelSourceIDs": []
+  },
+  "sink": {
+    "ethereum": {
+      "endpoint": "ws://127.0.0.1:8546"
```
doubledup, line 21: `gas-limit` was removed. Don't we need `"gas-limit": null` here, like in [`beefy-relay.json`](https://github.com/Snowfork/snowbridge/pull/777/files/dd12d04c34f67438160824ed77965e6444b6067b..6b9c725f04fa0736ee2b8076ec4e15358649d8f1#diff-7073e2b92ca88949ecf2c3fcbfe1ddfbdcdecafb89901f7689b78310ffa35099)?

Or could we remove the `gas-limit` line there?

core/packages/test/scripts/configure-contracts.sh
```diff
@@ -21,12 +21,16 @@ configure_contracts()
     current_root=$(sed -e 's/^"//' -e 's/"$//' <<< $current_root)
     next_root=$(sed -e 's/^"//' -e 's/"$//' <<< $next_root)
 
+    echo "Transact call to initialize BeefyClient"
+    # sometimes send transact with cast will fail in goerli network
+    # temporary resolved by manually setting --gas-price and --gas-limit
```
doubledup, line 26:
```suggestion
    # sometimes sending transact with cast will fail on the goerli network
    # temporarily resolved by manually setting --gas-price and --gas-limit
```

core/packages/test/scripts/set-env.sh
```diff
@@ -24,6 +25,7 @@ export PATH="$output_bin_dir:$PATH"
 eth_network="${ETH_NETWORK:-localhost}"
 eth_endpoint_http="${ETH_RPC_ENDPOINT:-http://127.0.0.1:8545}/${INFURA_PROJECT_ID:-}"
 eth_endpoint_ws="${ETH_WS_ENDPOINT:-ws://127.0.0.1:8546}/${INFURA_PROJECT_ID:-}"
+eth_gas_limt="${ETH_GAS_LIMIT:-5000000}"
```
doubledup, line 28:
```suggestion
eth_gas_limit="${ETH_GAS_LIMIT:-5000000}"
```

core/packages/test/scripts/start-relayer.sh
```diff
@@ -4,18 +4,17 @@ set -eu
...
-start_relayer()
-{
-    echo "Starting relay services"
-
+config_relayer(){
     # Configure beefy relay
     jq \
         --arg k1 "$(address_for BeefyClient)" \
         --arg eth_endpoint_ws $eth_endpoint_ws \
+        --arg eth_gas_limit $eth_gas_limt \
```
doubledup, line 12:
```suggestion
        --arg eth_gas_limit $eth_gas_limit \
```

core/init.sh
```diff
@@ -0,0 +1,11 @@
...
+#!/usr/bin/env bash
+
+echo "Hack SLOTS_IN_EPOCH in lodestar"
+cd packages/test && ./scripts/hack-ethereum.sh && cd -
+
+echo "Install husky hook"
+cd .. && ./core/node_modules/.bin/husky install && cd -
+
+echo "Initialize foundry libraries"
+cd packages/contracts && (forge install||true) && cd -
```
doubledup, line 11: Similar to Vincent's comment in flake.nix, would subshells also work here?
```suggestion
echo "Hack SLOTS_IN_EPOCH in lodestar"
(cd packages/test && ./scripts/hack-ethereum.sh)

echo "Install husky hook"
(cd .. && ./core/node_modules/.bin/husky install)

echo "Initialize foundry libraries"
(cd packages/contracts && (forge install||true))
```

doubledup (approved):
LGTM! All minor comments

core/packages/test/scripts/configure-contracts.sh
yrong: Good catch! IMO though foundry is fast but not stable enough, weird issue sometimes. So I'll keep it temprarily and test in a seperate PR.

core/packages/test/.envrc-example
yrong: Nice.

core/packages/test/config/parachain-relay.json
yrong: Yes, we need to add it. Test shows still required for local setup.

core/packages/test/scripts/set-env.sh
yrong: Good catch

core/init.sh
yrong: Exactly!

PR #778, opened 2023-02-17
Add NativeTokens Contract

alistair-singh:
Fixes: SNO-366, SNO-392, SNO-393, SNO-394 and SNO-395.


core/packages/contracts/contracts/utils/MessageProtocol.sol
```diff
@@ -0,0 +1,17 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity ^0.8.9;
+
+library MessageProtocol {
```
doubledup, line 4: I'll use this library after this PR is merged.

core/packages/contracts/contracts/NativeTokens.sol
```diff
@@ -0,0 +1,123 @@
...
+    /// @dev Processes messages from inbound channel.
+    /// @param origin The hashed substrate sovereign account.
+    /// @param message The message enqueued from substrate.
+    function handle(bytes32 origin, bytes calldata message) external onlyOwner {
+        /// TODO: require origin is statemint.
+        MessageProtocol.Message memory decoded = abi.decode(message, (MessageProtocol.Message));
+        if (decoded.action == MessageProtocol.Action.Unlock) {
+            unlock(origin, abi.decode(decoded.payload, (UnlockPayload)));
+        } else {
+            revert("NativeTokens: unknown action");
```
vgeddes, line 110: Rather raise a custom error here: https://blog.soliditylang.org/2021/04/21/custom-errors/

core/packages/contracts/contracts/utils/MessageProtocol.sol
vgeddes: I don't think we need a common `MessageProtocol` library, since each "app" will have its own specific message and action types. So for example, we could just move `Action` and `Message` to `NativeTokens`.

core/packages/contracts/contracts/NativeTokens.sol
```diff
@@ -0,0 +1,123 @@
...
+        }
+    }
+
+    /// @notice Unlocks funds from the vault and sends it to recipient.
+    /// @param origin The hashed substrate sovereign account.
+    /// @param payload A decoded unlock payload.
+    function unlock(bytes32 origin, UnlockPayload memory payload) private {
```
vgeddes, line 117: I'd prefer to call this `doUnlock`, which kinda implies that its an internal method for handling something.

core/packages/contracts/contracts/NativeTokens.sol
```diff
@@ -0,0 +1,123 @@
...
+        outboundChannel = _outboundChannel;
+    }
+
+    /// @notice Locks tokens to mint on substrate.
+    /// @dev Locks an amount of ERC20 Tokens in the vault and enqueues a mint message. Requires the allowance to be set on the ERC20 token where the spender is the vault.
+    /// @param token The token to lock.
+    /// @param recipient The recipient on the substrate side.
+    /// @param amount The amount to lock.
+    function lock(address token, bytes32 recipient, uint256 amount) public {
+        require(amount > 0, "NativeTokes: non zero amount");
```
vgeddes, line 63: Custom errors here please.

core/packages/contracts/contracts/NativeTokens.sol
```diff
@@ -0,0 +1,123 @@
...
+    /// @dev Enqueues a create native token message to substrate.
+    /// @param token The ERC20 token address.
+    /// @param name The name of the ERC20 token.
+    /// @param symbol The symbol of the ERC20 token.
+    /// @param decimals The decimals for the ERC20 token.
+    function create(
+        address token,
+        string calldata name,
+        string calldata symbol,
+        uint8 decimals
```
vgeddes, line 89: Actually we need to read this metadata from the ERC20 contract, as we can't trust input supplied by an arbitrary user.

https://docs.openzeppelin.com/contracts/4.x/api/token/erc20#IERC20Metadata

vgeddes (commented):
Looks good so far 🚀 

Just one thing - I would not spend too much time on the Natspec docstrings, as the interface is liable to evolve a fair bit over the next few weeks. Once we've settled on the API, it makes more sense to add natspec docstrings.

core/packages/contracts/contracts/utils/MessageProtocol.sol
alistair-singh: Ok so will we duplicate `Action` and `Message` types in each contract that implements `handle`?

core/packages/contracts/contracts/NativeTokens.sol
```diff
@@ -0,0 +1,148 @@
...
+        emit Created(token, name, symbol, decimals);
+        outboundChannel.submit(msg.sender, call, weight);
+    }
+
+    /// @notice Handles messages coming in over the bridge.
+    /// @dev Processes messages from inbound channel.
+    /// @param origin The hashed substrate sovereign account.
+    /// @param message The message enqueued from substrate.
+    function handle(bytes32 origin, bytes calldata message) external onlyOwner {
+        if (origin != allowedOrigin) revert UnauthorizedOrigin();
```
vgeddes, line 129: please don't use 1-liners like these. IMO, they make the code harder to audit

core/packages/contracts/contracts/NativeTokens.sol
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    /// @notice Locks tokens to mint on substrate.
+    /// @dev Locks an amount of ERC20 Tokens in the vault and enqueues a mint message. Requires the allowance to be set on the ERC20 token where the spender is the vault.
+    /// @param token The token to lock.
+    /// @param recipient The recipient on the substrate side.
+    /// @param amount The amount to lock.
+    function lock(address token, bytes32 recipient, uint256 amount) public {
+        if (amount == 0) revert ZeroAmount();
+        if (token == address(0)) revert ZeroAddressToken();
```
vgeddes, line 88: ditto, no 1-liners please.

core/packages/contracts/contracts/NativeTokens.sol
```diff
@@ -0,0 +1,148 @@
...
+        outboundChannel = _outboundChannel;
+        allowedOrigin = _allowedOrigin;
+    }
+
+    /// @notice Locks tokens to mint on substrate.
+    /// @dev Locks an amount of ERC20 Tokens in the vault and enqueues a mint message. Requires the allowance to be set on the ERC20 token where the spender is the vault.
+    /// @param token The token to lock.
+    /// @param recipient The recipient on the substrate side.
+    /// @param amount The amount to lock.
+    function lock(address token, bytes32 recipient, uint256 amount) public {
```
vgeddes, line 86: Remember from the previous ERC20 app that users should only be allowed to lock `type(uint128).max` amount of tokens. 

Parachains on the Polkadot side don't support 256-bits of precision.

core/packages/contracts/contracts/NativeTokens.sol
```diff
@@ -0,0 +1,148 @@
...
+
+import "./utils/MessageProtocol.sol";
+import "./ERC20Vault.sol";
+import "./OutboundChannel.sol";
+
+/// @title Native Tokens
+/// @notice A contract for managing ethereum native tokens.
+/// @dev Manages locking, unlocking ERC20 tokens in the vault. Initializes ethereum native tokens on the substrate side via create.
+contract NativeTokens is Ownable {
+    /// @notice Unlock payload format.
```
vgeddes, line 15: Rather just use `@dev` instead of both `@dev` and `@notice`. We don't need to be more specific.

core/packages/contracts/contracts/utils/MessageProtocol.sol
vgeddes: yup 👍

core/packages/contracts/contracts/NativeTokens.sol
vgeddes: Also we need to look back at your previous vault PR to see if we left the vault at 256-bits of precision. I think we did, but should confirm.

core/packages/contracts/contracts/NativeTokens.sol
```diff
@@ -0,0 +1,148 @@
...
+    struct UnlockPayload {
+        /// @notice The ERC20 token to unlock.
+        address token;
+        /// @notice The destination address that will receive unlocked funds.
+        address recipient;
+        /// @notice The amount to unlock.
+        uint256 amount;
+    }
+
+    /// @notice Funds where locked.
```
vgeddes, line 25:
```suggestion
    /// @dev Funds where locked.
```

core/packages/contracts/contracts/NativeTokens.sol
alistair-singh: It was mixed. I think I will leave the vault as uint256 and change `lock` and `unlock` code paths in NativeTokens to only use uint128 and validate this.

vgeddes: Sounds good 👌

core/packages/contracts/contracts/NativeTokens.sol
```diff
@@ -0,0 +1,166 @@
...
+    ERC20Vault public immutable vault;
+
+    /// @dev The channel used to enqueue messages for lock and create functions.
+    OutboundChannel public immutable outboundChannel;
+
+    /// @dev Initializes the NativeTokens contract with a vault and channels.
+    /// @param _vault The vault to use to `lock`/`unlock` tokens.
+    /// @param _outboundChannel The channel used to queue lock and create messages.
+    /// @param _allowedOrigin The origin allowed to call handle.
+    constructor(ERC20Vault _vault, OutboundChannel _outboundChannel, bytes32 _allowedOrigin) {
```
vgeddes, line 86: We need to resurrect the `ChannelRegistry` contract. Is necessary to allow the outbound channel to be upgraded. Can you pull it from our git history?

core/packages/contracts/contracts/NativeTokens.sol
```diff
@@ -0,0 +1,166 @@
...
+    function lock(address token, bytes32 recipient, uint128 amount) public {
+        if (amount == 0) {
+            revert ZeroAmount();
+        }
+        if (token == address(0)) {
+            revert ZeroAddressToken();
+        }
+        if (recipient == bytes32(0)) {
+            revert ZeroAddressRecipient();
+        }
```
vgeddes, line 105: These checks are unnecessary, I think. If the token is 0, then the transfer functions will revert. If the recipient is zero, the user definitely provided invalid input and their funds will be lost.

core/packages/contracts/contracts/NativeTokens.sol
```diff
@@ -0,0 +1,166 @@
...
+
+        Message memory decoded = abi.decode(message, (Message));
+        if (decoded.action == Action.Unlock) {
+            doUnlock(origin, abi.decode(decoded.payload, (UnlockPayload)));
+        } else {
+            revert UnsupportedMessageAction();
+        }
+    }
+
+    /// @dev Unlocks funds from the vault and sends it to recipient.
```
vgeddes, line 157: We don't need docstrings for internal methods. Please remove. They just add extra noise.

vgeddes: As I said before, lets not focus on docstrings at this stage.

core/packages/contracts/contracts/NativeTokens.sol
alistair-singh: addressed in 77bd470fe95a39.

core/packages/contracts/contracts/NativeTokens.sol
```diff
@@ -0,0 +1,166 @@
...
+import "./ERC20Vault.sol";
+import "./OutboundChannel.sol";
+
+/// @title Native Tokens
+/// @dev Manages locking, unlocking ERC20 tokens in the vault. Initializes ethereum native tokens on the substrate side via create.
+contract NativeTokens is Ownable {
+    /// @dev Describes the type of message.
+    enum Action {
+        /// @dev The default value for messages created.
+        Uninitialized,
```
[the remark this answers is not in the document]
alistair-singh: Initially, I needed this to test [this code branch](https://github.com/Snowfork/snowbridge/pull/778/files#diff-b553b52c9c70a372ce0b2dc5c9183a832e650c77ecb2a176c1d492647d692e90R134-R138). This was before I learned that `abi.decode` will validate enums are within a range of values. So the above enum value and related code branch are not needed.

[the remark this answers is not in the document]
alistair-singh: Resolved in 02a7bbe34e6e0dca1b6f1

core/packages/contracts/contracts/NativeTokens.sol
alistair-singh: I will create a new PR for this.

vgeddes (approved):
Looks good!

PR #781, opened 2023-02-25
Contracts for XCMv3

vgeddes:


core/packages/contracts/contracts/NativeTokens.sol
```diff
@@ -35,111 +38,90 @@ contract NativeTokens is Ownable {
...
-    OutboundChannel public immutable outboundChannel;
+    error InvalidAmount();
+    error Unauthorized();
 
-    /// @dev Initializes the NativeTokens contract with a vault and channels.
-    /// @param _vault The vault to use to `lock`/`unlock` tokens.
-    /// @param _outboundChannel The channel used to queue lock and create messages.
-    /// @param _allowedOrigin The origin allowed to call handle.
-    constructor(ERC20Vault _vault, OutboundChannel _outboundChannel, bytes32 _allowedOrigin) {
+    constructor(TokenVault _vault, IOutboundChannel _outboundChannel, bytes memory _peer) {
```
alistair-singh, line 65: Is `peer` a scale encoded multi-location to Statemint?

core/packages/contracts/contracts/NativeTokens.sol
```diff
@@ -35,111 +38,90 @@ contract NativeTokens is Ownable {
...
 
-    /// @dev the origin provided is unauthorized.
-    error UnauthorizedOrigin();
+    bytes32 public constant ADMIN_ROLE = keccak256("ADMIN_ROLE");
+    bytes32 public constant SENDER_ROLE = keccak256("SENDER_ROLE");
 
-    /// @dev the message action is unsupported.
-    error UnsupportedMessageAction();
+    bytes32 public immutable peerID;
+    bytes public peer;
```
alistair-singh, line 55: `peer` can be immutable?
```suggestion
    bytes public immutable peer;
```

vgeddes: yeah, good catch!

core/packages/contracts/src/NativeTokens.sol
```diff
@@ -55,7 +53,7 @@ contract NativeTokens is AccessControl {
     bytes32 public constant SENDER_ROLE = keccak256("SENDER_ROLE");
 
     bytes32 public immutable peerID;
-    bytes public immutable peer;
+    bytes public peer;
```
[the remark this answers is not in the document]
vgeddes: It's due to a compiler constraint. The bytes type cannot be immutable.

core/packages/contracts/src/test/BeefyClient.t.sol
```diff
@@ -51,67 +53,44 @@ contract BeefyClientTest is Test {
 
         // Allocate for input variables
         string[] memory inputs = new string[](10);
-        inputs[0] = "test/ffiWrapper.js";
-        // Always add type of command as first arguments
-        inputs[1] = "GenerateInitialSet";
+        inputs[0] = "node_modules/.bin/ts-node";
```
vgeddes, line 56: @yrong I simplified the compilation of the ffiWrapper.ts script, by using `ts-node`, which automatically compiles and runs the script. This means we no longer need to compile the script manually via the package.json script.

.husky/pre-commit
```diff
@@ -6,7 +6,7 @@ set -e
 echo 'Running pre-commit hook...'
 
 # check typos
-chronic typos .
+chronic typos --exclude core/packages/contracts/lib .
```
yrong, line 9: https://github.com/Snowfork/snowbridge/blob/be58054456a890b46574d5a8483b520ac11b09ef/_typos.toml#L2

could also add into this file

core/package.json
```diff
@@ -8,12 +8,12 @@
...
-    "test": "turbo run test",
-    "lint": "turbo run lint",
-    "format": "turbo run format",
-    "size": "turbo run size",
-    "coverage": "turbo run coverage",
+    "build": "turbo run build --filter '!./packages/contracts/lib/*'",
+    "test": "turbo run test --filter '!./packages/contracts/lib/*'",
+    "lint": "turbo run lint --filter '!./packages/contracts/lib/*'",
+    "format": "turbo run format --filter '!./packages/contracts/lib/*'",
+    "size": "turbo run size --filter '!./packages/contracts/lib/*'",
```
yrong, line 15: If we remove hardhat thoroughly from our stack, we need also remove `size` entry script here which depends on hardhat which is used to calculate contract size to avoid exceeds limit.

core/package.json
```diff
@@ -8,12 +8,12 @@
...
-    "build": "turbo run build",
-    "test": "turbo run test",
-    "lint": "turbo run lint",
-    "format": "turbo run format",
-    "size": "turbo run size",
-    "coverage": "turbo run coverage",
+    "build": "turbo run build --filter '!./packages/contracts/lib/*'",
+    "test": "turbo run test --filter '!./packages/contracts/lib/*'",
+    "lint": "turbo run lint --filter '!./packages/contracts/lib/*'",
+    "format": "turbo run format --filter '!./packages/contracts/lib/*'",
```
yrong, line 14: Is the `filter` here nessessary?

vgeddes: yeah, unfortunately. Otherwise turborepo will discover all npm script commands in the lib dir and run them.

But I am thinking we should remove turborepo since its a bit overkill for the few packages we have.

core/package.json
vgeddes: Fixed in latest commit.

core/packages/contracts/package.json
```diff
@@ -11,34 +11,26 @@
...
+    "build": "pnpm build:sol && pnpm build:types",
     "build:sol": "forge build",
-    "build:ffi": "esbuild ./test/ffiWrapper.ts --bundle --platform=node --outfile=./test/ffiWrapper.js",
-    "lint": "pnpm lint:solhint && pnpm lint:eslint",
-    "lint:solhint": "solhint 'contracts/**/*.sol'",
-    "lint:eslint": "eslint .",
-    "format": "prettier --write 'contracts/**/*.sol' 'test/**/*.ts'",
-    "size": "hardhat size-contracts",
-    "coverage": "hardhat coverage --testfiles 'test/*.ts'",
+    "build:types": "typechain --target ethers-v5 'out/**/!(*.t|test).sol/!(*.abi).json' --out-dir types",
```
yrong, line 16: Is `typechain` here nessessary in foundry stack? `forge build` will also generate abi files.

core/packages/contracts/src/NativeTokens.sol
```diff
@@ -0,0 +1,149 @@
...
+        bytes memory payload = NativeTokensTypes.Create(peer, token, name, symbol, decimals);
+        outboundChannel.submit{ value: msg.value }(peer, payload);
+
+        emit Created(token);
+    }
+
+    /// @dev Processes messages from inbound channel.
+    /// @param origin The multilocation of the source parachain
+    /// @param message The message enqueued from substrate.
+    function handle(bytes calldata origin, bytes calldata message) external onlyRole(SENDER_ROLE) {
```
yrong, line 132: would assume `handle` can be invoked only by `InboundChannel` with  `SENDER_ROLE`, seems we did not explicit initialize it in `construct`?

core/packages/contracts/src/NativeTokens.sol
```diff
@@ -0,0 +1,149 @@
...
+    constructor(
+        TokenVault _vault,
+        IOutboundChannel _outboundChannel,
+        bytes memory _peer,
+        uint256 _createTokenFee
+    ) {
+        _grantRole(ADMIN_ROLE, msg.sender);
+        _setRoleAdmin(SENDER_ROLE, ADMIN_ROLE);
+        vault = _vault;
+        outboundChannel = _outboundChannel;
```
yrong, line 80: are we gonna also to add `inboundChannel` here with `SENDER_ROLE`? 
detail comment here.
https://github.com/Snowfork/snowbridge/pull/781/files#r1135097810

core/packages/contracts/src/NativeTokens.sol
```diff
@@ -0,0 +1,149 @@
...
+
+        bytes memory payload = NativeTokensTypes.Mint(peer, token, recipient, amount);
+        outboundChannel.submit{ value: msg.value }(peer, payload);
+
+        emit Locked(recipient, token, amount);
+    }
+
+    /// @dev Enqueues a create native token message to substrate.
+    /// @param token The ERC20 token address.
+    function create(address token) external payable {
```
yrong, line 106: Is `create` here can execute by anyone? is it better we add a `CREATE_ROLE` for it?

.husky/pre-commit
vgeddes: done

core/packages/contracts/package.json
vgeddes: Yes, it is very much needed. [Typechain](https://github.com/dethcrypto/TypeChain) produces Typescript bindings for our contracts. We need these in bindings for any future JS/TS libs that depend on our contracts.

core/packages/contracts/src/NativeTokens.sol
vgeddes: Yeah, that's by design. Its permissionless. So anyone can call this to register a new token on Statemint.

This is to prevent us from being overwhelmed by governance requests to register new ERC20 tokens with the bridge.

We do make sure to charge a high fee for calling `create`, to avoid spamming attacks.

core/packages/contracts/src/NativeTokens.sol
vgeddes: yes, that's correct. We set that up here in `Deploy.sol`:
https://github.com/Snowfork/snowbridge/pull/781/files#diff-5b435e0b60b4b5dae9b667651a36e777d2de6372627be4a9fea0d9d7986565adR55

core/packages/contracts/src/NativeTokens.sol
vgeddes: We grant the authorization in Deploy.sol:
https://github.com/Snowfork/snowbridge/pull/781/files#diff-5b435e0b60b4b5dae9b667651a36e777d2de6372627be4a9fea0d9d7986565adR55

core/packages/contracts/src/TokenVault.sol
```diff
@@ -0,0 +1,41 @@
...
+    /// @dev Emitted when funds are withdrawn.
+    event Withdraw(address recipient, address token, uint256 amount);
+
+    /// @dev Not enough funds to transfer.
+    error InsufficientBalance();
+
+    /// @dev stores the total balance of each token locked in the vault.
+    mapping(address => uint128) public balance;
+
+    function deposit(address sender, address token, uint128 amount) external onlyOwner {
```
yrong, line 25: Is `onlyOwner` nessessary here for deposit?

vgeddes: Its not necessary. I think I will remove it.

Also, am realizing that for consistency with the rest of the code, I should change this contract to use AccessControl instead of Ownable.

alistair-singh: I think this should be `onlyOwner`. In an upgrade scenario, only one version of the `NativeTokens` contract should own the vault, and such should be the only version allowed to deposit. So if ownership has changed to a v2 of `NativeTokens`, if someone tries to use the v1 contract the `lock` would be rejected.

alistair-singh (approved):
Looks good!

yrong (approved):
Awesome!

alistair-singh (approved):

core/packages/contracts/foundry.toml
```diff
@@ -1,14 +1,18 @@
...
     'forge-std=lib/forge-std/src/',
-    '@openzeppelin=node_modules/@openzeppelin'
+    'openzeppelin=lib/openzeppelin-contracts/contracts/',
 ]
 ffi = true
+
+
+ignored_error_codes = [
+    # prevrandao not supported by Anvil VM
+    9432
```
musnit, line 17: is this used? don't see it in the anvil docs

if so, what are the implications of this? we can't test randomness properly on Anvil, just on Geth? Or?

core/packages/contracts/src/InboundChannel.sol
```diff
@@ -0,0 +1,155 @@
...
+    function submit(
+        Message calldata message,
+        bytes32[] calldata leafProof,
+        bytes calldata headerProof
+    ) external {
+        bytes32 leafHash = keccak256(abi.encode(message));
+        bytes32 commitment = MerkleProof.processProof(leafProof, leafHash);
+        if (!parachainClient.verifyCommitment(commitment, headerProof)) {
+            revert InvalidProof();
+        }
```
musnit, line 85: comment explaining each line would be helpful here.

What is the commitment in the first line?

What is being verified in the second line? That the commitment is in a parachain header or?

core/packages/contracts/src/InboundChannel.sol
```diff
@@ -0,0 +1,155 @@
...
+            revert InvalidNonce();
+        }
+
+        // Increment nonce for origin.
+        // This ensures messages are not replayed. It also ensures re-entrancy protection,
+        // as a re-entrant call will be detected by the nonce check above.
+        //
+        // Sources of re-entrancy:
+        // * The relayer which gets forwarded ETH as a reward for submission
+        // * XCM::Transact calls to arbitrary untrusted contracts (not in scope for initial release)
```
musnit, line 96: what if the re-entrant re-enters with a correct, valid message with the correct next nonce? is there some way they could exploit with that?

core/packages/contracts/src/NativeTokens.sol
```diff
@@ -0,0 +1,149 @@
...
+    /// @dev Set a new outbound channel.
+    event OutboundChannelUpdated(address newOutBoundChannel);
+
+    /* State */
+
+    bytes32 public constant ADMIN_ROLE = keccak256("ADMIN_ROLE");
+    bytes32 public constant SENDER_ROLE = keccak256("SENDER_ROLE");
+
+    bytes32 public immutable peerID;
+    bytes public peer;
```
musnit, line 58: comment to define/explain these 2 variables I think needed. Or maybe could also have some kinda PeeredApp base class that includes these fields which NativeTokens inherits from and which defines these concepts more clearly? PeeredApp base class could also be useful for future devs when we become permissionless?

core/packages/contracts/src/NativeTokens.sol
```diff
@@ -0,0 +1,149 @@
...
+        emit Created(token);
+    }
+
+    /// @dev Processes messages from inbound channel.
+    /// @param origin The multilocation of the source parachain
+    /// @param message The message enqueued from substrate.
+    function handle(bytes calldata origin, bytes calldata message) external onlyRole(SENDER_ROLE) {
+        if (peerID != keccak256(origin)) {
+            revert Unauthorized();
+        }
```
musnit, line 135: Should these 3 lines go into a separate, clearly defined function like "verifyPeerOrigin" ? Especially if we go with the above mentioned idea to create a reusable PeeredApp base class that other apps or other devs can use, then this function could go in that base class since it would likely be used across any peered app?

core/packages/contracts/src/NativeTokens.sol
```diff
@@ -0,0 +1,149 @@
...
+        Unlock
+    }
+
+    /// @dev Message format.
+    struct Message {
+        /// @dev The action type.
+        Action action;
+        /// @dev The message payload.
+        bytes payload;
+    }
```
musnit, line 28: type could go in some kinda PeeredApp base class, as mentioned in https://github.com/Snowfork/snowbridge/pull/781/files#r1155020081 ?

core/packages/contracts/src/SovereignTreasury.sol
```diff
@@ -0,0 +1,43 @@
...
+contract SovereignTreasury is AccessControl {
+    bytes32 public constant ADMIN_ROLE = keccak256("ADMIN_ROLE");
+    bytes32 public constant SENDER_ROLE = keccak256("SENDER_ROLE");
+
+    IVault public vault;
+
+    struct Message {
+        Action action;
+        bytes payload;
+    }
```
musnit, line 16: type could go in some kinda PeeredApp base class, as mentioned in https://github.com/Snowfork/snowbridge/pull/781/files#r1155020081 ?

core/packages/contracts/src/NativeTokens.sol
```diff
@@ -0,0 +1,149 @@
...
+    /// @param message The message enqueued from substrate.
+    function handle(bytes calldata origin, bytes calldata message) external onlyRole(SENDER_ROLE) {
+        if (peerID != keccak256(origin)) {
+            revert Unauthorized();
+        }
+
+        Message memory decoded = abi.decode(message, (Message));
+        if (decoded.action == Action.Unlock) {
+            UnlockPayload memory payload = abi.decode(decoded.payload, (UnlockPayload));
+            vault.withdraw(payload.recipient, payload.token, payload.amount);
```
musnit, line 140: vault.withdraw(payload.recipient..) takes in a recipient field from the payload. Is this defined/set on the substrate side? Can you confirm how/why it can be trusted as a user-inputted field? Does the substrate side validate/authorize this field?

core/packages/contracts/src/InboundChannel.sol
musnit: One major thing happening here is also that we are verifying the origin here that was provided as input in the message.

This is critical to understand for a reader of the code, since all security flows from trust in that origin and the authentication being done.

Since this is a bit of a base axiom that one needs to accept while auditing/reviewing the code, may be worth to have a comment on it here, and definitely as part of a README/document we should have a clear explanation of this and of why the authentication process can be trusted.

core/packages/contracts/src/SovereignTreasury.sol
```diff
@@ -0,0 +1,43 @@
...
+        Withdraw
+    }
+
+    struct WithdrawPayload {
+        address payable recipient;
+        uint256 amount;
+    }
+
+    constructor(IVault _vault) {
+        _grantRole(ADMIN_ROLE, msg.sender);
```
musnit, line 30: ADMIN_ROLE is set to deployer account, and they can change sender role even after deployment? Should have a revokeRole there in the contract / deployment process?

core/packages/contracts/src/Vault.sol
```diff
@@ -0,0 +1,57 @@
...
+
+    bytes32 public constant ADMIN_ROLE = keccak256("ADMIN_ROLE");
+    bytes32 public constant WITHDRAW_ROLE = keccak256("WITHDRAW_ROLE");
+
+    // Mapping of sovereign to balance
+    mapping(bytes => uint256) public balances;
+
+    constructor() {
+        _grantRole(ADMIN_ROLE, msg.sender);
+        _setRoleAdmin(WITHDRAW_ROLE, ADMIN_ROLE);
```
musnit, line 23: ADMIN_ROLE is set to deployer account, and they can change withdrawal role even after deployment? Should have a revokeRole there in the contract / deployment process?

core/packages/contracts/src/Vault.sol
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    receive() external payable {
+        revert("Must use deposit function");
+    }
+
+    function deposit(bytes calldata sovereign) external payable {
```
musnit, line 30: - No DEPOSIT_ROLE, anyone can deposit? Is that okay? Also, what is this function used for? Looks used just to collect fees? Should add comment/description

core/packages/contracts/src/Vault.sol
```diff
@@ -0,0 +1,57 @@
...
+            revert ZeroAmount();
+        }
+
+        if (balances[sovereign] < amount) {
+            revert InsufficientBalance();
+        }
+
+        balances[sovereign] -= amount;
+
+        (bool success, ) = recipient.call{ value: amount }("");
```
musnit, line 50: This recipient here is basically the fee-sponsor, which should usually be an account a parachain controls right? Of course, a bad recipient provided here (eg: contract with an infinite loop contract) will cause the transaction to fail and revert. So, it's on the sovereign who submits this to ensure that a good recipient is provided, otherwise they could mess up their message/channel. Is that correct?

core/packages/contracts/src/InboundChannel.sol
```diff
@@ -0,0 +1,155 @@
...
+        } catch Panic(uint256 errorCode) {
+            result.errorPanicCode = errorCode;
+        } catch (bytes memory returnData) {
+            result.errorReturnData = returnData;
+        }
+
+        emit MessageDispatched(message.origin, message.nonce, result);
+    }
+
+    function updateHandler(uint16 id, Handler memory handler) external onlyRole(ADMIN_ROLE) {
```
musnit, line 134: Handlers are permissioned here, only updatable by admin. I presume that's intentional for initial launch, and when we move to permissionless in future then this could be opened up?

Also, I can't see where this is actually called and where the handlers for the SovereignTreasury and NativeTokens are set? Is this done in the code/deployment at all?

core/packages/contracts/src/InboundChannel.sol
```diff
@@ -0,0 +1,155 @@
...
+    IVault public vault;
+
+    // The relayer reward for submitting a message
+    uint256 public reward;
+
+    // The governance contract, which is a proxy for Polkadot governance, administers via this role
+    bytes32 public constant ADMIN_ROLE = keccak256("ADMIN_ROLE");
+
+    // Relayers must provide enough gas to cover message dispatch plus this buffer
+    uint256 public constant GAS_BUFFER = 24000;
```
musnit, line 30: Why this amount? Can you also explain reasoning around why this number is correct?

core/packages/contracts/src/InboundChannel.sol
```diff
@@ -0,0 +1,155 @@
...
+    event VaultUpdated(address vault);
+    event RewardUpdated(uint256 reward);
+
+    error InvalidProof();
+    error InvalidNonce();
+    error InvalidHandler();
+    error NotEnoughGas();
+
+    constructor(IParachainClient _parachainClient, IVault _vault, uint256 _reward) {
+        _grantRole(ADMIN_ROLE, msg.sender);
```
musnit, line 70: Admin is just deployer. Needs to be changed to Polkadot governance? How will this work? Seems like there are 2 options here, ie, either the InboundChannel becomes an IRecipient too and then calls handle on itself with the Polkadot governance origin and with the update functions checking the origin, or, we have a sovereign account for Polkadot governance deployed onto ethereum that does the calls?

core/packages/contracts/src/InboundChannel.sol
musnit: Another question: origin here is in practice basically a parachain, right?

What about in general - can it be a specific pallet or even specific account in a smart contract in the VM of a parachain?

What about parachains that have smart contracts where users can control code/messages, like Moonbeam where if they exposed this XCM to users then those users could send messages as if they were the parachain to withdraw other users funds? Is this the case? Will we expect Moonbeam to secure this on their own on their side in this case or what? How do then expect an integration/implementation inside their smart contract VM to look?

core/packages/contracts/src/TokenVault.sol
```diff
@@ -0,0 +1,55 @@
...
+
+    bytes32 public constant ADMIN_ROLE = keccak256("ADMIN_ROLE");
+    bytes32 public constant WITHDRAW_ROLE = keccak256("WITHDRAW_ROLE");
+    bytes32 public constant DEPOSIT_ROLE = keccak256("DEPOSIT_ROLE");
+
+    /// @dev stores the total balance of each token locked in the vault.
+    mapping(address => uint128) public balance;
+
+    constructor() {
+        _grantRole(ADMIN_ROLE, msg.sender);
```
musnit, line 30: ADMIN_ROLE is set to deployer account, and they can change sender/withdrawal role even after deployment? Should have a revokeRole there in the contract / deployment process?

core/packages/contracts/src/NativeTokens.sol
```diff
@@ -0,0 +1,149 @@
...
+    error NoFundsforCreateToken();
+
+    constructor(
+        TokenVault _vault,
+        IOutboundChannel _outboundChannel,
+        bytes memory _peer,
+        uint256 _createTokenFee
+    ) {
+        _grantRole(ADMIN_ROLE, msg.sender);
+        _setRoleAdmin(SENDER_ROLE, ADMIN_ROLE);
```
musnit, line 78: needs deployer admin revoke?

core/packages/contracts/src/NativeTokensTypes.sol
```diff
@@ -0,0 +1,52 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity ^0.8.19;
+
+import "./ScaleCodec.sol";
+import "./SubstrateTypes.sol";
+
+library NativeTokensTypes {
```
musnit, line 7: Haven't properly reviewed this file, nor the corresponding ScaleCode changes, SubstrateTypes changes or Bytes/Memory changes, since my knowledge of Substrate types and encodings is low - should make sure to get an additional review/audit of these files from someone more familiar with those bits.

core/packages/contracts/src/NativeTokens.sol
```diff
@@ -0,0 +1,149 @@
...
+        peerID = keccak256(_peer);
+        createTokenFee = _createTokenFee;
+    }
+
+    /// @dev Locks an amount of ERC20 Tokens in the vault and enqueues a mint message.
+    /// Requires the allowance to be set on the ERC20 token where the spender is the vault.
+    /// @param token The token to lock.
+    /// @param recipient The recipient on the substrate side.
+    /// @param amount The amount to lock.
+    function lock(address token, bytes calldata recipient, uint128 amount) external payable {
```
musnit, line 91: what happens if i do a lock of a token before it's been created on the substrate side?

core/packages/contracts/src/NativeTokens.sol
```diff
@@ -0,0 +1,149 @@
...
+        bytes memory payload = NativeTokensTypes.Mint(peer, token, recipient, amount);
+        outboundChannel.submit{ value: msg.value }(peer, payload);
+
+        emit Locked(recipient, token, amount);
+    }
+
+    /// @dev Enqueues a create native token message to substrate.
+    /// @param token The ERC20 token address.
+    function create(address token) external payable {
+        // to avoid spam, charge a fee for creating a new token
```
musnit, line 107: what does spam mean in this context? why do we need to avoid spam? is there a possibility that the state on substrate can get bloated or not scale properly computation/storage wise if an attacker spends a lot to create lots of tokens?

core/packages/contracts/.envrc-example
```diff
@@ -1,11 +1,20 @@
...
+
+## OutboundChannel
+export RELAYER_FEE=100
+
+## InboundChannel
+export RELAYER_REWARD=100
+
+## NativeTokens
+export STATEMINT_LOCATION=0x09b636bf387d256bd35ebf301c2da34fdee53b42f6a4648ad126156034d4a2db
+export CREATE_TOKEN_FEE="1000000000000000000"
```
musnit, line 20: why this amount? any idea what it should actually be?

core/packages/contracts/.envrc-example
```diff
@@ -1,11 +1,20 @@
...
+export RANDAO_COMMIT_EXP=8
 
+## ParachainClient
+export PARAID=1000
+
+## OutboundChannel
+export RELAYER_FEE=100
+
+## InboundChannel
+export RELAYER_REWARD=100
```
musnit, line 16: why this amount? any idea what it should actually be?

core/packages/contracts/src/NativeTokens.sol
```diff
@@ -0,0 +1,149 @@
...
+        if (msg.value < createTokenFee) {
+            revert NoFundsforCreateToken();
+        }
+
+        IERC20Metadata metadata = IERC20Metadata(token);
+        bytes memory name = bytes(metadata.name());
```
musnit, line 113: What if i run NativeTokens.create with 0x0000...? Will that just error out and basically be ignored?

Related, what if the token doesn't have a name/symbol function? do we still support it or will it just revert?

core/packages/contracts/src/NativeTokens.sol
```diff
@@ -0,0 +1,149 @@
...
+        if (name.length > 32) {
+            name = hex"";
+        }
+        bytes memory symbol = bytes(metadata.symbol());
+        if (symbol.length > 32) {
+            symbol = hex"";
+        }
```
musnit, line 120: Need comment to explain purpose of the symbol.length > 32 lines and hex prefix

core/packages/contracts/src/OutboundChannel.sol
```diff
@@ -0,0 +1,46 @@
...
+
+    event Message(bytes dest, uint64 nonce, bytes payload);
+
+    event FeeUpdated(uint256 fee);
+
+    error FeePaymentToLow();
+
+    constructor(IVault _vault, uint256 _fee) {
+        _grantRole(ADMIN_ROLE, msg.sender);
+        _setRoleAdmin(SUBMIT_ROLE, ADMIN_ROLE);
```
musnit, line 25: need admin revoke?

also, can't find where new SUBMIT_ROLE contracts are added? doesn't seem to be in Deploy? where does the NativeTokens get submit permission?

core/packages/contracts/src/OutboundChannel.sol
```diff
@@ -0,0 +1,46 @@
...
+        vault = _vault;
+        fee = _fee;
+    }
+
+    function submit(
+        bytes calldata dest,
+        bytes calldata payload
+    ) external payable onlyRole(SUBMIT_ROLE) {
```
musnit, line 33: not fully following how authentication works for this - looks like anyone with SUBMIT_ROLE can submit here.

Does the message also need to include msg.sender ?

How does, for example, the NativeTokens peer app on the substrate side verify/authorize that a message to it came from the NativeTokens smart contract?

Also, looks like this is a shared channel - I presume this is to-be-deprecated and replaced with a non-shared channel?

core/packages/contracts/src/InboundChannel.sol
```diff
@@ -0,0 +1,155 @@
...
+
+        // reward the relayer
+        // Should revert if there are not enough funds. In which case, the origin
+        // should top up the funds and have a relayer resend the message.
+        vault.withdraw(message.origin, payable(msg.sender), reward);
+
+        Handler memory handler = handlers[message.handler];
+        if (address(handler.recipient) == address(0)) {
+            revert InvalidHandler();
+        }
```
musnit, line 107: Should this revert, or continue and skip over the nonce? If it doesn't revert, then the channel could get blocked indefinitely if a bad handler is sent?

musnit (commented):
Review done. See inline comments for most questions.

There are definitely a lot of "stupid questions" in my review as I'm coming in with a murky memory of our exact architecture/design, so sorry about that, but I think those should make for a better review more like an audit nonetheless 

In addition to the inline questions, some extras:
 - Depositing ETH to mint wrapped ETH is not yet implemented, right?
 - It looks like the word "recipient" is used ambiguously in the code base. Sometimes it means an asset/token recipient, sometimes it's an address, sometimes it's a contract, sometimes its the name of an interface, sometimes it's a contract that specifically implements the IRecipient interface. Should we use different terms/words here to remove this ambiguity? Maybe IHandler instead or?
 - One thing I want to dig in a little further is around the inbound channel, specifically:
   - How do we handle nonce creation, on the substrate side, for the inbound channel?
   - If any line in the InboundChannel submit function can potentially revert, then there is a risk that a channel becomes blocked indefinitely/entirely? At a first glance, https://github.com/Snowfork/snowbridge/pull/781/files#r1155033294 looks like a potential risk that could cause this, but we should go over every line meticulously to ensure that this is not possible.
   - How are channel nonces shared? It looks like each message.origin gets their own nonce, but if a particular origin represents a shared set of users, one of whom may be malicious, then that malicious user could block the channel by sending a blocking message into the channel? 
   - Similarly, even if a nonce is not shared and is just owned entirely by a single human origin, is there still a risk that through some phishing-style attack, an attacker could cause them to insert a "blocking message" into their channel and brick their channel?

core/packages/contracts/src/InboundChannel.sol
```diff
@@ -0,0 +1,155 @@
...
+        bytes origin;
+        uint64 nonce;
+        uint16 handler;
+        bytes payload;
+    }
+
+    // Message handler entry
+    struct Handler {
+        // Address of message handler which implements IRecipient
+        address recipient;
```
musnit, line 43: should this be an IRecipient type or address type? i guess not a big deal either way?

core/packages/contracts/src/InboundChannel.sol
vgeddes: > Handlers are permissioned here, only updatable by admin. I presume that's intentional for initial launch, and when we move to permissionless in future then this could be opened up?

Nope! And this is very important. These handlers are simply components of the bridge and are within its trust domain. Equivalent to pallets in a substrate chain.

This narrowed design actually simplified a lot of things, for example around authorizing relationships between channels and handlers. You may remember previously we had that channel "operator" concept, where users would authorize apps to send messages on their behalf. Well all that complexity is now gone.

Over time we'll expand to include the following set of handlers:
* NativeTokens: Send Ethereum-native assets to Polkadot
* ForeignTokens: Send Polkadot-native assets to Ethereum
* SovereignTreasury: Accounts for sovereign parachains. Used to debit and credit fees/rewards for relayers
* Upgrades: Upgrade Support
* Transact: Support for XCM::Transact to arbitrary smart contracts.

core/packages/contracts/src/NativeTokens.sol
vgeddes: You lose your money 😄 

There's no way to know whether a token was successfully created on the substrate side, unless we support message receipts. That's out of scope for our initial launch.

Also there are plenty of ways users could shoot themselves in the foot with this `lock` call (which by the way supports more routing parameters in a PR I'm working on). For example by submitting a bad recipient or routing parameter.

I'm expecting that frontends will help users supply the right inputs.

musnit: Right - that sounds fine, as long as it doesn't mess with the future registration and locking of that token?

core/packages/contracts/src/Vault.sol
vgeddes: Yes, this is OK, anyone can deposit funds here, by design.

Parachain teams must periodically deposit funds to cover relayer rewards.

These funds can be raised by selling DOT from their sovereign account on BridgeHub. This account contains funds charged to users for sending messages across the bridge to Ethereum.

So in the BridgeHub->Ethereum direction, parachain teams control two sovereign accounts:
* On BridgeHub, funds collected by users for using the bridge
* On Ethereum, funds rewarded to message relayers.

core/packages/contracts/src/SovereignTreasury.sol
vgeddes: See my latest PR to support contract upgrades:

https://github.com/Snowfork/snowbridge/pull/801

I revoke the deployer account's admin privileges.

core/packages/contracts/src/TokenVault.sol
vgeddes: See my latest PR to support upgrades:

https://github.com/Snowfork/snowbridge/pull/801

The deployer's admin privileges are revoked.

core/packages/contracts/src/Vault.sol
vgeddes: See my latest PR to support upgrades:

https://github.com/Snowfork/snowbridge/pull/801

The deployer's admin privileges are revoked.

core/packages/contracts/.envrc-example
```diff
@@ -1,11 +1,20 @@
...
-export TOKENS_ALLOWED_ORIGIN=0x09b636bf387d256bd35ebf301c2da34fdee53b42f6a4648ad126156034d4a2db
+## BeefyClient
+export RANDAO_COMMIT_DELAY=3
+export RANDAO_COMMIT_EXP=8
 
+## ParachainClient
+export PARAID=1000
+
+## OutboundChannel
+export RELAYER_FEE=100
```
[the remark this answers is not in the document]
vgeddes: That amount is just for testing.

We should select an amount that:
* Covers transaction fees on BridgeHub
* Discourages spamming/grieving

core/packages/contracts/foundry.toml
vgeddes: Yes, just a workaround for the current version of Anvil, the EVM used in foundry unit tests.

Current Geth supports this.

core/packages/contracts/src/InboundChannel.sol
vgeddes: > Another question: origin here is in practice basically a parachain, right?

Correct

> What about in general - can it be a specific pallet or even specific account in a smart contract in the VM of a parachain?

Nope, must always be a parachain, whose governance must take responsibility for charging fees from sub-consensus systems (pallet, etc) fees for using their dedicated messaging lane.

For example an EVM contract on Acala must pay fees to Acala toplevel governance for using Acala's messaging lane.

As for malicious sub-consensus systems, the bridge is designed to be resilient against them. Our pallets on BridgeHub will filter messages to only allow specific kinds of messages to be sent to Ethereum.

vgeddes:
> How are channel nonces shared? It looks like each message.origin gets their own nonce, but if a particular origin represents a shared set of users, one of whom may be malicious, then that malicious user could block the channel by sending a blocking message into the channel?

Each parachain is allocated a dedicated messaging lane, for itself and any consensus systems under its control (ink! and evm pallets)

> Similarly, even if a nonce is not shared and is just owned entirely by a single human origin, is there still a risk that through some phishing-style attack, an attacker could cause them to insert a "blocking message" into their channel and brick their channel?

Channels have multiple messaging lanes, keyed by parachain Id, purely to account for relayer fees and rewards.

It's up to individual parachains to ensure that inner pallets and ink!/evm contracts pay fees to use the parachain's messaging lane.

Protection against malicious payloads sent by ink!/evm pallets is the responsibility of our trusted message handler contracts. In the case of our current set of handlers, NativeTokens, and SovereignTreasury, they are able to trust incoming messages, because our pallets on BridgeHub mediate, validate, and transform messages.

PR #783, opened 2023-03-01
Simplify execution header proof

claravanstaden:
For execution header imports, we used to send the whole beacon block body, so that we could verify that the block body hash tree root is what we expect it to be, so that we know the Execution Payload is valid. However, now that that FastSSZ Go lib can generate proofs, we only have to send the Execution Payload and a proof. Then we can leave out the Beacon Block from the beacon client entirely.


relayer/relays/beacon/header/syncer/api/api_response.go
```diff
@@ -515,6 +556,302 @@ func (b BeaconBlockResponse) ToScale() (scale.BeaconBlock, error) {
...
+	}
+
+	n := new(big.Int)
+	n, ok := n.SetString(executionPayload.BaseFeePerGas, 10)
+	if !ok {
+		return nil, err
+	}
+
+	// FastSSZ expects a little endian byte array
+	baseFeePerGas := util.ChangeByteOrder(n.Bytes())
```
yrong, line 691: Interesting, Is this  because `baseFeePerGas` is actually a `uint256` in spec?

yrong (approved):
Awesome! Then I'll continue #786 based on this PR.

relayer/relays/beacon/header/syncer/api/api_response.go
claravanstaden: Yes, exactly! 😄 And in the spec unsigned ints are encoded to a 32 little endian byte, and big Int seems to use big endian. A bit of a pain, but we can remove this when we can convert the Lodestar block ssz bytes directly to the FastSSZ structs.

claravanstaden:
> Awesome! Then I'll continue #786 based on this PR.

Great! I'm going to merge it to unblock you, and then I'll open up a new PR if @vgeddes, @alistair-singh or @doubledup have comments.

PR #786, opened 2023-03-07
Shapella upgrade

yrong:


core/packages/test/scripts/deploy-ethereum.sh
```diff
@@ -15,7 +15,8 @@ start_geth() {
             --rpc.allow-unprotected-txs --mine --miner.threads=1 \
             --miner.etherbase=0xBe68fC2d8249eb60bfCf0e71D5A0d2F2e292c4eD \
             --authrpc.addr="127.0.0.1" \
-            --http.addr="127.0.0.1" \
+            --http.addr="0.0.0.0" \
+            --http.corsdomain '*' \
```
claravanstaden, line 19: Interesting, do you know why this is necessary at this point? 🤔

core/packages/test/scripts/deploy-ethereum.sh
```diff
@@ -52,6 +53,7 @@ start_lodestar() {
             --genesisEth1Hash $genesisHash \
             --params.ALTAIR_FORK_EPOCH 0 \
             --params.BELLATRIX_FORK_EPOCH 0 \
+            --params.CAPELLA_FORK_EPOCH 0 \
```
claravanstaden, line 56: 💪

core/packages/test/scripts/build-binary.sh
```diff
@@ -68,8 +68,22 @@ build_relayer()
...
+        echo "clone go ethereum project to $geth_dir"
+        git clone https://github.com/ethereum/go-ethereum.git $geth_dir
+    fi
+    pushd $geth_dir
+    git checkout tags/$geth_version
+    make geth
+    cp "$geth_dir/build/bin/geth" "$output_bin_dir"
+    popd
+}
+
```
claravanstaden, line 83: Will this re-download and install geth on every `start-services.sh` script run?

parachain/pallets/ethereum-beacon-client/src/lib.rs
```diff
@@ -572,6 +572,10 @@ pub mod pallet {
 			);
 		}
 
+		// todo: process_header will fail with Capella's BeaconBody and need some
+		// change in primitives, since we are also doing heavily refactoring in PR#783 simplify
+		// with execution_payload proof so disable checks here as workaround and will enable
+		// later after synchronized with that PR
```
claravanstaden, line 578: Perfect! 👌

relayer/relays/beacon/header/syncer/syncer.go
```diff
@@ -207,10 +207,10 @@ func (s *Syncer) GetBlockRoots(slot uint64) (scale.BlockRootProof, error) {
 
 	if s.activeSpec == config.Minimal {
 		blockRootsContainer = &state.BlockRootsContainerMinimal{}
-		beaconState = &state.BeaconStateBellatrixMinimal{}
+		beaconState = &state.BeaconStateCapellaMinimal{}
```
claravanstaden, line 210: I wonder if we shouldn't check the fork version here and use Bellatrix or Capella based on the fork version? Even if it's just to see how we will handle fork versions in production where we will need to support at least 2 hardfork versions at once (before upgrade and after upgrade version).

relayer/relays/beacon/state/beacon.go
```diff
@@ -168,6 +173,37 @@ type BeaconStateBellatrixMainnet struct {
...
+	FinalizedCheckpoint          *Checkpoint             `json:"finalized_checkpoint"`
+	InactivityScores             []uint64                `json:"inactivity_scores" ssz-max:"1099511627776"`
+	CurrentSyncCommittee         *SyncCommittee          `json:"current_sync_committee"`
+	NextSyncCommittee            *SyncCommittee          `json:"next_sync_committee"`
+	LatestExecutionPayloadHeader *ExecutionPayloadHeader `json:"latest_execution_payload_header"`
+	NextWithdrawalIndex          uint64                  `json:"next_withdrawal_index,omitempty"`
+	NextWithdrawalValidatorIndex uint64                  `json:"next_withdrawal_validator_index,omitempty"`
+	HistoricalSummaries          []*HistoricalSummary    `json:"historical_summaries,omitempty" ssz-max:"16777216"`
+}
+
```
claravanstaden, line 206: If the only changes are these 3 new fields, you could reuse the `BeaconStateBellatrixMainnet` struct:

```
type BeaconStateCapellaMainnet struct {
    BeaconStateBellatrixMainnet
    NextWithdrawalIndex          uint64                  `json:"next_withdrawal_index,omitempty"`
    NextWithdrawalValidatorIndex uint64                  `json:"next_withdrawal_validator_index,omitempty"`
    HistoricalSummaries          []*HistoricalSummary    `json:"historical_summaries,omitempty" ssz-max:"16777216"`
}
```

relayer/relays/beacon/state/beacon.go
```diff
@@ -259,6 +339,24 @@ type ExecutionPayload struct {
...
+	PrevRandao    [32]byte      `ssz-size:"32" json:"prev_randao"`
+	BlockNumber   uint64        `json:"block_number"`
+	GasLimit      uint64        `json:"gas_limit"`
+	GasUsed       uint64        `json:"gas_used"`
+	Timestamp     uint64        `json:"timestamp"`
+	ExtraData     []byte        `ssz-max:"32" json:"extra_data"`
+	BaseFeePerGas [32]byte      `ssz-size:"32" json:"base_fee_per_gas"`
+	BlockHash     [32]byte      `ssz-size:"32" json:"block_hash"`
+	Transactions  [][]byte      `ssz-max:"1048576,1073741824" ssz-size:"?,?" json:"transactions"`
+	Withdrawals   []*Withdrawal `ssz-max:"16" json:"withdrawals"`
```
claravanstaden, line 357: Similarly as with the `BeaconStateCapellaMainnet` struct, if `Withdrawals` is the only new field we can reuse the `ExecutionPayloadBellatrix` struct and only add the new `Withdrawals` field.

claravanstaden (commented):
Looks great so far! 👏

core/packages/test/scripts/build-binary.sh
alistair-singh: Is this because NIX does not support `v1.11.2` of `geth` in its `go-ethereum` package?

You should be able to install with `go` instead of checking out and building from source:
```cmd
go install github.com/ethereum/go-ethereum/cmd/geth@v1.11.2
```

yrong: no, the script here will ignore download if `go-ethereum` directory exist and `make geth` will ignore recompile if binary exist.

core/packages/test/scripts/deploy-ethereum.sh
yrong: Ah, actually nothing to do with this PR. Sometimes I use `remixd` remote connect to the local network for debug purpose, adding this allow me to do that. 
 Since it is only a dev setup, I would assume should be fine but correct me if I'am wrong.

relayer/relays/beacon/header/syncer/syncer.go
yrong: @claravanstaden Yes, absolute nessessary!   In recent commit I use `ForkEpoch` to check IMO maybe more straightforward  and with that we can even do some fork test with our local setup. I would assume same as use `ForkVersion` but correct me if wrong.

https://github.com/Snowfork/snowbridge/blob/38abe484f88fc11a2c64e653f23772b33a052791/relayer/relays/beacon/header/syncer/syncer.go#L219-L230

core/packages/test/scripts/deploy-ethereum.sh
alistair-singh: Very cool! Please link to a tutorial or docs of how to set this up. I have been copy-pasting solidity back and forth into remix website for debugging. So It would be helpful to my workflow.

yrong: Ok. Actually it's a websocket daemon connect web ide and I've added that in previous PR, just run `pnpm remixd` [here](https://github.com/Snowfork/snowbridge/blob/525126eabc92f2be27ce1b8ab7f580a5b6c01d02/core/package.json#L17) start the daemon and then connect to localhost as [following](https://remix-ide.readthedocs.io/en/latest/remixd.html)

I think other steps are same as what we use remix website as before, just compile code and connect to the local node with `External HTTP Provider`, you can deploy new contract or attach to an exist contract  and do some debug stuff then.

@alistair-singh Hope that is helpful.

relayer/relays/beacon/header/syncer/syncer.go
claravanstaden: Ah, I realized this may be a little confusing. The `spec` actually refers to the environment variables used in the network. So all the testnets (Goerli, Sepolia) and mainnet will use the `mainnet` spec. The minimal spec is only used for the local network with different environment variables like `SLOTS_PER_EPOCH`. So a Goerli specific spec might not be needed. 

I wonder if we shouldn't lookup the fork using `/eth/v1/beacon/states/{state_id}/fork`? Otherwise I guess we can hardcode the fork epoch.

core/packages/test/scripts/deploy-ethereum.sh
claravanstaden: > Ah, actually nothing to do with this PR. Sometimes I use `remixd` remote connect to the local geth node for debug purpose, adding this allow me to do that. Since it is only a dev setup, I would assume should be fine but correct me if I'am wrong.

Awesome, should be fine!

flake.nix
```diff
@@ -40,7 +40,6 @@
 
                     # ethereum
                     foundry-bin
-                    go-ethereum
```
vgeddes, line 43: I suppose this means that the updated version of Geth isn't packaged for nix yet?

relayer/testdata/curl.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
```
alistair-singh, line 1: We should probably name this script to reflect what it does. It looks like it fetches test data so maybe `fetchData.sh`?

relayer/relays/beefy/ethereum-writer.go
```diff
@@ -173,7 +173,7 @@ func (wr *EthereumWriter) submit(ctx context.Context, task Request) error {
 		return fmt.Errorf("transaction SubmitFinal failed (%v),handover (%v)", tx.Hash().Hex(), task.IsHandover)
 	}
 
-	log.WithFields(logrus.Fields{"tx": tx.Hash().Hex(), "handover": task.IsHandover}).Debug("Transaction SubmitFinal succeeded")
+	log.WithFields(logrus.Fields{"tx": tx.Hash().Hex(), "handover": task.IsHandover, "BlockNumber": task.SignedCommitment.Commitment.BlockNumber}).Debug("Transaction SubmitFinal succeeded")
```
alistair-singh, line 176: Camel-case log fields.
```suggestion
	log.WithFields(logrus.Fields{"tx": tx.Hash().Hex(), "handover": task.IsHandover, "blockNumber": task.SignedCommitment.Commitment.BlockNumber}).Debug("Transaction SubmitFinal succeeded")
```

flake.nix
yrong: Exactly, geth with shapella support required v11.x which is not packaged for nix, so we build from source code instead.

doubledup: I'm looking into packaging geth 11.0 and later for nix

vgeddes: Cool. Though for this PR lets continue to build the source code manually.

parachain/pallets/ethereum-beacon-client/src/lib.rs
```diff
@@ -353,6 +364,45 @@ pub mod pallet {
 
 			Ok(())
 		}
+
+		#[pallet::call_index(4)]
+		#[pallet::weight(T::WeightInfo::import_execution_header())]
+		#[transactional]
+		pub fn import_execution_header_capella(
```
vgeddes, line 371: I think it would be cleaner to support only one `import_execution_header()` extrinsic, instead of one for each version.

Then we can pass the following header type to `import_execution_header(origin, header)`:

```
enum VersionedHeaderUpdate {
    Capella(HeaderUpdateOfCapella)
}
```

And then the extrinsic can inspect `VersionedHeaderUpdate` to know which version needs to be processed.

parachain/pallets/ethereum-beacon-client/src/lib.rs
```diff
@@ -43,6 +44,14 @@ pub type HeaderUpdateOf<T> = HeaderUpdate<
 	<T as Config>::MaxProofBranchSize,
 	<T as Config>::MaxSyncCommitteeSize,
 >;
+pub type HeaderUpdateOfCapella<T> = HeaderUpdateCapella<
```
vgeddes, line 47: The `ThingOf<T>` pattern is always like this:

```suggestion
pub type HeaderUpdateCapellaOf<T> = HeaderUpdateCapella<
```

Because `Of` refers to the generic parameter `T`

parachain/primitives/beacon/src/lib.rs
```diff
@@ -268,6 +269,54 @@ pub struct HeaderUpdate<
...
+	ProofSize,
+	ProposerSlashingSize,
+	AttesterSlashingSize,
+	VoluntaryExitSize,
+	AttestationSize,
+	ValidatorCommitteeSize,
+	SyncCommitteeSize
+))]
+#[codec(mel_bound())]
+pub struct HeaderUpdateCapella<
```
vgeddes, line 303: Instead of creating a new `HeaderUpdate*` type for each new execution header format, can't we add the execution header type as a generic parameter?

I think this reduce code duplication in other parts of this PR.

parachain/pallets/ethereum-beacon-client/src/merkleization.rs
```diff
@@ -121,6 +156,18 @@ pub fn hash_tree_root_execution_header<
 	hash_tree_root(ssz_execution_payload)
 }
 
+pub fn hash_tree_root_execution_header_capella<
```
vgeddes, line 159: I also feel like we can have one version of `hash_tree_root_execution_header` that takes the execution header type as a generic parameter.

parachain/pallets/ethereum-beacon-client/src/lib.rs
```diff
@@ -303,41 +336,51 @@ pub mod pallet {
...
-					target: "ethereum-beacon-client",
-					"💫 Header update failed with error {:?}",
-					err
-				);
-				return Err(err)
+			match versioned_update {
+				VersionedHeaderUpdate::Bellatrix(update) => {
+					slot = update.beacon_header.slot;
+					block_hash = update.execution_header.block_hash;
+					result = Self::process_header(update);
```
vgeddes, line 357: There is a lot of duplicated code in `process_header` and `process_header_capella`. I think we can make better use of generics to avoid this.

vgeddes: For example, I analysed the diff between those two functions, and they differ by only 1 statement:

```diff
-    let execution_root = merkleization::hash_tree_root_execution_header(execution_payload.clone())
-        .map_err(|_| Error::<T>::BlockBodyHashTreeRootFailed)?;
+    let execution_root =
+        merkleization::hash_tree_root_execution_header_capella(execution_payload.clone())
+            .map_err(|_| Error::<T>::BlockBodyHashTreeRootFailed)?;
+
```

vgeddes (commented):
I took a deeper look at this and I think the current approach to versioning adds a lot of duplicated code.

Is there a way we can use generics to mitigate this problem?

parachain/pallets/ethereum-beacon-client/src/lib.rs
yrong: Yes, bad smell here. Thanks for point out and I'll do some refactoring accordingly.

yrong: @vgeddes some refactoring in commit
https://github.com/Snowfork/snowbridge/pull/786/commits/fff98fc2c968ffac3b23936c6cb4bb672af898d3

yrong:
@claravanstaden  After some sync with Vincent offline, upgrade from Bellatrix is not nessessary for us and the previous implementation with versioning support is somehow incomplete. To make things simple I change scope of this PR focus on Capella support only.

cvanstaden-godaddy:
Hey @yrong! That sounds good, I think keeping it simple is the correct way to go. However, even if we don't do the versioning now we still need to remember that it will be necessary for future prod cases when a hardfork is released.

claravanstaden:
Hey @yrong! That sounds good, I think keeping it simple is the correct way to go. However, even if we don't do the versioning now we still need to remember that it will be necessary for future prod cases when a hardfork is released.

parachain/primitives/beacon/src/lib.rs
```diff
@@ -385,7 +408,7 @@ pub struct SyncAggregate<SyncCommitteeSize: Get<u32>, SignatureSize: Get<u32>> {
 )]
 #[scale_info(skip_type_params(FeeRecipientSize, LogsBloomSize, ExtraDataSize))]
 #[codec(mel_bound())]
-pub struct ExecutionPayload<
+pub struct ExecutionPayloadHeaderCapella<
```
vgeddes, line 411: The name `ExecutionPayloadHeaderCapella` is a bit verbose. The `Capella` isn't necessary, for the reasons we spoke of offline.

```suggestion
pub struct ExecutionPayloadHeader<
```

parachain/primitives/beacon/src/lib.rs
```diff
@@ -260,7 +254,8 @@ pub struct HeaderUpdate<
 	SyncCommitteeSize: Get<u32>,
 > {
 	pub beacon_header: BeaconHeader,
-	pub execution_header: ExecutionPayload<FeeRecipientSize, LogsBloomSize, ExtraDataSize>,
+	pub execution_header:
+		ExecutionPayloadHeaderCapella<FeeRecipientSize, LogsBloomSize, ExtraDataSize>,
```
vgeddes, line 258:
```suggestion
		ExecutionPayloadHeader<FeeRecipientSize, LogsBloomSize, ExtraDataSize>,
```

parachain/pallets/ethereum-beacon-client/src/merkleization.rs
```diff
@@ -37,14 +38,19 @@ impl From<DeserializeError> for MerkleizationError {
 }
 
 impl<FeeRecipientSize: Get<u32>, LogsBloomSize: Get<u32>, ExtraDataSize: Get<u32>>
-	TryFrom<ExecutionPayload<FeeRecipientSize, LogsBloomSize, ExtraDataSize>> for SSZExecutionPayload
+	TryFrom<ExecutionPayloadHeaderCapella<FeeRecipientSize, LogsBloomSize, ExtraDataSize>>
```
vgeddes, line 41:
```suggestion
	TryFrom<ExecutionPayloadHeader<FeeRecipientSize, LogsBloomSize, ExtraDataSize>>
```

parachain/pallets/ethereum-beacon-client/src/ssz.rs
```diff
@@ -40,7 +40,7 @@ pub struct SSZSigningData {
 }
 
 #[derive(Default, SimpleSerialize, Clone, Debug)]
-pub struct SSZExecutionPayload {
+pub struct SSZExecutionPayloadHeaderCapella {
```
vgeddes, line 43:
```suggestion
pub struct SSZExecutionPayloadHeader {
```

vgeddes (commented):
Looks much better 🙌 

Just need to remove the unnecessary `Capella` from some struct names, please.

vgeddes (approved):

alistair-singh (approved):
🚢

parachain/pallets/ethereum-beacon-client/src/merkleization.rs
```diff
@@ -4,7 +4,7 @@ use crate::{config, ssz::*};
 use byte_slice_cast::AsByteSlice;
 use frame_support::{traits::Get, BoundedVec};
 use snowbridge_beacon_primitives::{
-	BeaconHeader, ExecutionPayload, ForkData, SigningData, SyncAggregate, SyncCommittee,
+	BeaconHeader, ExecutionPayloadHeader, ForkData, SigningData, SyncAggregate, SyncCommittee,
```
claravanstaden, line 7: Nit: This should probably just be `ExecutionHeader`. It is either `ExecutionPayload` (containing all the transactions) or `ExecutionHeader` containing the transactions root.

parachain/runtime/snowbase/src/lib.rs
```diff
@@ -639,7 +641,7 @@ impl ethereum_beacon_client::Config for Runtime {
 	type MaxFinalizedHeaderSlotArray = MaxFinalizedHeaderSlotArray;
 	type ForkVersions = ChainForkVersions;
 	type WeakSubjectivityPeriodSeconds = WeakSubjectivityPeriodSeconds;
-	type WeightInfo = weights::ethereum_beacon_client::SnowbridgeWeight<Self>;
+	type WeightInfo = ethereum_beacon_client::weights::SnowbridgeWeight<Self>;
```
claravanstaden, line 644: The weights used here should be snowbase/src/weights/ethereum-beacon-client (which looks like it is being deleted). Looks like this changes the weights to use the pallet weights. Why this change?

relayer/relays/beacon/config/config.go
```diff
@@ -56,10 +58,20 @@ type ActiveSpec string
 const (
 	Mainnet ActiveSpec = "mainnet"
 	Minimal ActiveSpec = "minimal"
+	GOERLI  ActiveSpec = "goerli"
```
claravanstaden, line 61: Nit: Should this not be `Goerli`? As far as I know Go constants' convention is still camel case.

relayer/relays/beacon/config/config.go
```diff
@@ -56,10 +58,20 @@ type ActiveSpec string
...
+	Minimal_CapellaForkEpoch uint64 = 0
+	Goerli_CapellaForkEpoch  uint64 = 162304
+	// TODO: change when timeline of Capella upgrade in mainnet finalized
+	Mainnet_CapellaForkEpoch uint64 = math.MaxUint64
 )
 
 func (c Config) GetActiveSpec() ActiveSpec {
 	switch c.Source.Beacon.ActiveSpec {
+	case string(GOERLI):
+		return GOERLI
```
claravanstaden, line 74: I'm also wondering why we need to add Goerli as a spec if Goerli's network config is the same as mainnet?

claravanstaden (commented):
Looks great @yrong! 🥳

parachain/runtime/snowbase/src/lib.rs
yrong: Yes, but seems we're still missing benchmarked weights for runtime [snowblink](https://github.com/Snowfork/snowbridge/blob/c3bd614dffad1dae5996b8b67e6cfd3023669111/parachain/runtime/snowblink/src/lib.rs#L640) and [snowbridge](https://github.com/Snowfork/snowbridge/blob/c3bd614dffad1dae5996b8b67e6cfd3023669111/parachain/runtime/snowbridge/src/lib.rs#L640) which I would assume more important than `snowbase` the test runtime. 

And also as benchmark required, we need to do it in a specific hardware same as deploy in production which I'm not sure finalized yet.

So I would prefer we make things easy as now and rebenchmark it in a more strict and consist manner.

relayer/relays/beacon/config/config.go
yrong: sorry I missed your comment [before](`https://github.com/Snowfork/snowbridge/pull/786#discussion_r1131957837`).
Yes, we can find a better way to config `ForkingEpoch` for respective network without polluting `ActiveSpec` here.

parachain/pallets/ethereum-beacon-client/src/merkleization.rs
yrong: Yes, the previous name `ExectionPayLoad` is a little confusing which actually contains `transactions_root` [here](https://github.com/Snowfork/snowbridge/blob/c3bd614dffad1dae5996b8b67e6cfd3023669111/parachain/primitives/beacon/src/lib.rs#L410) so I rename it as `ExectionPayLoadHeader`. I would assume after #783 we do not need `ExecutionPayload`  (containing all the transactions) any more but correct me if I'am wrong.

And about `ExecutionHeader`, AFAIK not all fields in `ExectionPayLoadHeader` are required for proof verification. e.g only `receipts_root` is required [here](https://github.com/Snowfork/snowbridge/blob/f5dc322860015c6e711c82e53a273d9b9d6764c2/parachain/pallets/ethereum-beacon-client/src/lib.rs#L1129) for execution layer message, so I just simplify [ExecutionHeader](https://github.com/Snowfork/snowbridge/blob/f5dc322860015c6e711c82e53a273d9b9d6764c2/parachain/primitives/beacon/src/lib.rs#L291) as a subset of `ExectionPayLoadHeader` and make it an version insensitive/agnostic data structure.

relayer/relays/beacon/config/config.go
yrong: https://github.com/Snowfork/snowbridge/pull/786/commits/39a9d7ede9f7cfa25afbc848042da08cf98dc783

parachain/pallets/ethereum-beacon-client/src/ssz.rs
yrong: @vgeddes @claravanstaden Though I remove the versioned suffix in this PR I'am still a little hesitate about the idea reuse same `SSZExecutionPayload` for multiple verions with some `Option` field. 

The main reason is that we use some ssz libraries like [fastssz](https://github.com/ferranbt/fastssz) in relayer and  [ssz-rs](https://github.com/ralexstokes/ssz-rs) in parachain to generate ssz-binding for encoding&merkelization.

Not sure these libraies will handle `Option` type correctly as expected. Though checked with ralexstokes,author of `ssz-rs` should be fine for it. But There is no `Option` type I would assume use `Pointer` type instead in golang. But [sszgen](https://github.com/Snowfork/snowbridge/blob/a0f32d4f1079510e769257bfef4d907798e3e04a/relayer/README.md?plain=1#L49) can not regonize it and will just fail.

so I'll prefer explicitly add exclusive versioned structures without the `Option` workaround, that is also the normal way most light client like [prysm](https://github.com/prysmaticlabs/prysm/blob/797cc360c77c650751499e375da2c96be9b32b69/api/client/builder/types.go#L467) does, even means some duplication introduced but more straightforward for me.


Anyway, it's not the scope of this PR, I've archived the previous [Impl](https://github.com/Snowfork/snowbridge/tree/ron/versioning-capella-archive), could be still worth to check when we gonna implement versioning support later.

vgeddes: This makes sense Ron. Thanks for investigating this 🙌

parachain/pallets/ethereum-beacon-client/src/merkleization.rs
vgeddes: > And about ExecutionHeader, AFAIK not all fields in ExectionPayLoadHeader are required for proof verification. e.g only receipts_root is required [here](https://github.com/Snowfork/snowbridge/blob/f5dc322860015c6e711c82e53a273d9b9d6764c2/parachain/pallets/ethereum-beacon-client/src/lib.rs#L1129) for execution layer message, so in this PR I just simplify [ExecutionHeader](https://github.com/Snowfork/snowbridge/blob/f5dc322860015c6e711c82e53a273d9b9d6764c2/parachain/primitives/beacon/src/lib.rs#L291) as a subset of ExectionPayLoadHeader and make it an version insensitive/agnostic data structure.

I think this is a good optimization to have, as it will cut down on storage costs in the BridgeHub parachain. So I support it.

parachain/runtime/snowbase/src/lib.rs
vgeddes: Sounds good to me.

vgeddes (approved):

parachain/pallets/ethereum-beacon-client/src/ssz.rs
claravanstaden: > So I'll prefer explicitly add exclusive versioned structures without the Option workaround, that is also the normal way most light client like [prysm](https://github.com/prysmaticlabs/prysm/blob/797cc360c77c650751499e375da2c96be9b32b69/api/client/builder/types.go#L467) does, even means some duplication introduced but more straightforward.

I agree! Thanks @yrong!

parachain/pallets/ethereum-beacon-client/src/merkleization.rs
claravanstaden: Agreed, thanks @yrong!

parachain/runtime/snowbase/src/lib.rs
claravanstaden: Makes sense, thank you!

claravanstaden (approved):
Good to go! 🚀

PR #791, opened 2023-03-23
Reworked pallets for XCMv3

vgeddes:
Changes:
* Channels renamed to Queues
  * OutboundChannel -> OutboundQueue
  * InboundChannel -> InboundQueue
* Inbound and Outbound queue pallets are now in separate crates 
* Nonces or "Lanes" in queues are know identified by parachain ID rather than parachain MultiLocation
* In the solidity unit tests, `TestToken` replaced with WETH9 (Ether wrapped in an ERC20 token).

I also disabled the parachain message relayer. It still needs to be updated to be compatible with these refactorings over the past month.

The ethereum execution message relayer is operational, but still needs more work to be production ready.

Fixes: SNO-396, SNO-402


alistair-singh (approved):
LGTM!

parachain/pallets/inbound-queue/src/lib.rs
```diff
@@ -36,11 +51,12 @@ pub mod pallet {
...
 		type Verifier: Verifier;
 
-		/// Verifier module for message verification.
-		type MessageDispatch: MessageDispatch<Self, MessageId>;
+		type Token: Mutate<Self::AccountId>;
 
 		/// Weight information for extrinsics in this pallet
 		type WeightInfo: WeightInfo;
+
+		type Reward: Get<BalanceOf<Self>>;
```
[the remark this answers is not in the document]
vgeddes: Nice catch, added ticket in https://linear.app/snowfork/issue/SNO-402

.husky/pre-commit
```diff
@@ -9,7 +9,7 @@ echo 'Running pre-commit hook...'
 chronic typos .
 
 # lint and format for core contracts and typescript codes
-(cd core && chronic pnpm lint && chronic pnpm format)
+(cd core && chronic pnpm lint)
```
yrong, line 12: is there anything break we remove pnpm format here? seems will just run `forge fmt` so should be fine?

_typos.toml
```diff
@@ -7,5 +7,7 @@ extend-exclude = [
     "**/*ffi*.js",
     "**/*mock*.rs",
     "**/*data*.rs",
-    "cumulus/**"
+    "cumulus/**",
+    "smoketest/src/parachains",
+    "smoketest/src/contracts",
```
yrong, line 12: Do we have these smoketest scripts ready? can't find in this PR.

parachain/pallets/inbound-queue/src/lib.rs
```diff
@@ -117,10 +137,9 @@ pub mod pallet {
 				}
 			})?;
 
-			let message_id = MessageId::new(envelope.account, envelope.nonce);
-			T::MessageDispatch::dispatch(envelope.source, message_id, &envelope.payload);
-
-			<LatestVerifiedBlockNumber<T>>::set(block_number);
+			// Reward relayer from the sovereign account of the destination parachain
+			let dest_account = envelope.dest.into_account_truncating();
+			T::Token::transfer(&who, &dest_account, T::Reward::get(), Preservation::Expendable)?;
```
yrong, line 142: Seems a little confused. Is the transfer logic here correct or just some temp placeholder?  Will the impl here supposed to be a xcm call sent to asset chain?

relayer/relays/execution/main.go
```diff
@@ -50,34 +56,97 @@ func (r *Relay) Start(ctx context.Context, eg *errgroup.Group) error {
...
+	messages := make(chan *contracts.OutboundQueueMessage)
+
+	key, err := types.EncodeToBytes(r.config.Source.LaneID)
 	if err != nil {
-		return err
+		return fmt.Errorf("encode to bytes: %w", err)
 	}
 
-	err = messages.Sync(ctx, eg)
+	sub, err := contract.WatchMessage(&opts, messages, [][]byte{key})
```
yrong, line 83: So to confirm, `WatchMessage` will only subscribe to the recent message events and we still need some scanner logic to find messages with missing nonces?

parachain/pallets/inbound-queue/src/lib.rs
```diff
@@ -117,10 +137,9 @@ pub mod pallet {
...
```
yrong, line 143: Is `LatestVerifiedBlockNumber` safe to remove here? I would assume we still need some scanner logic to find the unsynced messages and it is useful to narrow the scan scope down.

parachain/pallets/inbound-queue/src/lib.rs
yrong: And also seems we remove the previous logic `<LatestVerifiedBlockNumber<T>>::set(block_number);` Is it safe to remove? I would assume we still need some scanner logic to find the unsynced messages and it is useful to narrow the scan scope down.

relayer/relays/execution/main.go
vgeddes: yeah, this is just a very simple, incomplete implementation that allows us to test the on-chain code. We will need to rewrite it in a future PR.

parachain/pallets/inbound-queue/src/lib.rs
vgeddes: > Seems a little confused. Is the transfer logic here correct or just some temp placeholder? Is impl supposed to be some xcm call sent to asset chain?

That's just for rewarding the relayer. We definitely still need to implement the routing of the XCM call.

vgeddes: I temporarily removed the `LatestVerifiedBlockNumber` logic because the checkpoint logic in the relayer didn't work with multiple lanes (i.e each parachain has its lane and nonce).

So I am may add it back once I rewrite the relayer properly.

_typos.toml
vgeddes: I experimented with writing smoketest scripts in Rust, but deleted the experiment as they were too slow to compile, and Rust's test frameworks are not as good as Golang's.

But see this in case we ever want to generate Rust clients for our pallets and contracts: https://www.notion.so/snowfork/Rust-bindings-b6566f13dd894ac1ac9a347f9bbbf160?pvs=4

relayer/relays/execution/main.go
vgeddes: Also we may have to modify the execution message relayer to support the optimizations we talked about on slack (execution headers, etc).

yrong: Agree.  And not sure if `subscribe` will work here because we can not verify the execution block if it is not finized in beacon chain yet. so maybe the previous filter with `FilterOpts`(i.e. a bounded range) make more sense, the start of the range should be the `LatestVerifiedBlockNumber` in message layer(we remove it in this pr) and the end is the `LastestFinalized` in beacon layer?

.husky/pre-commit
vgeddes: I found it a bit confusing for a pre-commit hook to alter the code automatically. I disabled it for that reason.

I would rather we enable "Format on Save" option in VS code. For Solidity:
```
{
		"[solidity]": {
			"editor.formatOnSave": true,
			"editor.defaultFormatter": "JuanBlanco.solidity" 
		},
		"solidity.formatter": "forge",
}
```

relayer/relays/execution/main.go
vgeddes: I check here that the execution header has already been imported: https://github.com/Snowfork/snowbridge/pull/791/files#diff-ba222152cf61a695305ad44e12ccc757ffa95a1f6c1c8ee698e21a7650155229R102

But yes, when we re-implement the relayer properly, we will likely use FilterMessage along with the following algorithm:
1. Check message nonce on Parachain side
2. Check message nonce on Ethereum side
3. Compare nonces to determine which messages need to be synced.
4. Scan Ethereum blocks using `FilterMessage` to find the `OutboundQueue.Message` event logs. This is fast, as Ethereum indexes events, so we can almost lookup event logs by nonce: https://github.com/Snowfork/snowbridge/blob/2bf31bfbef0d1d3092533668017c3ad31aa8c680/core/packages/contracts/src/IOutboundQueue.sol#L7

Then as you say, we will use `LatestExecutionHeaderState` to check that execution header has been imported first.

PR #792, opened 2023-03-24
SNO-328

yrong:
companion with
https://github.com/Snowfork/cumulus/tree/dev/snowbridge-v0.9.38


parachain/primitives/beacon/src/lib.rs
```diff
@@ -125,7 +125,6 @@ impl<'de> Deserialize<'de> for PublicKey {
 
 #[derive(Default, Encode, Decode, TypeInfo, MaxEncodedLen)]
 pub struct ExecutionHeaderState {
-	pub beacon_block_root: H256,
```
yrong, line 128: @claravanstaden  Seems this field is not in use yet?  IMHO maybe some redundancy with `beacon_slot`, we can still retrieve from beacon-api later if nessessary. For saving runtime storage I remove it but correct me if I'm wrong.

.gitmodules
```diff
@@ -9,3 +9,6 @@
 	path = core/packages/contracts/lib/openzeppelin-contracts
 	url = https://github.com/openzeppelin/openzeppelin-contracts
 	branch = v4.8.2
+[submodule "cumulus"]
+	path = cumulus
+	url = https://github.com/Snowfork/cumulus
```
yrong, line 14: @alistair-singh 
In this PR I change the cumulus integration as submodule here and then in this [commit](https://github.com/Snowfork/cumulus/commit/4ae67639c2e1ebe1e048432ab71ed074bdc0530d) make snowbridge as local dependency. 
IMHO could make more sense for dev workflow when we're doing heavily refactoring at this moment. Please let me know what you think.

alistair-singh: I am not opposed to this change. I am using vscode to develop cumulus and snowbridge by having them in the same vscode workspace. So this change will not affect me very much. Let's see what the rest of the team says.

PR #793, opened 2023-03-24
Link cumulus as submodule

yrong:
Companion with https://github.com/Snowfork/cumulus/pull/4


alistair-singh (approved):
Our README might be out of date until.

See: https://github.com/Snowfork/snowbridge/blob/main/core/packages/test/README.md#using-a-custom-cumulus-build-for-bridgehub-and-statemine

We might want to remove that section for now and add it back later as with these changes it will not respect the CUMULUS_BIN parameter if set.

.gitmodules
```diff
@@ -9,3 +9,6 @@
 	path = core/packages/contracts/lib/openzeppelin-contracts
 	url = https://github.com/openzeppelin/openzeppelin-contracts
 	branch = v4.8.2
+[submodule "cumulus"]
+	path = cumulus
+	url = https://github.com/Snowfork/cumulus
```
vgeddes, line 14: Don't we need to specify the branch too?

core/packages/test/scripts/build-binary.sh
```diff
@@ -30,6 +30,18 @@ rebuild_cumulus(){
     popd
 }
 
+build_cumulus_from_source(){
+    cumulus_src=$root_dir/cumulus
+    if [ ! -d "$cumulus_src" ] ; then
+        git clone https://github.com/Snowfork/cumulus $cumulus_src
+    fi
+    pushd $cumulus_src
+    git fetch origin && git switch $cumulus_version
```
vgeddes, line 39: Aren't these steps typically performed using either `git submodule init` or `git submodule update`? Especially seeing as the branch name should already be defined in `.gitmodules`?

.gitmodules
yrong: I think we do not need,  PR here already contains matched [commit](https://github.com/Snowfork/cumulus/tree/07296e8f0f3bcaabd11aeb6e99a38e08ecbb1da2) in cumulus.

cumulus
yrong, line 1: matched cumulus here

vgeddes (approved):

PR #794, opened 2023-03-27
sno-328

yrong:
Companion with https://github.com/Snowfork/cumulus/pull/5


parachain/pallets/ethereum-beacon-client/src/lib.rs
```diff
@@ -155,6 +163,9 @@ pub mod pallet {
 		InvalidExecutionHeaderUpdate,
 		FinalizedBeaconHeaderSlotsExceeded,
 		ExecutionHeaderMappingFailed,
+		BoundedFinalityHeaderExceed,
+		BoundedExecutionHeaderExceed,
+		BoundedSyncCommitteeExceed,
```
alistair-singh, line 168:
```suggestion
		FinalityHeaderBoundsExceeded,
		ExecutionHeaderBoundsExceeded,
		SyncCommitteeBoundsExceeded,
```

parachain/pallets/ethereum-beacon-client/src/mock.rs
```diff
@@ -97,6 +97,8 @@ pub mod mock_minimal {
 				epoch: 0,
 			},
 		};
+		pub const MaxSyncCommitteeCachedSize: u32 = 256;
+		pub const MaxExecutionHeaderCachedSize: u32 = 8192;
```
alistair-singh, line 101: These values seem too large for mock configuration. I would expect really small numbers so that it is easy to test cases where the bounds are exceeded.

parachain/pallets/ethereum-beacon-client/src/tests_minimal.rs
```diff
@@ -217,6 +219,52 @@ mod beacon_minimal_tests {
 		});
 	}
 
+	#[test]
+	fn it_processes_pruning_finalized_header_update() {
+		let initial_sync = get_initial_sync::<mock_minimal::Test>();
```
alistair-singh, line 224: I am not sure if this is possible but maybe we should drop the bounds configured `mock_minimal::Test` to small amounts. And then just import some test data to reproduce states.

For instance, set `MaxExecutionHeaderCachedSize` to 2 and then try to import 3 headers. The same for `MaxSyncCommitteeCachedSize`. Unless it is too hard to get into those states.

yrong: Drop all bounds to 1 and all tests still passed as expected. Consider we'll import `initialSync`  so we have 2 headers to cover the case you mentioned above. 

Btw: actually in this test case we'll fill `FinalizedBeaconHeadersList` to `MaxBounds` first before `import_finalized_header` as following, so will also cover the case even with a big value.

https://github.com/Snowfork/snowbridge/blob/92e5f0f7478fe6c473d69a1ed329fe026df80ca2/parachain/pallets/ethereum-beacon-client/src/tests_minimal.rs#L242-L254

yrong: Drop all bounds and more tests: https://github.com/Snowfork/snowbridge/pull/794/commits/f6de4ca9df00c18f80464166749cfe32486a2797

parachain/pallets/ethereum-beacon-client/src/lib.rs
```diff
@@ -155,6 +163,9 @@ pub mod pallet {
 		InvalidExecutionHeaderUpdate,
 		FinalizedBeaconHeaderSlotsExceeded,
 		ExecutionHeaderMappingFailed,
+		FinalityHeaderBoundsExceed,
+		ExecutionHeaderBoundsExceed,
+		SyncCommitteeBoundsExceed,
```
alistair-singh, line 168: These need an 'ed' at the end.
```suggestion
		FinalityHeaderBoundsExceeded,
		ExecutionHeaderBoundsExceeded,
		SyncCommitteeBoundsExceeded,
```

alistair-singh (commented):
Looks good! Just a small typo made. I will approve once fixed.

parachain/pallets/ethereum-beacon-client/src/lib.rs
```diff
@@ -916,14 +973,27 @@ pub mod pallet {
...
 			beacon_slot: u64,
 			beacon_block_root: H256,
-		) {
+		) -> DispatchResult {
 			let block_number = header.block_number;
-
-			<ExecutionHeaders<T>>::insert(block_hash, header);
+			let block_hash = header.block_hash;
+
+			<ExecutionHeadersList<T>>::try_mutate(|vec| {
```
vgeddes, line 983: This is my understanding of the algorithm you're using:
1. Store a list of cached header hashes in a queue
2. Everytime a new header is added,
  a. Read the entire queue from storage
  b. Delete the header at the front of the queue
  c. Write the entire queue back to storage

I think the problem here is the performance overhead of writing and reading this queue. For example, if we want to cache headers for the last 3 weeks, the size of this queue is going to be 4MiB:
```
# assuming beacon block produced every 12 secs
((60 * 60 * 24 * 21) / 12) * 32
```

So this operation could use up all the block weight.

I want you to try and find an algorithm that has better performance. Ie O(1) or O(log2 N), where N is the number of cached items.

One idea is to make pruning the responsibility of an offchain worker:
1. Every N blocks:
2. The worker reads all headers from the `ExecutionHeaders` storage item in an offchain context, and finds the oldest one.
3. It calls the Light client extrinsic `prune_execution_header(header_hash: H256)`
4. The light verifies that the candidate is old enough to prune (ie. its block number is differs in height from the most recently finalized header, by certain margin).

yrong: @vgeddes Thanks for the review and yes that's what current implementation is. Your propsal here definitely a better way to do the pruning but I'am still a little hesitate to do so. 

The main reason is somehow related to [SNO-356](https://linear.app/snowfork/issue/SNO-356) in which @claravanstaden put forward a better way to sync execution header, instead of relaying every block we relay execution block which contains unprocessed messages only. My gut feeling though some impact on current implementation it is still an improvement worth checking again which will not only save much more weight consuming but also simplify workflow a bit.

There is also another reason commented [here](https://linear.app/snowfork/issue/SNO-356#comment-0be45547) why I would like to prioritize it.

Assume on demand relay is done then I would suspect we do not need to cache such a long list of execution headers(e.g. 3 weeks as you mentioned above). If so the pruning improvement seems not that important so I would prefer we keep things simple at this moment.

Please let me know what you think.

yrong: Btw, notice that you address the importance of BLS verification in the thread with some convincing benchmark result. I totally agree with that BLS is another important aspect we need to pay more attention to.  But I would assume the BLS improvement focus on single update optimization while relay on demand focus on the whole so IMO there is no conflict in both improvements.

vgeddes: Yeah I like the optimization Clara suggested. I think we should prioritize [SNO-356](https://linear.app/snowfork/issue/SNO-356) and then come back to this pruning work

yrong: Or we can use ringbuffer storage instead without pruning.

PR #795, opened 2023-03-29
For sno-405

yrong:
Companion with https://github.com/Snowfork/cumulus/pull/8


claravanstaden:
Hey @yrong! You are correct, in practice the weak subjectivity period is actually much longer than 27 hours. However, if I understand it correctly, that in the event where a massive number of validators get slashed, the subjectivity period could become shorter than the subjectivity period that we have set. Perhaps I also overemphasised the importance of [this statement](https://notes.ethereum.org/@adiasg/weak-subjectvity-eth2#:~:text=The%20maximum%20safe%20weak%20subjectivity%20period%20is%20always%20at%20least%2027%20hours).

I see that the article also takes into account a safety decay and the 2 weeks that you have set the weak subjectivity period is for about 200k validators, and on mainnet there are about 560k validators, so perhaps 2 weeks is a reasonable setting.

claravanstaden:
With regards to the backfill, I suspect this will still be necessary, because if we don't do the backfill for the missing sync committee periods and beacon headers, we will not be able to verify messages that occurred in the offline period. But will defer to @vgeddes here if it needs to be done now. 😄

yrong:
> With regards to the backfill, I suspect this will still be necessary, because if we don't do the backfill for the missing sync committee periods and beacon headers, we will not be able to verify messages that occurred in the offline period.

@claravanstaden Many thanks for the review! One misunderstanding I would like to address is that `backfill` what I refer to here is the reverse sync mechanism you described in notion page,  to make things easy at this moment what I would prefer is adding missing sync committee periods through unblock call directly as following.
https://github.com/Snowfork/snowbridge/blob/218cf483c365df8bbd05b1846881366bcb2ba57e/parachain/pallets/ethereum-beacon-client/src/lib.rs#L392-L410

Based on `Weak-Subjectivity` is extended to 14 days, then actually there is only a small chance that we need to do such backfill recovery.  For normal cases AFAIK our beacon relayer will just recover smoothly from breakpoint without intervention but correct me if I'm wrong.

Yes,  I'll wait to see more comments from team.

claravanstaden:
@yrong I see! Sure, that makes sense. I guess we'd just have to get an easy way to provide all the checkpoints to governance through a script or something, but that is certainly more simple.

> Based on WeakSubjectivityPeriod is extended to 14 days, then actually there is only a small chance that we need to do such backfill recovery. For normal cases AFAIK our beacon relayer will just recover smoothly from breakpoint without intervention but correct me if I'm wrong.

Totally agree. 👌

parachain/pallets/ethereum-beacon-client/src/lib.rs
```diff
@@ -947,13 +967,16 @@ pub mod pallet {
...
 
 			<FinalizedBeaconHeaders<T>>::insert(block_root, header);
-			<FinalizedBeaconHeadersBlockRoot<T>>::insert(block_root, block_roots_root);
 			LatestFinalizedHeaderState::<T>::mutate(|s| {
 				s.import_time = T::TimeProvider::now().as_secs();
 				s.beacon_block_root = block_root;
 				s.beacon_slot = slot;
 			});
 
+			if block_roots_root.is_some() {
```
claravanstaden, line 976: Nice! ✨

claravanstaden (commented):
Awesome!

core/packages/test/scripts/fetch-beacon-data.sh
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/env bash
 set -eu
+# export BEACON_HTTP_ENDPOINT=https://lodestar-goerli.chainsafe.io
```
alistair-singh, line 3: Remove commented code.

parachain/pallets/ethereum-beacon-client/src/lib.rs
```diff
@@ -342,51 +349,69 @@ pub mod pallet {
...
+		pub fn unblock_bridge(
+			origin: OriginFor<T>,
+			check_points: Option<Vec<CheckpointSyncOf<T>>>,
+		) -> DispatchResult {
 			let _sender = ensure_root(origin)?;
 
+			log::info!(target: "ethereum-beacon-client","💫 unblock bridge from governance with provided checkpoints.");
+			if check_points.is_some() {
+				for check_point in check_points.unwrap() {
+					Self::process_checkpoint_sync(check_point)?;
```
alistair-singh, line 378: `process_checkpoint_sync` uses `verify_sync_committee` to verify checkpoint. My understanding is that it validates that the provided SyncCommittee exists within the beacon state at the header specified.

We provide a list of CheckPoints in `unblock_bridge` to process in order. It looks like we validate each checkpoint independently. We do not explicitly check that all these checkpoints lie on the canonical chain or any SyncCommitee handovers. Is this correct?

Will this be handled by governance in the sense that the community would make sure the checkpoints are correct before voting on a proposal that would enact the unblocking?

alistair-singh (commented):
Looks good!

parachain/pallets/ethereum-beacon-client/src/lib.rs
yrong: Exactly!

parachain/pallets/ethereum-beacon-client/src/lib.rs
```diff
@@ -342,51 +349,69 @@ pub mod pallet {
...
+			Ok(())
+		}
+
+		#[pallet::call_index(4)]
+		#[pallet::weight(10_000_000_000)]
 		#[transactional]
-		pub fn unblock_bridge(origin: OriginFor<T>) -> DispatchResult {
+		pub fn unblock_bridge(
+			origin: OriginFor<T>,
+			check_points: Option<Vec<CheckpointSyncOf<T>>>,
```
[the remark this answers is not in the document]
yrong: I would assume not nessessary all in in one extrinsic, multiple calls each with partial of the missing data should also be fine. 
And as mentioned above, with some monitoring components running(not scope of this PR) out-of-sync can be easily checked so that we can respond in time. For normal cases our beacon relayer will just recover smoothly from breakpoint without the `backfill`  here.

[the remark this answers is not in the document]
alistair-singh: Will multiple calls mean multiple governance proposals?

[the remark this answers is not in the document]
yrong: Probably. Not sure if one governance proposal which contains multiple calls(like `batch_all` but can be scheduled in multiple blocks) is possible or we'll need multiple proposals. And based on improvements introduced in [Gov2](https://medium.com/polkadot-network/gov2-polkadots-next-generation-of-decentralised-governance-4d9ef657d11b) multiple propsals/referendums can be fast-tracked and run simultaneously I would suspect it's not a big problem.

alistair-singh (approved):
🚀

parachain/pallets/ethereum-beacon-client/src/lib.rs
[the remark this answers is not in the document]
vgeddes: I have an idea: The governance proposal would include the _hashes_ of the checkpoints.

The proposal would be the following extrinsic `begin_recovery`:

```rust
fn begin_recovery(checkpoints: Vec<CheckPointHash>);
```

Then any user can permissionlessly sync individual checkpoints, as long as they are in the list above:
```rust
fn sync_recovery_checkpoint(checkpoint: CheckPoint);
```

Once all checkpoints have been synced,  call this extrinsic to unblock the bridge:
```rust
fn complete_recovery();
```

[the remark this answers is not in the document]
yrong: Cool, definitely more friendly for governance! I'll do some refactoring accordingly.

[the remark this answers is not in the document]
yrong: @vgeddes https://github.com/Snowfork/snowbridge/pull/795/commits/3fbe0eb1af318e56b2fed9b87b0a531369bc04fa

parachain/pallets/ethereum-beacon-client/src/lib.rs
```diff
@@ -342,51 +357,89 @@ pub mod pallet {
...
+		pub fn unblock_bridge(origin: OriginFor<T>) -> DispatchResult {
+			ensure_root(origin)?;
+			<Blocked<T>>::set(false);
+			Self::deposit_event(Event::Activated);
 			Ok(())
 		}
 	}
 
 	impl<T: Config> Pallet<T> {
 		fn process_initial_sync(initial_sync: InitialSyncOf<T>) -> DispatchResult {
```
alistair-singh, line 419: Discussion: We can unify the `process_initial_sync` and `unblock_bridge` code paths. Instead of having an initial sync step, we could start the bridge in a `Deactivated` state by default and then use the `unblock_bridge` extrinsic to bring it into an Activated state. This will simplify the deployment process as we will not have to set `initial_sync` field in the beacon client genesis state.

yrong: Good point!  Some changes accordingly.
https://github.com/Snowfork/snowbridge/pull/795/commits/4b1279a84029f407945df500c767d56b6b52deb3

alistair-singh (approved):

core/packages/test/scripts/fetch-beacon-data.sh
```diff
@@ -9,14 +9,26 @@ echo "finalized_block_number is: $finalized_block_number"
...
+echo "checkpoint_root is:" $checkpoint_root
+curl -s "$beacon_endpoint_http/eth/v1/beacon/light_client/bootstrap/$checkpoint_root" | jq -r "." > testdata/beacon_check_point.json
+
+# get beacon block
+echo "fetch beacon block..."
 curl -s "$beacon_endpoint_http/eth/v2/beacon/blocks/$finalized_block_root" | jq -r "." > testdata/beacon_block.json
 
-# get beacon state by block_number
-curl -s "$beacon_endpoint_http/eth/v2/debug/beacon/states/$finalized_block_number" | jq -r "." > testdata/beacon_state.json
+if [[ $beacon_endpoint_http == "http://127.0.0.1:9596" ]];then
```
claravanstaden, line 31: Why do we get the beacon state for local testnet only? 🤔

parachain/pallets/ethereum-beacon-client/src/lib.rs
claravanstaden: I wonder if the methods shouldn't then mirror the language used, for example, rename `block_bridge` to `deactivate` and `unblock_bridge` to `activate`?

parachain/pallets/ethereum-beacon-client/src/lib.rs
```diff
@@ -692,7 +752,7 @@ pub mod pallet {
 			let index_in_array = block_slot % max_slots_per_historical_root;
 			let leaf_index = max_slots_per_historical_root + index_in_array;
 
-			log::info!(
+			log::debug!(
 				target: "ethereum-beacon-client",
 				"💫 Depth: {} leaf_index: {}", config::BLOCK_ROOT_AT_INDEX_PROOF_DEPTH, leaf_index
 			);
```
claravanstaden, line 758: We can remove this, I think I missed removing this while testing the ancestry proofs. 😄 

```suggestion
```

parachain/pallets/ethereum-beacon-client/src/lib.rs
```diff
@@ -915,14 +984,12 @@ pub mod pallet {
 		}
 
 		fn store_execution_header(
-			block_hash: H256,
 			header: ExecutionHeader,
 			beacon_slot: u64,
 			beacon_block_root: H256,
-		) {
+		) -> DispatchResult {
```
claravanstaden, line 990: Why does this method need to return a dispatch result? I don't see an error condition that can arise in the method, unless I am missing it?

parachain/primitives/beacon/src/lib.rs
```diff
@@ -156,6 +157,37 @@ pub struct InitialSync<SyncCommitteeSize: Get<u32>, ProofSize: Get<u32>> {
...
+
+	fn try_from(init_sync: InitialSync<SyncCommitteeSize, ProofSize>) -> Result<Self, Self::Error> {
+		Ok(CheckpointSync {
+			header: init_sync.header,
+			current_sync_committee: init_sync.current_sync_committee,
+			current_sync_committee_branch: init_sync.current_sync_committee_branch,
+		})
+	}
+}
+
```
claravanstaden, line 190: This is just for tests, right? Perhaps you can just comment it is only being used in the tests. 😸

core/packages/test/scripts/fetch-beacon-data.sh
yrong: Since we've provided [script](https://github.com/Snowfork/snowbridge/blob/ron/sno-405/core/packages/test/scripts/start-goerli.sh) to setup goerli nodes from locally I would suggest we fetch beacon state of goerli also in this way.
Btw, I’ve never successfully get beacon state from public [endpoint](https://lodestar-goerli.chainsafe.io) before, :joy:  guess related to the bad network in China.

parachain/primitives/beacon/src/lib.rs
yrong: No, actually we use it in `process_initial_sync` [here](https://github.com/Snowfork/snowbridge/blob/0176d783df795ddaf38491a8c22caca1523ca692/parachain/pallets/ethereum-beacon-client/src/lib.rs#L435-L437)  so to reuse the logic in `process_checkpoint_sync`, make code here a bit cleaner IMHO.

PR #801, opened 2023-04-10
Add support for contract upgrades

vgeddes:
Design:

There is a new message handler called `UpgradeProxy`. It can receive upgrade instructions from Polkadot governance, and execute them with elevated privileges.

Steps:
1. Polkadot governance manually deploys a contract that implements `IUpgradeTask`, and which contains the necessary upgrade logic.
2. When `UpgradeProxy` receives a message from governance, it will execute the upgrade logic in the task using `delegatecall`.

I've updated the forge deployment script so that `UpgradeProxy` has admin privileges for all upgradable contracts.

### DIFF
--- a/core/packages/contracts/src/UpgradeProxy.sol
+++ b/core/packages/contracts/src/UpgradeProxy.sol
@@ -15,13 +15,12 @@ contract UpgradeProxy is AccessControl {
         bytes payload;
     }
 
-    enum Action
-    // Withdraw from sovereign account and transfer to recipient.
-    // Parachain teams will occasionally send this message to retrieve collected fees.
-    {Upgrade}
+    enum Action {
+        Upgrade
+    }
 
     struct UpgradePayload {
-        address upgrader;
+        address task;
     }
 
     error InvalidMessage();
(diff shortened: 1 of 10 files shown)


core/packages/contracts/src/UpgradeProxy.sol
```diff
@@ -0,0 +1,56 @@
...
+        }
+
+        Message memory decoded = abi.decode(message, (Message));
+        if (decoded.action != Action.Upgrade) {
+            revert InvalidMessage();
+        }
+
+        UpgradePayload memory payload = abi.decode(decoded.payload, (UpgradePayload));
+
+        (bool success,) = payload.task.delegatecall(abi.encodeCall(IUpgradeTask.run, ()));
```
yrong, line 51: may not be the scope of this PR,  just curious is the `IUpgradeTask` here meant only for Parameter-change like fee or reward from governance? For senarios like bug fixing or introduce new features do we need some [proxy pattern](https://docs.openzeppelin.com/upgrades-plugins/1.x/proxies) instead?

yrong (approved):
Awesome!

core/packages/contracts/src/UpgradeProxy.sol
vgeddes: Yeah that's exactly correct :+1:

vgeddes: Though to add, there are two stateful contracts which we don't plan on replacing ever:
* Vault.sol: Controls locked up ETH on behalf of parachains/users
* Tokenvault.sol: Controls locked up ERC20 tokens on behalf of users

The idea is these contracts are very simple, and don't much logic that would need upgrading or replacing.

Our `NativeTokens` contract acts as a shell over `TokenVault`, implementing all the messy business logic of interacting with users and the bridge. That's the kind of logic which we would expect to be upgraded.

So when we deploy a new NativeTokens contract, we would allow the replacement control to access the existing tokenVault.

PR #802, opened 2023-04-12
Benchmark with main spec

yrong:
Companion with https://github.com/Snowfork/cumulus/pull/14

### DIFF
--- a/core/packages/test/scripts/build-binary.sh
+++ b/core/packages/test/scripts/build-binary.sh
@@ -26,13 +26,17 @@ rebuild_cumulus(){
         --git https://github.com/Snowfork/cumulus \
         --branch "$cumulus_version" polkadot-parachain-bin \
         --locked \
-        --root $cumulus_dir #add version path to root to avoid recompiling when switch between versions 
+        --root $cumulus_dir #add version path to root to avoid recompiling when switch between versions
     popd
 }
 
 build_cumulus_from_source(){
     pushd $root_dir/cumulus
-    cargo build --release --bin polkadot-parachain
+    if [[ "$active_spec" == "minimal" ]]; then
+      cargo build --features "minimal" --release --bin polkadot-parachain
+    else
+      cargo build --release --bin polkadot-parachain
+    fi
     cp target/release/polkadot-parachain $output_bin_dir/polkadot-parachain
     popd
 }
@@ -44,7 +48,7 @@ rebuild_relaychain(){
         --git https://github.com/paritytech/polkadot \
         --tag "$relaychain_version" polkadot \
         --locked \
-        --root $relaychain_dir #add version path to root to avoid recompiling when switch between versions 
+        --root $relaychain_dir #add version path to root to avoid recompiling when switch between versions
     popd
 }
 
--- a/core/packages/test/scripts/set-env.sh
+++ b/core/packages/test/scripts/set-env.sh
@@ -24,6 +24,9 @@ export output_bin_dir="$output_dir/bin"
 ethereum_data_dir="$output_dir/geth"
 export PATH="$output_bin_dir:$PATH"
 
+# Because we can run a local node for public network like goerli or mainnet, add active_spec as a separate config here
+# to decouple with eth_network, explicit set ACTIVE_SPEC to mainnet when test with public network
+active_spec="${ACTIVE_SPEC:-minimal}"
 eth_network="${ETH_NETWORK:-localhost}"
 eth_endpoint_http="${ETH_RPC_ENDPOINT:-http://127.0.0.1:8545}/${INFURA_PROJECT_ID:-}"
 eth_endpoint_ws="${ETH_WS_ENDPOINT:-ws://127.0.0.1:8546}/${INFURA_PROJECT_ID:-}"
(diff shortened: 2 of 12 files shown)


core/packages/test/scripts/set-env.sh
```diff
@@ -19,6 +19,7 @@ relay_bin="$relay_dir/build/snowbridge-relay"
 [ "$bridge_hub_runtime" = "bridge-hub-rococo-local" ] &&
    seconds_per_slot=6 ||
    seconds_per_slot=12
+active_spec="${ACTIVE_SPEC:-minimal}"
```
vgeddes, line 22: Lets please add a comment explaining this field. We need to start improve the documentation for this script.

core/packages/test/scripts/build-binary.sh
```diff
@@ -32,7 +32,11 @@ rebuild_cumulus(){
 
 build_cumulus_from_source(){
     pushd $root_dir/cumulus
-    cargo build --release --bin polkadot-parachain
+    if [ "$active_spec" == "minimal" ]; then
```
vgeddes, line 35: Double-square brackets are preferable. See https://stackoverflow.com/a/669486.
```suggestion
    if [[ "$active_spec" == "minimal" ]]; then
```

vgeddes (approved):
Looks good, as a temporary measure to allow us to benchmark the main spec.

In the long-term though, I don't want to have to force the cumulus team to support a `minimal` feature. So we'll need to find a workaround for that somehow.

yrong:
@vgeddes Some preliminary benchmark here for your reference.
https://github.com/Snowfork/snowbridge/pull/804#issue-1664889164

parachain/pallets/ethereum-beacon-client/src/weights.rs
```diff
@@ -1,47 +1,34 @@
-//! Autogenerated weights for ethereum_beacon_client
+//! Autogenerated weights for `snowbridge_ethereum_beacon_client`
```
alistair-singh, line 1: In the long term this file belongs in the cumulus repo and is generated by their benchmarking scripts on the hardware they specified for bridge hub.

alistair-singh (approved):

parachain/pallets/ethereum-beacon-client/src/lib.rs
```diff
@@ -59,14 +60,70 @@ pub type FinalizedHeaderUpdateOf<T> = FinalizedHeaderUpdate<
 >;
 pub type SyncCommitteeOf<T> = SyncCommittee<<T as Config>::MaxSyncCommitteeSize>;
 
+/// Weight functions needed for ethereum_beacon_client.
+pub trait WeightInfo {
```
vgeddes, line 64: Why are you putting these weights here instead of `weights.rs`?

parachain/pallets/ethereum-beacon-client/src/lib.rs
```diff
@@ -59,14 +60,70 @@ pub type FinalizedHeaderUpdateOf<T> = FinalizedHeaderUpdate<
...
+	fn block_bridge() -> Weight;
+	fn unblock_bridge() -> Weight;
+	fn verify_signed_header() -> Weight;
+	fn bls_fast_aggregate_verify() -> Weight;
+	fn bls_aggregate_signature() -> Weight;
+	fn bls_aggregate_pubkey() -> Weight;
+	fn bls_verify_message() -> Weight;
+}
+
+impl WeightInfo for () {
```
vgeddes, line 77: Why are you putting these weights here instead of `weights.rs`?

parachain/pallets/ethereum-beacon-client/src/lib.rs
```diff
@@ -766,38 +825,43 @@ pub mod pallet {
 			message: H256,
 			signature: BoundedVec<u8, T::MaxSignatureSize>,
 		) -> DispatchResult {
-			let sig = Signature::from_bytes(&signature[..]);
-			if let Err(_e) = sig {
-				return Err(Error::<T>::InvalidSignature.into())
-			}
+			let agg_sig = Self::bls_aggregate_signature(signature)?;
```
vgeddes, line 828: I don't think this refactoring is an improvement. The old version was easier to understand as all the logic was in a single function, and the function wasn't too large. This helps auditors.

Refactoring the method would make sense if `bls_aggregate_signature`, etc, was going to be used in multiple places, but thats not the case here.

Substrate's benchmarking framework allows to benchmark arbitrary code in `{ ... }`, and so I would rather use that approach.

vgeddes (commented):
Hey Ron,

The PR looks better this time around, but still needs more work. I also see there is a lot of refactoring which isn't related to the goals of this PR. Looks like you've based the this branch on https://github.com/Snowfork/snowbridge/pull/795, which I don't want to merge in yet.

yrong:
>  I also see there is a lot of refactoring which isn't related to the goals of this PR. Looks like you've based the this branch on https://github.com/Snowfork/snowbridge/pull/795, which I don't want to merge in yet.

@vgeddes Yes, realize that refactoring here is unnessessary and cause trouble for review. Sorry for the inconvenience and I've forced push updates more focused without the mess.  Hope it will make things easier and please let me know if still any problem.

PR #806, opened 2023-04-14
SNO-425: For parachain -> eth messages, sync by parachain id instead of account id

doubledup:
This also makes the parachain relayer sync messages for a single configured parachain id instead of multiple.

[Issue](https://linear.app/snowfork/issue/SNO-425)
[Companion PR in Cumulus fork](https://github.com/Snowfork/cumulus/pull/18) (will bump the submodule here once the companion's merged)


parachain/pallets/outbound-queue/src/lib.rs
```diff
@@ -29,27 +30,21 @@ pub use weights::WeightInfo;
...
 
-impl<SourceId, M: Get<u32>> Into<Token> for Message<SourceId, M>
-where
-	SourceId: Decode + Parameter + Member + MaxEncodedLen, //+ TypeInfo,
-{
+impl<M: Get<u32>> Into<Token> for Message<M> {
 	fn into(self) -> Token {
 		Token::Tuple(vec![
-			Token::Bytes(self.source_id.encode()),
+			Token::Bytes(self.parachain_id.encode()),
```
vgeddes, line 47: Parachain ID is actually an ABI-encoded unsigned u32. That's what the ethereum side is expecting:
 1. https://github.com/Snowfork/snowbridge/blob/0acc8f2dc7d36b9ac008d3db84856d91a4198a82/core/packages/contracts/src/Types.sol#L4
2. https://github.com/Snowfork/snowbridge/blob/0acc8f2dc7d36b9ac008d3db84856d91a4198a82/core/packages/contracts/src/InboundQueue.sol#L36
```suggestion
			Token::Uint(self.origin.into()),
```

parachain/pallets/outbound-queue/src/lib.rs
```diff
@@ -169,7 +161,7 @@ pub mod pallet {
 
 	impl<T: Config> Pallet<T> {
 		/// Submit message on the outbound channel
-		pub fn submit(source_id: &T::SourceId, payload: &[u8]) -> DispatchResult {
+		pub fn submit(parachain_id: &ParaId, payload: &[u8]) -> DispatchResult {
```
vgeddes, line 164: Need to add the u16 handler here too.

vgeddes:
> The other is an unresolved import for `snowbridge_router_primitives::ConvertTokenAddress`, which doesn't sound related to this change.

This should be fixed in latest master.

parachain/pallets/outbound-queue/src/lib.rs
```diff
@@ -32,20 +32,22 @@ pub use weights::WeightInfo;
...
+	handler: u16,
 	/// Payload for target application.
 	payload: BoundedVec<u8, M>,
 }
 
 impl<M: Get<u32>> Into<Token> for Message<M> {
 	fn into(self) -> Token {
 		Token::Tuple(vec![
-			Token::Bytes(self.parachain_id.encode()),
+			Token::Uint(u32::from(self.origin).into()),
```
doubledup, line 48: Had to add an explicit conversion here because `Uint` has a `From<u32>` but no `From<polkadot_parachain::primitives::Id>`

relayer/relays/parachain/scanner.go
```diff
@@ -155,112 +144,100 @@ func (s *Scanner) findTasks(
...
+			return nil, fmt.Errorf("fetch header for block hash %v: %w", blockHash.Hex(), err)
 		}
 
 		digestItems, err := ExtractAuxiliaryDigestItems(header.Digest)
 		if err != nil {
 			return nil, err
 		}
-
 		if len(digestItems) == 0 {
 			currentBlockNumber--
```
yrong, line 183: nothing to do with change in this PR, just curious is this line required or a bug should be removed?

yrong, line 183: nothing to do with change in this PR, just curious is this line required or a bug should be removed?

yrong (approved):
LGTM!

relayer/relays/parachain/scanner.go
```diff
@@ -155,112 +144,100 @@ func (s *Scanner) findTasks(
...
 			continue
 		}
```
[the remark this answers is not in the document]
doubledup: Yes that's a bug, good catch! I missed it when changing the loop, so it is part of this PR

relayer/relays/parachain/config.go
```diff
@@ -16,6 +16,7 @@ type SourceConfig struct {
 	Contracts SourceContractsConfig  `mapstructure:"contracts"`
 	// Block number when Beefy was activated
 	BeefyActivationBlock uint64 `mapstructure:"beefy-activation-block"`
+	LaneID               uint32 `mapstructure:"lane-id"`
```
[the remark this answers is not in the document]
doubledup: Right, that could be confusing. This was based on the [execution relayer config](https://github.com/Snowfork/snowbridge/blob/main/relayer/relays/execution/config.go#L15), so I'd rather do it in a follow-up PR so we don't change the execution relayer here. LaneIDs only exist in our relayers, in the queues & messages we call it origin but paraID sounds fine for the relayers.

@vgeddes Are you happy with ParaID in the relayers?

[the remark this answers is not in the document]
doubledup: Semi-related: I noticed a bunch of sourceID references and changed them to laneID for consistency.

PR #810, opened 2023-04-20
Bound finalized header, execution header and sync committee by pruning older data beyond threshold

ParthDesai:
### Description
This PR enables pruning older `FinalizedHeader`, `ExecutionHeader` and `SyncCommittee`. This is required to put bounds on the storage of the pallet. As left unchecked it could be huge amount of data.

First three PR adds support for pruning of `FinalizedHeader`, `ExecutionHeader` and `SyncCommittee` respectively. Fourth commit adds comprehensive tests for the same. Fifth commit is just cargo fmt changes. And last commit introduces relayer changes.

#### ExecutionHeader
For `ExecutionHeader` there is a queue like simulated data maintained that allows pruning of older data as the new data gets added. 

Since execution headers are not contiguous, to maintain the queue, we need to maintain a mapping of `u64 -> ExecutionHeader` and also keep track of front and back of the queue. While we push new data to back, we remove data from front this is supported by `(oldest_bound, latest_bound)` storage.

#### FinalizedHeader
For `FinalizedHeader` we are pruning it as per `MaxFinalizedHeaderSlotArray`. As, `FinalizedHeader` and its associated data with slot less than `0`th element of the vector will not be used by relayer in anycase.

#### SyncCommittee
Since `SyncCommittee` is stored in an order, we can rely on it to prune oldest sync committee. 

### Configuration changes
Both `ExecutionHeader` and `SyncCommittee`'s pruning threshold is defined in pallet configuration. 

### Relayer changes
Due to change in `FinalizedSlotArray` storage, there is a change in relayer to maintain compatibility.

### DIFF
--- a/parachain/pallets/ethereum-beacon-client/src/lib.rs
+++ b/parachain/pallets/ethereum-beacon-client/src/lib.rs
@@ -102,6 +102,9 @@ pub mod pallet {
 		type ExecutionHeadersPruneThreshold: Get<u64>;
 		#[pallet::constant]
 		type ForkVersions: Get<ForkVersions>;
+		/// Maximum sync committees to be stored
+		#[pallet::constant]
+		type SyncCommitteePruneThreshold: Get<u64>;
 		type WeightInfo: WeightInfo;
 		type WeakSubjectivityPeriodSeconds: Get<u64>;
 	}
@@ -190,10 +193,9 @@ pub mod pallet {
 		CountedStorageMap<_, Identity, H256, ExecutionHeader, OptionQuery>;
 
 	/// Current sync committee corresponding to the active header.
-	/// TODO  prune older sync committees than xxx
 	#[pallet::storage]
 	pub(super) type SyncCommittees<T: Config> =
-		StorageMap<_, Identity, u64, SyncCommitteeOf<T>, ValueQuery>;
+		CountedStorageMap<_, Identity, u64, SyncCommitteeOf<T>, ValueQuery>;
 
 	#[pallet::storage]
 	pub(super) type ValidatorsRoot<T: Config> = StorageValue<_, H256, ValueQuery>;
@@ -878,7 +880,7 @@ pub mod pallet {
 			Ok(())
 		}
 
-		fn store_sync_committee(period: u64, sync_committee: SyncCommitteeOf<T>) {
+		pub(crate) fn store_sync_committee(period: u64, sync_committee: SyncCommitteeOf<T>) {
 			<SyncCommittees<T>>::insert(period, sync_committee);
 
 			log::trace!(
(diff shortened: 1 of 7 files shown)


parachain/pallets/ethereum-beacon-client/src/lib.rs
```diff
@@ -874,15 +890,39 @@ pub mod pallet {
...
+					stored_sync_committees as u64 - threshold;
+
+				while number_of_sync_committees_to_remove > 0 {
+					<SyncCommittees<T>>::remove(current_sync_committee_to_remove);
+					number_of_sync_committees_to_remove =
+						number_of_sync_committees_to_remove.saturating_sub(1);
+					current_sync_committee_to_remove =
+						current_sync_committee_to_remove.saturating_sub(1);
+				}
+			}
```
yrong, line 918: The problem here is the performance overhead of the while loop which could use up all the block weight.

parachain/pallets/ethereum-beacon-client/src/lib.rs
```diff
@@ -941,6 +993,37 @@ pub mod pallet {
...
+				let (mut oldest_mapping_to_delete, _) = Self::get_mapping_bound();
+				let execution_headers_to_delete =
+					stored_execution_headers.saturating_sub(threshold);
+				for _i in 0..execution_headers_to_delete {
+					let execution_header_hash =
+						<ExecutionHeadersMapping<T>>::get(oldest_mapping_to_delete);
+					<ExecutionHeadersMapping<T>>::remove(oldest_mapping_to_delete);
+					<ExecutionHeaders<T>>::remove(execution_header_hash);
+					oldest_mapping_to_delete += 1;
+				}
```
yrong, line 1022: https://github.com/Snowfork/snowbridge/pull/810/files#r1173313662

yrong:
@ParthDesai Cool and thanks for the contribution!  Though IMHO implementation with some ringbuffer structure without pruing could be more simple and effiecent. 
https://substrate-developer-hub.github.io/substrate-how-to-guides/docs/storage-migrations/ringbuffer/

parachain/pallets/ethereum-beacon-client/src/lib.rs
ParthDesai: `while` loop would be most likely run only 1 time. As it runs after every insert.

parachain/pallets/ethereum-beacon-client/src/lib.rs
ParthDesai: while loop would be most likely run only 1 time. As it runs after every insert.

ParthDesai:
> @ParthDesai Cool and thanks for the contribution! Though IMHO implementation with some ringbuffer structure without pruning could be more simple and effiecient. https://substrate-developer-hub.github.io/substrate-how-to-guides/docs/storage-migrations/ringbuffer/

The `while` loop here is not supposed to be running more than one time in current version of code base. As it tries to prune oldest entry after every insert.

ParthDesai:
> @ParthDesai Cool and thanks for the contribution! Though IMHO implementation with some ringbuffer structure without pruning could be more simple and effiecient. https://substrate-developer-hub.github.io/substrate-how-to-guides/docs/storage-migrations/ringbuffer/

But, yes. The Ringbuffer structure does look like something which can be more efficient and simple than this. I will open another PR with updated implementation.

PR #814, opened 2023-04-24
Bound finalized header, execution header and sync committee by RingBuffer and BoundedVec

ParthDesai:
### Description
This PR is improvement over https://github.com/Snowfork/snowbridge/pull/810 as suggested by @yrong.  Instead of using Pruning, now we are using modified version of RingBuffer to limit the runtime storage. 

### How `RingBuffer` works?
If number of items go beyond some threshold, the `RingBuffer::insert` will remove oldest entry. 

### Advantage of `RingBuffer`
There is no while loop. The complexity is `O(1)`. And since it is abstracted, there is no clutter in application logic.

### DIFF
--- a/parachain/pallets/ethereum-beacon-client/src/lib.rs
+++ b/parachain/pallets/ethereum-beacon-client/src/lib.rs
@@ -18,9 +18,12 @@ mod tests_minimal;
 #[cfg(feature = "runtime-benchmarks")]
 mod benchmarking;
 
-pub use weights::WeightInfo;
+mod ringbuffer;
 
-use crate::merkleization::get_sync_committee_bits;
+use crate::{
+	merkleization::get_sync_committee_bits,
+	ringbuffer::{RingBufferMap, RingBufferMapImpl},
+};
 use frame_support::{dispatch::DispatchResult, log, traits::UnixTime, transactional};
 use frame_system::ensure_signed;
 use snowbridge_beacon_primitives::{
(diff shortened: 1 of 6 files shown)


yrong:
@ParthDesai Awesome! Many thanks and I'll review it together with team.

parachain/pallets/ethereum-beacon-client/src/lib.rs
```diff
@@ -165,7 +175,7 @@ pub mod pallet {
 
 	#[pallet::storage]
 	pub(super) type FinalizedBeaconHeaderSlots<T: Config> =
-		StorageValue<_, BoundedVec<u64, T::MaxFinalizedHeaderSlotArray>, ValueQuery>;
+		StorageValue<_, BoundedVec<(u64, H256), T::MaxFinalizedHeaderSlotArray>, ValueQuery>;
```
yrong, line 178: Just reminder, require associated change in relayer side [here](https://github.com/Snowfork/snowbridge/blob/4f3b5910edca114059ace9de130e1524d0a20625/relayer/chain/parachain/writer.go#L228-L246)

yrong:
LGTM! Definitely a more elegant way for the pruing. Cost is some auxiliary storage and more r/w operations introduced so more weight will be consumed. Anyway, it's some O(1) operation so should be fine. Would be nice if we can attach some benchmark accordingly.

yrong:
required https://github.com/Snowfork/cumulus/pull/20 and benchmark in comparison
https://github.com/Snowfork/cumulus/pull/20/files#diff-db08d74f3fe259c221df3ac86f73c1e06ebafc90e2907e95b98ff9cf71018ced

ParthDesai:
> @ParthDesai There is some more changes required to run e2e stack, if you do not mind I'll draft this one and we can review #815 instead.

No problem. If you need any help with that, please let me know.

PR #815, opened 2023-04-27
Bound finalized header, execution header and sync committee by RingBuffer and BoundedVec

yrong:
Based on https://github.com/Snowfork/snowbridge/pull/814 from @ParthDesai 
Companion for cumulus: https://github.com/Snowfork/cumulus/pull/20

Fixes: SNO-328


vgeddes:
Looks good! The implementation seems sound. Will give a more thorough review next week.

vgeddes:
@yrong can you review https://github.com/Snowfork/snowbridge/pull/816, and if you agree it is good, then we can rebase this PR on top of it?

parachain/pallets/ethereum-beacon-client/src/lib.rs
```diff
@@ -175,12 +188,53 @@ pub mod pallet {
...
+	#[pallet::storage]
+	pub(crate) type ExecutionHeaderBufferIndex<T: Config> = StorageValue<_, u64, ValueQuery>;
+
+	/// Intermediate storage for execution header mapping
+	#[pallet::storage]
+	pub(crate) type ExecutionHeaderMapping<T: Config> =
+		StorageMap<_, Identity, u64, H256, ValueQuery>;
+
+	/// Ring buffer Map for Execution header
+	pub(crate) type ExecutionHeaderRingBufferMap<T> = RingBufferMapImpl<
```
vgeddes, line 203: I don't think its worth the extra complexity to prune execution headers. For sync committees it makes sense, since they are huge.

I would rather we remove unnecessary fields from `ExecutionHeader`. All we really need is the `receipts_root` and `storage_root`.

Also see this discussion here for the Polkadot<->Kusama bridge, where they take a similar approach: https://github.com/Snowfork/cumulus/blob/c9cca565af5f165db5fbea9a149f9594c9a76b6e/bridges/modules/parachains/src/lib.rs#L187

parachain/pallets/ethereum-beacon-client/src/lib.rs
```diff
@@ -175,12 +188,53 @@ pub mod pallet {
...
+	#[pallet::storage]
+	pub(crate) type SyncCommitteesBufferIndex<T: Config> = StorageValue<_, u64, ValueQuery>;
+
+	/// Intermediate storage for sync committee mapping
+	#[pallet::storage]
+	pub(crate) type SyncCommitteesMapping<T: Config> =
+		StorageMap<_, Identity, u64, u64, ValueQuery>;
+
+	/// Ring buffer Map for Sync committee
+	pub(crate) type SyncCommitteesRingBufferMap<T> = RingBufferMapImpl<
```
vgeddes, line 229: Seems `StorageMap` and `StorageDoubleMap` already supports ring buffers. 

See for example: https://github.com/Snowfork/cumulus/blob/c9cca565af5f165db5fbea9a149f9594c9a76b6e/bridges/modules/parachains/src/lib.rs#L268

parachain/pallets/ethereum-beacon-client/src/lib.rs
vgeddes: On second thoughts, since `StorageMap` and `StorageDoubleMap` support ring buffers, is ok to use it for execution headers.

yrong: @vgeddes 
> I don't think its worth the extra complexity to prune execution headers. 
Agree but for another reason [here](https://linear.app/snowfork/issue/SNO-356#comment-4074dc26)

> since StorageMap and StorageDoubleMap support ring buffers
Really? AFAIK it does not really support. Some tests [here](https://github.com/Snowfork/snowbridge/commit/c9754f26765b23fb8841513857156824e8f3c0e9)

parachain/pallets/ethereum-beacon-client/src/lib.rs
ParthDesai: No. It is just indication of Max number of items that are supposed to be in that map. They do not implement `RingBuffer`.

parachain/pallets/ethereum-beacon-client/src/lib.rs
ParthDesai: See my comment here: https://github.com/Snowfork/snowbridge/pull/815#discussion_r1179964218

ParthDesai: @vgeddes  Regarding `ExecutionHeader` we do sync large number of execution headers compared to `SyncCommittees`. So, while object size is small overall the map of `ExecutionHeaders` occupies somewhat similar space as `SyncCommittees` map.

yrong: Agree, I'm fine we keep execution headers pruning in this PR, [benchmark](https://github.com/Snowfork/cumulus/pull/20/files#diff-db08d74f3fe259c221df3ac86f73c1e06ebafc90e2907e95b98ff9cf71018ced) shows very limited overhead can be safely ignored.  For the long run I look forward [eip-4788](https://eips.ethereum.org/EIPS/eip-4788) then perhaps we can remove importing execution headers completely.

parachain/pallets/ethereum-beacon-client/src/lib.rs
vgeddes: Oops, my mistake. 🙏

parachain/pallets/ethereum-beacon-client/src/lib.rs
vgeddes: Ok lets prune execution headers then 👍

vgeddes (commented):
This is a good design ring buffer design, but I think we can do even better to optimize the overhead of the implementation.

It seems the mappings `ExecutionHeaderMapping` and `SyncCommitteesMapping` simply contain pointers to next inserted node.

So lets consider storing the `next` pointers with the actual nodes.

See this rough example which explains it better:

```rust
struct RingBufferNode<K, V> {
    value: V,
    next: Option<K>,
}

struct RingBufferCursor<K> {
    // last inserted node
    node: K,
    // number of nodes in ring buffer
    count: u32
}

struct RingBufferTransient<K, V, Nodes, Cursor> 
where
    Nodes: StorageMap<H256, RingBufferNode<K, V>>,
    Cursor: StorageValue<RingBufferCursor<K>>;

impl<K, V, Nodes, Cursor> RingBufferTransient<K, V, Nodes, Cursor> 
where
    Nodes: StorageMap<H256, RingBufferNode<K, V>>,
    Cursor: StorageValue<RingBufferCursor<K>>;
{
    fn insert(key: K, value: V) {
        // handle all cases here, much like a circular singly linked list    
    }
}

// Ring buffer nodes
#[pallet::storage]
pub(super) type SyncCommittees<T: Config> = StorageMap<_, Identity, H256, RingBufferNode<H256, SyncCommitteeOf<T>>, ValueQuery>;

// Ring buffer cursor
#[pallet::storage]
pub(crate) type SyncCommitteesCursor<T: Config> = StorageValue<_, RingBufferCursor<H256>, ValueQuery>;
```

With this design, we reduce the number of storage writes from 3 to 2, so uses less weight.

Let me know what you think.

vgeddes:
> This is a good design ring buffer design, but I think we can do even better to optimize the overhead of the implementation.
> 
> It seems the mappings `ExecutionHeaderMapping` and `SyncCommitteesMapping` simply contain pointers to next inserted node.
> 
> So lets consider storing the `next` pointers with the actual nodes.
> 
> See this rough example which explains it better:
> 
> ```rust
> struct RingBufferNode<K, V> {
>     value: V,
>     next: Option<K>,
> }
> 
> struct RingBufferCursor<K> {
>     // last inserted node
>     node: K,
>     // number of nodes in ring buffer
>     count: u32
> }
> 
> struct RingBufferTransient<K, V, Nodes, Cursor> 
> where
>     Nodes: StorageMap<K, RingBufferNode<K, V>>,
>     Cursor: StorageValue<RingBufferCursor<K>>;
> 
> impl<K, V, Nodes, Cursor> RingBufferTransient<K, V, Nodes, Cursor> 
> where
>     Nodes: StorageMap<K, RingBufferNode<K, V>>,
>     Cursor: StorageValue<RingBufferCursor<K>>;
> {
>     fn insert(key: K, value: V) {
>         // handle all cases here, much like a circular singly linked list    
>     }
> }
> 
> // Ring buffer nodes
> #[pallet::storage]
> pub(super) type SyncCommittees<T: Config> = StorageMap<_, Identity, H256, RingBufferNode<H256, SyncCommitteeOf<T>>, ValueQuery>;
> 
> // Ring buffer cursor
> #[pallet::storage]
> pub(crate) type SyncCommitteesCursor<T: Config> = StorageValue<_, RingBufferCursor<H256>, ValueQuery>;
> ```
> 
> With this design, we reduce the number of storage writes from 3 to 2, so uses less weight.
> 
> Let me know what you think.

Actually, just realised this idea will have worse performance, since every time we need to update the `node.next` pointers, we will end up writing sync committees to storage redundantly.

parachain/pallets/ethereum-beacon-client/src/lib.rs
vgeddes, line 203: I generally prefer less verbose naming:

```suggestion
	pub(crate) type ExecutionHeadersIndex<T: Config> = StorageValue<_, u64, ValueQuery>;

	/// Intermediate storage for execution header mapping
	#[pallet::storage]
	pub(crate) type ExecutionHeadersMapping<T: Config> =
		StorageMap<_, Identity, u64, H256, ValueQuery>;

	/// Ring buffer Map for Execution header
	pub(crate) type ExecutionHeadersBuffer<T> = RingBufferMapImpl<
```

parachain/pallets/ethereum-beacon-client/src/lib.rs
vgeddes, line 229: Lets reduce verbosity

```suggestion
	pub(crate) type SyncCommitteesIndex<T: Config> = StorageValue<_, u64, ValueQuery>;

	/// Intermediate storage for sync committee mapping
	#[pallet::storage]
	pub(crate) type SyncCommitteesMapping<T: Config> =
		StorageMap<_, Identity, u64, u64, ValueQuery>;

	/// Ring buffer Map for Sync committee
	pub(crate) type SyncCommitteesBuffer<T> = RingBufferMapImpl<
```

ParthDesai:
> This is a good design ring buffer design, but I think we can do even better to optimize the overhead of the implementation.
> 
> It seems the mappings `ExecutionHeaderMapping` and `SyncCommitteesMapping` simply contain pointers to next inserted node.
> 
> So lets consider storing the `next` pointers with the actual nodes.
> 
> See this rough example which explains it better:
> 
> ```rust
> struct RingBufferNode<K, V> {
>     value: V,
>     next: Option<K>,
> }
> 
> struct RingBufferCursor<K> {
>     // last inserted node
>     node: K,
>     // number of nodes in ring buffer
>     count: u32
> }
> 
> struct RingBufferTransient<K, V, Nodes, Cursor> 
> where
>     Nodes: StorageMap<K, RingBufferNode<K, V>>,
>     Cursor: StorageValue<RingBufferCursor<K>>;
> 
> impl<K, V, Nodes, Cursor> RingBufferTransient<K, V, Nodes, Cursor> 
> where
>     Nodes: StorageMap<K, RingBufferNode<K, V>>,
>     Cursor: StorageValue<RingBufferCursor<K>>;
> {
>     fn insert(key: K, value: V) {
>         // handle all cases here, much like a circular singly linked list    
>     }
> }
> 
> // Ring buffer nodes
> #[pallet::storage]
> pub(super) type SyncCommittees<T: Config> = StorageMap<_, Identity, H256, RingBufferNode<H256, SyncCommitteeOf<T>>, ValueQuery>;
> 
> // Ring buffer cursor
> #[pallet::storage]
> pub(crate) type SyncCommitteesCursor<T: Config> = StorageValue<_, RingBufferCursor<H256>, ValueQuery>;
> ```
> 
> With this design, we reduce the number of storage writes from 3 to 2, so uses less weight.
> 
> Let me know what you think.


> > This is a good design ring buffer design, but I think we can do even better to optimize the overhead of the implementation.
> > It seems the mappings `ExecutionHeaderMapping` and `SyncCommitteesMapping` simply contain pointers to next inserted node.
> > So lets consider storing the `next` pointers with the actual nodes.
> > See this rough example which explains it better:
> > ```rust
> > struct RingBufferNode<K, V> {
> >     value: V,
> >     next: Option<K>,
> > }
> > 
> > struct RingBufferCursor<K> {
> >     // last inserted node
> >     node: K,
> >     // number of nodes in ring buffer
> >     count: u32
> > }
> > 
> > struct RingBufferTransient<K, V, Nodes, Cursor> 
> > where
> >     Nodes: StorageMap<K, RingBufferNode<K, V>>,
> >     Cursor: StorageValue<RingBufferCursor<K>>;
> > 
> > impl<K, V, Nodes, Cursor> RingBufferTransient<K, V, Nodes, Cursor> 
> > where
> >     Nodes: StorageMap<K, RingBufferNode<K, V>>,
> >     Cursor: StorageValue<RingBufferCursor<K>>;
> > {
> >     fn insert(key: K, value: V) {
> >         // handle all cases here, much like a circular singly linked list    
> >     }
> > }
> > 
> > // Ring buffer nodes
> > #[pallet::storage]
> > pub(super) type SyncCommittees<T: Config> = StorageMap<_, Identity, H256, RingBufferNode<H256, SyncCommitteeOf<T>>, ValueQuery>;
> > 
> > // Ring buffer cursor
> > #[pallet::storage]
> > pub(crate) type SyncCommitteesCursor<T: Config> = StorageValue<_, RingBufferCursor<H256>, ValueQuery>;
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > With this design, we reduce the number of storage writes from 3 to 2, so uses less weight.
> > Let me know what you think.
> 
> Actually, just realised this idea will have worse performance, since every time we need to update the `node.next` pointers, we will end up writing sync committees to storage redundantly. Ignore this idea

Hmm. Make sense. One other concern is implemeting `RingBuffer` should be internal detail of the pallet. If we expose the structure via Storage, client can make assumptions on that which might not be good idea as if we modify the `RingBuffer` implementation then we will have to do storage migration.

parachain/pallets/ethereum-beacon-client/src/lib.rs
```diff
@@ -175,12 +188,53 @@ pub mod pallet {
 	pub(super) type ExecutionHeaders<T: Config> =
 		StorageMap<_, Identity, H256, ExecutionHeader, OptionQuery>;
 
+	/// Execution headers ring buffer map implementation
+
+	/// Index storage for execution header ring buffer map
+	#[pallet::storage]
+	pub(crate) type ExecutionHeaderBufferIndex<T: Config> = StorageValue<_, u64, ValueQuery>;
```
vgeddes, line 195: `u32` should more than sufficient, and save storage costs.

parachain/pallets/ethereum-beacon-client/src/lib.rs
```diff
@@ -175,12 +188,53 @@ pub mod pallet {
...
-	/// TODO  prune older sync committees than xxx
 	#[pallet::storage]
 	pub(super) type SyncCommittees<T: Config> =
 		StorageMap<_, Identity, u64, SyncCommitteeOf<T>, ValueQuery>;
 
+	/// Sync committee ring buffer map implementation
+
+	/// Index storage for sync committee ring buffer
+	#[pallet::storage]
+	pub(crate) type SyncCommitteesBufferIndex<T: Config> = StorageValue<_, u64, ValueQuery>;
```
vgeddes, line 221: `u32` should more than sufficient, and save storage costs.

parachain/pallets/ethereum-beacon-client/src/lib.rs
ParthDesai: This is the name that will be `STORAGE_PREFIX` and would be displayed in UI. So, its better to be verbose with the naming to give user precise idea about what exactly particular storage entry is used for.

parachain/pallets/ethereum-beacon-client/src/lib.rs
ParthDesai: @vgeddes Same as here: https://github.com/Snowfork/snowbridge/pull/815#discussion_r1180611132

parachain/pallets/ethereum-beacon-client/src/lib.rs
ParthDesai: Fair enough. Though I do not have access to this PR to make this change.

yrong: @ParthDesai No worry, actually we have a lot of refactoring recently.  I'll rebase the PR and make change required later.

ParthDesai: @yrong Thank you. If you need my help, let me know. For my project, we really need this capability.

parachain/pallets/ethereum-beacon-client/src/lib.rs
yrong: I prefer the concise.  Though it's still possible to add some storage macro ```#[pallet::storage_prefix = "***RingBuffer"]``` to explicit the prefix if nessessary IMO it's some internal detail end user may not really care.

vgeddes: yeah I still prefer the concise form. Its only developers who would be using the Polkadot-JS console anyway, and they can refer to the comments in the code/docs to see what the storage fields are about.

ParthDesai: Make sense. Let's make it concise.

ParthDesai (approved):
Overall make sense to me. Though, some unrelated changes are from the main, which I haven't reviewed as I do not have enough context for those.

alistair-singh (approved):

yrong:
> @yrong Is it possible to merge this PR now?

Yes, almost done from my side but will defer to @vgeddes if any other concern.

parachain/pallets/ethereum-beacon-client/src/lib.rs
```diff
@@ -83,6 +100,15 @@ pub mod pallet {
 		type ForkVersions: Get<ForkVersions>;
 		#[pallet::constant]
 		type WeakSubjectivityPeriodSeconds: Get<u64>;
+		/// Maximum finalized headers
+		#[pallet::constant]
+		type FinalizedHeaderPruneThreshold: Get<u32>;
```
vgeddes, line 105: Prefer the name MaxFinalizedHeadersToKeep

vgeddes (approved):

vgeddes:
@ParthDesai Note, you need at least rust 1.70 (nightly) to successfully compile the light client in all modes that are required (debug, release, etc).

Older versions of Rust struggle: https://github.com/paritytech/substrate/issues/14075

PR #816, opened 2023-04-27
Refactor beacon client

vgeddes:
Lots of cleanups & improvements, mainly around data structures.

For example, its not really necessary to for data structures to use generic types such as `MaxFeeRecipientSize`, `MaxPubKeySize` `MaxSignatureSize`, since those will always be 20, 48, and 96 respectively. These constants are very unlikely to change.

* `BoundedVec` is only needed for collections that are being written to storage. Otherwise, `Vec` is fine.
* Data structures only need have generic parameters for `SyncCommitteeSize` and `SyncCommitteeBitsSize` where needed.
* `SyncCommittee.pubkeys` is now a fixed array instead a vector, since the size of the sync committee is known statically.
* `SyncAggregate.sync_committee_bits` is now a fixed array instead a vector, since the size of the sync committee is known statically.
* Converting data types to the SSZ form can be done without raising errors. So I replaced `TryFrom` with `From` in many cases.


yrong:
Nice refactoring!  Just a minor fix and some polish in this PR
https://github.com/Snowfork/snowbridge/pull/817

yrong (approved):
LGTM!

PR #818, opened 2023-05-02
Solmate proofs

doubledup:
Needed to add a new function to get the root hash, as the only one available is for verifying a proof. It's currently in our [solmate fork](https://github.com/Snowfork/solmate). I'll work on pushing this upstream.

Also need to check the licenses: Solmate is released under AGPL 3 (though the files we're using have an MIT license identifier as their first line) and our contracts under Apache 2.

Added a large proof test (with data generated in a [benchmarks repo](https://github.com/doubledup/merkle-proof-contract-benchmarks/blob/a14cdb5162eb5cfadfe4042961e716a249969901/src/bin/generate_proof.rs)) and a fuzz test.

### DIFF
--- a/core/packages/contracts/src/InboundQueue.sol
+++ b/core/packages/contracts/src/InboundQueue.sol
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 pragma solidity ^0.8.19;
 
-import {MerkleProof} from "openzeppelin/utils/cryptography/MerkleProof.sol";
+import {MerkleProofLib} from "solmate/utils/MerkleProofLib.sol";
 import {AccessControl} from "openzeppelin/access/AccessControl.sol";
 import {IParachainClient} from "./ParachainClient.sol";
 import {IRecipient} from "./IRecipient.sol";
@@ -67,7 +67,7 @@ contract InboundQueue is AccessControl {
 
     function submit(Message calldata message, bytes32[] calldata leafProof, bytes calldata headerProof) external {
         bytes32 leafHash = keccak256(abi.encode(message));
-        bytes32 commitment = MerkleProof.processProof(leafProof, leafHash);
+        bytes32 commitment = MerkleProofLib.calculateRoot(leafProof, leafHash);
         if (!parachainClient.verifyCommitment(commitment, headerProof)) {
             revert InvalidProof();
         }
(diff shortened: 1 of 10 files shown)


vgeddes (approved):
Looks good! Lets merge it.

core/packages/contracts/src/InboundQueue.sol
```diff
@@ -67,7 +67,7 @@ contract InboundQueue is AccessControl {
 
     function submit(Message calldata message, bytes32[] calldata leafProof, bytes calldata headerProof) external {
         bytes32 leafHash = keccak256(abi.encode(message));
```
vgeddes, line 69: Lets add comment explaining the purpose of this hashing the leaf

core/packages/contracts/src/InboundQueue.sol
```diff
@@ -67,7 +67,7 @@ contract InboundQueue is AccessControl {
...
-        bytes32 commitment = MerkleProof.processProof(leafProof, leafHash);
+        bytes32 commitment = MerkleProofLib.calculateRoot(leafProof, leafHash);
```
vgeddes, line 70: Ditto, lets add a comment explaining why we're calculating the root.

core/packages/contracts/src/InboundQueue.sol
```diff
@@ -67,7 +67,7 @@ contract InboundQueue is AccessControl {
...
         if (!parachainClient.verifyCommitment(commitment, headerProof)) {
```
vgeddes, line 71: Ditto, and another comment explaining what is actually being verified here.

vgeddes (approved):

alistair-singh (approved):
🚀

yrong:
Are we missing the refactoring here?
https://github.com/Snowfork/snowbridge/blob/d001d7255ca1d8e81305795be19484e4a9fa42d3/core/packages/contracts/src/ParachainClient.sol#L4

yrong:
Since we've introduce fuzz test here, would it better to explicit declare in foundry.toml like this?
https://github.com/transmissions11/solmate/blob/2001af43aedb46fdc2335d2a7714fb2dae7cfcd1/foundry.toml#L6-L7

yrong (approved):
Awesome!

vgeddes:
> Are we missing refactoring here?
> 
> https://github.com/Snowfork/snowbridge/blob/d001d7255ca1d8e81305795be19484e4a9fa42d3/core/packages/contracts/src/ParachainClient.sol#L4

Yeah, good catch! @doubledup you need to update both ParachainClient and BeefyClient to use the new merkle proof verifier.

doubledup:
The default number of fuzz test runs is 256, I've set it to 1000 in our default profile.

PR #824, opened 2023-05-08
Use Nix's geth package & update nix lockfile

doubledup:
Had to switch to the unstable channel as that's got the new geth version. Running `nix flake update` after this bumped other packages like node and pnpm.


alistair-singh (approved):

vgeddes (approved):

core/packages/contracts/test/BeefyClient.t.sol
```diff
@@ -112,88 +112,88 @@ contract BeefyClientTest is Test {
...
+//
+//        beefyClient.submitInitial(commitHash, bitfield);
+//
+//        cheats.roll(block.number + randaoCommitDelay);
+//
+//        cheats.difficulty(difficulty);
+//
+//        beefyClient.commitPrevRandao(commitHash);
+//
+//        BeefyClient.Commitment memory commitment = BeefyClient.Commitment(blockNumber, setId, payload);
```
yrong, line 140: Are comments here for test CI breaking? Add some analysis here
https://snowfork.slack.com/archives/G01BGMGLAC9/p1683682087326629?thread_ts=1683621046.426719&cid=G01BGMGLAC9

core/packages/contracts/src/BeefyClient.sol
```diff
@@ -243,8 +244,14 @@ contract BeefyClient is Ownable {
             revert TicketExpired();
         }
 
+        console.log("prevRandao before commit:");
+        console.log(ticket.prevRandao);
```
yrong, line 248: would assume for debug only and better to remove for production.

core/packages/contracts/test/BeefyClient.t.sol
doubledup: Exactly, I was trying to figure out what was breaking but the Github incidents yesterday interrupted me. I'll revert the changes - best way forward is using Nix in CI to lock all of our dependencies.

yrong (approved):
LGTM!

PR #828, opened 2023-05-11
Initialize beacon light client from checkpoint

yrong:
Refactoring including:
- rename `InitialUpdate` to `CheckPointUpdate`
- rename `HeaderUpdate` to `ExecutionHeaderUpdate`

Extrinsic added:
- check_point_update
- deactivate_bridge

Essentially `InitialUpdate` is just a CheckPoint for initialization, it's possible we do checkpoint any time at runtime(e.g. checkpoint after a routine maintenance or emergency which cause bridge stopped for a long time).

https://github.com/Snowfork/cumulus/pull/23 required

Fixes: SNO-429


yrong:
Seems there is no `Sudo` or `Democracy` pallet in cumlus bridge, so would assume `check_point_update` which requires `sudo` authority can only be triggered from relaychain by XCM.

To make e2e test easy, we keep the `GenesisConfig`.

alistair-singh:
> requires `sudo` authority can only be triggered from relaychain by XCM

Working on a script for this. Will PR shortly.

alistair-singh:
> Working on a script for this. Will PR shortly.

https://github.com/Snowfork/snowbridge/pull/829

yrong:
@alistair-singh Thanks, that's exactly what I need.
https://github.com/Snowfork/snowbridge/blob/feceb0aef2fd0143ec5db68fc6f52eae15603766/core/packages/test/scripts/configure-statemine.sh#L30C25-L50

core/package.json
```diff
@@ -19,7 +19,6 @@
   "devDependencies": {
     "husky": "^8.0.1",
     "turbo": "^1.6.3",
-    "pnpm-deduplicate": "^0.4.1",
```
yrong, line 22: @doubledup pnpm v7.26.0 supports built-in `pnpm dedupe` command so just remove here.

.gitmodules
```diff
@@ -5,7 +5,7 @@
 [submodule "core/packages/contracts/lib/forge-std"]
 	path = core/packages/contracts/lib/forge-std
 	url = https://github.com/foundry-rs/forge-std
-	branch = v1.4.0
+	branch = master
```
vgeddes, line 8: Seems like this is unrelated to the beacon light client. Lets try to keep the PRs focused.

yrong:
> I would rather not reduce the slot time. Clara reverted that optimization for several reasons I can't remember.

It's different. Last time what we change is `SLOTS_PER_EPOCH` from 32 to 8 which will impact the generation of ancetry proof whereas change `SECONDS_PER_SLOT` won't.

.gitmodules
yrong: It's for fixing solidity tests use `prevrandao` to replace `difficulty` which is deprecated and cause CI breaking. I'll move it to another PR.

vgeddes:
> > I would rather not reduce the slot time. Clara reverted that optimization for several reasons I can't remember.
> 
> It's different. Last time what we change is `SLOTS_PER_EPOCH` from 32 to 8 which will impact the generation of ancestry proof whereas change `SECONDS_PER_SLOT` won't.

Does this mean geth will produce an execution header every 2 seconds? If so, this could cause the light client to fall behind importing execution headers every 2 seconds (due to the overhead of BLS signature verification).

core/packages/contracts/scripts/configure-beefy.ts
```diff
@@ -11,7 +11,7 @@ import fs from "fs"
 import path from "path"
 
 let endpoint = process.env.RELAYCHAIN_ENDPOINT || "ws://127.0.0.1:9944"
-const beefyStartBlock = process.env.BEEFY_START_BLOCK ? parseInt(process.env.BEEFY_START_BLOCK) : 15
+const beefyStartBlock = process.env.BEEFY_START_BLOCK ? parseInt(process.env.BEEFY_START_BLOCK) : 1
```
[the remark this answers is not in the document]
yrong: Yes, work as expected.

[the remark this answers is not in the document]
vgeddes: nice!

yrong:
> Does this mean geth will produce an execution header every 2 seconds? If so, this could cause the light client to fall behind importing execution headers every 2 seconds (due to the overhead of BLS signature verification).

Yes, that's true. But the hack is for minimal spec and dev use only and we will not deploy these hacked config to any public network. It did help for fast check basic function like beacon/beefy work as expected so I would prefer keep it, also add an env config in this commit and will not enable fast mode by default.

https://github.com/Snowfork/snowbridge/pull/828/commits/dc488f007d50e558b8593371d6124484c3910f63

vgeddes (approved):
+1

core/packages/test/scripts/xcm-helper.sh
```diff
@@ -0,0 +1,193 @@
...
+    echo "--------------------------------------------------"
+
+    npx polkadot-js-api \
+        --ws "${relaychain_ws_url?}" \
+        --seed "${relaychain_sudo_seed?}" \
+        --sudo \
+        tx.xcmPallet.send \
+            "${dest}" \
+            "${message}"
+}
```
alistair-singh, line 193: Nit: end file with new line.

alistair-singh (approved):

claravanstaden:
@vgeddes @yrong Just mentioning, I would rather we not deviate from the official[ spec minimal config](https://github.com/ethereum/consensus-specs/blob/dev/configs/minimal.yaml), since it a standard spec that all the client implementation uses. Speeding up those configs may save time with testing, but we lose time with development since it has wasted some time figuring out why some proofs fail, only to find because the spec is different. I'd wager that more cases like this will occur if we deviate from the standard spec in the future.

vgeddes:
> @vgeddes @yrong Just mentioning, I would rather we not deviate from the official[ spec minimal config](https://github.com/ethereum/consensus-specs/blob/dev/configs/minimal.yaml), since it a standard spec that all the client implementation uses. Speeding up those configs may save time with testing, but we lose time with development since it has wasted some time figuring out why some proofs fail, only to find because the spec is different. I'd wager that more cases like this will occur if we deviate from the standard spec in the future.

That's a good point @claravanstaden. I see that these config parameters are all tightly related, and decreasing the slot time may interfere with ancestry proofs and the `SLOTS_PER_HISTORICAL_ROOT` setting. 

@yrong can you revert this change please. Since it is disabled by default, I think no one will actually use it anyway.

PR #835, opened 2023-05-17
Refactor verification of updates in beacon client

vgeddes:
We've duplicated a lot of verification code across the code paths for processing different kinds of updates. This makes the code harder to understand and maintain.

In some cases, the duplicated code differed slightly, leading to various bugs.

Fixes: SNO-438, SNO-444


vgeddes:
@yrong Would you mind take this PR over please?

After fixing some bugs, I discovered that some of the mock data used by `tests_minimal.rs` is incorrect, causing these failures:
```
failures:
    tests_minimal::it_updates_a_committee_period_sync_update
    tests_minimal::it_updates_a_invalid_committee_period_sync_update_with_duplicate_entry
```

The problem is this:
1. In `minimal_initial_sync.json`, `header.slot` is 152.
2. In `minimal_sync_committee_update.json`, `finalized_header.slot` is 128.

This is incorrect, as the finalized header in the initial checkpoint should always be older than finalized headers in future updates.

My code improvements here detected the bug: https://github.com/Snowfork/snowbridge/blob/8decb84e253a55d837df6e3c6fbeac884be0f4f5/parachain/pallets/ethereum-beacon-client/src/lib.rs#L706-L709

Can you update the mock data to fix this problem?

vgeddes:
@yrong We also need to update the relayer, as I changed some field names in `parachain/primitives/beacon/src/updates.rs`

claravanstaden:
Thanks for this refactor - I agree it is necessary.

yrong:
> Can you update the mock data to fix this problem?

Actually fixture data used here is retrieved from lodestar directly with script [generateBeaconData](https://github.com/Snowfork/snowbridge/blob/5a6d1cf2e07ef748866c44feafa678217d520afc/relayer/cmd/generate_beacon_data.go#L126) and 128 in `minimal_sync_committee_update.json` is actually the first slot in that sync period(i.e. period 2) also correct. 

Seems the problem is we retrieve all these updates in one command, maybe we can split to multiple commands and generate `sync_committee_update` for the next period(i.e. period 3).

parachain/primitives/beacon/src/updates.rs
```diff
@@ -86,14 +80,12 @@ pub struct FinalizedHeaderUpdate<const COMMITTEE_SIZE: usize, const COMMITTEE_BI
 	derive(Deserialize),
 	serde(deny_unknown_fields, bound(serialize = ""), bound(deserialize = ""))
 )]
-#[scale_info(skip_type_params(SyncCommitteeSize))]
-#[codec(mel_bound())]
 pub struct ExecutionHeaderUpdate<const COMMITTEE_SIZE: usize, const COMMITTEE_BITS_SIZE: usize> {
-	pub beacon_header: BeaconHeader,
-	pub execution_header: ExecutionPayloadHeader,
-	pub execution_branch: Vec<H256>,
+	pub attested_header: BeaconHeader,
```
yrong, line 84: AFAIK we do execution sync from last sync to latest finalized so `beacon_header` in it should all be finalized,  so `attested_header` here does not make sense.

parachain/pallets/ethereum-beacon-client/src/lib.rs
```diff
@@ -183,13 +185,16 @@ pub mod pallet {
 		StorageMap<_, Identity, H256, CompactExecutionHeader, OptionQuery>;
 
 	#[pallet::storage]
+	#[pallet::getter(fn validators_root)]
 	pub(super) type ValidatorsRoot<T: Config> = StorageValue<_, H256, ValueQuery>;
 
 	#[pallet::storage]
+	#[pallet::getter(fn latest_finalized_header)]
 	pub(super) type LatestFinalizedHeaderState<T: Config> =
```
yrong, line 193: would prefer we also rename to `LatestFinalizedHeader`

parachain/pallets/ethereum-beacon-client/src/lib.rs
```diff
@@ -183,13 +185,16 @@ pub mod pallet {
...
 		StorageValue<_, FinalizedHeaderState, ValueQuery>;
 
 	#[pallet::storage]
+	#[pallet::getter(fn latest_execution_header)]
 	pub(super) type LatestExecutionHeaderState<T: Config> =
```
yrong, line 198: would prefer we also rename to `LatestExecutionHeader`

vgeddes: Agree! Can you make the change?

parachain/pallets/ethereum-beacon-client/src/lib.rs
vgeddes: Agree! Can you make the change?

parachain/primitives/beacon/src/updates.rs
vgeddes: It is true that the header is already consider finalized. However we are also verifying it using `verify_attested_header`,  as the sync committee has signed it. In that sense, I think it is good to be consistent with the rest of the code, and name it `attested_header`.

parachain/pallets/ethereum-beacon-client/src/lib.rs
```diff
@@ -471,6 +471,10 @@ pub mod pallet {
 				&update.sync_aggregate,
 				update.signature_slot,
 			)?;
+			ensure!(
+				update.finalized_header.slot > Self::latest_finalized_header().beacon_slot,
+				Error::<T>::DuplicateFinalizedHeaderUpdate
+			);
 			let finalized_block_root = Self::verify_finalized_header(
 				&update.attested_header,
 				&update.finalized_header,
```
yrong: In commit https://github.com/Snowfork/snowbridge/pull/835/commits/faa4526f342ee190f885f945f80d4aa2d1f6357a
just move this check out from `verify_finalized_header`, because it does not apply to `process_sync_committee_period_update`

vgeddes: Good catch Ron 🙌 

I see that in the consensus spec for [validate_light_client_update](https://github.com/ethereum/consensus-specs/blob/dev/specs/altair/light-client/sync-protocol.md#validate_light_client_update), they don't have such a check either.

Instead in [apply_light_client_update](https://github.com/ethereum/consensus-specs/blob/dev/specs/altair/light-client/sync-protocol.md#apply_light_client_update), the finalized header is only applied if it is newer than the latest finalized header in storage. So we should probably follow the same approach if we decide to unify `sync_committee_period_update` and `import_finalized_header` in SNO-446.

yrong:
require https://github.com/Snowfork/cumulus/pull/25
E2E test should be fine now.

PR #837, opened 2023-05-18
Split implementation into several files

vgeddes:
I think the beacon code is easier to read and understand if we split some of into other files:

* `types.rs`: Types defined for use by client, and which cannot go into the primitives crate 
* `functions.rs`: Functions defined for use by the client, and which cannot go into the primitives crate
* `impls.rs`: Implementation of `Verifier` trait for message verification.

Other misc changes:
* Renamed `compute_current_sync_period` to `sync_period`.
* Removed unused `initialise_storage` function from `Verifier` trait.

### DIFF
--- a/parachain/pallets/ethereum-beacon-client/src/benchmarking/mod.rs
+++ b/parachain/pallets/ethereum-beacon-client/src/benchmarking/mod.rs
@@ -22,7 +22,7 @@ benchmarks! {
 		let initial_sync_data = initial_sync();
 		let sync_committee_update = initialize_sync_committee::<T>()?;
 
-		let period = EthereumBeaconClient::<T>::compute_current_sync_period(sync_committee_update.attested_header.slot);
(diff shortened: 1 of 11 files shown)


yrong:
Would it be better also move [pallet impl functions](https://github.com/Snowfork/snowbridge/blob/f1723a27d698aafbd754312341f9aa482b78ed31/parachain/pallets/ethereum-beacon-client/src/lib.rs#L353) to impls.rs?

yrong (approved):

vgeddes:
> Nit: would it be better also move [pallet impl functions](https://github.com/Snowfork/snowbridge/blob/f1723a27d698aafbd754312341f9aa482b78ed31/parachain/pallets/ethereum-beacon-client/src/lib.rs#L353) to impls.rs?

I felt to best to just move the impls on traits unrelated to beacon chain stuff.

PR #840, opened 2023-05-22
Follow consensus-specs sync protocol more closely

yrong:
For SNO-446

strict follows ALC spec only `apply_light_client_update` after successfully `validate_light_client_update`


yrong:
More checks added as following

https://github.com/Snowfork/snowbridge/blob/a948332e2c6619ae14881112082fbf3a0ed53298/parachain/pallets/ethereum-beacon-client/src/lib.rs#L263-L267

https://github.com/Snowfork/snowbridge/blob/a948332e2c6619ae14881112082fbf3a0ed53298/parachain/pallets/ethereum-beacon-client/src/lib.rs#L331-L337

https://github.com/Snowfork/snowbridge/blob/a948332e2c6619ae14881112082fbf3a0ed53298/parachain/pallets/ethereum-beacon-client/src/lib.rs#L375-L380

parachain/primitives/beacon/src/merkle_proof.rs
```diff
@@ -9,13 +9,13 @@ pub fn verify_merkle_branch(
 	index: usize,
 	depth: usize,
 	root: H256,
-) -> Option<bool> {
+) -> bool {
```
vgeddes, line 12: I like this change. Much simpler 👍

parachain/pallets/ethereum-beacon-client/src/lib.rs
```diff
@@ -242,29 +236,37 @@ pub mod pallet {
...
 			// Verify update does not skip a sync committee period
 			ensure!(
-				update.signature_slot > update.attested_header.slot,
-				Error::<T>::InvalidSignatureSlot
+				update.signature_slot > update.attested_header.slot &&
+					update.attested_header.slot >= update.finalized_header.slot,
+				Error::<T>::InvalidUpdateSlot
 			);
-			let store_period = compute_period(Self::latest_finalized_header().beacon_slot);
+			let latest_state = Self::latest_finalized_header();
```
vgeddes, line 268: Be aware that in my latest commits, that the `LatestFinalizedHeader` storage item has been renamed to `LatestFinalizedBlockRoot` and only contains the block root. You need to make a second lookup with the block root in `FinalizedBeaconState` to get the slot number.

vgeddes (commented):
Thanks Ron, this is also a good improvement.

Just wanted to say, we don't have to strictly follow the design in ALC consensus-specs. We can also improve on it, and change the algorithm if it makes sense.

yrong:
> Just wanted to say, we don't have to strictly follow the design in ALC consensus-specs. We can also improve on it, and change the algorithm if it makes sense.

IMHO we first follow spec as closely as possible which make it easy to find missing checks as above, also make audit/review a bit easier. Then we do optimization based on that.

parachain/pallets/ethereum-beacon-client/src/lib.rs
```diff
@@ -290,14 +303,66 @@ pub mod pallet {
 				(update.next_sync_committee_update.is_some() &&
 					update_attested_period == store_period);
 			ensure!(
-				update.attested_header.slot > store_period || update_has_next_sync_committee,
```
yrong, line 293: bug fix here. should be `update.attested_header.slot > latest_finalized_state.slot`

parachain/pallets/ethereum-beacon-client/src/lib.rs
```diff
@@ -243,15 +243,15 @@ pub mod pallet {
...
-				.is_some_and(|x| x),
+				),
 				Error::<T>::InvalidBlockRootsRootMerkleProof
 			);
 
 			let sync_committee_prepared: SyncCommitteePrepared = (&update.current_sync_committee)
 				.try_into()
 				.map_err(|_| <Error<T>>::BLSPreparePublicKeysFailed)?;
 			<CurrentSyncCommittee<T>>::set(sync_committee_prepared);
+			<NextSyncCommittee<T>>::kill();
```
yrong, line 254: Kill `NextSyncCommittee` here.

relayer/cmd/generate_beacon_data.go
```diff
@@ -171,7 +172,6 @@ func generateBeaconData(cmd *cobra.Command, _ []string) error {
 			return fmt.Errorf("get initial sync: %w", err)
 		}
 		initialSync := initialSyncScale.ToJSON()
-		err = writeJSONToFile(initialSync, fmt.Sprintf("initial-checkpoint.%s.json", activeSpec.ToString()))
```
yrong, line 174: As mentioned before, we can not use `initialSync` generated in same command together with other updates.

vgeddes:
See this comment here https://github.com/Snowfork/snowbridge/pull/839#discussion_r1201595469. I don't want to prune FinalizedBeaconState right now, because it depends on `FinalizedCheckpoints`/`FinalizedHeaderSlotsCache`. I want to remove that storage item.

relayer/cmd/generate_beacon_data.go
vgeddes: How can we improve the command so that it generates all the updates? Surely this possible to implement.

yrong: Resolved in https://github.com/Snowfork/snowbridge/pull/840/commits/7fda5f4b26e785095ead35230eb9af043ef6b5d6

relayer/cmd/generate_beacon_data.go
```diff
@@ -172,11 +173,19 @@ func generateBeaconData(cmd *cobra.Command, _ []string) error {
...
 		initialSync := initialSyncScale.ToJSON()
+		err = writeJSONToFile(initialSync, fmt.Sprintf("initial-checkpoint.%s.json", activeSpec.ToString()))
 		initialSyncHeaderSlot := initialSync.Header.Slot
+		initialSyncPeriod := s.ComputeSyncPeriodAtSlot(initialSyncHeaderSlot)
 		log.Info("created initial sync file")
 
 		log.Info("downloading beacon state, this can take a few minutes...")
-		syncCommitteePeriod := s.ComputeSyncPeriodAtSlot(initialSyncHeaderSlot)
+		// wait for 5 blocks
+		time.Sleep(6 * time.Second * 5)
```
vgeddes, line 183: Is the block time 6s in our E2E stack? Curious if we configured this explicitly somewhere

You should probably make it wait some more time in case the 5th block hasn't been produced yet.

relayer/cmd/generate_beacon_data.go
```diff
@@ -172,11 +173,19 @@ func generateBeaconData(cmd *cobra.Command, _ []string) error {
...
 		}
 		initialSync := initialSyncScale.ToJSON()
+		err = writeJSONToFile(initialSync, fmt.Sprintf("initial-checkpoint.%s.json", activeSpec.ToString()))
 		initialSyncHeaderSlot := initialSync.Header.Slot
+		initialSyncPeriod := s.ComputeSyncPeriodAtSlot(initialSyncHeaderSlot)
 		log.Info("created initial sync file")
 
 		log.Info("downloading beacon state, this can take a few minutes...")
-		syncCommitteePeriod := s.ComputeSyncPeriodAtSlot(initialSyncHeaderSlot)
+		// wait for 5 blocks
```
vgeddes, line 182: Is 5 special number here?

relayer/cmd/generate_beacon_data.go
yrong: Yes, it's 6s for minimal spec and we did not hack it for this test. Waiting for 5 blocks is just a decent value to make sure `FinalizedUpdate` change and `SyncPeriod` does not change.  This is requied by recently added checks in our light client.
There is chance at some time we run command fail to get data meet requirements,  in that case we just fail with error as following.
https://github.com/Snowfork/snowbridge/blob/3accd341008bc0743fc972b8b10663ecc032dc09/relayer/cmd/generate_beacon_data.go#L214-L219

parachain/pallets/ethereum-beacon-client/src/lib.rs
vgeddes: Nice catch!

PR #842, opened 2023-05-24
Remove FinalizedHeaderSlotsCache

yrong:
For sno-449

Also some discussions around not finalized there https://linear.app/snowfork/issue/SNO-449#comment-2ce1536a

### DIFF
--- a/relayer/relays/beacon/header/header.go
+++ b/relayer/relays/beacon/header/header.go
@@ -48,12 +48,7 @@ func (h *Header) Sync(ctx context.Context, eg *errgroup.Group) error {
 
 	log.WithField("period", latestSyncedPeriod).Info("set cache: last beacon synced sync committee period")
 
-	finalizedSlots, err := h.writer.GetFinalizedSlots()
-	if err != nil {
-		return fmt.Errorf("fetch parachain finalized header slots: %w", err)
-	}
-
-	h.cache.AddCheckPointSlots(finalizedSlots)
+	h.cache.AddCheckPointSlots([]uint64{lastFinalizedHeaderState.BeaconSlot})
 
 	log.WithField("finalizedSlots", h.cache.Finalized.Checkpoints.Slots).Info("set cache: finalized checkpoint slots")
 
@@ -163,8 +158,6 @@ func (h *Header) SyncCommitteePeriodUpdate(ctx context.Context, period uint64) e
 		}
 	}
 
-	h.cache.AddCheckPoint(update.FinalizedHeaderBlockRoot, update.BlockRootsTree, uint64(update.Payload.FinalizedHeader.Slot))
-
 	log.WithFields(log.Fields{
 		"finalized_header_slot": update.Payload.FinalizedHeader.Slot,
 		"period":                period,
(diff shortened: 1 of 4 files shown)


relayer/relays/beacon/header/header.go
```diff
@@ -175,19 +168,10 @@ func (h *Header) SyncCommitteePeriodUpdate(ctx context.Context, period uint64) e
...
-	// lastFinalizedHeaderPeriod := h.syncer.ComputeSyncPeriodAtSlot(lastFinalizedHeader.BeaconSlot)
-
-	// Period + 1 because the sync committee update contains the next period's sync committee
-	// if lastFinalizedHeaderPeriod != period+1 {
-	// 	return fmt.Errorf("synced committee period %d not imported successfully", lastFinalizedHeaderPeriod)
-	// }
-
 	h.cache.SetLastSyncedSyncCommitteePeriod(period)
 
+	h.cache.AddCheckPoint(update.FinalizedHeaderBlockRoot, update.BlockRootsTree, uint64(update.Payload.FinalizedHeader.Slot))
```
yrong, line 173: Only add checkpoint after successfully submit update.

relayer/relays/beacon/header/header.go
```diff
@@ -448,6 +430,8 @@ func (h *Header) syncLaggingExecutionHeaders(ctx context.Context, lastFinalizedH
 		return fmt.Errorf("sync headers: %w", err)
 	}
 
+	h.cache.AddCheckPoint(lastFinalizedHeader, blockRootsProof.Tree, lastFinalizedSlot)
```
yrong, line 433: Only add checkpoint after successfully submit update.

relayer/relays/beacon/header/header.go
```diff
@@ -48,12 +48,7 @@ func (h *Header) Sync(ctx context.Context, eg *errgroup.Group) error {
 
 	log.WithField("period", latestSyncedPeriod).Info("set cache: last beacon synced sync committee period")
 
-	finalizedSlots, err := h.writer.GetFinalizedSlots()
-	if err != nil {
-		return fmt.Errorf("fetch parachain finalized header slots: %w", err)
-	}
-
-	h.cache.AddCheckPointSlots(finalizedSlots)
+	h.cache.AddCheckPointSlots([]uint64{lastFinalizedHeaderState.BeaconSlot})
```
yrong, line 51: `lastFinalizedSlot` is always the start checkpoint.

relayer/relays/beacon/header/header.go
yrong, line 173: Only cache Checkpoint after successfully submit update.

vgeddes (commented):
If the relayer is only adding the last finalized slot to its checkpoint cache, how it be able to find the closest checkpoint to produce ancestry proofs for much older execution headers?

Let me put it another way, when the relayer starts up, it will call `syncLaggingExecutionHeaders` to sync historical execution headers. So we need historical checkpoints in the cache, surely?

Most of the time, `syncLaggingExecutionHeaders` won't need to sync older execution headers. However, when there are exceptional circumstances, such as network errors, execution header syncing could lag behind perhaps.

vgeddes (approved):
Happy to merge this in now since the E2E tests work.

yrong:
> However, when there are exceptional circumstances, such as network errors, execution header syncing could lag behind perhaps.

That depends on how far execution header fall behind, if no more than 8192 blocks(i.e. 27.3 days) which I would assume unlikely to happen, then latest finalized slot should just be fine as checkpoint and there is no need to add more checkpoints in this case.

And we just need to add some check to make sure `LatestExecutionState` not fall behind `LatestFinalizedState` too much , I'm pretty sure we can find a way to achieve it but not the scope of this PR.

PR #845, opened 2023-05-25
Update foundry overlay

doubledup:
An [overlay commit](https://github.com/shazow/foundry.nix/commit/1435fd712725dfbfbe92293e1405b147bc92867b) was just created that references a [Foundry build that exists](https://github.com/foundry-rs/foundry/releases/tag/nightly-a26edce5d2e1ad28d833328b22e857ecb7075e63). We can use this for now, still not sure what caused it other than weird results from the Github API at the time of the overlay update.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -54,11 +54,11 @@
         "nixpkgs": "nixpkgs"
       },
       "locked": {
-        "lastModified": 1684228155,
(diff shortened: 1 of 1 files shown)


yrong (approved):
LGTM!

alistair-singh (approved):

PR #850, opened 2023-05-30
Merge from bko-transfer-assets-via-bridge

alistair-singh:
Cumulus Companion: https://github.com/Snowfork/cumulus/pull/30


vgeddes (approved):
+1

doubledup (approved):
Looks like the cumulus submodule needs to be updated from c0c986 to the companion's 35e005.

PR #780, opened 2023-02-24
Ron/polkadot v0.9.38 12857

yrong:
Refactoring for these prs:
https://github.com/paritytech/substrate/pull/12857
https://github.com/paritytech/polkadot/pull/6577

PR #851, opened 2023-05-31
Reimplement OutboundQueue pallet

vgeddes:
The existing implementation was deficient in that it dropped messages if the maximum number of pending messages was breached. Due to the nature of our bridging design and related weight calculations, there is a hard limit on the number of outgoing messages which can be committed per block. Any messages beyond that limit need to be buffered somewhere, or dropped.

Dropping messages is bad for many reasons:
* Potential loss of funds if user's messages are dropped rather than being sent over the bridge
* Cross-chain governance messages could be dropped, leading to loss of control over the bridge.

The new implementation takes advantage of a new generalised message queueing pallet in Substrate: [MessageQueue](https://paritytech.github.io/substrate/master/pallet_message_queue/index.html). Its main responsibility is to store pending messages in a PoV-efficient manner, and coordinate their processing by other pallets. It never drops messages.

At a very high-level, the messaging flow now looks like this
* Message is first submitted to `OutboundQueue` via `OutboundQueue::submit`
* `OutboundQueue` forwards message to `MessageQueue`
* `MessageQueue` hooks back into `OutboundQueue` to process the message when there is enough remaining weight in the block

Other improvements:
* The `snowbridge-query-events` tool is no longer needed, since the pallet stores the committed messages in a way that GSRPC can now query using normal storage queries.
* The final process of producing a message commitment is much more efficient and easier to benchmark, by doing some of the processing when a message is first initially submitted.

Todo:
- [x] Update cumulus runtimes
- [x] Update relayer

Cumulus companion PR: https://github.com/Snowfork/cumulus/pull/36

Fixes: SNO-488, SNO-511, SNO-474


parachain/pallets/outbound-queue/src/lib.rs
```diff
@@ -92,153 +111,184 @@ pub mod pallet {
...
+				AggregateMessageOrigin::Parachain(origin),
+			);
+			Self::deposit_event(Event::MessageQueued { xcm_hash });
+			Ok(())
 		}
+	}
+
+	impl<T: Config> ProcessMessage for Pallet<T> {
+		type Origin = AggregateMessageOrigin;
+		fn process_message(
```
yrong, line 275: Seems `ProcessMessage` is also invoked in `on_initialized` [here](
https://github.com/paritytech/substrate/blob/d9cffa6bd50f17beb6f392bd1266c3fe8ec71898/frame/message-queue/src/lib.rs#L560-L568),  seems we need to make sure `outbound-queue` pallet is defined before `messge-queue` in runtime, right?

parachain/pallets/outbound-queue/src/lib.rs
```diff
@@ -92,153 +111,184 @@ pub mod pallet {
...
-			));
-			if weight_remaining.ref_time() <= MINIMUM_WEIGHT_REMAIN_IN_BLOCK.ref_time() {
-				return total_weight
-			}
-			Self::commit(total_weight)
+		fn on_initialize(_: T::BlockNumber) -> Weight {
+			// Remove storage from previous block
+			Messages::<T>::kill();
+			MessageLeaves::<T>::kill();
+			return T::WeightInfo::on_finalize()
```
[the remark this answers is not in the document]
vgeddes: The way it works is that if a pallet has an `on_finalize` handler, it needs to return the weight of `on_finalize` in `on_initialize`. This reserves some block weight for `on_finalize` to run successfully.

For that reason, you'll notice that the `on_finalize` does not return the weight it consumed, since it's scheduled to run right at the end of block production.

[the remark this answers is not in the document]
yrong: then should it be `T::WeightInfo::on_finalize()` + 2 write in `on_initialize` ?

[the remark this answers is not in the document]
vgeddes: Not sure if deleting storage counts as a read, will check,

parachain/pallets/outbound-queue/src/lib.rs
vgeddes: Yeah, that's a good point! Nice catch. Will fix.

parachain/pallets/outbound-queue/src/lib.rs
```diff
@@ -92,153 +111,184 @@ pub mod pallet {
...
+		fn process_message(
+			message: &[u8],
+			_: Self::Origin,
+			meter: &mut frame_support::weights::WeightMeter,
+		) -> Result<bool, ProcessMessageError> {
+			ensure!(
+				MessageLeaves::<T>::decode_len().unwrap_or(0) <
+					T::MaxMessagesPerBlock::get() as usize,
+				ProcessMessageError::Yield
+			);
```
yrong, line 284: Is `MaxMessagesPerBlock` hard limit check still nessessary here? `MessageQueue` will just schedule dynamically based on current block limit assume we can provide accurate weight for `do_process_message`

yrong: I think so or it's not easy to benchmark `on_finalize` based on worst case.

parachain/pallets/outbound-queue/merkle-proof/src/lib.rs
yrong, line 1: Could we just remove it and reference externally from substrate upstream
https://github.com/paritytech/substrate/blob/master/utils/binary-merkle-tree/Cargo.toml

parachain/pallets/outbound-queue/src/lib.rs
vgeddes: yeah, correct 👍

core/packages/contracts/src/Registry.sol
```diff
@@ -0,0 +1,25 @@
...
+contract Registry is AccessControl {
+    bytes32 public constant ADMIN_ROLE = keccak256("ADMIN_ROLE");
+    bytes32 public constant REGISTER_ROLE = keccak256("REGISTER_ROLE");
+
+    mapping(bytes32 contractID => address) registry;
+
+    constructor() {
+        _grantRole(ADMIN_ROLE, msg.sender);
+        _setRoleAdmin(ADMIN_ROLE, ADMIN_ROLE);
+        _setRoleAdmin(REGISTER_ROLE, ADMIN_ROLE);
```
yrong, line 15: will `_grantRole(REGISTER_ROLE, msg.sender);` also required here?

core/packages/contracts/src/InboundQueue.sol
```diff
@@ -4,29 +4,27 @@ pragma solidity ^0.8.19;
...
+import {RegistryLookup} from "./RegistryLookup.sol";
+import {Auth} from "./Auth.sol";
+import {Vault} from "./Vault.sol";
+
 import {IRecipient} from "./IRecipient.sol";
-import {IVault} from "./IVault.sol";
 import {ParaID} from "./Types.sol";
 
-contract InboundQueue is AccessControl {
+contract InboundQueue is Auth, RegistryLookup {
```
yrong, line 15: Should it be inherited from `Registry` or `Auth`?

core/packages/contracts/src/InboundQueue.sol
```diff
@@ -51,15 +49,14 @@ contract InboundQueue is AccessControl {
...
     error InvalidNonce();
-    error InvalidHandler();
+    error InvalidRecipient();
     error NotEnoughGas();
 
-    constructor(IParachainClient _parachainClient, IVault _vault, uint256 _reward) {
-        _grantRole(ADMIN_ROLE, msg.sender);
-        _setRoleAdmin(ADMIN_ROLE, ADMIN_ROLE);
+    constructor(Registry registry, IParachainClient _parachainClient, Vault _vault, uint256 _reward)
+        RegistryLookup(registry)
```
yrong, line 58: Also need to add construct `Registry()` explicit construct from parent.

parachain/pallets/outbound-queue/src/lib.rs
vgeddes: The `MessageQueue` pallet is now registered after the `OutboundQueue` pallet

core/packages/contracts/src/OutboundQueue.sol
```diff
@@ -3,25 +3,25 @@ pragma solidity ^0.8.19;
...
     uint256 public fee;
 
     event FeeUpdated(uint256 fee);
 
     error FeePaymentToLow();
 
-    constructor(IVault _vault, uint256 _fee) {
-        _grantRole(ADMIN_ROLE, msg.sender);
-        _setRoleAdmin(ADMIN_ROLE, ADMIN_ROLE);
+    constructor(Registry registry, Vault _vault, uint256 _fee) RegistryLookup(registry) {
```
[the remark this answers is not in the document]
vgeddes: Its not being used yet, but could be in future.

core/packages/contracts/src/NativeTokens.sol
```diff
@@ -53,57 +52,55 @@ contract NativeTokens is AccessControl, IRecipient {
...
+    function lock(address token, ParaID dest, bytes calldata recipient, uint128 amount) external payable {
         if (amount == 0) {
             revert InvalidAmount();
         }
 
         vault.deposit(msg.sender, token, amount);
 
-        bytes memory payload = NativeTokensTypes.Mint(token, dest, recipient, amount);
-        outboundQueue.submit{ value: msg.value }(assetHubParaID, payload);
+        bytes memory payload = NativeTokensTypes.Mint(address(registry), token, dest, recipient, amount);
```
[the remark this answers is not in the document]
vgeddes: Yes, indeed.

All assets transferred across the bridge to Asset Hub should have a Location of the form: `/GlobalConsensus(Ethereum)/AccountKey20(Registry)/AccountKey20(TokenContractAddress)`

core/packages/contracts/src/Registry.sol
vgeddes: The deployer address does that in `DeployScript.sol`

core/packages/contracts/src/InboundQueue.sol
vgeddes: It needs to inherit from both (multiple inheritance). `Auth.sol` and `RegistryLookup` offer different capabilities to `InboundQueue`.

core/packages/contracts/src/InboundQueue.sol
vgeddes: I think you mean an explicit call to `RegistryLookup` constructor, which I am doing.

core/packages/contracts/src/utils/MerkleProof.sol
```diff
@@ -0,0 +1,60 @@
...
+            return false;
+        }
+        return root == computeRoot(leaf, position, width, proof);
+    }
+
+    function computeRoot(bytes32 leaf, uint256 position, uint256 width, bytes32[] memory proof)
+        internal
+        pure
+        returns (bytes32)
+    {
```
doubledup, line 36: Nice! I was just making this change as a follow-up and got merge conflicts. Should we keep some `require`s here? Might want a `MAX_INT` constant instead of hard-coding the max value.
```suggestion
    {
        require(position < width, "Merkle position is too high");
        require(position < 0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff);
```

doubledup:
@yrong That's exactly right, I missed that in #861. Thanks for catching it!

core/packages/contracts/src/utils/MerkleProof.sol
vgeddes: Also I moved said check to `verify`

vgeddes:
> @doubledup @vgeddes Seems #861 is merged here. I remember when coordinate with [paritytech/substrate#12857](https://github.com/paritytech/substrate/pull/12857) there are several changes in relayer side I made as following: https://github.com/Snowfork/snowbridge/pull/780/files#diff-f8e330734941339a0ff763ff8d35f39639595660b175b591697a38163a436ba2
> 
> But I can not find the counterpart revert in this PR, is that not required or we missed here?

Thanks Ron, I definitely overlooked those changes. As the relayer is currently failing when it tries to call `BeefyClient.submitFinal`.

core/packages/contracts/src/utils/MerkleProof.sol
doubledup: The intention was to check that the `position + 1` value in `computeRoot` doesn't overflow, but that should be implicitly covered by `position < width`

PR #852, opened 2023-06-02
Beacon Client comments

claravanstaden:
Adds some explanatory comments to the beacon client.

### DIFF
--- a/parachain/pallets/ethereum-beacon-client/src/lib.rs
+++ b/parachain/pallets/ethereum-beacon-client/src/lib.rs
@@ -202,6 +202,7 @@ pub mod pallet {
 		#[pallet::call_index(0)]
 		#[pallet::weight(T::WeightInfo::force_checkpoint())]
 		#[transactional]
+		// Used for pallet initialization and light client resetting.
 		pub fn force_checkpoint(origin: OriginFor<T>, update: CheckpointUpdate) -> DispatchResult {
 			ensure_root(origin)?;
 			Self::process_checkpoint_update(&update)?;
@@ -216,6 +217,7 @@ pub mod pallet {
 			}
 		})]
 		#[transactional]
+		// Submits a new finalized beacon header update.
 		pub fn submit(origin: OriginFor<T>, update: Update) -> DispatchResult {
 			ensure_signed(origin)?;
 			Self::process_update(&update)?;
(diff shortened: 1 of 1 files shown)


vgeddes (commented):
A good start. Remember to use doc comment syntax, to ensure we can convert the comments into rust docs.

Particularly function comments should start with a triple forward slash.

https://doc.rust-lang.org/rust-by-example/meta/doc.html#doc-comments

yrong (approved):
Awesome!

parachain/pallets/ethereum-beacon-client/src/impls.rs
```diff
@@ -68,9 +68,8 @@ impl<T: Config> Verifier for Pallet<T> {
 }
 
 impl<T: Config> Pallet<T> {
-	// Verifies that the receipt encoded in proof.data is included
-	// in the block given by proof.block_hash. Inclusion is only
-	// recognized if the block has been finalized.
+	/// Verifies that the receipt encoded in proof.data is included in the block given by
+	/// proof.block_hash. Inclusion is only recognized if the block has been finalized.
```
vgeddes, line 72:
```suggestion
	/// Verifies that the receipt encoded in `proof.data` is included in the block given by
	/// `proof.block_hash`. Inclusion is only recognized if the block has been finalized.
```

parachain/pallets/ethereum-beacon-client/src/lib.rs
```diff
@@ -291,16 +304,18 @@ pub mod pallet {
 			Ok(())
 		}
 
-		// Cross check to make sure ExecutionHeader not fall behind FinalizedHeader too much, if
-		// that happens just return error so to pause processing FinalizedHeader until
-		// ExecutionHeader catch up
+		/// Cross check to make sure the ExecutionHeader does not fall too far behind
+		/// FinalizedHeader updates. If that happens just return an error so to pause processing
+		/// FinalizedHeader until ExecutionHeader processing has caught up.
```
vgeddes, line 309: We don't really have a `FinalizedHeader`type, so I felt this would read better:

```suggestion
		/// Cross check to make sure that execution header import does not fall too far behind
		/// finalised beacon header import. If that happens just return an error and pause processing
		/// until execution header processing has caught up.
```

vgeddes (approved):
Thanks Clara, looks good.

PR #855, opened 2023-06-12
Improve e2e scripts

yrong:
Refactoring including

- [x] Upgrade polkadot to 0.9.42 & geth to V1.12
- [x] Refactoring for goerli setup
- [x] Verify contracts in etherscan


core/packages/test/scripts/set-env.sh
```diff
@@ -127,3 +126,14 @@ check_tool() {
         exit
     fi
 }
+
+wait_contract_deployed() {
+    ready=""
```
vgeddes, line 131: To ensure the variable isn't set in the global scope, use [local](https://tldp.org/LDP/abs/html/localvar.html):

```suggestion
    local ready=""
```

flake.nix
```diff
@@ -40,7 +40,7 @@
 
                     # ethereum
                     foundry-bin
-                    go-ethereum
+                    # go-ethereum
```
vgeddes, line 43: @doubledup can you try bump the version in Nix again please.

core/packages/test/deployment/goerli-cd9e04958cb72e3fb1026c80b0e6ed032c9e83a6-1686329993.json
```diff
@@ -0,0 +1,3925 @@
+{
```
vgeddes, line 1: Should we add this file to gitignore, or is it being used somewhere?

core/packages/contracts/src/DeployScript.sol
vgeddes, line 1: The foundry book recommends having all scripts in a `scripts` dir: https://book.getfoundry.sh/projects/project-layout.

I think putting the script in the `src` dir could confuse various tools such as test coverage, etc.

flake.nix
vgeddes: Actually @yrong, is it necessary to upgrade to go-ethereum 1.12? Can we wait for a bit longer?

yrong: Seems nix packages for 1.12 is not ready yet, For upstreams components(e.g. Polkadot,Cumulus,Geth,Lodestar) we heavily depend on I would prefer to sync more closely to recent release without waiting nix counterpart to be ready. 

Btw, Geth client used here is only for dev setup and I've tested E2E still work as expected. For production usage I think we can still connect to infura endpoint as we did before.

core/packages/test/deployment/goerli-cd9e04958cb72e3fb1026c80b0e6ed032c9e83a6-1686329993.json
yrong: Yes, the deployment file here is just for placeholder and I can delete it if nessessary.  But IMO saving deployment history somewhere could be useful for production usage when we would like to debug on chain or upgrade from previous. The naming rule here is `${network}-${commitId}-${timestamp}`.

core/packages/contracts/src/DeployScript.sol
yrong: Met an issue as following when verify in etherscan and resolved after moving to `src`
https://github.com/foundry-rs/foundry/issues/2645#issuecomment-1212320518

flake.nix
vgeddes: ok, sounds fine :+1:

core/packages/contracts/src/DeployScript.sol
vgeddes: Ok, thanks for background info. Actually now I prefer to put it into `src`, since I don't want two directories `script` and `scripts`.

vgeddes (approved):
+1

claravanstaden (approved):
Looks great! I'll test this out as soon as my local setup works again.

doubledup (approved):

flake.nix
yrong: Exactly, but I would prefer these locally running clients are only readable for fast api response(reading from public endpoint could be slow), for transactions we can still submit to the infura endpoint as before.

PR #856, opened 2023-06-12
Downgrade Rust to 1.69.0

doubledup:
Opened as a draft until it's successfully built locally.

This avoids an issue where the sign-ext WASM feature is enabled in Rust >= 1.70.0.

[cargo-contract issue](https://github.com/paritytech/cargo-contract/issues/1139)
[rust issue](https://github.com/rust-lang/rust/issues/109807)

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -127,11 +127,11 @@
         "nixpkgs": "nixpkgs_3"
       },
       "locked": {
-        "lastModified": 1684376381,
-        "narHash": "sha256-XVFTXADfvBXKwo4boqfg80awUbT+JgQvlQ8uZ+Xgo1s=",
+        "lastModified": 1686537156,
+        "narHash": "sha256-mJD80brS6h6P4jzwdKID0S9RvfyiruxgJbXvPPIDqF0=",
         "owner": "oxalica",
         "repo": "rust-overlay",
-        "rev": "7c9a265c2eaa5783bc18593b1aec39a85653c076",
+        "rev": "e75da5cfc7da874401decaa88f4ccb3b4d64d20d",
         "type": "github"
       },
       "original": {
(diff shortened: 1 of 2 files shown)


claravanstaden (approved):
🥳 Thanks @doubledup!

ascjones:
You guys are using `ink` `pallet-contracts`? :eyes:

doubledup:
1.69.0 gives the same opcode error. Trying 1.67.0 nightly...

vgeddes:
> You guys are using `ink` `pallet-contracts`? 👀

Hi @ascjones, nope we aren't using ink contracts. We were just referencing other projects which were also having issues with bleeding edge rust.

vgeddes:
@yrong is this downgrade also fine for you? It seems all the eth light client checks above have passed.

yrong:
> is this downgrade also fine for you? It seems all the eth light client checks above have passed.

I can not reproduce the issue though. But since we've box wrapped `PreparedPublicKey` previous issue https://github.com/paritytech/substrate/issues/14075  does not block me anymore. So I'm fine if the downgrade can solve problem here.

vgeddes:
As said on Slack, I want us to try 1.70 stable and see if that works for us.

yrong:
> I want us to try 1.70 stable and see if that works for us.

Though the issue here seems not rust toolchain related I did test some stable releases including v1.70, all work without problem so I think should be fine to switch to 1.70 stable.

yrong:
> I want us to try 1.70 stable and see if that works for us.

I did test with 1.70 stable and just work without problem. But as https://github.com/paritytech/cargo-contract/issues/1139 suggested 1.69 maybe safer.

doubledup:
Closing this in favour of #855

PR #862, opened 2023-06-17
Configure InboundQueue allow list

alistair-singh:
Resolves: SNO-506
Cumulus companion: https://github.com/Snowfork/cumulus/pull/37


yrong (approved):
LGTM!

PR #868, opened 2023-06-21
SNO-519: Prevent collisions on mmrRootHash

doubledup:
[SNO-519](https://linear.app/snowfork/issue/SNO-519)

Check for the [MMR_ROOT_ID](https://github.com/paritytech/substrate/blob/34134935dd894d8322c1de72e95cc5d37780ff32/primitives/consensus/beefy/src/payload.rs#L33) and SCALE-encoded vector length of the hash (32 << 2) when encoding commitments. This prevents relayers from submitting an arbitrary hash in the commitment as the MMR root. If they submit a hash that wasn't originally encoded as an MMR root, the BEEFY client's encoding of the commitment payload will change by prepending the payload id & vector length to the submitted hash. This encoding will have a different keccak256 hash: one which was not signed by the validators.


core/packages/contracts/src/BeefyClient.sol
```diff
@@ -463,6 +463,8 @@ contract BeefyClient is Ownable {
     function encodeCommitment(Commitment calldata commitment) internal pure returns (bytes memory) {
         return bytes.concat(
             commitment.payload.prefix,
+            hex"6d68", // "mh" (MMR_ROOT_ID)
+            hex"80", // 128 (32 << 2, compact encoding)
```
alistair-singh, line 467: I think a more meaningful comment for auditors would be to say that this compact u64 represents the length of the mmrRootHash.

vgeddes (commented):
I've realised that we need to require that relayers pass the [whole commitment structure](https://github.com/paritytech/substrate/blob/fe1f8ba1c4f23931ae89c1ada35efb3d908b50f5/primitives/consensus/beefy/src/commitment.rs#L31) to the light client (both in submitInitial and submitFinal).

This is so that a future off-chain equivocation detecting agent can report equivocations back to the Polkadot relay chain. In the equivocation report, the whole commitment needs to be included.

core/packages/contracts/src/BeefyClient.sol
doubledup: Tried out actually encoding `"mh"` and `32`, left the id's encoding in and hard-coded the length.

alistair-singh (approved):

yrong (approved):
LGTM!

doubledup:
@vgeddes Passed the full commitment structure in #872

PR #871, opened 2023-06-22
Improve upgrade pattern

yrong:
Some context:

Currently upgrade pattern requires temporarily contract deployed with params hardcoded which is a bit inflexible and not easy to manage, so in this PR we improve this pattern to be parameterized.

Currently beefy checkpoint allows only one-off initialized first time deployed, we may need to force checkpoint anytime at run time(i.e. same as we do for beacon checkpoint).

Currently in `snowbridge-control` pallet we allow upgrade call from any signed account which is risky. In this PR we ensure only root origin allowed and wrap script for demonstrate `UpdateOutboundFee` with the new upgrade pattern(also avaiable as an E2E test case to demonstrate message relayed from polkadot to ethereum).

### DIFF
--- a/core/packages/contracts/src/BeefyClient.sol
+++ b/core/packages/contracts/src/BeefyClient.sol
@@ -3,7 +3,7 @@
 pragma solidity 0.8.20;
 
 import {ECDSA} from "openzeppelin/utils/cryptography/ECDSA.sol";
-import {Ownable} from "openzeppelin/access/Ownable.sol";
+import {Auth} from "./Auth.sol";
 import {MerkleProof} from "./utils/MerkleProof.sol";
 import {Bitfield} from "./utils/Bitfield.sol";
 import {MMRProof} from "./utils/MMRProof.sol";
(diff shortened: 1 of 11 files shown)


vgeddes:
I've been having a lot of chats with Parity devs over the past week about the problem of upgrading Beefy. Looks like we are coming up with a solution, and will share more with you about it after the conference.

But in summary it will work like this: When BEEFY protocol is updated on the relay chain (for example, changing data structure of BEEFY commitment or MMR), the protocol will:
1) Emit commitments using the new data structures.
2) For compatibility with existing light clients, will continue to emit commitments using the old data structures (at least for temporary period of time)

This means that after a beefy protocol upgrade, our light client will continue to work using the previous version commitments, keeping the bridge working. That allows us to use cross-chain governance to upgrade the light client.


There are also some other design changes to upgrades in the pipeline - so I wouldn't continue working on this PR until we can sync again properly next week.

yrong:
Yes, also realize to upgrade beefy is not as easy as imagine before so remove it from PR to limit the scope. Just prefer not `renounceOwnership` when first time initialized, especially at this early phase when not stable enough(e.g could be helpful for issues like https://github.com/paritytech/substrate/issues/14436 without restarting the whole E2E setup).

And what do you think other [improvements](https://github.com/Snowfork/snowbridge/pull/871#issue-1769720707) as mentioned above?

PR #872, opened 2023-06-23
SNO-536: Pass full commitment to submitInitial

doubledup:
[SNO-536](https://linear.app/snowfork/issue/SNO-536)


vgeddes (commented):
Looks good! Though I realise I wasn't very clear. I also want us to pass `commitment.payload` in its original, unmodified form.

So instead of:
```
    struct Payload {
        bytes32 mmrRootHash;
        bytes prefix;
        bytes suffix;
    }
```

relayers must pass:
```
PayloadItem[] payload
```

where `PayloadItem` is:
```
struct PayloadItem {
    bytes2 id
    bytes data
}
```

As per the types defined in the Substrate repo: https://github.com/paritytech/substrate/blob/fe1f8ba1c4f23931ae89c1ada35efb3d908b50f5/primitives/consensus/beefy/src/payload.rs#L43

In the [slack thread](https://snowfork.slack.com/archives/G01BGMGLAC9/p1687419189770919) I gave some tips on how to SCALE-encode `Vec<PayloadItem>`

doubledup:
ah ok, I'll swap the types. This will increase gas costs and makes the code cleaner. Are there other benefits or costs I'm missing?

yrong (approved):
LGTM!
@doubledup Would you please also check https://github.com/Snowfork/snowbridge/pull/875 if nessessary to be included or not?

core/packages/contracts/src/BeefyClient.sol
```diff
@@ -473,18 +491,45 @@ contract BeefyClient is Ownable {
         }
     }
 
+    // TODO: do we need to handle duplicate MMR_ROOT_ID entries in the payload?
+    // Unnecessary if we have a guarantee of unique payloadIDs per item.
+    // eg. if payload was a key-value mapping like a hash table.
```
doubledup, line 496: @vgeddes We're taking the first MMR root payload item for now. Guessing this is fine, as multiple MMR roots for the same (finalised) block number is invalid (and probably grounds for slashing).

alistair-singh (approved):

core/packages/contracts/src/BeefyClient.sol
```diff
@@ -473,18 +503,39 @@ contract BeefyClient is Ownable {
...
-            hex"80", // ScaleCodec.encodeU64(commitment.payload.mmrRootHash.length)
-            commitment.payload.mmrRootHash,
-            commitment.payload.suffix,
+            encodePayload(commitment.payload),
             ScaleCodec.encodeU32(commitment.blockNumber),
             ScaleCodec.encodeU64(commitment.validatorSetID)
         );
     }
 
+    function encodePayload(PayloadItem[] calldata items) internal pure returns (bytes memory) {
```
vgeddes, line 528:
```suggestion
    function encodeCommitmentPayload(PayloadItem[] calldata items) internal pure returns (bytes memory) {
```

vgeddes (approved):

PR #874, opened 2023-06-26
Add format into pre-commit hook

yrong:
There is no logic but only format change.

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -14,7 +14,7 @@ chronic typos .
 # lint and format for relayer codes
 (cd relayer && chronic mage lint && chronic go fmt ./...)
 
-# cargo fmt
-(cd parachain && SKIP_WASM_BUILD= cargo fmt --all)
+# comment here and run `cargo +nightly fmt` manually without nix for https://github.com/oxalica/rust-overlay/issues/66#issuecomment-993816000
+#(cd parachain && SKIP_WASM_BUILD= cargo +nightly fmt --all)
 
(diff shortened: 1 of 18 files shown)


.husky/pre-commit
```diff
@@ -9,12 +9,12 @@ echo 'Running pre-commit hook...'
...
 # lint and format for core contracts and typescript codes
-(cd core && chronic pnpm lint)
+(cd core && chronic pnpm lint && pnpm format)
 
 # lint and format for relayer codes
 (cd relayer && chronic mage lint && chronic go fmt ./...)
 
 # cargo fmt
-#(cd parachain && SKIP_WASM_BUILD= cargo fmt --all)
+(cd parachain && SKIP_WASM_BUILD= cargo +nightly fmt --all)
```
doubledup, line 18: This gives me an error:
```
error: no such command: `+nightly`

	Cargo does not handle `+toolchain` directives.
	Did you mean to invoke `cargo` through `rustup` instead?
husky - pre-commit hook exited with code 101 (error)
```

Looks like it works outside of Nix, so likely something the overlay does. Have you tested this in a Nix shell?

doubledup: Happy to merge if we remove `nightly` for now. This skips a lot of checks, but adding `nightly` breaks the pre-commit hook in a Nix shell.
```suggestion
(cd parachain && SKIP_WASM_BUILD= cargo fmt --all)
```

doubledup (approved):

PR #875, opened 2023-06-26
Beefy Commitment with static common prefix

yrong:
From logic here there is only 1 entry in `Payload` vec struct
https://github.com/paritytech/substrate/blob/75be6e2a6cc86cfa11390a05f113a00c270cdb24/primitives/consensus/beefy/src/mmr.rs#L185

So prefix can be furthurly restricted and suffix can be removed.


doubledup (changes requested):
The MMR root is the only payload in the protocol, but we expect another payload on handover: https://github.com/paritytech/substrate/blob/75be6e2a6cc86cfa11390a05f113a00c270cdb24/primitives/consensus/beefy/src/commitment.rs#L35-L36

I'll look into this next, need to check our validator set handover process.

Even if we don't have multiple payload items right now, we need to build for them so that Parity can iterate on BEEFY without risking stalling our bridge:
This change saves gas now but prevents the bridge from handling any changes to commitment payloads. If parity deploys changes to commitment payloads, the bridge will stall until we deploy support for them - I'd rather not put us in that position.

yrong:
Good to know that. Though only `MmrRoot` is included seems it's possible to add something more into the Payload.

e.g. It's pretty easy to add `MmrProof`(which contains `BeefyNextAuthoritySet`) with runtime api `self.runtime.runtime_api().generate_proof(header.hash())` into current impl of [PayloadProvider](https://github.com/paritytech/substrate/blob/75be6e2a6cc86cfa11390a05f113a00c270cdb24/primitives/consensus/beefy/src/mmr.rs#L185),  and our relayer logic could be simplified a lot then.

IMO BeefyPayload is a very low level primitive and if somehow changed from Parity we can not seamless handle it without an upgrade. I agree the magic prefix here a bit obscure, but it’s tradeoff whether we choose gas efficient or more clear code logic and I prefer first more.

doubledup:
You're right there, it's only the MMR root included now. Looks like that comment I referenced is out of date.

We definitely can't use new payloads without an upgrade, but the bridge should keep working when new payloads are introduced.

PR #876, opened 2023-06-26
Inbound queue benchmarks

claravanstaden:
Resolves: [#SNO-515](https://linear.app/snowfork/issue/SNO-515)
Cumulus companion: https://github.com/Snowfork/cumulus/pull/39


parachain/pallets/inbound-queue/src/benchmarking.rs
```diff
@@ -1,2 +1,46 @@
...
+    use snowbridge_core::Proof;
+    use super::*;
+    use hex_literal::hex;
+    use sp_core::H256;
+
+    #[benchmark]
+    fn submit() -> Result<(), BenchmarkError> {
+        let caller: T::AccountId = whitelisted_caller();
+
+        let payload = hex!("f9011c94ee9170abfbf9421ad6dd07f6bdec9d89f2b581e0f863a01b11dcf133cc240f682dab2d3a8e4cd35c5da8c9cf99adac4336f8512584c5ada000000000000000000000000000000000000000000000000000000000000003e8a00000000000000000000000000000000000000000000000000000000000000001b8a00000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000000000000000000000000000000000004b000f000000000000000100d184c103f7acc340847eee82a0b909e3358bc28d440edffa1352b13227e8ee646f3ea37456dec701345772617070656420457468657210574554481235003511000000000000000000000000000000000000000000");
```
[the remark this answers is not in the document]
claravanstaden: Correct. 😄

parachain/pallets/inbound-queue/src/benchmarking.rs
```diff
@@ -1,2 +1,46 @@
...
+
+        let execution_header = CompactExecutionHeader::default();
+        let block_hash = H256::default();
+
+        T::Verifier::initialize_storage(block_hash, execution_header);
```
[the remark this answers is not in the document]
claravanstaden: Cool, will do!

parachain/pallets/inbound-queue/src/test.rs
```diff
@@ -89,6 +91,35 @@ impl pallet_balances::Config for Test {
...
+
+impl snowbridge_ethereum_beacon_client::Config for Test {
+	type RuntimeEvent = RuntimeEvent;
+	type ForkVersions = ChainForkVersions;
+	type MaxExecutionHeadersToKeep = ExecutionHeadersPruneThreshold;
+	type WeightInfo = ();
+}
+
 // Mock verifier
 pub struct MockVerifier;
```
yrong, line 125: Would prefer not use `mockVerifier` here because it's the core logic and maybe the most expensive call in `submit` and should be benchmarked as true as possible.

parachain/pallets/inbound-queue/src/test.rs
```diff
@@ -111,6 +142,8 @@ impl inbound_queue::Config for Test {
 	type XcmSender = ();
```
yrong, line 142: Maybe you can use a `mockXcmSender` here just need to implement the `SendXcm` trait return some mock value [here](https://github.com/paritytech/polkadot/blob/ff869d0cda604f9adf2b6f886a813552b6fb848f/xcm/src/v3/traits.rs#L512-L533)

parachain/pallets/inbound-queue/src/test.rs
claravanstaden: `MockVerifier` is only used for running the benchmark tests as tests, not running the actual benchmark. The actual benchmark executes the `verify` method in `impls`.

claravanstaden:
The inbound-queue-benchmark is functional now. I'm busy generating a test fixture with many transactions for a real-life representative receipts trie.

parachain/pallets/inbound-queue/src/benchmarking/mod.rs
```diff
@@ -0,0 +1,70 @@
...
+		<AllowList<T>>::put(create_allowlist::<T>());
+
+		let dest_para: ParaId = 1000u32.into();
+		let sovereign_account = dest_para.into_account_truncating();
+
+		// So that the receiving account exists
+		let _ = T::Token::mint_into(&caller, T::Token::minimum_balance().into());
+		// Fund the sovereign account (parachain sovereign account) so it can transfer a reward
+		// fee to the caller account
+		let _ = T::Token::mint_into(&sovereign_account, 10000u32.into());
```
yrong, line 39: Seems like some initialize operation, maybe better move to  where we construct the test runtime [here](https://github.com/Snowfork/snowbridge/blob/9577c65a740479c484062fca4f19eaa835fe74ab/parachain/pallets/inbound-queue/src/test.rs#L167-L175)

claravanstaden: Vincent mentioned this in yesterday's call, https://github.com/Snowfork/snowbridge/pull/876/files/9577c65a740479c484062fca4f19eaa835fe74ab#diff-d9161fdb08bb2142f833c3f0a33e6592196395f95bb183ba7c372cc432cf741bR52 only gets executed when running the benchmark as a test, not when executing the benchmark. So moving these operations will break the benchmark because the initialize method is only called when executing the tests.

parachain/pallets/inbound-queue/src/test.rs
yrong: Corret. But since you've already linked  `EthereumBeaconClient` as direct dependendy and initialized `ExecutionHeaderBuffer` storage so IMO `MockVerifier` is not required anymore. 

I would assume just config `type Verifier = snowbridge_ethereum_beacon_client::Pallet<Test>;` in test runtime should be fine. It will also increase the test coverage seems we don't have tests for verify logic yet.

parachain/pallets/ethereum-beacon-client/src/lib.rs
```diff
@@ -37,9 +37,10 @@ use functions::{
 	compute_epoch, compute_period, decompress_sync_committee_bits, sync_committee_sum,
 };
 use types::{
-	CheckpointUpdate, ExecutionHeaderBuffer, ExecutionHeaderUpdate, FinalizedBeaconStateBuffer,
+	CheckpointUpdate, ExecutionHeaderUpdate, FinalizedBeaconStateBuffer,
 	SyncCommitteePrepared, Update,
 };
+pub use types::ExecutionHeaderBuffer;
```
claravanstaden, line 43: This needs to be public to be set in the inbound-queue pallet.

claravanstaden (commented):
A comparison of weights by transaction count:
- 1 transaction - `112_000_000`
- 221 transactions - `121_000_000` - this is value produced by running the benchmark submitted in this PR.
- 421 transactions - `121_000_000`

Since nearly done the transactions does not seem to have an effect on the weight, I think the current weight should be sufficient. I would like to do another test with about 900 transactions, but will do so async and update the PR if needed. I encountered a problem with the beacon relayer while running the E2E stack, so I'll need to look at that first before running more tests.

alistair-singh (approved):

parachain/pallets/inbound-queue/Cargo.toml
```diff
@@ -30,6 +30,8 @@ xcm = { git = "https://github.com/paritytech/polkadot.git", branch = "master", d
 snowbridge-core = { path = "../../primitives/core", default-features = false }
 snowbridge-ethereum = { path = "../../primitives/ethereum", default-features = false }
 snowbridge-router-primitives = { path = "../../primitives/router", default-features = false }
+snowbridge-beacon-primitives = { path = "../../primitives/beacon", default-features = false }
+snowbridge-ethereum-beacon-client = { path = "../../pallets/ethereum-beacon-client", default-features = false }
```
vgeddes, line 34: See my previous comment, we don't want the pallet to directly depend on the beacon light client crate.

You can move these to `dev-dependencies` to though, as its fine to depend on these for tests.

parachain/pallets/inbound-queue/Cargo.toml
```diff
@@ -30,6 +30,8 @@ xcm = { git = "https://github.com/paritytech/polkadot.git", branch = "master", d
 snowbridge-core = { path = "../../primitives/core", default-features = false }
 snowbridge-ethereum = { path = "../../primitives/ethereum", default-features = false }
 snowbridge-router-primitives = { path = "../../primitives/router", default-features = false }
+snowbridge-beacon-primitives = { path = "../../primitives/beacon", default-features = false }
```
vgeddes, line 33: Can also try moving this dependency to the "runtime-benchmarks" feature below.

vgeddes (commented):
Looks good!

Though you still need to update the `bridge-hub-rococo` runtime to be aware of the benchmarks: https://github.com/Snowfork/cumulus/tree/1a8a815c224fa51a5df00a92eb28235b42defda6/parachains/runtimes/bridge-hubs/bridge-hub-rococo

I think @alistair-singh should be able to help you with this.

claravanstaden:
> By defining it here in the pallet, you're making it have a hard dependency on the beacon light client pallet.

@vgeddes thanks mentioning the dependency issue, I will have a look at that.

> Though you still need to update the bridge-hub-rococo runtime to be aware of the benchmarks:

I updated the bridge-hub-rococo runtime in the Cumulus companion (https://github.com/Snowfork/cumulus/pull/39). Is something else needed there?

vgeddes:
> I updated the bridge-hub-rococo runtime in the Cumulus companion ([Snowfork/cumulus#39](https://github.com/Snowfork/cumulus/pull/39)). Is something else needed there?

Oops I overlooked the companion PR.

yrong (approved):
LGTM!

vgeddes (commented):
Also need to update `weights.rs` in the pallet.

claravanstaden:
@vgeddes added weights in the inbound queue pallet. 😄

yrong:
@claravanstaden  Would suggest merge cumulus change together with changes here. Or other prs based on latest cumulus is likely to fail as following:
https://github.com/Snowfork/snowbridge/actions/runs/5586563561/jobs/10210752797?pr=883

claravanstaden:
@yrong apologies! It looked like I had to merge the cumulus companion to get my `check-cumulus-bridgehub` GA to pass. Is it OK if it fails until the companion is merged?

vgeddes (approved):
+1

PR #878, opened 2023-07-06
Remove hardcoded accounts

alistair-singh:
Removes hardcoded sovereign and relayer accounts. Accounts are funded via `start-services.sh`.

Resolves: SNO-528
Cumulus companion: https://github.com/Snowfork/cumulus/pull/40


yrong (approved):
LGTM!

PR #879, opened 2023-07-12
Distinct validator for submitInitial

yrong:


core/packages/contracts/src/BeefyClient.sol
```diff
@@ -257,6 +261,12 @@ contract BeefyClient is Ownable {
             revert NotEnoughClaims();
         }
 
+        uint256 claimerKey = uint256(keccak256(abi.encodePacked(msg.sender, proof.account, vset.id)));
```
doubledup, line 264: Note that scoping by `vset.id` here means a single validator address isn't allowed to relay more than `vset.length` commitments per era. This makes the average time for a relayer to relay each commitment from Polkadot at least `24*60 m / 297 ~= 4.85 m`. So this increases the minimum latency to about 5 minutes if we're running a single relayer.

But, we don't have to run a single relayer: n additional relayers divide this latency by n. Also as the number of validators on Polkadot increase, this latency decreases for each relayer. So this latency bound decreases both as we add more relayers and as the Polkadot network grows.

At the target 1000 validators (assuming an era remains 24h), average latency per relayer is `24*60 m / 1000 = 1.44 m = 87 s`. So with just 8 relayers the latency could drop below 12s.

yrong: Yes, this internally requires more relayers which is not bad IMO. And I would assume not a big deal even some non-handover commitments failed for `ClaimerAlreadyCaptured`.

PR #880, opened 2023-07-13
Upgrade ssz_rs crate.

claravanstaden:
Updates the ssz_rs crate to 0.9.0. The crate name was also changed again from `ssz-rs` back to `ssz_rs`.
Cumulus companion: https://github.com/Snowfork/cumulus/pull/42

### DIFF
--- a/parachain/Cargo.lock
+++ b/parachain/Cargo.lock
@@ -2585,8 +2585,8 @@ dependencies = [
  "sp-io",
  "sp-runtime",
  "sp-std",
- "ssz-rs",
- "ssz-rs-derive",
+ "ssz_rs",
+ "ssz_rs_derive",
  "static_assertions",
 ]
 
@@ -3255,28 +3255,6 @@ dependencies = [
  "serde",
  "serde_json",
  "unicode-xid",
-]
-
-[[package]]
-name = "ssz-rs"
-version = "0.8.0"
-source = "git+https://github.com/ralexstokes/ssz-rs?rev=d18af912abacbf84219be37ab3b42a9abcf10d2a#d18af912abacbf84219be37ab3b42a9abcf10d2a"
-dependencies = [
- "bitvec",
- "num-bigint",
- "sha2 0.9.9",
- "ssz-rs-derive",
- "thiserror",
-]
-
-[[package]]
-name = "ssz-rs-derive"
-version = "0.8.0"
-source = "git+https://github.com/ralexstokes/ssz-rs?rev=d18af912abacbf84219be37ab3b42a9abcf10d2a#d18af912abacbf84219be37ab3b42a9abcf10d2a"
-dependencies = [
- "proc-macro2",
- "quote",
- "syn 1.0.109",
 ]
 
 [[package]]
--- a/parachain/Cargo.lock
+++ b/parachain/Cargo.lock
@@ -2676,8 +2676,8 @@ dependencies = [
  "sp-keyring",
  "sp-runtime",
  "sp-std",
- "ssz-rs",
- "ssz-rs-derive",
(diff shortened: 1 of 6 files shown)


parachain/pallets/ethereum-beacon-client/Cargo.toml
```diff
@@ -14,8 +14,8 @@ targets = ["x86_64-unknown-linux-gnu"]
 serde = { version = "1.0.164", optional = true }
 codec = { version = "3.1.5", package = "parity-scale-codec", default-features = false, features = [ "derive" ] }
 scale-info = { version = "2.7.0", default-features = false, features = [ "derive" ] }
-ssz-rs = { git = "https://github.com/ralexstokes/ssz-rs", default-features = false, rev="d18af912abacbf84219be37ab3b42a9abcf10d2a" }
```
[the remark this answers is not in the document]
claravanstaden: Ah nice catch, thanks! 👍

claravanstaden:
I missed the primitive crate and that's where all the changes are - so putting this back into draft.

parachain/Cargo.lock
```diff
@@ -2676,8 +2676,8 @@ dependencies = [
  "sp-keyring",
  "sp-runtime",
  "sp-std",
- "ssz-rs",
- "ssz-rs-derive",
+ "ssz_rs",
```
doubledup, line 2679: Also, from the looks of the test scripts, we're not actually using this lock file any more (except maybe in language servers). We only build in the `polkadot` & `cumulus` repos now, so the [lockfile in our Cumulus fork](https://github.com/Snowfork/cumulus/blob/1a8a815c224fa51a5df00a92eb28235b42defda6/Cargo.lock) needs this update in a companion PR on that repo.

doubledup: This shouldn't impact testing, as the Cumulus build will still use the updated `Cargo.toml`s here. But it'll change the lockfile in Cumulus, so we should commit it there.

parachain/pallets/outbound-queue/src/lib.rs
```diff
@@ -79,6 +79,12 @@ impl Into<Token> for Message {
 	}
 }
 
+impl From<u32> for AggregateMessageOrigin {
+	fn from(value: u32) -> Self {
+		AggregateMessageOrigin::Parachain(value.into())
+	}
+}
+
```
claravanstaden, line 87: Unrelated to this PR, but this breaks when building Cumulus when enabling the benchmarks feature.

parachain/primitives/beacon/src/ssz.rs
```diff
@@ -161,17 +178,17 @@ impl From<ExecutionPayloadHeader> for SSZExecutionPayloadHeader {
 			block_hash: payload.block_hash.to_fixed_bytes(),
 			transactions_root: payload.transactions_root.to_fixed_bytes(),
 			withdrawals_root: payload.withdrawals_root.to_fixed_bytes(),
-		}
+		})
 	}
 }
 
-pub fn hash_tree_root<T: SimpleSerialize>(mut object: T) -> Result<H256, MerkleizationError> {
+pub fn hash_tree_root<T: SimpleSerialize>(mut object: T) -> Result<H256, SimpleSerializeError> {
```
claravanstaden, line 185: Changed from `MerklizationError` to `SimpleSerializeError` since the error returned here could now also be a `DeserializeError`.

claravanstaden:
OK @doubledup, ready for another 👁️ ! Thanks!

yrong (approved):
LGTM!

doubledup (approved):

PR #881, opened 2023-07-13
Release relayer

yrong:
Resolves: SNO-558

Also include some fix for a non-local setup(i.e. use non-local/public endpoints for substrate chain nodes).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -12,7 +12,7 @@ on:
 
 env:
   REGISTRY: ghcr.io
-  IMAGE_NAME: ${{ github.repository }}
+  IMAGE_NAME: snowfork/snowbridge-relay
         
 jobs:
   release-binary:
--- a/web/packages/test/docker-compose.yml
+++ b/web/packages/test/docker-compose.yml
@@ -3,7 +3,7 @@ version: '3'
 services:
   beacon:
     platform: linux/amd64
-    image: ronyang/snowbridge-relay # change to snowfork
+    image: ghcr.io/snowfork/snowbridge-relay
     command: run beacon --config /config/beacon-relay.json --substrate.private-key "//BeaconRelay"
     volumes:
       - $output_dir:/config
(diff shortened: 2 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -3,36 +3,53 @@ name: release
...
+        uses: docker/login-action@v2.2.0
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+      - name: Build and push
+        run: |
+          docker buildx build \
+          --platform linux/amd64 \
+          -f relayer/Dockerfile \
+          -t "ronyang/snowbridge-relay:${{ github.event.inputs.tag || github.sha }}" \
```
yrong, line 53: @vgeddes Except for the binary format I would suggest we add docker image as another option for running relay service. This is also the recomended way for running most of the other components in our stack.

Problem is that I don't have authority to config the repo so use my own account here temporarily. We need to register an account in docker hub as orgnization and add secrets here so to push image.

.github/workflows/release.yml
```diff
@@ -3,36 +3,53 @@ name: release
 on:
   push:
     tags:
-      - v0.*
-      - v1.*
+      - "v*"
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: tag (ex. v1.0) to publish
```
claravanstaden, line 10:
```suggestion
        description: tag (e.g. v1.0) to publish
```

.github/workflows/release.yml
vgeddes: Actually in the past we've used GitHub Packages as a docker repository: https://github.com/Snowfork/snowbridge/pkgs/container/snowbridge-relay

Can you try and upload the image there to test?

yrong: Cool! Seems also require to config `GITHUB_TOKEN` for [Authenticating to the Container registry](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry#authenticating-to-the-container-registry), so is that already configured in this repo before?

yrong: Some revamp in https://github.com/Snowfork/snowbridge/pull/881/commits/81f3bf07519bdc432ed55168128edd747d131415 also add a [workflow_dispatch](https://github.com/Snowfork/snowbridge/blob/765709754952dcabd263ca3ece19af5fd3987a6c/.github/workflows/release.yml#L7-L11) for manually trigger the action for test.

But seems without first merging to main it's not easy to test the newly added action.

alistair-singh (approved):

PR #882, opened 2023-07-15
Contracts V3

vgeddes:
Several major changes:

1. Instead of a dozen contracts managing their own state, the bridge is now centred around a single contract design: `Gateway.sol`.
2. [Agent contracts](https://hackmd.io/aoXYv74lQVKciPh_e7uZlg?edit) for Polkadot parachains as discussed with Parity
3. I've switched over to the proxy upgrade pattern, and using a special technique to implement upgradable storage.

All in all, I think this is a major improvement over the old code. There's less code, and its easier to reason about.


core/packages/contracts/src/storage/CoreStorage.sol
```diff
@@ -0,0 +1,37 @@
...
+        uint256 defaultFee;
+        // The reward given to relayers for submitting inbound messages from Polkadot
+        uint256 defaultReward;
+        // Relayers must provide enough gas to cover message dispatch plus a buffer
+        uint256 gasToForward;
+        // Parachain ID of BridgeHub
+        ParaID bridgeHubParaID;
+    }
+
+    bytes32 internal constant SLOT = keccak256("org.snowbridge.storage.core");
```
musnit, line 29: should b versioned?

core/packages/contracts/src/AgentExecutor.sol
```diff
@@ -0,0 +1,50 @@
...
+
+    bytes32 public constant ACTION_UNLOCK_TOKENS = keccak256("unlockTokens");
+
+    event NativeTokensUnlocked(address token, address recipient, uint256 amount);
+
+    error InsufficientBalance();
+    error TransferEthFailed();
+    error Unauthorized();
+
+    function execute(address, bytes memory data) external {
```
yrong, line 28: Is it nessessary to check address here only allowed from Gateway?

core/packages/contracts/src/storage/CoreStorage.sol
```diff
@@ -0,0 +1,37 @@
...
+import {AgentExecutor} from "../AgentExecutor.sol";
+import {Agent} from "../Agent.sol";
+import {Channel, OperatingMode, ParaID} from "../Types.sol";
+
+library CoreStorage {
+    struct Layout {
+        // Operating mode:
+        OperatingMode mode;
+        // Message channels
+        mapping(ParaID paraID => Channel) channels;
```
musnit, line 14: as discussed, maybe make this mapping from an agentId instead of paraId?

core/packages/contracts/src/Gateway.sol
```diff
@@ -2,34 +2,445 @@
...
+            revert NotEnoughGas();
+        }
+
+        bool success = true;
+
+        if (message.command == COMMAND_AGENT_EXECUTE) {
+            try Gateway(this).handleAgentExecute{gas: $.gasToForward}(message.params) {}
+            catch {
+                success = false;
+            }
```
yrong, line 176: Just curious why not wrap `handle*` as internal call here?

core/packages/contracts/src/Gateway.sol
```diff
@@ -2,34 +2,445 @@
...
+        CoreStorage.Layout storage $ = CoreStorage.layout();
+        return $.agents[agentID];
+    }
+
+    /**
+     * Handlers
+     */
+
+    // Execute code within an agent
+    function handleAgentExecute(bytes calldata params) external onlySelf {
```
yrong, line 245: https://github.com/Snowfork/snowbridge/pull/882/files#r1265249658

core/packages/contracts/src/Gateway.sol
```diff
@@ -2,35 +2,476 @@
...
+        bytes32 agentID;
     }
 
-    function ensureOrigin(ParaID a, ParaID b) internal pure {
-        if (a != b) {
+    // Create an agent for a consensus system on Polkadot
+    function createAgent(bytes calldata data) external onlySelf {
+        CoreStorage.Layout storage $ = CoreStorage.layout();
+
+        CreateAgentParams memory params = abi.decode(data, (CreateAgentParams));
```
doubledup, line 230: Can we remove the separate decode step?
```suggestion
    function createAgent(CreateAgentParams calldata params) external onlySelf {
        CoreStorage.Layout storage $ = CoreStorage.layout();
```

core/packages/contracts/src/IGateway.sol
```diff
@@ -0,0 +1,63 @@
...
+
+    // Register a new fungible asset in the `ForeignAssets` pallet on AssetHub. This new asset
+    // is a wrapped token correspondig to the ERC20 token `token`.
+    //
+    // This instruction is idempotent, and will fail if `ForeignAssets`
+    function registerNativeToken(address token) external payable;
+
+    // Send ERC20 tokens to Polkadot. The bridged assets will be minted on AssetHub
+    // and then reserve transferred to `recipient` on `finalDestPara`.
+    function lockNativeTokens(address token, ParaID finalDestPara, bytes calldata recipient, uint128 amount)
```
doubledup, line 60: What's the path for transferring ETH? Is there a token address we can use, or should we have a function like `lockEth(ParaID finalDestPara, bytes calldata recipient, uint128 amount)`?

core/packages/contracts/src/Types.sol
```diff
@@ -2,18 +2,36 @@
 // SPDX-FileCopyrightText: 2023 Snowfork <hello@snowfork.com>
 pragma solidity 0.8.20;
 
-type ParaID is uint32;
+type ParaID is uint256;
```
doubledup, line 5: Is this type correct? `ParaId` is still a `u32` in the parachain: https://github.com/Snowfork/snowbridge/blob/a99fc50c1ebb23a3fd74e4c28b8bcb08b209b869/parachain/primitives/core/src/lib.rs#L21

doubledup: (resolves to this: https://github.com/paritytech/polkadot/blob/6a71a7ef7/parachain/src/primitives.rs#L154)

core/packages/contracts/src/AgentExecutor.sol
```diff
@@ -0,0 +1,32 @@
...
+            _transferToken(token, recipient, amount);
+        }
+    }
+
+    function transferNative(address payable recipient, uint256 amount) external {
+        recipient.safeNativeTransfer(amount);
+    }
+
+    function _transferToken(address token, address recipient, uint128 amount) internal {
+        IERC20(token).safeTransfer(recipient, amount);
```
yrong, line 30: Will this be
`IERC20(token).safeTransferFrom(assetHubAgent, recipient, amount);` ?

.gitmodules
```diff
@@ -12,7 +12,9 @@
 [submodule "core/packages/contracts/lib/openzeppelin-contracts"]
 	path = core/packages/contracts/lib/openzeppelin-contracts
 	url = https://github.com/openzeppelin/openzeppelin-contracts
-	branch = v4.8.2
 [submodule "core/packages/contracts/lib/canonical-weth"]
 	path = core/packages/contracts/lib/canonical-weth
 	url = https://github.com/snowfork/canonical-weth
+[submodule "lib/openzeppelin-contracts"]
+	path = lib/openzeppelin-contracts
```
alistair-singh, line 19: Should this be in `core/packages/contracts`?
```suggestion
	path =  core/packages/contracts/lib/openzeppelin-contracts
```

alistair-singh:
> @alistair-singh Seems you've introduced some format change in [d43b5ad](https://github.com/Snowfork/snowbridge/commit/d43b5ad4e3b8d486701569755a7e00121a73a750), is that required? Since current code is already formatted with `cargo +nightly fmt` so would suggest not change it.
> 
> And one minor comment [here](https://github.com/Snowfork/snowbridge/pull/892#discussion_r1269099068) FYI.

PR to revert:
https://github.com/Snowfork/snowbridge/pull/896

parachain/pallets/control/src/lib.rs
```diff
@@ -41,58 +44,64 @@ pub mod pallet {
...
 	}
 
 	#[pallet::error]
 	pub enum Error<T> {
+		UpgradeDataTooLarge,
 		SubmissionFailed,
 	}
 
+	#[pallet::storage]
+	pub type Agents<T: Config> = StorageMap<_, Twox64Concat, H256, (), OptionQuery>;
```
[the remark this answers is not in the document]
vgeddes: yeah, pretty much.

core/packages/contracts/src/Gateway.sol
vgeddes: `createAgent` is actually called by `submitInbound`. Since `abi.decode` can error and revert the entire transaction, it's safer to decode within `createAgent`.

Notice that `submitInbound` calls `createAgent` within a try catch block. However in Solidity, try catch is not able to handle errors from abi.decode.

core/packages/contracts/src/Gateway.sol
```diff
@@ -2,35 +2,476 @@
...
+        // Otherwise malicious relayers can break the bridge by allowing handlers to run out gas.
+        // Resubmission of the message by honest relayers will fail as the tracked nonce
+        // has already been updated.
+        if (gasleft() < GAS_FOR_DISPATCH + GAS_BUFFER) {
+            revert NotEnoughGas();
+        }
+
+        bool success = true;
+
+        if (message.command == COMMAND_AGENT_EXECUTE) {
```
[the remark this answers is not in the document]
vgeddes: Yeah good point. Will consider it.

core/packages/contracts/src/GatewayProxy.sol
```diff
@@ -0,0 +1,36 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2023 Snowfork <hello@snowfork.com>
+pragma solidity 0.8.20;
+
+import {ERC1967} from "./utils/ERC1967.sol";
+
+contract GatewayProxy {
```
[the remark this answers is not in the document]
vgeddes: Depends on the specific application of xcm. If we are sending Transact to Ethereum, then origin will be the external smart contract that initiated the transact.

For sending native assets to the ForeignAssets pallet of AssetHub, I think the origin should now be the address of the agent contract for AssetHub.

core/packages/contracts/src/AgentExecutor.sol
vgeddes: Nope.

The agent will invoke the `AgentExecutor.execute` function via `delegatecall`, and so `IERC20(token).safeTransfer` means that the tokens will be transferred from the agent to recipient.

See the `Gateway.agentExecute` function that implements this.

core/packages/contracts/src/Gateway.sol
vgeddes: The reason we use an external call is because its not possible to catch errors in internal calls using `try catch`.

If an internal call reverted, it would revert the entire transaction. And that could be problematic, as the channel would get blocked.

core/packages/contracts/src/AgentExecutor.sol
vgeddes: This code is always run via `Agent.invoke`, which does have a check for gateway.

core/packages/contracts/src/storage/CoreStorage.sol
vgeddes: I don't think there's enough time before the audit to properly spec the change and implement it.

core/packages/contracts/src/AgentExecutor.sol
vgeddes: Users definitely need to call `IERC20.approve` to allow the gateway contract to deposit the tokens into the agent.

See the test for sendNativeToken here: https://github.com/Snowfork/snowbridge/blob/7b7a037296a7698af28703cd525fd630ed282d8e/core/packages/contracts/test/Gateway.t.sol#L458

core/packages/contracts/src/Gateway.sol
yrong: Gotcha! Then will it make more sense if we make these `handle*` functions return some particular error code instead of the revert?  The error code could be also included in the event emitted.  In this way we can make these functions internal and also make client/relayer has more knowlege about the internal detail in ethereum side.

core/packages/contracts/src/AgentExecutor.sol
yrong: Yes that's true. But the `execute` function here is marked as external, without the address check how to differ from an EOA or from an attacker contract?

core/packages/contracts/src/Types.sol
```diff
@@ -2,18 +2,56 @@
...
+    ParaID origin;
+    uint64 nonce;
+    Command command;
+    bytes params;
+}
+
+enum OperatingMode {
+    Normal,
+    RejectingOutboundMessages
+}
```
yrong, line 37: Is mode `RejectingInboundMessages` nessessary here?

parachain/primitives/router/src/outbound/mod.rs
```diff
@@ -27,7 +25,7 @@ where
 	OutboundQueue: OutboundQueueTrait,
 	OutboundQueue::Ticket: Encode + Decode,
 {
-	type Ticket = (Vec<u8>, XcmHash);
+	type Ticket = Vec<u8>;
```
yrong, line 28: Is removing XcmHash nessessary here? Seems is required in BridgeHubRococoOrBridgeHubWococoSwitchExporter and will cause cumulus fail to build?

core/packages/contracts/test/Gateway.t.sol
```diff
@@ -0,0 +1,652 @@
...
+        ParaID destPara = ParaID.wrap(2043);
+        bytes32 destAddress = keccak256("/Alice");
+
+        vm.expectEmit(true, true, false, true);
+        emit TokenSent(address(this), address(token), destPara, abi.encodePacked(destAddress), 1);
+
+        // Expect the gateway to emit `OutboundMessageAccepted`
+        vm.expectEmit(true, false, false, false);
+        emit OutboundMessageAccepted(
+            assetHubParaID, 1, SubstrateTypes.SendToken(address(gateway), address(token), destPara, destAddress, 1)
```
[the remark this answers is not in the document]
vgeddes: As discussed on Slack, the gateway address will now form part of the asset multilocation.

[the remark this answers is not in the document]
musnit: link to Slack convo please? this sounds weird/off to me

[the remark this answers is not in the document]
alistair-singh: When executing xcm on bridgehub we `DescendOrigin` to a sovereign account. Before this PR, we created a sovereign for the Registry contract from its address `../../GlobalConsensus(Ethereum{ chain_id }/AccountKey20{ $REGISTRY_CONTRACT_ADDRESS }`.
So this question is me asking if we would use the AssetHubAgent or GatewayProxy contract address for the `DescendOrigin`. I think in earlier conversations we decided on using AssetHubAgent but the tests use GatewayProxy in some places so I was unclear. So we decided to use GatewayProxy as the address for creating a sovereign and I am busy implementing that now.

core/packages/contracts/src/Types.sol
vgeddes: My thinking is that if on BridgeHub we update the pallet's operating mode to `RejectingOutboundMessages`, it will have the same effect of preventing any more messages being sent to Ethereum.

core/packages/contracts/test/Gateway.t.sol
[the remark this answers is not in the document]
vgeddes: Just adding more background info here:

Tokens bridged over from Ethereum to Polkadot will be stored on the AssetHub (Statemint) parachain. AssetHub considers these assets to be "foreign" and each one is uniquely identified by a MultiLocation.

This MultiLocation must be of the form `/<PREFIX>/<ID>`, where `<PREFIX>` is the consensus system that owns or controls the asset, and `<ID>` is an arbitrary asset id within the consensus system.

AssetHub sets up access controls so that only `<PREFIX>` can mint/burn foreign assets deposited on AssetHub.

What this means is that we need to use a MultiLocation like the following for ERC20s bridged over to AssetHub:
```
/GlobalConsensus(Ethereum)/AccountKey20(Gateway.address)/AccountKey20(Token.address)
```

This could be decomposed into PREFIX and ID
```
PREFIX=/GlobalConsensus(Ethereum)/AccountKey20(Gateway.address)
ID=/AccountKey20(Token.address)
```

Effectively this allows our Gateway contract on Ethereum to administer assets on AssetHub.

core/packages/contracts/src/Types.sol
yrong: My concern is assuming some security bug exist in our bridge contract which allows malicious relayer/attacker to construct transaction manually(i.e. not from BridgeHub pallets) and steal money from the protocol.

core/packages/contracts/test/Gateway.t.sol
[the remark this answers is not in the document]
musnit: I'm not sure this is a good idea. This ties the Asset on Polkadot to our particular Gateway. Shouldn't the prefix ideally should just be the asset, ie, 

```
PREFIX=/GlobalConsensus(Ethereum)
ID=/AccountKey20(Token.address)
```

It should be up to the bridge to encapsulate that the Gateway is the only way to act as Ethereum, not baked into the asset?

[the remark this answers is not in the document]
musnit: How does AssetHub validate the PREFIX? Is the prefix controlled by Polkadot itself or?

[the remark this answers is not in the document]
yrong: Trust asset allowed is defined in AssetHub runtime as
https://github.com/Snowfork/cumulus/blob/9b83c7325df5ec50fec2571f8ff304d40a7e88fe/parachains/runtimes/assets/asset-hub-kusama/src/xcm_config.rs#L399
with trusted reserve locations(i.e. PREFIX) in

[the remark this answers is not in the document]
yrong: > This ties the Asset on Polkadot to our particular Gateway

The Gateway is now a proxy so should not change. And I guess it's useful to differ the same asset(e.g. weth) from  other bridge providers(e.g. From Axelar).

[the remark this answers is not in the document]
musnit: Thanks @yrong! Yeah we had a quick call and discussed this and basically:
 - We'll have the same issue with XCM Transact and the Identity/Multilocation of agents, since they'll be tied to our particular gateway too. We'll likely need a Gateway prefix for those too
 - On the other hand, there is a disadvantage to fragmenting liquidity and fragmenting identities across different bridges. Ideally the Polkadot community can come to some wider consensus on what their primary trust model is for bridges and pick a bridge or composition of bridges to control the entire Ethereum PREFIX directly, instead of just a particular suffix for a particular contract. If desired, this will likely have to come later via wider Polkadot governance, so it's not something we should implement now but we should be thinking about it and ensuring we'll have future compatibility if this does become needed.

alistair-singh (approved):
🚀

core/packages/contracts/src/Types.sol
vgeddes: Assuming there a bug in code, there is no way to guard against this kind of attack, since if a malicious attacker can manually construct a transaction and have it pass verification by the Gateway, then they can take over entire control of the Gateway using the `upgrade` handler.

PR #884, opened 2023-07-17
Use hashed description

alistair-singh:
https://linear.app/snowfork/issue/SNO-529
Cumulus companion: https://github.com/Snowfork/cumulus/pull/44


parachain/primitives/router/src/inbound/mod.rs
```diff
@@ -101,10 +103,21 @@ impl NativeTokensMessage {
...
+								GlobalConsensus(Ethereum { chain_id }),
+								AccountKey20 { network: None, key: *origin.as_fixed_bytes() },
+							),
+						),
+					);
+
+				let owner = match some_owner {
+					Some(owner) => owner,
+					None => return Err(ConvertError::BadOrigin),
+				};
```
yrong, line 120: could be simplified with `ok_or`

parachain/primitives/router/src/inbound/mod.rs
```diff
@@ -101,10 +103,17 @@ impl NativeTokensMessage {
...
-				let owner = GlobalConsensusEthereumAccountConvertsFor::<[u8; 32]>::from_params(
-					&chain_id,
-					origin.as_fixed_bytes(),
-				);
+				let owner =
+					GlobalConsensusEthereumAccountConvertsFor::<[u8; 32]>::convert_location(
+						&MultiLocation::new(
+							2,
+							(
+								GlobalConsensus(Ethereum { chain_id }),
```
doubledup, line 111: We can reuse the network id:
```suggestion
								GlobalConsensus(network),
```

parachain/primitives/router/src/inbound/mod.rs
```diff
@@ -101,10 +103,17 @@ impl NativeTokensMessage {
...
+								AccountKey20 { network: None, key: *origin.as_fixed_bytes() },
```
doubledup, line 112: When would we set a network id in an account key?

doubledup (approved):

parachain/primitives/router/src/inbound/mod.rs
yrong: Guess already set in `GlobalConsensus(network)` above so can just ignore here.

parachain/primitives/router/src/inbound/mod.rs
```diff
@@ -224,6 +233,21 @@ impl NativeTokensMessage {
 	}
 }
 
+pub struct DescribeEthereumAccountKey20Terminal;
+impl DescribeLocation for DescribeEthereumAccountKey20Terminal {
```
yrong, line 237: Maybe a dump question. So what's the benifit  for introduce `DescribeLocation` here?  Seems the internal convert logic still `(b"ethereum", chain_id, key).using_encoded(blake2_256)` not changed?

yrong (approved):

claravanstaden (approved):

parachain/primitives/router/src/inbound/mod.rs
alistair-singh: An example of where we "could" use network id is if we are specifying an ERC20 token contract on an EVM based parachain and potentially transferring it over the bridge and we would want a way to specify its Network as `Kusama` so we could handle it specially in some way. But I say "could" here because it's hypothetical.

If you look at some places in the source code, we often ignore network id. Such as [here](https://github.com/paritytech/polkadot/blame/master/xcm/xcm-builder/src/location_conversion.rs#L67-L85).
There are also some places where it makes sense to validate it if present like [here](https://github.com/Snowfork/snowbridge/blob/main/parachain/primitives/router/src/outbound/mod.rs#L272-L274).

In this case yrong is correct in that it is set in GlobalConsensus so we can set it as None on the key itself.

parachain/primitives/router/src/inbound/mod.rs
alistair-singh: This is just to bring our implementation in line with the new way to convert locations by implementing via the `DescribeLocation` trait. Currently they do not have an implementation in [location_conversion.rs](https://github.com/paritytech/polkadot/blame/master/xcm/xcm-builder/src/location_conversion.rs) for ethereum addresses. We would probably merge our implementation of `DescribeEthereumAccountKey20Terminal` into `xcm-builder` at some point.

PR #885, opened 2023-07-17
Bump nixpkgs to use its geth package

doubledup:
This comes with Node & pnpm version bumps because we're updating nixpkgs, but they're unlikely to cause issues. Still tracking down the BEEFY relayer failure as it affects this section of the upgrades.

### DIFF
--- a/core/package.json
+++ b/core/package.json
@@ -2,8 +2,8 @@
   "name": "@snowbridge/core",
   "private": true,
   "engines": {
-    "node": "v18.16.0",
-    "pnpm": "8.4.0"
+    "node": "v18.16.1",
+    "pnpm": "8.6.6"
   },
   "scripts": {
     "preinstall": "npx only-allow pnpm",
--- a/core/packages/test/scripts/build-binary.sh
+++ b/core/packages/test/scripts/build-binary.sh
@@ -59,19 +59,9 @@ build_relayer()
     cp $relay_bin "$output_bin_dir"
 }
 
-build_geth() {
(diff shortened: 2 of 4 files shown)


core/package.json
```diff
@@ -2,8 +2,8 @@
   "name": "@snowbridge/core",
   "private": true,
   "engines": {
-    "node": "v18.16.0",
-    "pnpm": "8.4.0"
+    "node": "v18.16.1",
+    "pnpm": "8.6.6"
```
yrong, line 6: We may also sync node&pnpm in github action.
https://github.com/Snowfork/snowbridge/blob/a9d3df4e065622bde119d4a8ef953a1b5a1c883e/.github/workflows/ethereum.yml#L23-L29

yrong:
> tracking down the BEEFY relayer failure
@doubledup So what's the failure in your setup?  Just tried e2e setup with this update and seems beefy relayer works fine.

doubledup:
@yrong The relayer consistently fails after the first handover commitment when waiting for randao, here: https://github.com/Snowfork/snowbridge/blob/main/relayer/relays/beefy/ethereum-writer.go#L139. I've tried prior commits on `main` and they're failing on `submitInitial`.

Let's debug this separately to prevent noise on the PR.

doubledup:
Confirmed that the problem was an old `.envrc` file I had from tracking down a previous issue. I've updated the README with a Troubleshooting section in case we have these sorts of mismatches again.

yrong (approved):
LGTM!

PR #886, opened 2023-07-18
Update cumulus submodule

doubledup:
We missed this in #880 after the cumulus companion was merged.


claravanstaden (approved):
Thanks for catching this @doubledup!

PR #887, opened 2023-07-18
Clara/sno 552

claravanstaden:
Super tiny changes after comparing light client with spec.


parachain/pallets/ethereum-beacon-client/src/lib.rs
```diff
@@ -320,12 +320,10 @@ pub mod pallet {
 					update.attested_header.slot >= update.finalized_header.slot,
 				Error::<T>::InvalidUpdateSlot
 			);
-
```
claravanstaden, line 323: Remove these spaces so it's clear the code is part of `// Verify update does not skip a sync committee period.`

yrong (approved):
LGTM!

PR #888, opened 2023-07-18
SNO-565: Make parachain relayers use Gateway contract

doubledup:
Opening as a PR into Vincent's branch as I need those changes as a base.

Fixed the parachain build failures on the base branch in #889.

We don't have an extrinsic that sends messages yet, so can't fully test this yet. I've made sure that the relayer compiles and launches without crashing. The execution relayer still crashes on startup for comparison.


relayer/contracts/generate.go
```diff
@@ -1,6 +1,5 @@
 //go:generate bash -c "jq .abi ../../core/packages/contracts/out/OpaqueProof.sol/OpaqueProof.json | abigen --abi - --type OpaqueProof --pkg contracts --out opaque_proof.go"
 //go:generate bash -c "jq .abi ../../core/packages/contracts/out/BeefyClient.sol/BeefyClient.json | abigen --abi - --type BeefyClient --pkg contracts --out beefy_client.go"
-//go:generate bash -c "jq .abi ../../core/packages/contracts/out/InboundQueue.sol/InboundQueue.json | abigen --abi - --type InboundQueue --pkg contracts --out inbound_queue.go"
-//go:generate bash -c "jq .abi ../../core/packages/contracts/out/OutboundQueue.sol/OutboundQueue.json | abigen --abi - --type OutboundQueue --pkg contracts --out outbound_queue.go"
+//go:generate bash -c "jq .abi ../../core/packages/contracts/out/Gateway.sol/Gateway.json | abigen --abi - --type Gateway --pkg contracts --out gateway.go"
```
vgeddes, line 3: Actually, we should be generating bindings for `contracts/src/interfaces/IGateway.sol`.

This is how `GatewayProxy.sol`, `Gateway.sol`, and `IGateway.sol` all relate to each other:

* `GatewayProxy.sol`: The contract at the "front of the house" that users interact with. This is the proxy contract.
* `Gateway.sol`: `GatewayProxy.sol` forwards messages to this implementation or "logic" contract.
* `IGateway.sol`: The API exposed by the proxy & implementation contract.

When configuring the relayer, make sure to use the address of `GatewayProxy.sol` as the gateway's address.

relayer/contracts/generate.go
```diff
@@ -1,6 +1,5 @@
 //go:generate bash -c "jq .abi ../../core/packages/contracts/out/OpaqueProof.sol/OpaqueProof.json | abigen --abi - --type OpaqueProof --pkg contracts --out opaque_proof.go"
```
vgeddes, line 1: Obsolete. No longer needed since we pass the proof directly to `IGateway.submitInbound` without converting it to opaque bytes

```suggestion
```

relayer/relays/parachain/ethereum-writer.go
```diff
@@ -28,13 +28,12 @@ import (
...
-	inboundQueue     *contracts.InboundQueue
-	tasks            <-chan *Task
-	abiPacker        abi.Arguments
-	abiBasicUnpacker abi.Arguments
-	inboundQueueABI  abi.ABI
+	config     *SinkConfig
+	conn       *ethereum.Connection
+	gateway    *contracts.Gateway
+	tasks      <-chan *Task
+	abiPacker  abi.Arguments
```
vgeddes, line 35: Obsolete. Was previously used to convert proof to opaque bytes.

```suggestion
```

relayer/contracts/generate.go
doubledup: Right, will use the interface. We're using the proxy address [here](https://github.com/Snowfork/snowbridge/pull/888/commits/3bca8e8981373833b6b2dfa2fa1ee17b572c8b7b)

doubledup: I've kept the name `Gateway` for the Go bindings, even though they're generated from the IGateway interface. Seems sensible as the interface is the only thing we'll use from the relayers.

relayer/relays/parachain/config.go
```diff
@@ -9,17 +9,18 @@ type Config struct {
 	Sink   SinkConfig   `mapstructure:"sink"`
 }
 
+// TODO: check whether ChannelID should be uint32 (as in the parachain) or big.Int (uint256, as in the Gateway contract)
```
doubledup, line 12: @vgeddes How big are the channelIDs?

yrong: Guess uint32 as ParaId should be enough. But remember Aidan mentioned last time we'll probably allow multiple channels for each Parachain?

yrong (approved):
LGTM!

relayer/relays/parachain/config.go
doubledup: Ah right, we can expand the channel id on Polkadot to get multiple channels per parachain

PR #889, opened 2023-07-18
Fix parachain build in proxy-contracts branch

doubledup:
Got blocked on the parachain build when testing #888, but I've found a way around the failures for now. Everything builds, though the relayers won't send messages until they're updated to use the Gateway contract (#888 is for the parachain relayer, execution relayer is coming in a future PR).

This starts with a merge from `main` - the later commits are the intended changes.

One caveat is that this reverts to the deprecated `GenesisBuild` trait instead of `BuildGenesisConfig`. 

I haven't included a Cumulus companion PR as this is merging into the `proxy-contracts` feature branch. Instead I've pushed a corresponding `proxy-contracts` branch to our cumulus fork that we can use for the feature branch's companion PR.


alistair-singh (approved):
+1 this was blocking me as well. LGTM

yrong:
Seems polkadot dependency has been upgraded to `v0.9.43` while cumulus is still `v0.9.42`, will that be a problem?

doubledup:
@yrong Where are you seeing cumulus use `v0.9.42`? The cumulus submodule is at [d29494f4](https://github.com/Snowfork/cumulus/tree/d29494f4e2bd60011b529edd36d0792e9e66fb14), which has all of the Polkadot crates locked at `0.9.43`.

The break in CI is from a benchmark that doesn't have a matching extrinsic yet - I'm assuming Vincent's still going to add `create_agent` to the control pallet.

yrong:
> Where are you seeing cumulus use v0.9.42? The cumulus submodule is at [d29494f4](https://github.com/Snowfork/cumulus/tree/d29494f4e2bd60011b529edd36d0792e9e66fb14), which has all of the Polkadot crates locked at 0.9.43.

https://github.com/Snowfork/cumulus/blob/d29494f4e2bd60011b529edd36d0792e9e66fb14/Cargo.lock#L9925

doubledup:
ah ok, `0.9.420`. That's on the `snowbridge` branch and used in `main` - if `main` is breaking we should fix it separately. Haven't seen any problems on `main` though, so I'd guess it's fine.

doubledup:
@alistair-singh noticed that even though we're using `0.9.43` in both the parachain and cumulus, the hashes don't match: the parachain version's ahead of cumulus. This can happen when we add or update crates but don't double-check that the parachain lockfile still matches cumulus. He's downgraded the parachain version as we have a hard dependency on the versions in cumulus.

PR #890, opened 2023-07-19
sno-472 rebased

yrong:
https://github.com/Snowfork/cumulus/pull/46


doubledup:
ah damn, I just merged that branch and that seems to have caused a mess here, sorry. This should be fine if you drop the merge commit here from `david/fix-parachain-build` and do one more merge from `proxy-contracts`.

core/packages/test/scripts/set-env.sh
```diff
@@ -46,6 +46,7 @@ beacon_endpoint_http="${BEACON_HTTP_ENDPOINT:-http://127.0.0.1:9596}"
 bridgehub_ws_url="${BRIDGEHUB_WS_URL:-ws://127.0.0.1:11144}"
 bridgehub_para_id="${BRIDGEHUB_PARA_ID:-1013}"
 bridgehub_seed="${BRIDGEHUB_SEED:-//Alice}"
+bridgehub_pallets_owner="${BRIDGEHUB_PALLETS_OWNER:-0xd43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d}"
```
doubledup, line 49: Let's add a comment saying which account this is
```suggestion
# Parachain account that owns the BridgeHub pallets. Defaults to the Alice test account.
bridgehub_pallets_owner="${BRIDGEHUB_PALLETS_OWNER:-0xd43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d}"
```

core/packages/test/scripts/configure-bridgehub.sh
```diff
@@ -17,7 +17,31 @@ config_inbound_queue()
...
+    send_governance_transact_from_relaychain $bridgehub_para_id "$payload"
+}
+
+config_pallet_owner()
+{
+    local owner=$(echo $bridgehub_pallets_owner | cut -c3-)
+    local option="01"
+
+    # config owner of inbound queue
+    local pallet="30"
```
doubledup, line 29: Which runtime do these indexes come from? I'm only seeing Rococo, but that looks like the index is 48: https://github.com/Snowfork/cumulus/blob/be6d97cd/parachains/runtimes/bridge-hubs/bridge-hub-rococo/src/lib.rs#L740

alistair-singh (approved):

doubledup (approved):

vgeddes:
+1

parachain/pallets/inbound-queue/Cargo.toml
```diff
@@ -60,8 +61,7 @@ std = [
     "snowbridge-router-primitives/std",
     "ethabi/std",
     "xcm/std",
-    "snowbridge-beacon-primitives/std",
-    "snowbridge-ethereum-beacon-client/std",
```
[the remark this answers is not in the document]
yrong: Yes, It's only used in test so not required here.

parachain/pallets/outbound-queue/src/lib.rs
```diff
@@ -272,6 +316,7 @@ pub mod pallet {
 		}
 
 		fn submit(ticket: Self::Ticket) -> Result<(), SubmitError> {
+			Self::ensure_not_halted().map_err(|_| SubmitError::BridgeHalted)?;
```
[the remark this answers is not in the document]
yrong: This is a trait method which expect `SubmitError` here.

parachain/pallets/inbound-queue/src/test.rs
```diff
@@ -162,16 +165,16 @@ fn expect_events(e: Vec<RuntimeEvent>) {
 	assert_eq!(last_events(e.len()), e);
 }
 
-pub fn new_tester(gateway: H160) -> sp_io::TestExternalities {
-	new_tester_with_config(inbound_queue::GenesisConfig { gateway })
+pub fn new_tester<T: Config>(gateway: H160) -> sp_io::TestExternalities {
```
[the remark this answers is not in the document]
yrong: Previously I add `owner` to genenis config which required `T: Confg`, but since the default owner account from Parity team may not the one as we want, IMO account from our team make more sense here so remove it.

core/packages/test/scripts/set-env.sh
yrong: I would prefer we add comment for `BRIDGEHUB_PALLETS_OWNER` in `.envrc-example`.

core/packages/test/scripts/configure-bridgehub.sh
yrong: Yes, `30` is just the hex value of 48.

parachain/pallets/outbound-queue/src/lib.rs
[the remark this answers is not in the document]
doubledup: I meant adding the `OwnedBridgeModuleError` to the `SubmitError`, like this:
```suggestion
			Self::ensure_not_halted().map_err(|e| SubmitError::BridgeHalted(e))?;
```

core/packages/test/scripts/configure-bridgehub.sh
doubledup: Oh, it's a hex value 🤦 Sorry

PR #892, opened 2023-07-19
Proxy contracts Tests fixes

alistair-singh:
There is some overlap with changes made in https://github.com/Snowfork/snowbridge/pull/890. But this PR renamed tests and fixes for the XcmConverter tests.

### DIFF
--- a/parachain/pallets/control/Cargo.toml
+++ b/parachain/pallets/control/Cargo.toml
@@ -32,7 +32,7 @@ xcm = { git = "https://github.com/paritytech/polkadot.git", branch = "master", d
 ethabi = { git = "https://github.com/Snowfork/ethabi-decode.git", package = "ethabi-decode", branch = "master", default-features = false }
 
 [dev-dependencies]
-xcm-builder = { git = "https://github.com/paritytech/polkadot.git", branch = "master" }
+xcm-builder = { git = "https://github.com/paritytech/polkadot.git", branch = "master", default-features = false }
 
 [features]
 default = ["std"]
@@ -47,6 +47,7 @@ std = [
 	"sp-io/std",
 	"sp-runtime/std",
 	"xcm/std",
+	"xcm-builder/std",
 	"ethabi/std"
 ]
 runtime-benchmarks = [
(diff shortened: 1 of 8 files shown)


parachain/pallets/control/src/lib.rs
```diff
@@ -33,7 +33,6 @@ pub mod pallet {
 	use super::*;
 	use frame_support::{pallet_prelude::*, traits::EnsureOrigin};
 	use frame_system::pallet_prelude::*;
-	use xcm::v3::MultiLocation;
```
doubledup, line 36: #890 Removes `xcm::prelude::*` above. If we merge both we'll probably end up with no `MultiLocation` here. Could we instead remove the prelude above?

doubledup (approved):
One merge-conflict-free error case to watch out for. Otherwise we just need to be careful when merging.

parachain/pallets/control/Cargo.toml
```diff
@@ -50,7 +47,11 @@ std = [
 	"sp-io/std",
 	"sp-runtime/std",
 	"xcm/std",
+	"xcm-builder/std",
```
yrong, line 50: Seems `xcm-builder` is only used in test codes and not requied here.

yrong:
@alistair-singh 
I've merged this PR and my previous #890 together into branch `proxy-contracts`. So could you help to double-check `proxy-contracts` if anything missing or wrong.

yrong (approved):

parachain/pallets/control/Cargo.toml
alistair-singh: this failed when running `cargo test --features runtime-benchmarks` so I had to add this line.

parachain/pallets/control/src/lib.rs
alistair-singh: #890 is merged and this is rebased on top of that.

parachain/pallets/control/Cargo.toml
yrong: also require remove `default-features = false` in
`xcm-builder = { git = "https://github.com/paritytech/polkadot.git", branch = "master", default-features = false }` above

Already merged into `proxy-contracts` and seems CI tests for `runtime-benchmarks` just works fine.

PR #897, opened 2023-07-20
SNO-566: Execution relayer watching for gateway messages

doubledup:
The smoke test fails gateway validation in the inbound channel, but this is because the gateway isn't set in the channel. I've tried via the test script that should execute as root (with `--sudo`), but the XCM execution there fails due to a lack of permissions.

This is worth merging at this point - we need to set the inbound queue's gateway properly to process the message further, but the message at least passes validation.

Fixes: SNO-566


vgeddes (approved):
Looks good!

alistair-singh (approved):
🚀

PR #901, opened 2023-07-25
Improve beefy tests

yrong:
Resolves: SNO-576

PR #903, opened 2023-07-26
SNO-517: Check validator set size in submitFinal

doubledup:
Fixes: SNO-517

### DIFF
--- a/core/packages/contracts/src/BeefyClient.sol
+++ b/core/packages/contracts/src/BeefyClient.sol
@@ -169,22 +169,22 @@ contract BeefyClient {
     uint256 public immutable randaoCommitExpiration;
 
     /* Errors */
+    error InvalidBitfield();
+    error InvalidBitfieldLength();
     error InvalidCommitment();
-    error StaleCommitment();
-    error InvalidValidatorProof();
-    error InvalidSignature();
-    error NotEnoughClaims();
     error InvalidMMRLeaf();
     error InvalidMMRLeafProof();
-    error InvalidBitfield();
-    error WaitPeriodNotOver();
-    error TicketExpired();
+    error InvalidMMRRootLength();
+    error InvalidSignature();
+    error InvalidTicket();
+    error InvalidValidatorProof();
+    error NoMMRRootInCommitment();
+    error NotEnoughClaims();
     error PrevRandaoAlreadyCaptured();
     error PrevRandaoNotCaptured();
-    error InvalidBitfieldLength();
-    error InvalidTicket();
-    error InvalidMMRRootLength();
-    error NoMMRRootInCommitment();
+    error StaleCommitment();
+    error TicketExpired();
+    error WaitPeriodNotOver();
 
     constructor(
         uint256 _randaoCommitDelay,
--- a/core/packages/contracts/src/BeefyClient.sol
+++ b/core/packages/contracts/src/BeefyClient.sol
@@ -297,6 +297,10 @@ contract BeefyClient {
 
         Ticket storage ticket = tickets[ticketID];
 
+        if (ticket.validatorSetLen != currentValidatorSet.length) {
+            revert InvalidTicket();
+        }
+
         if (commitment.validatorSetID != currentValidatorSet.id) {
             revert InvalidCommitment();
         }
@@ -330,6 +334,10 @@ contract BeefyClient {
         (bytes32 commitmentHash, bytes32 ticketID) = validate(commitment, bitfield);
 
         Ticket storage ticket = tickets[ticketID];
+
+        if (ticket.validatorSetLen != nextValidatorSet.length) {
+            revert InvalidTicket();
+        }
 
         if (commitment.validatorSetID != nextValidatorSet.id) {
             revert InvalidCommitment();
--- a/core/packages/contracts/src/BeefyClient.sol
+++ b/core/packages/contracts/src/BeefyClient.sol
@@ -240,6 +240,10 @@ contract BeefyClient {
             revert InvalidValidatorProof();
         }
 
+        if (commitment.validatorSetID != vset.id) {
+            revert InvalidCommitment();
+        }
+
         // Check if validatorSignature is correct, ie. check if it matches
         // the signature of senderPublicKey on the commitmentHash
         bytes32 commitmentHash = keccak256(encodeCommitment(commitment));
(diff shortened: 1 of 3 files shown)


yrong (approved):
LGTM! Though add some comment [here](https://linear.app/snowfork/issue/SNO-517#comment-725703f1) I'm still with adding this check for more security and assuming gas cost here will be cheap.

yrong:
May not be direct related to this PR, just notice current structure definition of `Ticket` is
https://github.com/Snowfork/snowbridge/blob/ecb847a1bb36de5737376288d413e17b4c4b273b/core/packages/contracts/src/BeefyClient.sol#L97-L103

Will change to structure as following a better layout packed into 2 slots?
```
struct Ticket {
        address account;
        uint32 blockNumber;
        uint32 validatorSetLen;
        bytes32 bitfieldHash;
        uint256 prevRandao;
}
```

doubledup:
@yrong I think that'll still use 3 slots, as `bytes32` and `uint256` both take 1 slot. We should look at optimising our struct layout at some point though, eg. `ValidatorProof` is
```
    struct ValidatorProof {
        uint8 v;
        bytes32 r;
        bytes32 s;
        uint256 index;
        address account;
        bytes32[] proof;
    }
```
which uses `3+proof.length` slots, but swapping `index` and `account` would use `2+proof.length` slots.

Let's wait until the proxy refactor is done before optimising. We'll need to check our tests' coverage is good enough before we can use them for benchmarks to measure & validate these improvements.

core/packages/contracts/test/BeefyClient.t.sol
```diff
@@ -413,4 +415,36 @@ contract BeefyClientTest is Test {
         vm.expectRevert(BeefyClient.NotEnoughClaims.selector);
         beefyClient.submitInitial(commitment, initialBits, finalValidatorProofs[0]);
     }
+
+    function testSubmitFailWithIncorrectValidatorSetLengthOnTicket() public {
+        BeefyClient.Commitment memory commitment = initialize(setId);
+
+        beefyClient.submitInitial(commitment, bitfield, finalValidatorProofs[0]);
+
+        beefyClient.setTicketValidatorSetLen(keccak256(beefyClient.encodeCommitment_public(commitment)), setSize / 2);
```
doubledup, line 424: Manually editing tickets like this isn't a great way to test - we should rather have test data with different validator set lengths. But let's get the updated tests & fixture data in #901 merged first.

core/packages/contracts/src/BeefyClient.sol
```diff
@@ -301,6 +336,10 @@ contract BeefyClient is Ownable {
 
         Ticket storage ticket = tickets[ticketID];
 
+        if (ticket.validatorSetLen != currentValidatorSet.length) {
+            revert InvalidTicket();
+        }
+
         if (commitment.validatorSetID != currentValidatorSet.id) {
```
vgeddes, line 343: Hmm, doesn't this check here prevent relayers from submitting commitments signed by an invalid validator set?

Validator set ids are monotonically increasing. So `nextValidatorSet.id + 1 == currentValidatorSet.id`.

doubledup: It does, but the attack we're defending against here already uses the correct validator set in the `submitFinal` call:

- The attacker uses the `submitInitial` call for the smaller validator set, so they only need to supply a bitfield with 2/3 of the smaller set's size in set bits.
(They also need to compromise one validator in the smaller set, and ensure that they've compromised the validator with the corresponding index in the larger validator set. The compromised validator also needs to sign the commitment for another era, so we should be able to slash them afterwards.)
- When the attacker gets to `submitFinal`, they can use the larger (& valid/correct) validator set. The id check here passes, because the set ids match. But now we're sampling from a bitfield that only has set bits equal to 2/3 of the smaller set's size, not of the larger as it should.

Come to think of it, if we did this id check in `submitInitial*` we'd also defend against this attack. The commitment hash is used to look up the ticket, so the commitment needs to be the same across submitInitial & submitFinal. But the commitment has to have a single validator set id - either current or next - so mixing handover & non-handover calls for the same commitment would fail one of the commitment id checks.

core/packages/contracts/test/BeefyClient.t.sol
yrong: Already merged.

core/packages/contracts/src/BeefyClient.sol
yrong: > The attacker uses the submitInitial call for the smaller validator set

AFAIK active validators are elected from polkadot staking with a fixed size(297)  now and will increased to 1000 in long run. So is that possible we add another check when update validator set here.

https://github.com/Snowfork/snowbridge/blob/d26678debcac6048c2c7dd2db54c06687575661e/core/packages/contracts/src/BeefyClient.sol#L355

Something like `assert(nextValidatorSet.length>=currentValidatorSet.length)`?

vgeddes: > It does, but the attack we're defending against here already uses the correct validator set in the submitFinal call:

ah I see, I understand better now.

> Come to think of it, if we did this id check in submitInitial* we'd also defend against this attack. The commitment hash is used to look up the ticket, so the commitment needs to be the same across submitInitial & submitFinal. But the commitment has to have a single validator set id - either current or next - so mixing handover & non-handover calls for the same commitment would fail one of the commitment id checks.

I agree, I think this this a more comprehensive check. Let's go this route rather.

> AFAIK active validators are elected from polkadot [staking](https://polkadot.js.org/apps/?rpc=wss%3A%2F%2Fpolkadot-rpc-tn.dwellir.com#/staking) with a fixed size(297) now and will increased to 1000 in long run. So is that possible we add another check when update validator set [here](https://github.com/Snowfork/snowbridge/blob/d26678debcac6048c2c7dd2db54c06687575661e/core/packages/contracts/src/BeefyClient.sol#L355) something like assert(nextValidatorSet.length>=currentValidatorSet.length)?

This would not be needed if we check the validator set id in submitInitial, as per David's suggestion

core/packages/contracts/src/BeefyClient.sol
```diff
@@ -288,6 +288,41 @@ contract BeefyClient is Ownable {
         ticket.prevRandao = block.prevrandao;
     }
 
+    /**
```
vgeddes, line 291: I preferred to have these functions where they were originally (at the end of the list of public functions), since they are really just helpers.

vgeddes (commented):
Looks good! Can you change the base to `proxy-contracts`?

alistair-singh (approved):

PR #904, opened 2023-07-26
Asset transfer

alistair-singh:
Resolves: SNO-569
Cumulus companion: https://github.com/Snowfork/cumulus/pull/52

### TODO
- [x] Register token
- [x] Send token Eth -> Substrate
- [x] Send token back Substrate -> Eth


core/packages/contracts/src/Assets.sol
```diff
@@ -74,7 +76,7 @@ library Assets {
 
         _transferToAgent(assetHubAgent, token, sender, amount);
 
-        payload = SubstrateTypes.SendToken(assetHubAgent, token, destinationChain, destinationAddress, amount);
+        payload = SubstrateTypes.SendToken(gateway, token, destinationChain, destinationAddress, amount);
```
vgeddes, line 79: nice catch! Though it's not necessary to pass `gateway` as a parameter to `Assets.sendToken`. Since `Assets.sol` is a library used by `Gateway`, you can just get the address of the gateway via `address(this)`.

```suggestion
        payload = SubstrateTypes.SendToken(address(this), token, destinationChain, destinationAddress, amount);
```

core/packages/contracts/src/Assets.sol
```diff
@@ -48,16 +49,17 @@ library Assets {
...
-            payload = SubstrateTypes.SendToken(assetHubAgent, token, destinationChain, destinationAddress, amount);
+            payload = SubstrateTypes.SendToken(gateway, token, destinationChain, destinationAddress, amount);
         }
         extraFee = $.sendTokenFee;
 
         emit TokenSent(sender, token, destinationChain, abi.encodePacked(destinationAddress), amount);
     }
 
     function sendToken(
+        address gateway,
```
vgeddes, line 62: See my comment below.

```suggestion
```

core/packages/contracts/src/Gateway.sol
```diff
@@ -384,10 +384,7 @@ contract Gateway is IGateway {
 
     // Register a token on AssetHub
     function registerToken(address token) external payable {
-        CoreStorage.Layout storage $ = CoreStorage.layout();
-        address assetHubAgent = $.agents[ASSET_HUB_AGENT_ID];
-
-        (bytes memory payload, uint256 extraFee) = Assets.registerToken(assetHubAgent, token, CREATE_TOKEN_CALL_ID);
+        (bytes memory payload, uint256 extraFee) = Assets.registerToken(address(this), token, CREATE_TOKEN_CALL_ID);
```
vgeddes, line 387:
```suggestion
        (bytes memory payload, uint256 extraFee) = Assets.registerToken(token, CREATE_TOKEN_CALL_ID);
```

core/packages/contracts/src/Gateway.sol
```diff
@@ -416,7 +420,14 @@ contract Gateway is IGateway {
         address assetHubAgent = $.agents[ASSET_HUB_AGENT_ID];
 
         (bytes memory payload, uint256 extraFee) = Assets.sendToken(
-            ASSET_HUB_PARA_ID, assetHubAgent, token, msg.sender, destinationChain, destinationAddress, amount
+            address(this),
```
vgeddes, line 423: See my comments above. Libraries can get address of gateway via `address(this)`.

```suggestion
```

vgeddes (commented):
Looks good! Added a few minor comments

smoketest/tests/send_tokens.rs
```diff
@@ -71,19 +64,10 @@ async fn send_tokens() {
...
-
     // Lock tokens into vault
     let value1: u128 = U256::from(value).low_u128();
-    native_tokens
-        .lock(weth.address(), 0, recipient.into(), value1)
+    gateway
+        .send_token(weth.address(), 1000.into(), FERDIE, value1)
         .value(1000)
         .send()
         .await
```
yrong, line 73: Would suggest we log the transaction hash here.

smoketest/tests/send_tokens.rs
```diff
@@ -71,19 +64,10 @@ async fn send_tokens() {
...
-        parents: 0,
-        interior: X1(Junction::AccountId32 {
-            network: None,
-            id: FERDIE,
-        }),
-    }
-    .encode();
-
     // Lock tokens into vault
     let value1: u128 = U256::from(value).low_u128();
```
yrong, line 68: Remember I met `Insufficient***` error for launch this test multiple times, so would suggest  use a lower value here. Meanwhile some change in https://github.com/Snowfork/snowbridge/pull/905/files#r1276100691

core/packages/contracts/src/Gateway.sol
alistair-singh: addressed in https://github.com/Snowfork/snowbridge/pull/904/commits/8289362594510c9672b87bd44bfbf09dcbe3f596

core/packages/contracts/src/Gateway.sol
alistair-singh: addressed in https://github.com/Snowfork/snowbridge/pull/904/commits/8289362594510c9672b87bd44bfbf09dcbe3f596

core/packages/contracts/src/Assets.sol
alistair-singh: addressed in https://github.com/Snowfork/snowbridge/pull/904/commits/8289362594510c9672b87bd44bfbf09dcbe3f596

core/packages/contracts/src/Assets.sol
alistair-singh: addressed in https://github.com/Snowfork/snowbridge/pull/904/commits/8289362594510c9672b87bd44bfbf09dcbe3f596

smoketest/tests/send_tokens.rs
alistair-singh: Resolved in [0c4bdcc](https://github.com/Snowfork/snowbridge/pull/904/commits/0c4bdccbe776bfeb53b2d24c4437b4a0c74b6fa2)

smoketest/tests/send_tokens.rs
alistair-singh: Resolved in [0c4bdcc](https://github.com/Snowfork/snowbridge/pull/904/commits/0c4bdccbe776bfeb53b2d24c4437b4a0c74b6fa2)

vgeddes (approved):
Looks good!

yrong (approved):

parachain/primitives/core/src/lib.rs
```diff
@@ -143,13 +143,21 @@ pub enum AgentExecuteCommand {
...
+	}
 	pub fn abi_encode(&self) -> Vec<u8> {
 		match self {
 			AgentExecuteCommand::TransferToken { token, recipient, amount } => {
 				ethabi::encode(&vec![
-					Token::Address(*token),
-					Token::Address(*token),
-					Token::Uint(U256::from(*amount)),
+					Token::Uint(self.index().into()),
+					Token::Bytes(ethabi::encode(&vec![Token::Tuple(vec![
```
vgeddes, line 156: The Solidity side isn't expecting a tuple (struct) type though.

i.e `abi.decode(data, (address, address, uint128))` is equivalent to: 
```rust
ethabi::encode(&vec![Token::Address(*token), Token::Address(*recipient), Token::Uint(U256::from(*amount))]);
````

alistair-singh: I think implicitly it's encoded to the same thing because `(address, address, uint128)` is an implicit tuple. I have tested both versions and they work. The tuple and just a `vec![Tokens]`. I used the tuple version because it feels more correct. So I think this is more of style thing so which would you prefer?

vgeddes: I prefer the original form.

We can't say for sure whether they encode to the same thing for all input parameter types, without more investigation and testing. Also `Token::Tuple` was introduced purely to encode structs, and we're not using structs in this case.

PR #906, opened 2023-07-27
Revert "Use hashed description (#884)"

alistair-singh:
This reverts commit d26678debcac6048c2c7dd2db54c06687575661e.

Reverting because it makes `proxy-contracts` merge harder. This commit introduced the use of HashedDescription trait which may return error. Merging with `proxy-contracts` will mean plumbing the error all the way up to the call site which is a significant change. The actual hash produced using the legacy api's is still the same.


vgeddes (commented):
+1

PR #910, opened 2023-07-28
Testing the upgrade process

alistair-singh:
Resolves: SNO-580
Cumulus companion: https://github.com/Snowfork/cumulus/pull/54

TODO:
- [x] Create unit tests for upgrade
- [x] Run benchmarks and generate weights
- [x] Create script to test the upgrade path


vgeddes (approved):
Looks great.

contracts/test/GatewayUpgradeMock.sol
vgeddes, line 1: Should rather place this file in `contracts/test/mocks`.

alistair-singh: Resolved

PR #915, opened 2023-07-31
Improve dev setup

yrong:
Resolves: SNO-583


claravanstaden (approved):
Nice! 👍

alistair-singh (approved):

doubledup (approved):
Brilliant!

web/init.sh
```diff
@@ -1,13 +0,0 @@
...
-(cd .. && (git submodule update --init --recursive||true))
-
-echo "Install husky hook"
-(cd .. && ./web/node_modules/.bin/husky install)
-
-echo "Installing sszgen"
-go install github.com/ferranbt/fastssz/sszgen@latest
-
-echo "Initialize foundry libraries"
-(cd ../contracts && forge install)
```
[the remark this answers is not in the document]
yrong: It's running automaticlly when first time initialize the setup. And in https://github.com/Snowfork/snowbridge/blob/0bbee53545011a40ff982ccf689efd0485895cbe/scripts/generate_bindings.sh#L5-L6 `forge build` will install the library if not exist and then compile.

parachain/pallets/ethereum-beacon-client/src/lib.rs
vgeddes, line 1: Lets rather focus this PR on dev improvements rather than code formatting

scripts/init.sh
```diff
@@ -2,7 +2,16 @@
 
 set -e
 
+echo "Setup git hooks"
+git config core.hooksPath hooks/
```
vgeddes, line 6: I like to be explicit here:
```suggestion
git config --local core.hooksPath hooks/
```

scripts/init.sh
```diff
@@ -2,7 +2,16 @@
...
+
+echo "Update submodules"
 git submodule update --init --recursive || true
+
+echo "Installing dev tools"
+command -v sszgen || go install github.com/ferranbt/fastssz/sszgen@latest
```
vgeddes, line 12: This `init.sh` script only needs to be called once when cloning the repo, so we don't to test whether `sszgen` exists.

```suggestion
go install github.com/ferranbt/fastssz/sszgen@latest
```

flake.nix
```diff
@@ -82,12 +83,16 @@
                     export HOME=~
                     export GOPATH=$PWD/go
                     export PATH=$GOPATH/bin:$PATH
+                    export PATH=''${CARGO_HOME:-~/.cargo}/bin:$PATH
```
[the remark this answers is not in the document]
yrong: Since we've replaced `rust-overlay` with `rustup` as the management tool in Nix env, I would think the rust version is explicitly declared in `rust-toolchain.toml` no matter what outside of Nix is.

But yes, this line is unnessessary so just remove it.

scripts/init.sh
```diff
@@ -2,7 +2,16 @@
...
+
+echo "Installing dev tools"
+go install github.com/ferranbt/fastssz/sszgen@v0.1.3
+
+echo "Install node packages"
 (cd web && pnpm install)
-go install github.com/ferranbt/fastssz/sszgen@latest
-ln -sf pre-commit.sh .git/hooks/pre-commit
+
+./scripts/generate_bindings.sh
```
vgeddes, line 17: There isn't a need to regenerate all bindings when initialising the repo for the first time. For example:
1. The contract bindings used by the relayer are already checked into the repo
2. The bindings used by the smoketests can be generated when required.

```suggestion
```

scripts/init.sh
```diff
@@ -2,7 +2,17 @@
...
+echo "Installing dev tools"
+go install github.com/ferranbt/fastssz/sszgen@v0.1.3
+
+echo "Install node packages"
 (cd web && pnpm install)
-go install github.com/ferranbt/fastssz/sszgen@latest
-ln -sf pre-commit.sh .git/hooks/pre-commit
+
+echo "Initialize foundry libraries"
+(cd contracts && forge install)
```
vgeddes, line 18: Also explicitly calling `forge install` is not necessary, since the `git submodule update` command above initialises the submodules already

smoketest/make-bindings.sh
```diff
@@ -15,11 +15,16 @@ command -v subxt || cargo install subxt-cli \
...
 
+if lsof -Pi :11144 -sTCP:LISTEN -t >/dev/null ; then
+    echo "substrate nodes running, fetch metadata and generate api bindings"
+    # Fetch metadata from BridgeHub and generate client
+    subxt codegen --url ws://localhost:11144 | rustfmt --edition 2021 --emit=stdout > src/parachains/bridgehub.rs
+    subxt codegen --url ws://localhost:12144 | rustfmt --edition 2021 --emit=stdout > src/parachains/assethub.rs
+    subxt codegen --url ws://localhost:9944  | rustfmt --edition 2021 --emit=stdout > src/parachains/relaychain.rs
+else
+    echo "substrate nodes not running, please start with the e2e setup and rerun this script"
+fi
```
vgeddes, line 26: The script should exit early if the bindings cannot be generated:

```suggestion
if ! lsof -Pi :11144 -sTCP:LISTEN -t >/dev/null; then
      echo "substrate nodes not running, please start with the e2e setup and rerun this script"
      exit 1
fi

# Fetch metadata from BridgeHub and generate client
subxt codegen --url ws://localhost:11144 | rustfmt --edition 2021 --emit=stdout > src/parachains/bridgehub.rs
subxt codegen --url ws://localhost:12144 | rustfmt --edition 2021 --emit=stdout > src/parachains/assethub.rs
subxt codegen --url ws://localhost:9944  | rustfmt --edition 2021 --emit=stdout > src/parachains/relaychain.rs
```

vgeddes (approved):
Looks good!

PR #916, opened 2023-07-31
Replace rust overlay with rustup package

doubledup:
Turns out things work just fine without the overlay. Please test this on your machines though in case I missed something when cleaning.

After a `git clean` and `rustup self uninstall`, the E2E env starts up in a pure shell (with no local Rust installation) and smoketests build & pass. The only errors I got were from forgetting to run the `init.sh` and `make-bindings.sh` scripts, but those are already mentioned in the READMEs 😅 

One issue I can think of is if the Nix rustup version in an impure shell behaves in a way that conflicts with a local one. This seems unlikely, but I've reinstalled Rust locally & compiled other projects, then switched back to Snowbridge in an impure shell and the E2E stack still works.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -30,24 +30,6 @@
         "type": "github"
       }
     },
-    "flake-utils_3": {
-      "inputs": {
-        "systems": "systems"
-      },
-      "locked": {
-        "lastModified": 1681202837,
-        "narHash": "sha256-H+Rh19JDwRtpVPAWp64F+rlEtxUWBAQW28eAi3SRSzg=",
-        "owner": "numtide",
-        "repo": "flake-utils",
-        "rev": "cfacdce06f30d2b68473a46042957675eebb3401",
-        "type": "github"
-      },
-      "original": {
-        "owner": "numtide",
-        "repo": "flake-utils",
-        "type": "github"
-      }
-    },
     "foundry": {
       "inputs": {
         "flake-utils": "flake-utils_2",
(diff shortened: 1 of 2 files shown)


yrong (approved):
Nice!

yrong:
> an impure shell behaves in a way that conflicts with a local one

With the `rust-toolchain.toml` explicitly declared this is not a problem IMO.

vgeddes:
@doubledup This is still missing a few things:

* The `scripts/init.sh` script needs to explicitly install the rust toolchain as well a nightly for `rustfmt`.
* Rustup is going to install the toolchains outside the nix env into the `$HOME/.rustup`, along with other toolchains that have nothing to do with the nix env. You should set `RUSTUP_HOME` in `flake.nix` so that the env has its own rustup dir.

vgeddes:
> We automatically install the stable toolchain when running `web/packages/test/scripts/start-services.sh` (I've tested this by removing my local rustup installation), but we do need to install the nightly version. Might as well make the installation explicit in `init.sh`.
> 

Yes, I'd rather be explicit about installing the toolchain after initialising the nix env.

> Installing toolchains in the project directory doesn't affect which versions are selected: the stable toolchain version is selected by the toolchain file and we don't have a way to lock the nightly version. This might help if there's some state kept in the rustup directory though.

Well the other problem is that if you run `rustup toolchain list` inside the nix env, it's going to show all toolchains, even those installed outside of the nix env. that's why I'd like to completely segregate Nix's rustup state from the outside world.

doubledup:
Unfortunately there's no explicit command for installing the version in the toolchain file: https://github.com/rust-lang/rustup/issues/2686

There are other commands that will implicitly install from the toolchain file, but there's talk of removing that behaviour: https://github.com/rust-lang/rustup/issues/1397

I'm seeing two options:
- We can explicitly install a nightly version, but we need to duplicate the stable version in `init.sh` in order to install it explicitly. Here we need to update the Rust version in 2 places when upgrading Rust. If we miss this, fresh machines will install the old Rust version but will still use the correct version from the toolchain file.
- We can use something like `rustup show` to implicitly install the version in the toolchain file. Then we need to look out for `rustup`'s behaviour changing during an upgrade (via upgrading nixpkgs). If we miss this, fresh machines won't install the required toolchain in `init.sh` but it's likely that the `cargo` and `rustc` wrappers will.

In both cases, we use the toolchain file's version.

I'd prefer to duplicate the versions for now: it looks like inconsistencies in explicitly mentioned Rust versions are easier to detect than changes to `rustup`'s behaviour that only affect fresh environments.

PR #917, opened 2023-08-01
Complete Control Pallet Implementation

yrong:
Resolves: SNO-584

Requires: https://github.com/Snowfork/cumulus/pull/56


yrong:
@alistair-singh 

Seems a wide range of `XcmOriginToTransactDispatchOrigin` allowed in bridge-hub runtime [here](https://github.com/Snowfork/cumulus/blob/5f9f6cd8eb6d779d8b713dd7f931a9521e93b132/parachains/runtimes/bridge-hubs/bridge-hub-rococo/src/xcm_config.rs#L120-L139) for the xcm transact. There is also a unit test in our code base with mock data [here](https://github.com/Snowfork/snowbridge/blob/3dad79b95b0ee26d8e8f62973e071a767a386d5b/parachain/pallets/control/src/mock.rs#L79) and test code [here](https://github.com/Snowfork/snowbridge/blob/3dad79b95b0ee26d8e8f62973e071a767a386d5b/parachain/pallets/control/src/tests.rs#L28-L44) to demonstrate that transact origin from relaychain is allowed.

But I'm a little confused here do we really have a use case `create_agent` for relaychain itself?  Because in that case there is no `para_id` which is required for `create_channel`  and AFAIK our beefy light client is only able to handle message from parachain.

So is that as expected or a bug we should fix?

yrong:
https://github.com/Snowfork/snowbridge/pull/917/commits/1f0e818aea84d448c9db32af8e4a19e12e5456db
Some workaround to fix the breaking tests. I think maybe we can limit the scope of origins allowed(e.g. with a customized impl of `AgentHashedDescription`) but fine as now to make things simeple.

contracts/src/Gateway.sol
```diff
@@ -141,6 +141,9 @@ contract Gateway is IGateway, IInitializable {
 
         // Dispatch message to a handler
         if (message.command == Command.AgentExecute) {
+            if (channel.mode == OperatingMode.RejectingInboundMessages) {
+                revert Disabled();
+            }
```
vgeddes, line 146: Ron, we need to think about this a bit more. Since setting the mode of the outbound-queue pallet on BridgeHub to `RejectingOutboundMessages` can have the same effect.

So we need to analyse the difference in the approaches, and consider the following issues among others:
* Which one works better when used for upgrades.
* Which approach results in users losing messages/funds? If messages rejected at BridgeHub, or at the Gateway?
* Which one works better for emergency shutdowns?

I would suggest rather leaving this work for another PR until we have that analysis above.

contracts/src/Types.sol
```diff
@@ -45,7 +45,8 @@ struct InboundMessage {
 
 enum OperatingMode {
```
vgeddes, line 46: Assuming we do go ahead with this change (see above comment), you must also make sure update https://github.com/Snowfork/snowbridge/blob/3dad79b95b0ee26d8e8f62973e071a767a386d5b/parachain/primitives/core/src/outbound.rs#L53

parachain/pallets/control/src/lib.rs
```diff
@@ -158,6 +165,91 @@ pub mod pallet {
...
+		/// - `origin`: Must be `MultiLocation`
+		#[pallet::call_index(3)]
+		#[pallet::weight(T::WeightInfo::create_agent())]
+		pub fn update_channel(
+			origin: OriginFor<T>,
+			mode: OperatingMode,
+			fee: u128,
+			reward: u128,
+		) -> DispatchResult {
+			let location: MultiLocation = T::CreateAgentOrigin::ensure_origin(origin)?;
```
alistair-singh, line 209: Maybe we should rename this to `AgentOrigin` as its used for more than just create agent.
```suggestion
			let location: MultiLocation = T::AgentOrigin::ensure_origin(origin)?;
```

alistair-singh:
> @alistair-singh
> 
> Seems a wide range of [XcmOriginToTransactDispatchOrigin](https://github.com/Snowfork/cumulus/blob/5f9f6cd8eb6d779d8b713dd7f931a9521e93b132/parachains/runtimes/bridge-hubs/bridge-hub-rococo/src/xcm_config.rs#L120-L139) allowed in `bridge-hub`. There is also a unit test in our code base with mock data [here](https://github.com/Snowfork/snowbridge/blob/3dad79b95b0ee26d8e8f62973e071a767a386d5b/parachain/pallets/control/src/mock.rs#L79) and test code [here](https://github.com/Snowfork/snowbridge/blob/3dad79b95b0ee26d8e8f62973e071a767a386d5b/parachain/pallets/control/src/tests.rs#L28-L44) to demonstrate that transact origin as relaychain is allowed.
> 
> But I'm a little confused here do we really have a use case `create_agent` for relaychain itself? Because in that case there is no `para_id` which is required for `create_channel` and AFAIK our beefy light client is only able to handle `InboundMessage` from parachain.
> 
> Failing test in https://github.com/Snowfork/snowbridge/actions/runs/5727950221/job/15521537960?pr=917 when extract `para_id` FYI.
> 
> So is that as expected there is some context I missed or a bug we should fix?

The issue with the tests is that it is using the relay chain for the benchmark tests. Looking at these lines: https://github.com/Snowfork/snowbridge/blob/main/parachain/pallets/control/src/mock.rs#L152-L157
I chose the relaychain for this, because it was the first test case that i wrote. You can choose another case that will suit all examples such as a mulit-location with parachain id. That will allow the benchmarks to run for both create agent and the other examples.

alistair-singh:
> [1f0e818](https://github.com/Snowfork/snowbridge/commit/1f0e818aea84d448c9db32af8e4a19e12e5456db) Some workaround to fix the breaking tests. I think maybe we can limit the scope of origins allowed(e.g. with a customized impl of `AgentHashedDescription`) but fine as now to make things simple.

I think having a more custom implementation of `AgentHashedDescription` is a good idea. But if you just need to get around the benchmark failures than we can just choose another multi-location for the benchmarks. https://github.com/Snowfork/snowbridge/blob/main/parachain/pallets/control/src/mock.rs#L152-L157

yrong:
> Since setting the mode of the outbound-queue pallet on BridgeHub to RejectingOutboundMessages can have the same effect as setting the mode of InboundChannel to OperatingMode.RejectingInboundMessages

IMHO it's not an either-or situation. I prefer to close both entry points at the same time to achieve max security, concern based on some previous comment [here](https://github.com/Snowfork/snowbridge/pull/882#discussion_r1274840347).

And I'm fine to move it out to make this PR more focused.

contracts/src/Gateway.sol
yrong: > Since setting the mode of the outbound-queue pallet on BridgeHub to RejectingOutboundMessages can have the same effect as setting the mode of InboundChannel to OperatingMode.RejectingInboundMessages

IMHO it's not an either-or situation. I prefer to close both entry points at the same time to achieve max security, concern based on some previous comment [here](https://github.com/Snowfork/snowbridge/pull/882#discussion_r1274840347).

And I'm fine to move it out to make this PR more focused.

contracts/src/Types.sol
yrong: It's coverd in https://github.com/Snowfork/snowbridge/pull/917/commits/abf2f97dac370b59e4a34ee9574185067041ff12

parachain/pallets/control/src/lib.rs
yrong: I would prefer to rename as `ControlOrigin` as now applies also to `create_channel`.

alistair-singh: I like that name.

contracts/src/Gateway.sol
vgeddes: Yes, lets please move it out.

I added a comment on https://github.com/Snowfork/snowbridge/pull/882#discussion_r1274840347

> Assuming there a bug in code, there is no way to guard against this kind of attack, since if a malicious attacker can manually construct a transaction and have it pass verification by the Gateway, then they can take over entire control of the Gateway using the upgrade handler.

yrong:
Since there is limitation not allowed to send xcm with `XcmPallet` from `asset_hub` directly, https://github.com/Snowfork/cumulus/blob/5f9f6cd8eb6d779d8b713dd7f931a9521e93b132/parachains/runtimes/assets/asset-hub-kusama/src/xcm_config.rs#L465-L470

So we add another parachain which allows it for Control pallet tests which is xcm based.

web/packages/test/scripts/deploy-ethereum.sh
```diff
@@ -76,6 +87,10 @@ deploy_local()
     echo "Waiting for geth API to be ready"
     sleep 3
 
+    if [ "$eth_fast_mode" == "true" ]; then
+      hack_beacon_client
+    fi
```
yrong, line 92: Seems in previous commit we remove `fast_mode`, would prefer we add it back since it's useful to quick test something without waiting for a long time, meanwhile it's just an option not enabled by default.

contracts/src/Gateway.sol
```diff
@@ -260,7 +262,7 @@ contract Gateway is IGateway, IInitializable {
             revert AgentAlreadyCreated();
         }
 
-        address payable agent = payable(new Agent(params.agentID));
+        address payable agent = payable(new Agent(params.agentID,params.paraID));
```
claravanstaden, line 265:
```suggestion
        address payable agent = payable(new Agent(params.agentID, params.paraID));
```

(nit)

claravanstaden (approved):
Nice!

contracts/src/Agent.sol
```diff
@@ -14,9 +16,13 @@ contract Agent {
     /// @dev The gateway contract controlling this agent
     address public immutable GATEWAY;
 
-    constructor(bytes32 agentID) {
+    /// @dev The para_id of the agent
+    ParaID public immutable PARA_ID;
```
vgeddes, line 20: Not all agents are toplevel parachains. Agents can also correspond to nested consensus systems within a parachain, such as pallets or ink! smart contracts.

For this reason, it does not make much sense for an agent to have a paraid field.

parachain/pallets/control/src/lib.rs
```diff
@@ -131,31 +156,156 @@ pub mod pallet {
...
+			Ok(())
+		}
+
+		/// Sends a message to the Gateway contract to create a new Agent representing `origin`
+		///
+		/// - `origin`: Must be `MultiLocation`
+		#[pallet::call_index(2)]
+		#[pallet::weight(T::WeightInfo::create_channel())]
+		pub fn create_channel(origin: OriginFor<T>) -> DispatchResult {
+			let location: MultiLocation = T::ControlOrigin::ensure_origin(origin)?;
```
vgeddes, line 205: Must also verify that the origin is a toplevel parachain, I.e "/Parachain(X)", and not some nested account within a parachain.

Ie only Parachains can create channels, not ink! smart contracts, etc.

parachain/pallets/control/src/lib.rs
```diff
@@ -131,31 +156,156 @@ pub mod pallet {
...
+			ensure!(!Agents::<T>::contains_key(agent_id), Error::<T>::AgentAlreadyCreated);
+
+			Agents::<T>::insert(
+				agent_id,
+				AgentInfo { origin: VersionedMultiLocation::from(location), para_id },
+			);
+
+			let message = Message {
+				origin: T::OwnParaId::get(),
+				command: Command::CreateAgent { agent_id, para_id },
```
vgeddes, line 181: It's not really useful to include the parachain id when creating the agent.

See my comment above.

contracts/src/Agent.sol
yrong: Yes, It's not required for this PR.  But when start with some design for arbitrary transact from Ethereum -> Parachain, 
a new external function in https://linear.app/snowfork/issue/SNO-615 in `Gateway` we need to find the matching ParaId for the agent.

yrong: Yes, it's not required for this PR, but seems nessessary for arbitrary transact from Ethereum to Polkadot some initial design in https://linear.app/snowfork/issue/SNO-615
For `sendTransact(bytes32 agentId, bytes memory payload, uint256 extraFee)` , we need to find the matching ParaId for the particular agent.

vgeddes: Cool, though I added some advice on that in https://linear.app/snowfork/issue/SNO-615#comment-48a711d1

parachain/pallets/control/src/lib.rs
yrong: That depends on how source chain implemente `SendXcmOrigin`. e.g. `xcm_pallet` in `asset_hub_westmint` we use for launching xcm transact tests in this PR use `LocalOriginToLocation` in https://github.com/Snowfork/cumulus/blob/5f9f6cd8eb6d779d8b713dd7f931a9521e93b132/parachains/runtimes/assets/asset-hub-westend/src/xcm_config.rs#L457, so origin becomes `(1,X2(Parachain(x),AccountId(y))) when it reachs bridge_hub.

We may implement some a customize xcm call or a cutomize runtime which encode to /Parachain(x) but I would assume it's not required for this PR and I've some other thought regard this in https://github.com/Snowfork/snowbridge/blob/dc9c4ff0cbd7004cfdc74f3964b2a32055aaad4f/parachain/pallets/control/src/lib.rs#L325-L330

parachain/pallets/control/src/lib.rs
yrong: https://github.com/Snowfork/snowbridge/pull/917#discussion_r1287119996

contracts/src/Agent.sol
yrong: Reverted the change about paraid

contracts/src/Gateway.sol
yrong: Revert the change here for https://linear.app/snowfork/issue/SNO-615#comment-48a711d1

parachain/pallets/control/src/lib.rs
```diff
@@ -55,7 +57,7 @@ pub mod pallet {
 		type MaxUpgradeDataSize: Get<u32>;
 
 		/// Origin check for `create_agent`
```
vgeddes, line 59:
```suggestion
		/// EnsureOrigin implementation that ensures origin is an XCM location 
```

parachain/pallets/control/src/lib.rs
```diff
@@ -76,14 +78,37 @@ pub mod pallet {
 		Upgrade { impl_address: H160, impl_code_hash: H256, params_hash: Option<H256> },
 		/// An CreateAgent message was sent to the Gateway
 		CreateAgent { location: Box<MultiLocation>, agent_id: AgentId },
+		/// An CreateChannel message was sent to the Gateway
+		CreateChannel { location: Box<MultiLocation>, para_id: ParaId, agent_id: AgentId },
```
vgeddes, line 82: Not necessary to log the origin location. That is only required for CreateAgent

```suggestion
		CreateChannel { para_id: ParaId, agent_id: AgentId },
```

parachain/pallets/control/src/lib.rs
```diff
@@ -76,14 +78,37 @@ pub mod pallet {
...
+		/// An UpdateChannel message was sent to the Gateway
+		UpdateChannel {
+			location: Box<MultiLocation>,
```
vgeddes, line 85: Ditto, not required to log the origin location. That is only required for CreateAgent.

parachain/pallets/control/src/lib.rs
```diff
@@ -76,14 +78,37 @@ pub mod pallet {
...
+			agent_id: AgentId,
+			mode: OperatingMode,
+			fee: u128,
+			reward: u128,
+		},
+		/// An SetOperatingMode message was sent to the Gateway
+		SetOperatingMode { mode: OperatingMode },
+		/// An TransferNativeFromAgent message was sent to the Gateway
+		TransferNativeFromAgent {
+			location: Box<MultiLocation>,
```
vgeddes, line 96: Ditto, not required to log the origin location. That is only required for CreateAgent.

parachain/pallets/control/src/lib.rs
```diff
@@ -131,33 +156,153 @@ pub mod pallet {
...
+				target: LOG_TARGET,
+				"💫 Create Agent request processed with outbound message {:?}",
+				message
+			);
 
 			Self::deposit_event(Event::<T>::CreateAgent { location: Box::new(location), agent_id });
 			Ok(())
 		}
+
+		/// Sends a message to the Gateway contract to create a new Agent representing `origin`
```
vgeddes, line 190: The comment needs to be updated.

parachain/pallets/control/src/lib.rs
vgeddes: I see the problem. Thanks Ron.

Anyway, I don't think its a huge  issue since create_channel is idempotent and is not destructive.

So the current code looks good.

parachain/pallets/control/src/lib.rs
```diff
@@ -131,33 +156,153 @@ pub mod pallet {
...
+		///
+		/// - `origin`: Must be `MultiLocation`
+		#[pallet::call_index(2)]
+		#[pallet::weight(T::WeightInfo::create_channel())]
+		pub fn create_channel(origin: OriginFor<T>) -> DispatchResult {
+			let location: MultiLocation = T::ControlOrigin::ensure_origin(origin)?;
+
+			let (agent_id, some_para_id, location) = Self::convert_location(location)?;
+
+			ensure!(Agents::<T>::contains_key(agent_id), Error::<T>::AgentNotExist);
```
vgeddes, line 200: Should probably also add a check to make sure the channel hasn't already been created before.

parachain/pallets/control/src/lib.rs
vgeddes: That technique works great for smoketests that test Polkadot->Ethereum messaging. However it doesn't help us in the opposite direction where for example we need to verify that assets were correctly deposited into the final destination parachain.

parachain/pallets/control/src/lib.rs
```diff
@@ -76,19 +78,41 @@ pub mod pallet {
...
+		AgentNotExist,
+		ChannelAlreadyCreated,
+		ChannelNotExist,
 	}
 
 	#[pallet::storage]
 	pub type Agents<T: Config> = StorageMap<_, Twox64Concat, AgentId, (), OptionQuery>;
 
+	#[pallet::storage]
+	pub type Channels<T: Config> = StorageMap<_, Twox64Concat, ParaId, AgentId, OptionQuery>;
```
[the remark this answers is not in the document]
yrong: I would prefer to keep it to have some view of relationship between Agent<->Channel.  IMO there will not be many channels and the `AgentId` is just a `H256` so will not cost much.

smoketest/tests/create_agent.rs
```diff
@@ -0,0 +1,29 @@
...
+async fn create_agent() {
+    let test_clients = initial_clients().await.expect("initialize clients");
+
+    let message = construct_xcm_message(
+        construct_create_agent_call(&test_clients.bridge_hub_client)
+            .await
+            .expect("construct innner call."),
+    );
+
+    let result = send_xcm_transact(&test_clients.template_client, message)
```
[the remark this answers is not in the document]
yrong: I would prefer to avoid sudo operation from relaychain and let parachain self-governing to create the channel. Btw, the new parachain is also a pre-step for the ongoing arbitratry transact tests.

parachain/pallets/control/src/lib.rs
yrong: https://github.com/Snowfork/snowbridge/pull/917#discussion_r1297081685

parachain/pallets/control/src/lib.rs
```diff
@@ -131,33 +155,147 @@ pub mod pallet {
...
+		/// - `origin`: Must be `MultiLocation`
+		#[pallet::call_index(3)]
+		#[pallet::weight(T::WeightInfo::update_channel())]
+		pub fn update_channel(
+			origin: OriginFor<T>,
+			mode: OperatingMode,
+			fee: u128,
+			reward: u128,
+		) -> DispatchResult {
+			let location: MultiLocation = T::ControlOrigin::ensure_origin(origin)?;
```
vgeddes, line 229: As per my most recent comment in the thread above, we need to ensure that only the origin `/Parachain(X)` can call update channel.

parachain/pallets/control/src/lib.rs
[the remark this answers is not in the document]
vgeddes: We can always derive the agent id on-demand by hashing the origin multilocation though. Given that there is always a 1-1 relationship between para_ids and channels, this is on-demand derivation is always possible.

Given this property, storing agent ids doesn't really add value.

parachain/pallets/control/src/lib.rs
yrong: > we really must enforce that the origin is /Parachain(X) and not /Parachain(X)/AccountId32(Y) or anything else

Actually this is already enforced in `create_channel`  and we check existence of agent when `create_channel` so implicitly applies to `create_agent`.  Reason I did not enforce that in `create_agent` is  that will break most of the test cases in Control pallet with initial intention to allow diversity of origins.

That's also the reason why I asked in https://linear.app/snowfork/issue/SNO-615#comment-445ddfb8. Aidan gave some comment about the scope for longer term and I agree with you to explictily enforce it for now to make less confusion.

yrong: About the helper pallet, I would assume it's just some improvement and would suggest to do it in another PR.

Btw: there is already some low level modules referenced in https://github.com/open-web3-stack/open-runtime-module-library/blob/master/xcm/src/lib.rs

yrong: Some revamp in https://github.com/Snowfork/snowbridge/pull/917/commits/eb484f72e178b96f0e078e2582f0a539bb9bfa82

parachain/pallets/control/src/lib.rs
yrong: Gotcha! It's better to add such limit in `ControlOrigin`.

doubledup (approved):

parachain/pallets/control/src/lib.rs
```diff
@@ -167,5 +259,29 @@ pub mod pallet {
 			T::OutboundQueue::submit(ticket).map_err(|_| Error::<T>::SubmissionFailed)?;
 			Ok(())
 		}
+
+		fn convert_location(
+			mut location: MultiLocation,
+		) -> Result<(H256, ParaId, MultiLocation), DispatchError> {
```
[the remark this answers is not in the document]
yrong: @alistair-singh  We explicit enforce all extrinsics including `CreateAgent` from  `/Parachain(X)` as Vincent mentioned in https://github.com/Snowfork/snowbridge/pull/917#discussion_r1298317752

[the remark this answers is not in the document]
vgeddes: Ah, just to clarify, we must only enforce /Parachain(X) for create_channel and update_channel.

For create_agent, it can accept any xcm origin. This is necessary to support Transact. For example an ink! pallet on an origin parachain must create an agent before it can send transact messages over the bridge.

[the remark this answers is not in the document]
yrong: My bad. revert in https://github.com/Snowfork/snowbridge/pull/917/commits/9d279e97a204451f268581f513ebd53f6b5d4bc4

alistair-singh (approved):
🚀

PR #918, opened 2023-08-02
Separate rust toolchains from global install

doubledup:
As mentioned in [#916](https://github.com/Snowfork/snowbridge/pull/916#issuecomment-1659878334) we need to be more explicit about separating the dev environment's rustup installation and about how we select Rust toolchains.

Using the `.toml` extension in `.cargo` is the preferred form, eg. `.cargo/config.toml`: https://doc.rust-lang.org/cargo/reference/config.html


smoketest/make-bindings.sh
```diff
@@ -15,11 +15,10 @@ command -v subxt || cargo install subxt-cli \
...
     --root $cargo_dir
 
-
 # Fetch metadata from BridgeHub and generate client
-subxt codegen --url ws://localhost:11144 | rustfmt --edition 2021 --emit=stdout > src/parachains/bridgehub.rs
-subxt codegen --url ws://localhost:12144 | rustfmt --edition 2021 --emit=stdout > src/parachains/assethub.rs
-subxt codegen --url ws://localhost:9944  | rustfmt --edition 2021 --emit=stdout > src/parachains/relaychain.rs
+subxt codegen --url ws://localhost:11144 | rustfmt +nightly-2023-07-31 --edition 2021 --emit=stdout > src/parachains/bridgehub.rs
+subxt codegen --url ws://localhost:12144 | rustfmt +nightly-2023-07-31 --edition 2021 --emit=stdout > src/parachains/assethub.rs
+subxt codegen --url ws://localhost:9944  | rustfmt +nightly-2023-07-31 --edition 2021 --emit=stdout > src/parachains/relaychain.rs
```
yrong, line 21: Seems codes above are just ignored in https://github.com/Snowfork/snowbridge/blob/599f14930b34d077d1f8d12ec5931a3a16c61f60/smoketest/.gitignore#L14-L16 
so may not require fmt.

Btw: In previous commit  in PR#915 there is a workaround here 
https://github.com/Snowfork/snowbridge/blob/c2ae10a22140fe0a110d78c6f54251fa8e0d5109/hooks/pre-commit#L16-L35
by adding another `rust-toolchain.toml.nightly` into project and switch between the two toolchains as required.

Then no need to explict add `nightly-2023-07-31`  multiple times. But it seems a bit tricky so I'm not strongly request the change.

yrong (approved):
LGTM!

smoketest/make-bindings.sh
doubledup: I think the formatting is just in case we need to read the code when writing smoke tests - you're right that there's no need for it, but it'll help a bit when writing tests.

There isn't really a good path for our toolchain management right now. Swapping out toolchain files means we need a separate command to set the toolchain before every set of Rust commands. We're removing the toolchain file after the standup discussion, but now users could modify their toolchain and see build behaviour change without a change in the repo. This is fixed by rerunning `scripts/init.sh`, but ideally we would commit a stable and a nightly version (with relevant components & targets) that would be used automatically.

The installation behaviour in `rustup show` will likely be replaced by a subcommand for explicit installation at some point:
https://github.com/rust-lang/rustup/issues/2686 (in particular [this comment](https://github.com/rust-lang/rustup/issues/2686#issuecomment-1084179901))
https://github.com/rust-lang/rustup/issues/1397

yrong: > There isn't really a good path for our toolchain management right now. Swapping out toolchain files means we need a separate command to set the toolchain before every set of Rust commands

Yes, agree. We're doing some workaround here to make both toolchains to work in same repo and It's a bit tricky. TBH I prefer we stick to nightly version for the dev setup.

I would assume actually Parity encourage the use of nightly rust and their docker image to build production release is nigthly also. https://github.com/paritytech/scripts/blob/f901682d2d1b2fa5ed046864f36c2da481532843/dockerfiles/ci-linux/Dockerfile#L7

We can still manually build binary with stable rust to check if it works(I think it should work as we're not binding to nightly specific feature).

yrong: > The Dockerfile mentions that the default is stable

Yes, you're right as defined here
https://github.com/paritytech/scripts/blob/f901682d2d1b2fa5ed046864f36c2da481532843/dockerfiles/base-ci-linux/Dockerfile#L66

rust-toolchain-nightly.toml
```diff
@@ -0,0 +1,7 @@
+[toolchain]
+channel = "nightly-2023-07-31"
```
yrong, line 2: better to sync nightly version with Parity 
https://github.com/paritytech/scripts/blob/f901682d2d1b2fa5ed046864f36c2da481532843/dockerfiles/ci-linux/Dockerfile#L7

scripts/init.sh
```diff
@@ -2,7 +2,26 @@
...
+# NOTE: This ensures that the toolchain in rust-toolchain.toml is installed.
+# rustup has no subcommand (yet) for installing the toolchain in rust-toolchain.toml:
+# https://github.com/rust-lang/rustup/issues/2686
+# The auto-installation behaviour in rustup will likely be removed:
+# https://github.com/rust-lang/rustup/issues/1397
+cp rust-toolchain-stable.toml rust-toolchain.toml
+rustup show
+cp rust-toolchain-nightly.toml rust-toolchain.toml
+rustup show
+cp rust-toolchain-stable.toml rust-toolchain.toml
```
vgeddes, line 21: Hey David, what I meant on Slack is that its OK if rustup automatically installs the stable toolchain using `rust-toolchain.toml`.

These `rust-toolchain-stable.toml` and `rust-toolchain-nightly.toml` files complicate things a bit.

vgeddes: So be more clear:

1. Don't install rust stable in `init.sh`, since rustup will automatically install it.

2. Install `rustfmt` in `init.sh` using `rustup toolchain install $SNOWBRIDGE_RUST_NIGHTLY`, etc.

alistair-singh (approved):

vgeddes (approved):
Looks good now!

PR #919, opened 2023-08-03
Uniform Interface submit beefy commitment

yrong:
### Context

Currently in beefy light client there are two different interfaces for submit beefy commitment with or without handover. I would assume actually there is no big difference between the two branches except for the mmrLeaf verification. 

Furthurmore, IMO it's more secure to also verify mmrLeaf for the non-handover case. So this PR unify the two interfaces to one and it will also make logic in beefy relayer a bit more simple.

The idea here could be immature with something important missing so would like to get some feedback from team before furthur refactor(i.e. revamp and simplify beefy relayer accordingly).

### DIFF
--- a/contracts/src/BeefyClient.sol
+++ b/contracts/src/BeefyClient.sol
@@ -200,10 +200,10 @@ contract BeefyClient {
         nextValidatorSet = _nextValidatorSet;
     }
 
-    /* Public Functions */
+    /* External Functions */
 
     /**
-     * @dev Begin submission of commitment that was signed by the current validator set
+     * @dev Begin submission of commitment
      * @param commitment contains the commitment signed by the validators
      * @param bitfield a bitfield claiming which validators have signed the commitment
      * @param proof a proof that a single validator from currentValidatorSet has signed the commitment
@@ -211,29 +211,13 @@ contract BeefyClient {
     function submitInitial(Commitment calldata commitment, uint256[] calldata bitfield, ValidatorProof calldata proof)
         external
     {
-        doSubmitInitial(currentValidatorSet, commitment, bitfield, proof);
-    }
+        if (commitment.validatorSetID != currentValidatorSet.id && commitment.validatorSetID != nextValidatorSet.id) {
+            revert InvalidCommitment();
+        }
 
-    /**
-     * @dev Begin submission of commitment that was signed by the next validator set
-     * @param commitment contains the commitment signed by the validators
-     * @param bitfield a bitfield claiming which validators have signed the commitment
-     * @param proof a proof that a single validator from nextValidatorSet has signed the commitment
-     */
-    function submitInitialWithHandover(
-        Commitment calldata commitment,
-        uint256[] calldata bitfield,
-        ValidatorProof calldata proof
-    ) external {
-        doSubmitInitial(nextValidatorSet, commitment, bitfield, proof);
-    }
+        bool is_next_validator_set = commitment.validatorSetID == nextValidatorSet.id;
+        ValidatorSet memory vset = is_next_validator_set ? nextValidatorSet : currentValidatorSet;
 
-    function doSubmitInitial(
-        ValidatorSet memory vset,
-        Commitment calldata commitment,
-        uint256[] calldata bitfield,
-        ValidatorProof calldata proof
-    ) internal {
         // Check if merkle proof is valid based on the validatorSetRoot and if proof is included in bitfield
         if (!isValidatorInSet(vset, proof.account, proof.index, proof.proof) || !Bitfield.isSet(bitfield, proof.index))
         {
@@ -289,32 +273,6 @@ contract BeefyClient {
     }
 
     /**
-     * @dev Submit a commitment for final verification
-     * @param commitment contains the full commitment that was used for the commitmentHash
-     * @param bitfield a bitfield claiming which validators have signed the commitment
-     * @param proofs a struct containing the data needed to verify all validator signatures
-     */
-    function submitFinal(Commitment calldata commitment, uint256[] calldata bitfield, ValidatorProof[] calldata proofs)
-        public
-    {
-        (bytes32 commitmentHash, bytes32 ticketID) = validate(commitment, bitfield);
-
-        Ticket storage ticket = tickets[ticketID];
-
-        if (commitment.validatorSetID != currentValidatorSet.id) {
-            revert InvalidCommitment();
-        }
-        verifyCommitment(commitmentHash, bitfield, currentValidatorSet, ticket, proofs);
-
-        bytes32 newMMRRoot = getFirstMMRRoot(commitment);
-        latestMMRRoot = newMMRRoot;
-        latestBeefyBlock = commitment.blockNumber;
-
-        emit NewMMRRoot(newMMRRoot, commitment.blockNumber);
-        delete tickets[ticketID];
-    }
-
-    /**
      * @dev Submit a commitment and leaf for final verification
      * @param commitment contains the full commitment that was used for the commitmentHash
      * @param bitfield claiming which validators have signed the commitment
@@ -323,26 +281,32 @@ contract BeefyClient {
      * @param leafProof an MMR leaf proof
      * @param leafProofOrder a bitfield describing the order of each item (left vs right)
      */
-    function submitFinalWithHandover(
+    function submitFinal(
         Commitment calldata commitment,
         uint256[] calldata bitfield,
         ValidatorProof[] calldata proofs,
         MMRLeaf calldata leaf,
         bytes32[] calldata leafProof,
         uint256 leafProofOrder
-    ) public {
+    ) external {
         (bytes32 commitmentHash, bytes32 ticketID) = validate(commitment, bitfield);
 
         Ticket storage ticket = tickets[ticketID];
 
-        if (commitment.validatorSetID != nextValidatorSet.id) {
+        if (commitment.validatorSetID != currentValidatorSet.id && commitment.validatorSetID != nextValidatorSet.id) {
             revert InvalidCommitment();
         }
-        verifyCommitment(commitmentHash, bitfield, nextValidatorSet, ticket, proofs);
 
-        if (leaf.nextAuthoritySetID != nextValidatorSet.id + 1) {
+        bool is_next_validator_set = commitment.validatorSetID == nextValidatorSet.id;
+
+        verifyCommitment(
+            commitmentHash, bitfield, is_next_validator_set ? nextValidatorSet : currentValidatorSet, ticket, proofs
+        );
+
+        if (is_next_validator_set && leaf.nextAuthoritySetID != nextValidatorSet.id + 1) {
             revert InvalidMMRLeaf();
         }
+
         bytes32 newMMRRoot = getFirstMMRRoot(commitment);
         bool leafIsValid =
             MMRProof.verifyLeafProof(newMMRRoot, keccak256(encodeMMRLeaf(leaf)), leafProof, leafProofOrder);
@@ -350,16 +314,18 @@ contract BeefyClient {
             revert InvalidMMRLeafProof();
         }
 
-        currentValidatorSet = nextValidatorSet;
-        nextValidatorSet.id = leaf.nextAuthoritySetID;
-        nextValidatorSet.length = leaf.nextAuthoritySetLen;
-        nextValidatorSet.root = leaf.nextAuthoritySetRoot;
+        if (is_next_validator_set) {
+            currentValidatorSet = nextValidatorSet;
+            nextValidatorSet.id = leaf.nextAuthoritySetID;
+            nextValidatorSet.length = leaf.nextAuthoritySetLen;
+            nextValidatorSet.root = leaf.nextAuthoritySetRoot;
+        }
 
         latestMMRRoot = newMMRRoot;
         latestBeefyBlock = commitment.blockNumber;
+        delete tickets[ticketID];
 
         emit NewMMRRoot(newMMRRoot, commitment.blockNumber);
-        delete tickets[ticketID];
     }
 
     /**
@@ -411,7 +377,7 @@ contract BeefyClient {
         );
     }
 
-    /* Private Functions */
+    /* Internal Functions */
 
     // Creates a unique ticket ID for a new interactive prover-verifier session
     function createTicketID(address account, bytes32 commitmentHash) internal pure returns (bytes32 value) {
--- a/contracts/test/BeefyClient.t.sol
+++ b/contracts/test/BeefyClient.t.sol
@@ -174,7 +174,7 @@ contract BeefyClientTest is Test {
 
         createFinalProofs();
 
-        beefyClient.submitFinal(commitment, bitfield, finalValidatorProofs);
+        beefyClient.submitFinal(commitment, bitfield, finalValidatorProofs, mmrLeaf, mmrLeafProofs, leafProofOrder);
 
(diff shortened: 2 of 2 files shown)


vgeddes (commented):
Looks good Ron. I think this a good idea. Though lets prioritising completing https://github.com/Snowfork/snowbridge/pull/917.

alistair-singh (approved):

vgeddes (commented):
@yrong Before I give this a thorough review, can you run the E2E stack overnight to exercise the relayer and all the new submission code paths.

I am thinking at least 12 hours, followed by grepping relayer logs for any errors.

yrong:
> I am thinking at least 12 hours, followed by grepping relayer logs for any errors.

Setup kept running during the weekend, from log seems beefy commitment synced as expected after long time period(25809 blocks more than 24 hours as following)
```
{"@timestamp":"2023-08-20T22:34:01.158240339+08:00","blockNumber":25809,"level":"debug","message":"Transaction SubmitFinal succeeded","tx":"0xe6600caa241fb09616d742ceddedf19764c88d4a212f7865b5dad5950631f63e"}
```

contracts/src/BeefyClient.sol
```diff
@@ -200,40 +200,24 @@ contract BeefyClient {
...
      * @param commitment contains the commitment signed by the validators
      * @param bitfield a bitfield claiming which validators have signed the commitment
      * @param proof a proof that a single validator from currentValidatorSet has signed the commitment
      */
     function submitInitial(Commitment calldata commitment, uint256[] calldata bitfield, ValidatorProof calldata proof)
         external
     {
-        doSubmitInitial(currentValidatorSet, commitment, bitfield, proof);
-    }
+        if (commitment.validatorSetID != currentValidatorSet.id && commitment.validatorSetID != nextValidatorSet.id) {
```
vgeddes, line 214: I think we should somehow combine this check with the code below. Because its duplicated arithmetic and an extra storage load which costs extra gas.

```solidity
bool is_next_validator_set = commitment.validatorSetID == nextValidatorSet.id;
```

vgeddes: There's a duplicate, redundant vset.id check on line 243.

contracts/src/BeefyClient.sol
```diff
@@ -323,43 +281,51 @@ contract BeefyClient {
...
         Ticket storage ticket = tickets[ticketID];
 
-        if (commitment.validatorSetID != nextValidatorSet.id) {
+        if (commitment.validatorSetID != currentValidatorSet.id && commitment.validatorSetID != nextValidatorSet.id) {
             revert InvalidCommitment();
         }
-        verifyCommitment(commitmentHash, bitfield, nextValidatorSet, ticket, proofs);
 
-        if (leaf.nextAuthoritySetID != nextValidatorSet.id + 1) {
+        bool is_next_validator_set = commitment.validatorSetID == nextValidatorSet.id;
```
vgeddes, line 300: Alternative, more efficient approach, which combines both checks:

```solidity
is_next_validator_set = false;
if (commitment.validatorSetID == nextValidatorSet.id) {
    is_next_validator_set = true;
else if (commitment.validatorSetID != currentValidatorSet.id) {
    revert InvalidCommitment();
}
```

contracts/src/BeefyClient.sol
```diff
@@ -323,43 +281,51 @@ contract BeefyClient {
...
             MMRProof.verifyLeafProof(newMMRRoot, keccak256(encodeMMRLeaf(leaf)), leafProof, leafProofOrder);
         if (!leafIsValid) {
             revert InvalidMMRLeafProof();
         }
 
-        currentValidatorSet = nextValidatorSet;
-        nextValidatorSet.id = leaf.nextAuthoritySetID;
-        nextValidatorSet.length = leaf.nextAuthoritySetLen;
-        nextValidatorSet.root = leaf.nextAuthoritySetRoot;
+        if (is_next_validator_set) {
```
vgeddes, line 317: I think we should only verify the leaf within this block. Otherwise we're just wasting a huge amount of gas.

vgeddes (commented):
Looks good, added first round of comments.

contracts/src/BeefyClient.sol
yrong: https://github.com/Snowfork/snowbridge/pull/919/commits/002016f6fae5f391c03c049d9f18c9ee328b25be

contracts/src/BeefyClient.sol
yrong: As mentioned, the initial intention for this PR is  

> Also verify MMRLeaf for the non-handover case before updating MMRRoot

mainly for security considerations, gas consumed did increase but seems acceptable as follows(e.g. 3.790% for the non-handover testSubmit):

```
Test result: ok. 23 passed; 0 failed; finished in 29.97ms
testCreateInitialBitfield() (gas: -44 (-0.007%))
testCreateInitialBitfieldInvalid() (gas: -44 (-0.008%))
testRegenerateBitField() (gas: -22 (-0.018%))
testSubmitWithHandoverFailStaleCommitment() (gas: 430 (0.063%))
testScaleEncodeCommit() (gas: -22 (-0.067%))
testSubmitWithHandoverFailWithoutPrevRandao() (gas: 430 (0.077%))
testSubmitFailWithInvalidMMRLeafProof() (gas: 874 (0.116%))
testSubmitWithHandover() (gas: 857 (0.116%))
testSubmitFailWithInvalidMMRLeaf() (gas: 897 (0.120%))
testSubmitFailForPrevRandaoCapturedMoreThanOnce() (gas: 344 (0.128%))
testCreateFinalBitfield() (gas: 366 (0.131%))
testCreateFinalBitfieldInvalid() (gas: 366 (0.137%))
testSubmitWithHandoverFailWithInvalidValidatorSet() (gas: 1078 (0.161%))
testSubmitFailForPrevRandaoTooEarlyOrTooLate() (gas: 344 (0.165%))
testSubmitFailWithNotEnoughClaims() (gas: 367 (0.184%))
testSubmitFailWithStaleCommitment() (gas: 29291 (3.494%))
testSubmitFailInvalidSignature() (gas: 24694 (3.749%))
testSubmit():((uint32,uint64,(bytes2,bytes)[])) (gas: 26916 (3.790%))
testSubmitFailWithInvalidTicket() (gas: 24397 (3.813%))
testSubmitFailValidatorNotInBitfield() (gas: 24694 (3.819%))
testSubmitFailWithInvalidBitfield() (gas: 24429 (3.841%))
testSubmitFailWithInvalidValidatorSet() (gas: 24964 (3.871%))
testSubmitFailWithoutPrevRandao() (gas: 24407 (4.574%))
Overall gas change: 210013 (1.741%)
```

contracts/src/BeefyClient.sol
```diff
@@ -323,43 +283,50 @@ contract BeefyClient {
...
         uint256 leafProofOrder
-    ) public {
+    ) external {
         (bytes32 commitmentHash, bytes32 ticketID) = validate(commitment, bitfield);
 
-        Ticket storage ticket = tickets[ticketID];
-
-        if (commitment.validatorSetID != nextValidatorSet.id) {
+        bool is_next_session = false;
+        ValidatorSet memory vset = currentValidatorSet;
```
vgeddes, line 297: See my above comment, we can save gas costs by not loading `currentValidatorSet` unnecessarily.

contracts/src/BeefyClient.sol
vgeddes: It's still not clear to me why there it improves security to verify the MMR leaf in the non-handover case.  In the non-handover case, the data in the MMR leaf is not used at all, so the leaf's validity is of no importance, I'd say.

Regarding gas costs. An increase of 4% is roughly equivalent to an increase of $40,000 in yearly operating costs. So I would like to avoid extra costs if at all possible.

yrong: > An increase of 4% is roughly equivalent to an increase of $40,000 in yearly operating costs. 

Did not realize that much! Good to know and yes we need to be super careful in gas consumption.

relayer/relays/beefy/polkadot-listener.go
```diff
@@ -82,36 +82,36 @@ func (li *PolkadotListener) scanCommitments(
 			validatorSetID := result.SignedCommitment.Commitment.ValidatorSetID
 			nextValidatorSetID := uint64(result.MMRProof.Leaf.BeefyNextAuthoritySet.ID)
 
+			if validatorSetID < currentValidatorSet || validatorSetID > currentValidatorSet+1 {
+				return fmt.Errorf("commitment has unexpected validatorSetID: blockNumber=%v validatorSetID=%v expectedValidatorSetID=%v",
+					committedBeefyBlock,
+					validatorSetID,
+					currentValidatorSet,
+				)
```
doubledup, line 90: It might be clearer to phrase this as the exact 2 cases we expect:
```suggestion
			if validatorSetID != currentValidatorSet && validatorSetID != currentValidatorSet+1 {
				return fmt.Errorf("commitment has unexpected validatorSetID: blockNumber=%v validatorSetID=%v currentValidatorSetID=%v",
					committedBeefyBlock,
					validatorSetID,
					currentValidatorSet,
				)
```

doubledup (approved):
LGTM! Main thing is we can save some gas by continuing to skip the leaf MMR proof in the non-handover case.

contracts/src/BeefyClient.sol
```diff
@@ -323,43 +283,50 @@ contract BeefyClient {
...
+        if (is_next_session && leaf.nextAuthoritySetID != nextValidatorSet.id + 1) {
             revert InvalidMMRLeaf();
         }
+
         bytes32 newMMRRoot = getFirstMMRRoot(commitment);
         bool leafIsValid =
             MMRProof.verifyLeafProof(newMMRRoot, keccak256(encodeMMRLeaf(leaf)), leafProof, leafProofOrder);
         if (!leafIsValid) {
             revert InvalidMMRLeafProof();
         }
```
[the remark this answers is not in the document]
doubledup: Duplicate of https://github.com/Snowfork/snowbridge/pull/919#discussion_r1300514409

contracts/src/BeefyClient.sol
doubledup: I believe it's because the handover case updates `nextValidatorSet` with values from the `leaf`: https://github.com/Snowfork/snowbridge/pull/919/files#diff-473fead6cd500a4a05f06d0cea4d9ef166c9bdbe2a2e1a06a33d2e1aedcc428eR320-R325

Leaf proof verification checks that the leaf is part of a Merkle tree - it only verifies the leaf, not the `newMMRRoot`. Given that we don't use the leaf in the non-handover case, there's no need to verify it.

yrong: That makes sense, thanks for the explanation. Revamp in https://github.com/Snowfork/snowbridge/pull/919/commits/4d7ddc3259adfde193a1a7b3b3b36737bc23c0f2

contracts/test/BeefyClient.t.sol
```diff
@@ -396,27 +390,25 @@ contract BeefyClientTest is Test {
...
+        beefyClient.submitInitial(commitment, bitfield, finalValidatorProofs[0]);
 
         vm.roll(block.number + randaoCommitDelay);
         commitPrevRandao();
 
         createFinalProofs();
 
         //reinitialize with next validator set
-        initialize(setId);
+        initialize(setId + 1);
```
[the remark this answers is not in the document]
yrong: Exactly.

contracts/src/BeefyClient.sol
yrong: Actually my previous thought is that in `MMRLeaf` there are fields like `parentHash` and ` parachainHeadsRoot` which will be used to verify the new `MMRRoot` even in non-handover case.   I would assume though the proof can be faked but adds some complexity for bad relayers.

If you all think that does not make much sense, then I would prefer to abandon the improvement here. Because now always require `MMRLeaf` as call data which costs much more gas(3%-4%).

doubledup (approved):

vgeddes (approved):
+1

contracts/src/BeefyClient.sol
vgeddes: I don't think we can trust the "gas diff". because it also includes gas executed during test setup.

You should run `forge test --match-path test/BeefyClient.t.sol --gas-report`, which gives more info:
```
 test/mocks/BeefyClientMock.sol:BeefyClientMock contract |                 |       |        |        |         |
|---------------------------------------------------------|-----------------|-------|--------|--------|---------|
| Deployment Cost                                         | Deployment Size |       |        |        |         |
| 2186482                                                 | 11250           |       |        |        |         |
| Function Name                                           | min             | avg   | median | max    | # calls |
| commitPrevRandao                                        | 785             | 17943 | 21119  | 21119  | 19      |
| createFinalBitfield                                     | 1134            | 9966  | 10597  | 10597  | 15      |
| createInitialBitfield                                   | 544             | 52136 | 52804  | 70348  | 52      |
| encodeCommitment_public                                 | 5556            | 5556  | 5556   | 5556   | 1       |
| initialize_public                                       | 2281            | 84897 | 92081  | 92081  | 25      |
| latestBeefyBlock                                        | 466             | 466   | 466    | 466    | 3       |
| minimumSignatureThreshold_public                        | 267             | 267   | 267    | 267    | 1       |
| setLatestBeefyBlock                                     | 20364           | 20364 | 20364  | 20364  | 1       |
| submitFinal                                             | 4569            | 31392 | 6444   | 108887 | 9       |
| submitFinalWithHandover                                 | 5071            | 52439 | 48615  | 112192 | 6       |
| submitInitial                                           | 5625            | 53973 | 61254  | 61254  | 15      |
| submitInitialWithHandover                               | 5515            | 53197 | 61144  | 61144  | 7       |
```

vgeddes: I have an idea. According to the Ethereum spec:

> The Ethereum Virtual Machine imposes a cost of 4 gas for every byte set to zero in calldata, and 16 gas for every non-zero byte (reference: [Ethereum Yellow Paper](https://ethereum.github.io/yellowpaper/paper.pdf)).

If for the non-handover case, the relayer passes **zero values** for `leaf` and `leafProof` when calling `submitFinal`, then the increase in gast cost would be minimal.

Can you benchmark this approach to find out?

yrong: |---------------------------------------------------------|-----------------|-------|--------|--------|---------|
| Deployment Cost                                         | Deployment Size |       |        |        |         |
| 2197297                                                 | 11304           |       |        |        |         |
| Function Name                                           | min             | avg   | median | max    | # calls |
| commitPrevRandao                                        | 763             | 17921 | 21097  | 21097  | 19      |
| createFinalBitfield                                     | 1134            | 9966  | 10597  | 10597  | 15      |
| createInitialBitfield                                   | 544             | 52080 | 52804  | 70348  | 48      |
| encodeCommitment_public                                 | 5556            | 5556  | 5556   | 5556   | 1       |
| initialize_public                                       | 2259            | 84250 | 92059  | 92059  | 23      |
| latestBeefyBlock                                        | 488             | 488   | 488    | 488    | 3       |
| minimumSignatureThreshold_public                        | 267             | 267   | 267    | 267    | 1       |
| setLatestBeefyBlock                                     | 20364           | 20364 | 20364  | 20364  | 1       |
| submitFinal                                             | 4829            | 40169 | 6704   | 112156 | 15      |
| submitInitial                                           | 14308           | 58987 | 61586  | 61853  | 20      |

PR #921, opened 2023-08-04
Avoid global pnpm store

doubledup:
The root cause of the weird behaviour I was seeing on my machine was a default global cache used by pnpm. A package in the cache had a binary that referenced dynamic libraries in Nix derivations that were removed in garbage collection.

This change makes pnpm use a store in the project directory, which isolates our dev environments more. The cost is that we're not using packages cached across projects, though our monorepo approach limits the usefulness of a global cache. Also with external dependencies like the dynamic library found here, the isolation's more useful.

Additional changes:
- Post-checkout hook to keep the cumulus submodule updated.
- Fix for Rust formatting in the pre-commit hook (with formatting changes in the parachain).
- Fix for relayer contract bindings: compile contracts before the relayer, so the generated bindings are up-to-date. This blocks fresh environments, as they have no contract artifacts to use when generating go bindings. Existing envs may get out-of-date contract bindings, causing relayer requests to contracts to fail until the relayer is rebuilt with the updated bindings (usually via `start-services`).


yrong (commented):
Cool!  Seems there are some rust format change also included, would prefer to move it out to make this pr focused and actually there is https://github.com/Snowfork/snowbridge/pull/920 specific for that.

hooks/post-checkout
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+
+set -e
+
+git submodule update
```
yrong, line 5: Is the check nessessary? Seems already coverd in init script.

hooks/pre-commit
```diff
@@ -23,7 +23,7 @@ chronic typos .
 (cd relayer && chronic mage lint && chronic go fmt ./...)
 
 # format rust
-(cd parachain && chronic cargo +nightly fmt)
+(cd parachain && chronic cargo +nightly-"$SNOWBRIDGE_RUST_NIGHTLY" fmt)
```
yrong, line 26: Nice, I'll add that to https://github.com/Snowfork/snowbridge/pull/920

hooks/post-checkout
doubledup: The init script is only run once, when setting up the repo. This runs after switching branches, to ensure that the cumulus version is correct on the new branch.

hooks/pre-commit
doubledup: I'll remove the formatting here & leave this change to #920

yrong (approved):
LGTM!

claravanstaden (approved):
Nice! 👍

PR #923, opened 2023-08-08
SNO-604: Add ERC20 contract

doubledup:
Uses the [Axelar ERC20 contract](https://github.com/axelarnetwork/axelar-cgp-solidity/blob/main/contracts/ERC20.sol), but with a mint method restricted to an owner instead of a separate [supply mechanism](https://forum.openzeppelin.com/t/how-to-implement-erc20-supply-mechanisms/226). Idea being that each parachain will register its own tokens via its agent, so minting tokens must be restricted to that agent.

[SNO-604](https://linear.app/snowfork/issue/SNO-604)


yrong:
>  each parachain will register its own tokens via its agent, so minting tokens must be restricted to that agent
So will it make sense add `agent_id` into construct of the token so that we can check the sender as expected?

Also in [MintableCappedERC20](https://github.com/axelarnetwork/axelar-cgp-solidity/blob/main/contracts/MintableCappedERC20.sol) there is a cap could be useful in some cases.

doubledup:
Not sure about supply caps for wrapped tokens. I'd rather leave them out for now as we don't have a need for them yet afaik:

The bridge should never accept messages on Polkadot to mint tokens that aren't backed 1-1. So if this cap becomes relevant (because we're over-minting), the bridge has already failed (because the wrapped assets aren't fully backed). The cap could mitigate this failure for tokens that use it, but some won't (eg. DOT) so we have to avoid this failure case anyway.

This could be an argument for separating `mint` from the main `ERC20` contract, if we find a case for another `mint` implementation.

yrong (approved):
LGTM!

contracts/src/ERC20.sol
```diff
@@ -0,0 +1,235 @@
...
+ * and does not conflict with the expectations of ERC20 applications.
+ *
+ * Additionally, an {Approval} event is emitted on calls to {transferFrom}.
+ * This allows applications to reconstruct the allowance for all accounts just
+ * by listening to said events. Other implementations of the EIP may not emit
+ * these events, as it isn't required by the specification.
+ *
+ * Finally, the non-standard {decreaseAllowance} and {increaseAllowance}
+ * functions have been added to mitigate the well-known issues around setting
+ * allowances. See {IERC20-approve}.
```
musnit, line 31: cleanup comment plz

contracts/src/ERC20.sol
```diff
@@ -0,0 +1,235 @@
+// SPDX-License-Identifier: MIT
+// SPDX-FileCopyrightText: 2023 Axelar Network
```
musnit, line 2: i think remove this is we've made significant changes?

doubledup: Only the license identifier is in the original files, so that's all we need to keep

vgeddes: The MIT license, like most permissive free software licenses, requires that the original copyright be maintained in derived works:

> The above copyright notice and this permission notice shall be included in all
> copies or substantial portions of the Software.

So rather add a second `SPDX-FileCopyrightText` line with our copyright.

contracts/src/ERC20Permit.sol
```diff
@@ -0,0 +1,77 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity 0.8.20;
+
+import {IERC20} from "./interfaces/IERC20.sol";
+import {IERC20Permit} from "./interfaces/IERC20Permit.sol";
+
+import {ERC20} from "./ERC20.sol";
+import {Ownable} from "openzeppelin/access/Ownable.sol";
```
vgeddes, line 9: I don't really want to rely on Openzeppelin. I've removed most of its use from the codebase. See my preferred solution in the comments below.

contracts/src/ERC20Permit.sol
```diff
@@ -0,0 +1,77 @@
...
+    bytes32 private constant DOMAIN_TYPE_SIGNATURE_HASH =
+        bytes32(0x8b73c3c69bb8fe3d512ecc4cf759cc79239f7b179b0ffacaa9a75d522b39400f);
+
+    // keccak256('Permit(address owner,address spender,uint256 value,uint256 nonce,uint256 deadline)')
+    bytes32 private constant PERMIT_SIGNATURE_HASH =
+        bytes32(0x6e71edae12b1b97f4d1f60370fef10105fa2faae0126114a169c64845d6126c9);
+
+    mapping(address => uint256) public nonces;
+
+    constructor(string memory name_, string memory symbol_, uint8 decimals_) ERC20(name_, symbol_, decimals_) {
```
vgeddes, line 31: Add an `address immutable owner` variable, and initialize it with `owner = msg.sender`. Similar to the pattern used in the constructor for `Agent.sol`.

Then you can also introduce a modifier similar to the following to implement owner checks:
https://github.com/Snowfork/snowbridge/blob/60a82e0d6bc6ff214124e8aa2f29d3ec566bc0e5/contracts/src/Gateway.sol#L67

contracts/src/ERC20Permit.sol
vgeddes, line 1: Can you merge `ERC20.sol` and `ERC20Permit.sol` into a single file named `ERC20.sol`? We don't really use inheritance in our codebase as I believe it makes the code harder to understand and audit.

contracts/src/interfaces/IERC20Permit.sol
```diff
@@ -0,0 +1,13 @@
+// SPDX-License-Identifier: MIT
+// SPDX-FileCopyrightText: 2023 Axelar Network
+
+pragma solidity ^0.8.9;
```
vgeddes, line 4: This should match our currently used compiler version.

alistair-singh (approved):

PR #924, opened 2023-08-08
Arbitrary call from Polkadot to Ethereum contract

claravanstaden:
- [x] Smoke test to send xcm to template parachain
- [x] Template chain forwards message to bridge hub
- [x] Bridge hub forwards message to outbound router
- [x] Outbound router message transform to Gateway contract
- [x] Execute contract in AgentExecute
- [ ] Figure out how to do the dynamic fees

Cumulus companion: https://github.com/Snowfork/cumulus/pull/57


parachain/primitives/core/src/outbound.rs
```diff
@@ -225,6 +235,14 @@ impl AgentExecuteCommand {
 					Token::Uint(U256::from(*amount)),
 				])),
 			]),
+			AgentExecuteCommand::Transact { target, payload, dynamic_gas } => ethabi::encode(&[
+				Token::Uint(self.index().into()), // TODO not sure what this does
```
yrong, line 239: It's the abi encode of the Transact Enum matching what you defined above in solidity types.

contracts/src/AgentExecutor.sol
```diff
@@ -38,4 +43,9 @@ contract AgentExecutor {
     function _transferToken(address token, address recipient, uint128 amount) internal {
         IERC20(token).safeTransfer(recipient, amount);
     }
+
+    /// @dev Call a contract at the given address, with provided bytes as payload.
+    function _executeCall(address target, bytes memory payload, uint256 dynamicGas) internal {
+       target.call{gas: dynamicGas}(payload);
```
yrong, line 49: Should be `delegateCall` based.

yrong, line 49: Should check gas left and handle error here.

parachain/primitives/core/src/outbound.rs
claravanstaden: @yrong I'm specifically wondering what the `index` does?

yrong: Just the enum index in solidity as you defined.
https://github.com/Snowfork/snowbridge/blob/5b87c0615c7b2e11cd77727b1ddde5b2b988c421/contracts/src/Types.sol#L76

contracts/src/AgentExecutor.sol
yrong: Yes, we need to check result and revert if error happens, something like 
https://github.com/axelarnetwork/axelar-cgp-solidity/blob/main/contracts/util/Caller.sol

parachain/primitives/router/src/outbound/mod.rs
```diff
@@ -197,7 +200,7 @@ impl<'a, Call> XcmConverter<'a, Call> {
 		let max_target_fee = self.fee_info()?;
 
 		// Get withdraw/deposit and make native tokens create message.
-		let result = self.native_tokens_unlock_message()?;
+		let result = self.agent_execute_message()?;
```
yrong, line 203: Nice! And just realize that the original xcm transact should be constructed from another parachain and sent through a `BridgeXcmSender`(e.g.  take [BridgeXcmSender](https://github.com/Snowfork/cumulus/blob/5f9f6cd8eb6d779d8b713dd7f931a9521e93b132/parachains/runtimes/assets/asset-hub-kusama/src/lib.rs#L727) for `pallet_bridge_transfer` in `asset_hub_runtime` as reference).

Since a template runtime is introduced in https://github.com/Snowfork/cumulus/pull/56,  we can wrap a transact extrinsic into it and then do some e2e tests accordingly.

yrong: Currently `asset_hub_runtime` use [UnpaidRemoteExporter](https://github.com/Snowfork/cumulus/blob/5f9f6cd8eb6d779d8b713dd7f931a9521e93b132/parachains/runtimes/assets/asset-hub-kusama/src/xcm_config.rs#L635-L636)  for fee of  asset transfer.  But when it comes to arbitrary message from arbitrary sibling parachain,  [SovereignPaidRemoteExporter](https://github.com/paritytech/polkadot/blob/8f05479e4bd61341af69f0721e617f01cbad8bb2/xcm/xcm-builder/src/universal_exports.rs#L217) make more sense and the delivery fee [here] should also be carefully addressed.

claravanstaden: > Nice! And just realize that the original xcm transact should be constructed from another parachain and sent through a BridgeXcmSender(e.g. take [BridgeXcmSender](https://github.com/Snowfork/cumulus/blob/5f9f6cd8eb6d779d8b713dd7f931a9521e93b132/parachains/runtimes/assets/asset-hub-kusama/src/lib.rs#L727) for pallet_bridge_transfer in asset_hub_runtime as reference).

Should `BridgeXcmSender` be added to the bridge hub runtime? Or to the template pallet runtime?

claravanstaden:
@yrong I have added a new extrinsic in the [template parachain](https://github.com/Snowfork/cumulus/pull/57/files#diff-b8badc3d2955238ef54eb6d7544ff6daca104a9cc24b256097f8d9d035c84208R150). To use the [BridgeXcmSender](https://github.com/Snowfork/cumulus/blob/5f9f6cd8eb6d779d8b713dd7f931a9521e93b132/parachains/runtimes/assets/asset-hub-kusama/src/lib.rs#L727), I need to add a bunch of config from the `pallet-bridge-transfer-primitives` crate. I'm not sure including the `pallet-bridge-transfer-primitives` as a dependency (which, incidentally, seems to have some no-std problem) in the template parachain is the right move? Let me know your thoughts and if this seems on the right track.

yrong:
> I'm not sure including the pallet-bridge-transfer-primitives as a dependency (which, incidentally, seems to have some no-std problem) in the template parachain is the right move? 

Yes, actually it's not required. But since you've done some integration already so I just continue from current branch and did some fix in https://github.com/Snowfork/snowbridge/pull/924/commits/faccaee25f129b34d1cce628ac64351c4d3251a5

```2023-08-15 10:17:42.071  INFO tokio-runtime-worker xcm::ethereum_blob_exporter: [Parachain] 🤩 converting.```

Logs shows it comes to `router-primitives` as expected, I would assume will not block you to continue with the unfinished `Convert` logic.

File: parachain/primitives/router/src/outbound/mod.rs [hunk 4]
```diff
@@ -303,6 +365,43 @@ impl<'a, Call> XcmConverter<'a, Call> {
...
+		} = self.next()?
+		{
+			call
+		} else {
+			return Err(TransactExpected)
+		};
+
+		Ok(AgentExecuteCommand::Transact {
+			target: contract_address.into(),
+			payload: data.encode(),
```
claravanstaden, line 400: @yrong The `data` to be sent to Ethereum is of type `DoubledEncoded<Call>`. I am struggling to get it into a `Vec<u8>` type. I have used `.encode()` here, but I do not think this is correct. It also prepends an extra byte to the data array, as seen by a test here: https://github.com/Snowfork/snowbridge/pull/924/files#diff-8de255c76ce8e2d34cf5bc5a3d0a46919ff8bf9f7af384d29a85dad048b95ec7R1701

parachain/primitives/router/src/outbound/mod.rs
```diff
@@ -225,11 +264,34 @@ impl<'a, Call> XcmConverter<'a, Call> {
...
 		};
+		log::trace!(target: "xcm::ethereum_blob_exporter", "execution_fee is {execution_fee:?}.");
 		Ok(execution_fee)
 	}
 
+	fn is_transact(&self) -> bool {
+		for instruction in self.iter.clone() {
+			// TODO not certain if this is OK to differentiate between TransferToken and Transact
+			// messages
+			if let Transact { .. } = instruction {
```
claravanstaden, line 275: @yrong I need to distinguish between a `TransferToken` and `Transact` message in the router. Here, I just look if there is a Transact message in the XCM messages. I am not sure this is correct, please let me know.

parachain/primitives/router/src/outbound/mod.rs
```diff
@@ -1492,4 +1591,108 @@ mod tests {
...
+
+		let fee = MultiAsset { id: Concrete(Here.into()), fun: Fungible(1000) };
+		let fees: MultiAssets = vec![fee.clone()].into();
+
+		let message: Xcm<()> = vec![
+			UnpaidExecution { weight_limit: Unlimited, check_origin: None },
+			DescendOrigin(X1(AccountKey20 {
+				network: Some(NetworkId::Ethereum{ chain_id: 15 }),
+				key: hex!("1000000000000000000000000000000000000000"),
+			})),
```
claravanstaden, line 1678: @yrong  I need to send the Ethereum contract address as part of the XCM message. It does not look like it would form part of the Transact XCM message. `DescendOrigin` looked like the closest thing that matched, let me know if this is OK.

parachain/primitives/router/src/outbound/mod.rs
yrong: > prepends an extra byte to the data array

Not running the test yet but guess the extra byte is the length of the array which is how scale encode array. So we can just remove the extra byte here.

parachain/primitives/router/src/outbound/mod.rs
yrong: Should be fine as a workaround since there is not much `AgentExecuteCommand` to distinguish, and it's unlikely to include a `Transact` into the `Transfer*` Command.

I'll also check if there is some xcm primitive for user customized data(like the type of Command here).

parachain/primitives/router/src/outbound/mod.rs
yrong: Yes, you can do that. But I think the `Gateway` address is already hardcode as a const into the runtime [here](https://github.com/Snowfork/snowbridge/blob/30be20dca6a505fe7c3c02f1d496b60c967f4cb5/parachain/primitives/router/src/outbound/mod.rs#L31) so you may not need it.

rust-toolchain.toml
```diff
@@ -3,7 +3,7 @@
 # https://github.com/rust-lang/rustup/issues/2686
 # The auto-installation behaviour in rustup will likely be removed:
 # https://github.com/rust-lang/rustup/issues/1397
-channel = "1.70.0"
+channel = "nightly-2023-05-23"
```
yrong, line 6: Why change to nightly here?

parachain/primitives/router/src/outbound/mod.rs
```diff
@@ -37,14 +37,19 @@ where
...
 		message: &mut Option<Xcm<()>>,
 	) -> SendResult<Self::Ticket> {
 		let gateway_location = GatewayLocation::get();
 		let universal_location = UniversalLocation::get();
 
+		log::info!(target: "xcm::ethereum_blob_exporter", "🤩 validate message {:?}", message);
+		log::info!(target: "xcm::ethereum_blob_exporter", "🤩 network {:?}", network);
+		log::info!(target: "xcm::ethereum_blob_exporter", "🤩 universal_source {:?}", universal_source);
+		log::info!(target: "xcm::ethereum_blob_exporter", "🤩 channel {:?}", channel);
+
```
yrong, line 52: Can be merged into one log.

rust-toolchain.toml
claravanstaden: Just for local testing - to run Intellij test runner - it uses `rust-toolchain.toml`. Will r4evert before I merge. 😄

parachain/primitives/router/src/outbound/mod.rs
yrong: Ah, it's the destination contract not the gateway. Yes, definitely need it.

parachain/primitives/router/src/outbound/mod.rs
```diff
@@ -303,6 +364,46 @@ impl<'a, Call> XcmConverter<'a, Call> {
...
+				return Err(InvalidJunction)
+			}
+		} else {
+			return Err(DescendOriginExpected)
+		};
+
+		log::trace!(target: "xcm::ethereum_blob_exporter", "before transact.");
+
+		let data = if let Transact {
+			origin_kind: _origin_kind,
```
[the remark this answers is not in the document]
claravanstaden: Addressed in https://github.com/Snowfork/snowbridge/pull/924/commits/6b3b09de254ead22ffb279eeb8db5b98976477fc.

parachain/primitives/router/src/outbound/mod.rs
```diff
@@ -225,11 +264,34 @@ impl<'a, Call> XcmConverter<'a, Call> {
...
+			if let Transact { .. } = instruction {
+				log::trace!(target: "xcm::ethereum_blob_exporter", "is transact message.");
+				return true
+			}
+		}
+		log::trace!(target: "xcm::ethereum_blob_exporter", "is not transact message.");
+		false
+	}
+
+	fn agent_execute_message(&mut self) -> Result<AgentExecuteCommand, XcmConverterError> {
```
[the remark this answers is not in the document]
claravanstaden: Good idea. Addressed here: https://github.com/Snowfork/snowbridge/pull/924/commits/efdf43428df49b5db486479029ce45efbba142b6#diff-8de255c76ce8e2d34cf5bc5a3d0a46919ff8bf9f7af384d29a85dad048b95ec7R238.

smoketest/tests/transact_polkadot_to_ethereum.rs
```diff
@@ -0,0 +1,130 @@
...
+    println!("data is {}", hex::encode(encoded_data.clone()));
+
+    let contract_location = Junctions::X1(AccountKey20 {
+        network: Some(Ethereum { chain_id: 15 }),
+        key: HELLO_WORLD_CONTRACT.into(),
+    });
+
+    let inner_message = Box::new(Xcm(vec![
+        Instruction::UnpaidExecution { weight_limit: Unlimited, check_origin: None },// TODO update to paid
+        Instruction::DescendOrigin(contract_location), // TODO not sure if this is right, want to pass the contract address
```
claravanstaden, line 71: @alistair-singh the `Transact` instruction doesn't have a field to specify the contract to be executed on Ethereum. Is `DescendOrigin` an OK way to send across the address of the arbitrary contract to be executed?

yrong:
@claravanstaden For the issue you mentioned yesterday I just made a fix in https://github.com/Snowfork/snowbridge/pull/924/commits/5db97866bfedcd51f45473c9f1296d451998feb1 and the smoke test shows e2e flow work as expected.

contracts/src/Gateway.sol
```diff
@@ -133,45 +133,48 @@ contract Gateway is IGateway, IInitializable {
         // Otherwise malicious relayers can break the bridge by allowing the message handlers below to run out gas and fail silently.
         // In this scenario case, the channel's state would have been updated to accept the message (by virtue of the nonce increment), yet the actual message
         // dispatch would have failed
-        if (gasleft() < DISPATCH_GAS + BUFFER_GAS) {
+
+        uint256 computedGas = computeGas(message.command, message.params);
+
+        if (gasleft() < computedGas + BUFFER_GAS) {
             revert NotEnoughGas();
         }
```
yrong, line 141: Since `dynamicGas` is provided by the end-user manipulated with an unrealistic high value, then the check here will always fail, and the bridge stuck.

claravanstaden: That's a good point, thanks @yrong!

claravanstaden: @yrong although, `dynamicGas` is only checked for AgentExecute.Transact: https://github.com/Snowfork/snowbridge/pull/924/files#diff-74470ba8f35b42dbc8805739e92708c0696343e0b1819feb3e7a5ebb77cc623cR471. So while the user can manipulate it with an unrealistic high value, it will only work for `Transact` requests, so I think this will not be a problem. Let me know if you agree.

yrong: AFAIK currently we require each message to be handled properly without error or subsequent message will fail in `nonce` check.

So I think maybe require some extra check in `bridgehub` when forwarding the transact command, make sure not over the upper limit here.
https://github.com/Snowfork/snowbridge/blob/1201e1db5aa6ae580123a745bc7387c2019c4321/web/packages/test/scripts/set-env.sh#L27C1-L27C14

claravanstaden:
Closing, will be re-done in [SNO-866](https://linear.app/snowfork/issue/SNO-866).

PR #926, opened 2023-08-11
SNO-611: Add handlers for RegisterToken & MintToken

doubledup:
Stacked on #923.

[SNO-611](https://linear.app/snowfork/issue/SNO-611)

PR #927, opened 2023-08-16
SNO-612: Extend sendToken for Polkadot-native assets

doubledup:
Stacking this on #926. I'm not sure about initiating burn within `Assets`, given the need to invoke an agent. Would it be more appropriate to burn in `Gateway`? (where we invoke agents & register Polkadot tokens) Should we rather move Polkadot token registration into `Assets` as well?

[SNO-612](https://linear.app/snowfork/issue/SNO-612)

### DIFF
--- a/contracts/src/Assets.sol
+++ b/contracts/src/Assets.sol
@@ -53,8 +53,10 @@ library Assets {
         AssetsStorage.Layout storage $ = AssetsStorage.layout();
 
         if (tokenOwnerAgentID == bytes32(0)) {
+            // Ethereum-native token: transfer to agent
             _transferToAgent(agent, token, sender, amount);
         } else {
+            // Polkadot-native token: burn wrapped token
             _burn(agentExecutor, agent, token, sender, amount);
         }
 
@@ -85,8 +87,10 @@ library Assets {
         }
 
         if (tokenOwnerAgentID == bytes32(0)) {
+            // Ethereum-native token: transfer to agent
             _transferToAgent(agent, token, sender, amount);
         } else {
+            // Polkadot-native token: burn wrapped token
             _burn(agentExecutor, agent, token, sender, amount);
         }
 
--- a/contracts/src/Assets.sol
+++ b/contracts/src/Assets.sol
@@ -6,11 +6,14 @@ import {IERC20} from "./interfaces/IERC20.sol";
 import {IGateway} from "./interfaces/IGateway.sol";
 
 import {SafeTokenTransferFrom} from "./utils/SafeTransfer.sol";
+import {Agent} from "./Agent.sol";
+import {AgentExecutor} from "./AgentExecutor.sol";
 
 import {AssetsStorage} from "./storage/AssetsStorage.sol";
 import {SubstrateTypes} from "./SubstrateTypes.sol";
 import {ParaID, Config} from "./Types.sol";
 import {Address} from "./utils/Address.sol";
+import {Call} from "./utils/Call.sol";
 
 /// @title Library for implementing Ethereum->Polkadot ERC20 transfers.
 library Assets {
@@ -38,7 +41,9 @@ library Assets {
 
     function sendToken(
         ParaID assetHubParaID,
-        address assetHubAgent,
+        address agent,
+        address agentExecutor,
+        bytes32 tokenOwnerAgentID,
         address token,
         address sender,
         ParaID destinationChain,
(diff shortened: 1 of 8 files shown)


contracts/src/Gateway.sol
```diff
@@ -410,6 +415,35 @@ contract Gateway is IGateway, IInitializable {
...
+        if (agent == address(0)) {
+            revert AgentDoesNotExist();
+        }
+
+        bytes memory call = abi.encodeCall(AgentExecutor.registerToken, (params.name, params.symbol, params.decimals));
+        bytes memory response = _invokeOnAgent(agent, call);
+
+        address token = address(bytes20(response));
+        $.ownerAgentIDs[token] = params.agentID;
+    }
```
vgeddes, line 445: The gateway already has an `Command.AgentExecute` command for sending arbitrary messages to Agents. Lets use that rather than adding special-purpose commands for higher-level application stuff. 

I understand we do need to keep a token registry in the gateway. So you could make the agent call back into the gateway to update the registry.

vgeddes: Oops, my mistake, your approach here is correct. My memory was foggy as I just got back from leave.

contracts/src/Assets.sol
```diff
@@ -73,15 +84,19 @@ library Assets {
             revert InvalidDestination();
         }
 
-        _transferToAgent(assetHubAgent, token, sender, amount);
+        if (tokenOwnerAgentID == bytes32(0)) {
+            _transferToAgent(agent, token, sender, amount);
+        } else {
+            _burn(agentExecutor, agent, token, amount);
+        }
```
yrong, line 91: Would prefer add another external function to distinguish the two cases? (e.g. `sendNativeToken` vs `repayForeignToken` more explicitly).

alistair-singh: I think this would make the user experience worse. A user or API would have to know when to call which method per token type. And even if you had two methods with two distinct cases, you would have to guard against each method being used against the wrong case with the same lookup of `tokenOwnerAgentId`. Maybe it would be better to have two internal functions which distinguish the two cases and branch in the top-level `sendToken` method.

contracts/src/storage/CoreStorage.sol
```diff
@@ -14,6 +12,8 @@ library CoreStorage {
         mapping(ParaID paraID => Channel) channels;
         // Agents
         mapping(bytes32 agentID => address) agents;
+        // AgentIDs of token owners
+        mapping(address token => bytes32) ownerAgentIDs;
```
alistair-singh, line 16: +1

contracts/src/Assets.sol
doubledup: @yrong The thinking behind extending sendToken is that, in both cases, it's dealing with ERC20 tokens that should look the same externally. The only distinction between native vs wrapped tokens is in the Gateway's storage. The intention is "send this ERC20 token to Polkadot" - whether the token is Polkadot-native or not shouldn't matter for the caller.

As Alistair says, a caller would have to figure out which function to call, potentially by querying the `ownerAgentIDs` mapping in the Gateway. We'd then query the same mapping in sendToken anyway, to check if the token has been registered and if the correct function has been called.

doubledup: @alistair-singh Separate internal functions would differ only in the line above, ie. whether we send the token to an agent or burn it. I'm open to separating these cases but  don't think we should until there's a larger difference between them.

yrong: Yes, that make sense!

contracts/src/Gateway.sol
vgeddes: Actually, yes, we should be reusing using 'Command.AgentExecute' instead of creating a new toplevel 'Command.RegisterToken'.

doubledup: The call back into the gateway opens a door to attackers: they can change the agentID for any token. This won't let them control the token (as they're not the owner recorded in the token contract), but it can be used to grief legitimate owners by changing the token's agentID so that calls to the token fail.

So we'll need some kind of authorisation when updating the token registry. It might be enough to require that the token hasn't been registered before.

contracts/src/Gateway.sol
```diff
@@ -410,6 +415,35 @@ contract Gateway is IGateway, IInitializable {
         _transferNativeFromAgent(agent, payable(params.recipient), params.amount);
     }
 
+    struct RegisterTokenParams {
```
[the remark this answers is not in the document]
yrong: Agree!  Arbitrary transact also depends on the mapping here(asset_id <-> location) to support dynamic fee, currently we only support `DOT` to be paid as fee of `BuyExecution`.

contracts/src/Gateway.sol
```diff
@@ -412,13 +412,14 @@ contract Gateway is IGateway, IInitializable {
     }
 
     // @dev Register a new fungible Polkadot token for an agent
-    function setTokenOwnerAgentID(address token, bytes32 agentID) external {
+    function registerToken(bytes32 tokenID, address token, bytes32 agentID) external {
```
vgeddes, line 415: I've thought of a way we can prevent malicious parachains from using `Transact` to call back into gateway via an agent.

In AgentExecutor.sol, we make sure the destination of the transact is not the gateway contract. If it the destination is the gateway, we disallow the transact.

What we still need for this `registerToken` is a guard to ensure it's only being called by a registered agent. We may need to introduce another storage field which allows to perform this lookup. Unfortunately the storage field below only allows us to lookup by agentID:

https://github.com/Snowfork/snowbridge/blob/8b8f1b5af3fc0a622f207d8069a04c0821ad67ce/contracts/src/storage/CoreStorage.sol#L16

doubledup: We receive the relevant `agentID` as the 3rd parameter here, so we can check that `msg.sender == agents[agentID]`. This means that only a registered agent can call `registerToken` and that they must register tokens for themselves. It's slightly stricter than just "registered agents only", but seems reasonable as the intention is for agents to register their own tokens anyway.

yrong: > prevent malicious parachains from using Transact to call back into gateway via an agent. In AgentExecutor.sol, we make sure the destination of the transact is not the gateway contract. If it the destination is the gateway, we disallow the transact.

@vgeddes Don't think we can prevent arbitrary transact from reentering Gateway without a [ReentrencyGuard](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/security/ReentrancyGuard.sol),  demonstrated in https://github.com/Snowfork/snowbridge/pull/933

claravanstaden:
Closing, will reopen if it becomes relevant.

PR #936, opened 2023-08-28
Fix for issue 10

claravanstaden:
The generalized_index_bit should ideally match the spec version. Because our version returned a `0` and the spec version returns a `1`, I had to update the merkle proof left and right node as well.

Resolves: https://linear.app/snowfork/issue/SNO-637


claravanstaden:
Closing to create a new branch off tag `oak-security-audit`.

PR #939, opened 2023-08-29
Ensure outbound queue stop processing messages when PalletOperatingMode is halted

yrong:
Fix for audit issue (9).
Resolves: SNO-636


parachain/pallets/outbound-queue/src/lib.rs
```diff
@@ -338,7 +340,6 @@ pub mod pallet {
 		}
 
 		fn submit(ticket: Self::Ticket) -> Result<MessageHash, SubmitError> {
-			Self::ensure_not_halted().map_err(|_| SubmitError::BridgeHalted)?;
```
claravanstaden, line 341: Removing this check would mean that messages can still be enqueued, even though the bridge is halted. I understand check in `process_message` would prevent the message from being processed, but I am wondering if we shouldn't even enqueue the message in the first place? The current implementation would allow the message to be enqueued, and then not processed in `process_message`. Should we not perhaps leave this check here?

parachain/pallets/outbound-queue/src/lib.rs
```diff
@@ -356,6 +357,8 @@ pub mod pallet {
 			meter: &mut frame_support::weights::WeightMeter,
 			_: &mut [u8; 32],
 		) -> Result<bool, ProcessMessageError> {
+			// Make sure the bridge not halted
+			Self::ensure_not_halted().map_err(|_| ProcessMessageError::Yield)?;
```
claravanstaden, line 361: I looked the `Yield` error description:

> The queue wants to give up its current processing slot. Hints the message processor to cease servicing this queue and proceed to the next one. This is seen as a hint, not an instruction. Implementations must therefore handle the case that a queue is re-serviced within the same block after yielding. A queue is not required to yield again when it is being re-serviced withing the same block.

Does the implementation of this message queue discard this message?

claravanstaden (commented):
Hey @yrong! I have a couple of questions (mainly because I am unsure and just want to understand better) 😄

parachain/pallets/outbound-queue/src/lib.rs
yrong: No, it's not discarded and will be reattempted later. There are some other errors like `BadFormat` or `Corrupt`
https://github.com/paritytech/polkadot-sdk/blob/b728724318d65bf8d61831aac35c286727b41b60/substrate/frame/support/src/traits/messages.rs#L32-L36
will be discarded.

alistair-singh (approved):

PR #940, opened 2023-08-29
SNO-624: Charge fee for outbound operations

yrong:
Fix for audit issue (1).
Resolves: SNO-624
Requires:  https://github.com/Snowfork/cumulus/pull/63

# Context

It makes sense to let third-party Parachain who would like to use the bridge know what the fee is in advance, so added the [Estimate API](https://github.com/Snowfork/snowbridge/blob/1d160b08d7f8edfb78e9c72df0068cc4759adfd6/parachain/pallets/outbound-queue/runtime-api/src/lib.rs#L14) into the runtime.  The extra fee is related to 
- dispatch gas for each command
- gas price
- swap price between Ether and DOT

So we've made all these factors configurable through extrinsic. Basically for most of the commands gas cost should be static, the only exception is maybe the arbitrary transact we'll add later, the dispatch gas should be some dynamic input from third party. Nonetheless we need to ensure all the gas cost in a reasonable range. 

Based on that we've added a [Configuring Extrinsic](https://github.com/Snowfork/snowbridge/blob/1d160b08d7f8edfb78e9c72df0068cc4759adfd6/parachain/pallets/outbound-queue/src/lib.rs#L214) for all the factors above. And also there are two smoke tests added accordingly.
- estimate_fee
- configure_base_fee


parachain/pallets/control/src/lib.rs
```diff
@@ -170,6 +208,11 @@ pub mod pallet {
 				location
 			);
 
+			Self::reserve_deposit(
+				para_id,
```
vgeddes, line 212: Specifically for `CreateAgent`, we should be transferring the fee from the sovereign account of `origin_location` rather than the parachain that `origin_location` lives on.

Often the governance of parachains won't have control over which sub-accounts make XCM calls into BridgeHub, and so for example, malicious EVM contracts on Moonbeam could can spam `create_agent` calls and they won't have to pay for it, as the fees will be deducted from the parachain's sovereign account.

So you'll need some kind of converter that converts `origin_location` into a signed sovereign account.

vgeddes (commented):
Looks good. Added one comment.

parachain/pallets/control/src/lib.rs
```diff
@@ -170,6 +209,12 @@ pub mod pallet {
 				location
 			);
 
+			let agent_owner = Self::agent_account_id(agent_id)?;
```
vgeddes, line 212: Just to clarify, we need to convert `origin_location` into a signed sovereign account. So `MultiLocation` to  `AccountId`.

Why? because this allows the origin top up their sovereign account on BridgeHub using normal XCM transfers/teleports. This allows them to pay for the deposit when calling `create_agent`.

For example,  see how pallet-xcm uses this converter:
https://github.com/paritytech/polkadot-sdk/blob/430edd75351e82ea3f814c05fbbbe0cb860e6d27/polkadot/xcm/pallet-xcm/src/lib.rs#L251

https://github.com/paritytech/polkadot-sdk/blob/430edd75351e82ea3f814c05fbbbe0cb860e6d27/polkadot/xcm/pallet-xcm/src/lib.rs#L1758

yrong: > need to convert origin_location into a signed sovereign account

 I would assume the signed sovereign you mentioned above refer in particular to [AccountId32Aliases](https://github.com/paritytech/polkadot-sdk/blob/430edd75351e82ea3f814c05fbbbe0cb860e6d27/cumulus/parachains/runtimes/bridge-hubs/bridge-hub-rococo/src/xcm_config.rs#L85C1-L85C1)  which applies to `AccountId32` only.

AFAIK in our design there is no limitation for the xcm origin, actually there are several other use cases in unit tests which allows from [PalletInstance](https://github.com/Snowfork/snowbridge/blob/1201e1db5aa6ae580123a745bc7387c2019c4321/parachain/pallets/control/src/mock.rs#L100-L106) or [AccountKey20](https://github.com/Snowfork/snowbridge/blob/1201e1db5aa6ae580123a745bc7387c2019c4321/parachain/pallets/control/src/mock.rs#L119-L128),  would assume these are all valid use cases that we need to support.

>  allows the origin top up their sovereign account on BridgeHub using normal XCM transfers/teleports.

 I would assume actually the derivated sovereign account is keyless and does not require that particular signed origin to top up it, actually it's the dest account anyone can fund it with a xcm transfer(e.g. check how we fund the gateway sovereign from relaychain in https://github.com/Snowfork/snowbridge/blob/1201e1db5aa6ae580123a745bc7387c2019c4321/web/packages/test/scripts/configure-bridgehub.sh#L42)





Btw:  I would assume does not require that particular signed origin to top up the derivated sovereign account, could be from arbitrary account as we initial fund the  

https://github.com/Snowfork/snowbridge/blob/1201e1db5aa6ae580123a745bc7387c2019c4321/web/packages/test/scripts/configure-bridgehub.sh#L39

yrong: Since `agent_id` is also a unique identifier derivated from `MultiLocation` with [AgentHashedDescription](https://github.com/Snowfork/snowbridge/blob/1201e1db5aa6ae580123a745bc7387c2019c4321/parachain/pallets/control/src/lib.rs#L66), so I would like to reuse it as the derivated sovereign account.

yrong: https://linear.app/snowfork/issue/SNO-625#comment-c20c5032

vgeddes: > > need to convert origin_location into a signed sovereign account
> 

Oops, I meant convert into a just a sovereign account. Is keyless as you say.

> Since `agent_id` is already a unique identifier derivated from `MultiLocation` with [AgentHashedDescription](https://github.com/Snowfork/snowbridge/blob/1201e1db5aa6ae580123a745bc7387c2019c4321/parachain/pallets/control/src/lib.rs#L66), so I would prefer to reuse it rather than introducing another origin converter.

Agents, agent addresses, and agent ids are mainly for use on the Ethereum side of the bridge, where using `MultiLocation` is not possible. On the Polkadot side of the bridge we shouldn't be complicating the matter of identity by using agent ids to refer to Polkadot origins. It also introduces an unnecessary 2-step conversion to transform origins to sovereign accounts: `MultiLocation -> AgentID -> SovereignAccount`.

vgeddes: > I would assume actually the derivated sovereign account is keyless and does not require some particular signed origin to fund it, actually it's a dest account anyone can fund through xcm transfer(e.g. check how we [fund](https://github.com/Snowfork/snowbridge/blob/1201e1db5aa6ae580123a745bc7387c2019c4321/web/packages/test/scripts/configure-bridgehub.sh#L42) the gateway sovereign from relaychain).

Sure, anyone can fund the sovereign account derived from the agent, but no one is able to withdraw funds from the account, unless we add a special extrinsic to allow it. 

If we just create a sovereign account based on the `MultiLocation`, then the origin is able to withdraw using XCM without us adding any special logic

parachain/pallets/control/src/lib.rs
```diff
@@ -299,6 +356,27 @@ pub mod pallet {
 
 			Ok(())
 		}
+
+		/// - `update`: (ControlOperation, Fee).
+		#[pallet::call_index(6)]
+		#[pallet::weight(T::WeightInfo::update_operation_fee())]
+		pub fn update_operation_fee(
```
vgeddes, line 363: I think these fees will be updated very rarely. By making them static config, we would not need to maintain and audit this `update_operation_fee` extrinsic.

parachain/pallets/control/src/lib.rs
```diff
@@ -231,14 +279,17 @@ pub mod pallet {
...
+			let (agent_id, para_id, _) = Self::convert_location(location)?;
 
 			ensure!(Agents::<T>::contains_key(agent_id), Error::<T>::AgentNotExist);
 
-			let para_id = some_para_id.ok_or(Error::<T>::LocationToParaIdConversionFailed)?;
-
 			ensure!(Channels::<T>::contains_key(para_id), Error::<T>::ChannelNotExist);
 
+			Self::reserve_deposit(
+				para_id.into_account_truncating(),
```
vgeddes, line 289: I think specifically for the deposits, the origin converter used to the obtain the sovereign account should be the same across `create_agent`, `create_create_channel`, `update_channel`, and `transfer_native_from_agent`.

As I described in another comment, the origin converter in the following example is able to handle all multilocations passed to it:
https://github.com/paritytech/polkadot-sdk/blob/430edd75351e82ea3f814c05fbbbe0cb860e6d27/polkadot/xcm/pallet-xcm/src/lib.rs#L251

parachain/pallets/control/src/lib.rs
vgeddes: For example, the Pallet config, can have a `Get<BTreeMap<ControlOperation, BalanceOf<T>>` which contains the fees.

https://paritytech.github.io/substrate/master/sp_std/collections/btree_map/struct.BTreeMap.html

parachain/pallets/control/src/lib.rs
```diff
@@ -311,25 +389,49 @@ pub mod pallet {
...
+
+		pub fn account_id() -> T::AccountId {
+			PalletId(*b"snow/ctl").into_account_truncating()
+		}
+
+		pub fn reserve_deposit(
+			payer: T::AccountId,
+			reserve_deposit: Option<BalanceOf<T>>,
+		) -> DispatchResult {
+			if reserve_deposit.is_some() {
```
vgeddes, line 420: I think we can simplify this code by just passing a fee of `0` instead of Option. Then `T::Token::transfer` may just handle the zero fee case gracefully.

parachain/pallets/control/src/lib.rs
yrong: > Agents, agent addresses, and agent ids are mainly for use on the Ethereum side of the bridge, where using MultiLocation is not possible. On the Polkadot side of the bridge we shouldn't be complicating the matter of identity by using agent ids to refer to Polkadot origins

Yes, `agent_address` is used on the Ethereum side to represent the agent, but I would assume `agent_id` is not, it's derivated through [AgentHashedDescription](https://github.com/Snowfork/snowbridge/blob/1201e1db5aa6ae580123a745bc7387c2019c4321/parachain/pallets/control/src/lib.rs#L66) purely from the `Multilocation`, check the internal implementation [here](https://github.com/paritytech/polkadot-sdk/blob/43c0c09bcb9a4db0856598dd07357c0b7292bb84/polkadot/xcm/xcm-builder/src/location_conversion.rs#L120) nothing to do with the Ethereum side things.

And in current implementation the sovereign account is identical as `agent_id`(both 32 bytes).

yrong: >  If we just create a sovereign account based on the MultiLocation, then the origin is able to withdraw using XCM without us adding any special logic

AFAIK `bridgeHub` is configured not allow to send arbitray xcm messages through `xcmPallet`, check the config [here](https://github.com/Snowfork/cumulus/blob/38c2e26a4ead087d0837b5abdc3cdc3e0e6e0d8e/parachains/runtimes/bridge-hubs/bridge-hub-rococo/src/xcm_config.rs#L311-L312)
which disable the send extrinsic and config [here](https://github.com/Snowfork/cumulus/blob/38c2e26a4ead087d0837b5abdc3cdc3e0e6e0d8e/parachains/runtimes/bridge-hubs/bridge-hub-rococo/src/xcm_config.rs#L315) which disable the execute extrinsic.

So I think we do need some special logic(a custom call) to withdraw from sovereign anyway.

parachain/pallets/control/src/lib.rs
yrong: > I think specifically for the deposits, the origin converter used to the obtain the sovereign account should be the same across create_agent, create_create_channel, update_channel, and transfer_native_from_agent.

Agree, I'll use the same derivated account for channel operations.

>  the origin converter in the following example is able to handle all multilocations passed to it

https://github.com/Snowfork/snowbridge/pull/940#discussion_r1310522554 and by the way as I mentioned before the current [SovereignAccountOf] (https://github.com/Snowfork/cumulus/blob/38c2e26a4ead087d0837b5abdc3cdc3e0e6e0d8e/parachains/runtimes/bridge-hubs/bridge-hub-rococo/src/xcm_config.rs#L332C7-L332C25) in `bridgeHub` only allow three kinds of the origin as defined [here](https://github.com/Snowfork/cumulus/blob/38c2e26a4ead087d0837b5abdc3cdc3e0e6e0d8e/parachains/runtimes/bridge-hubs/bridge-hub-rococo/src/xcm_config.rs#L94) which does not cover the use case we need to support.

parachain/pallets/control/src/lib.rs
yrong: > I think these fees will be updated very rarely. By making them static config, we would not need to maintain and audit this update_operation_fee extrinsic.

I would assume the fee charged here is supposed to cover the cost in ethereum side, considering the volatile gas price and fee model so I would prefer to make it configurable and not a static config.

vgeddes: ok good point! Yes, I realise the fee charged in the pallet needs to cover the reward given to the relayer on Ethereum side (from funds in BridgeHub's agent contract)

parachain/pallets/control/src/lib.rs
vgeddes: > So I think we do need some special logic(a custom call) allow to withdraw from sovereign anyway.

Sure, but only bridge hub governance to withdraw its funds. Other origins on Polkadot can withdraw funds from their sovereign accounts without using any special logic, as long as we derive the sovereign account from the MultiLocation.

I really would like to follow the convention that other parachains are using to create sovereign accounts, which is to use some [converter](https://github.com/Snowfork/cumulus/blob/38c2e26a4ead087d0837b5abdc3cdc3e0e6e0d8e/parachains/runtimes/bridge-hubs/bridge-hub-rococo/src/xcm_config.rs#L94) already available in the polkadot xcm libs. It's going to reduce surprises in future.

vgeddes: Also in the latest Polkadot code, you can use this instead of `LocationToAccountId`: https://github.com/paritytech/polkadot-sdk/blob/73915d3459a0723e2ca80a5ca5d0a11a3736b18d/polkadot/xcm/xcm-builder/src/location_conversion.rs#L121

vgeddes: OK, I took a deeper look at the implementation of [AgentHashedDescription](https://github.com/Snowfork/snowbridge/blob/1201e1db5aa6ae580123a745bc7387c2019c4321/parachain/pallets/control/src/lib.rs#L66) and I see that it basically is what we need to use to convert MultiLocations.

As you say, we cannot use `LocationToAccountId`, since it doesn't support converting AccountId32 junctions.

So I was technically wrong on a few points.

parachain/pallets/control/src/lib.rs
```diff
@@ -311,25 +389,49 @@ pub mod pallet {
...
+				)?;
+			}
+			Ok(())
+		}
+
+		pub fn agent_account_id(agent_id: H256) -> Result<T::AccountId, DispatchError> {
+			let agent_owner = T::AccountId::decode(&mut &agent_id.as_fixed_bytes()[..])
+				.map_err(|_| Error::<T>::AgentIdToAccountConversionFailed)?;
+			Ok(agent_owner)
+		}
```
vgeddes, line 435: There is already a converter in Polkadot that we can reuse here:

https://github.com/paritytech/polkadot-sdk/blob/73915d3459a0723e2ca80a5ca5d0a11a3736b18d/polkadot/xcm/xcm-builder/src/location_conversion.rs#L120

So you can add a type to the config like: `SovereignAccountOf: ConvertLocation<AccountId>`

parachain/pallets/control/src/lib.rs
vgeddes: See https://github.com/Snowfork/snowbridge/pull/940/files#r1310744745, we can use this converter in Polkadot. It already uses some of the same code as [AgentHashedDescription](https://github.com/Snowfork/snowbridge/blob/1201e1db5aa6ae580123a745bc7387c2019c4321/parachain/pallets/control/src/lib.rs#L66), but returns `AccountId` instead of `H256`.

parachain/pallets/control/src/lib.rs
yrong: Exactly, essentially [AgentHashedDescription](https://github.com/Snowfork/snowbridge/blob/ffa44a7cce7388d5bb32211922c12f2626d22870/parachain/primitives/router/src/outbound/mod.rs#L18) is just an implementation of trait `ConvertLocation` though we provide `H256` here.

And I've added another [AgentAccountDescription](https://github.com/Snowfork/snowbridge/blob/ffa44a7cce7388d5bb32211922c12f2626d22870/parachain/primitives/router/src/outbound/mod.rs#L20) convert to `AccountId32` to make it less confusing.

But actually just the same as before and identical as the `agent_id`,  check the test [here](https://github.com/Snowfork/snowbridge/blob/ffa44a7cce7388d5bb32211922c12f2626d22870/parachain/pallets/control/src/tests.rs#L42) FYI.

vgeddes: I feel differently. I would rather have two separate converters defined in the config:

```rust
    type AgentIdOf: ConvertLocation<H256>;
    type SovereignAccountOf: ConvertLocation<AccountId>
```

This makes it clear that the converters are being used for different purposes.

yrong: Make sense! some revamp in https://github.com/Snowfork/snowbridge/pull/940/commits/6cc82e5ed1bd849694e863e24f5674ba6c87cbbd please check.

vgeddes: Looks much better!

parachain/pallets/control/src/lib.rs
vgeddes, line 363: I think it is simpler to have a reconfigurable _base fee_, and then a static per-operation fee multiplier. This makes the updating the fees simpler, as there is only one variable to adjust.

For example:

```
BaseFee=10 DOT
Fee(CreateAgent) = BaseFee * StaticMultiplier[CreateAgent]
Fee(UpdateChannel) = BaseFee * StaticMultiplier[UpdateChannel]
Fee(TransferToken) = BaseFee * StaticMultiplier[TransferToken]
```

@claravanstaden is also working on fees charged by the outbound router for the `TransferToken` operation, and so I think both your work needs to align on the implementation somehow.

Assuming we go for the BaseFee approach, we also need to store the base fee somewhere. I think the `outbound-router` pallet is the best place. The pallet could implement a trait like the following, so that other pallets can access the base fee.

```rust
trait FeeProvider {
  fn base_fee() -> Balance;
}
```

yrong: That makes sense! I'll handle that and make it reusable for other outbound operations like transfer.

vgeddes: I think it's also important to define `BaseFee` as something we can reason about, for example:

_Cost in DOT for 500,000 GAS_

vgeddes: So from the previous comment, even BaseFee is composed of several factors:

* DOT/ETH price
* Gas Price (Ether)
* Gas Amount

It may make sense to thus define BaseFee in terms of those three sub-parameters.

yrong: Yes, the DOT/ETH price and Gas Price (Ether) are some common parameters that apply to all operations,  and the gas amount for each operation can be extracted from the gas report.

Btw, the `outbound-router` currently in our repo is some primitives and not a pallet. So I would like to add another pallet for that.  On-chain storage in that pallet including:

- a dynamic map from operation type to gas amount(e.g. CreateAgent->500 gas, TransferToken->100 gas,...)
- a configurable price ratio(reflect the two price above).

I would assume the trait should be modified to something like

```
trait FeeProvider {
  fn base_fee(operation Operation) -> Balance;
}
```
and the default implemenation in the pallet for `base_fee` is simply to be `price_ratio` multiply `gas_amount` for that operation.

@vgeddes Please let me know what you think.

vgeddes: > Btw, the outbound-router currently in our repo is some primitives and not a pallet. So I would like to add another pallet for that.

Oops, I meant the `outbound-queue` pallet. This is the right place for it.

vgeddes: > a dynamic map from operation type to gas amount(e.g. CreateAgent->500 gas, TransferToken->100 gas,...)

I don't want to introduce this level complexity just yet, as we need to have this PR ready by Wednesday at the latest.

We just need an approximate solution that satisfies the audit review.

So the `control` pallet computes the fee for `CreateAgent` by executing:

```
let fee = T::FeeProvider::base_fee() * CREATE_AGENT_MULTIPLIER
```

vgeddes: > So from the previous comment, even BaseFee is composed of several factors:
> 
> * DOT/ETH price
> * Gas Price (Ether)
> * Gas Amount
> 
> It may make sense to thus define BaseFee in terms of those three sub-parameters.

So to be more clear, I suggest calculating a global `BaseFee` using a _static_ gas amount of "100,000".

Then if `CreateAgent` uses 200,000 gas, then `CREATE_AGENT_MULTIPLIER` is `2`.

claravanstaden: All of this sounds good 😄

yrong: https://github.com/Snowfork/snowbridge/pull/940/commits/cb6f6aaa5d615a53e4f53d0ef53566c71b9ff75e

parachain/primitives/core/src/outbound.rs
```diff
@@ -46,9 +48,12 @@ pub struct Message {
 	pub origin: ParaId,
 	/// The stable ID for a receiving gateway contract
 	pub command: Command,
+	/// The multilocation the message comes from
+	pub location: MultiLocation,
```
yrong, line 52: add `location` here to derivate sovereign account for the agent.

parachain/primitives/core/src/outbound.rs
```diff
@@ -129,18 +134,48 @@ impl Command {
...
+			Command::Upgrade { .. } => 0,
+		}
+	}
+
+	// Todo: configurable gas price, for now set as 20 Gwei
+	// TBC: 4/5 to relayer and 1/5 left for protocol maintenance
+	pub fn reward(&self) -> u128 {
+		let gas_price: u128 = 20_000_000_000;
+		self.dispatch_gas() * gas_price * 4 / 5
+	}
```
yrong, line 163: Make `dispatch_gas` and  `reward` inherent attribute of the command. The `reward` is linear correlation with the gas cost.

contracts/src/Gateway.sol
```diff
@@ -136,45 +137,46 @@ contract Gateway is IGateway, IInitializable {
         // Otherwise malicious relayers can break the bridge by allowing the message handlers below to run out gas and fail silently.
         // In this scenario case, the channel's state would have been updated to accept the message (by virtue of the nonce increment), yet the actual message
         // dispatch would have failed
-        if (gasleft() < DISPATCH_GAS + BUFFER_GAS) {
+        uint256 dispatchGas = (message.dispatchGas > 0 ? message.dispatchGas : DISPATCH_GAS);
+        if (gasleft() < dispatchGas + BUFFER_GAS) {
```
yrong, line 141: Is the `BUFFER_GAS` still nessessary here? as we've already left some margin for the substrate config.

parachain/pallets/control/src/lib.rs
```diff
@@ -198,19 +202,18 @@ pub mod pallet {
...
 
-			let para_id = some_para_id.ok_or(Error::<T>::LocationToParaIdConversionFailed)?;
-
 			ensure!(!Channels::<T>::contains_key(para_id), Error::<T>::ChannelAlreadyCreated);
 
 			Channels::<T>::insert(para_id, ());
 
 			let message = Message {
-				origin: T::OwnParaId::get(),
+				origin: para_id,
```
yrong, line 214: This is a fix for other control operations except `create_agent`, the origin should be the original Parachain, not bridgeHub.

parachain/pallets/outbound-queue/src/lib.rs
```diff
@@ -344,6 +418,7 @@ pub mod pallet {
 				ticket.message.as_bounded_slice(),
 				AggregateMessageOrigin::Parachain(ticket.origin),
 			);
+			Self::charge_fee(&ticket).map_err(|_| SubmitError::ChargeFeeFailed)?;
```
vgeddes, line 421: As discussed on Slack, we should charge a static fee so that malicious parachains don't DOS the outbound-queue.

However we should use the mechanisms provided by the XCM executor to do this, which involves:

1. In our [implementation](https://github.com/Snowfork/snowbridge/blob/1ffd01ec450ba06c5d7562e431b03b50c3c55c2c/parachain/primitives/router/src/outbound/mod.rs#L140) of `ExportXcm::validate`, we must specify the fee for the export.
2. Wait until this PR is ready so that the XCM executor automatically handles the fee for us: https://github.com/paritytech/polkadot-sdk/pull/1234

We can still use the code in `charge_fee` for charging a deposit for using these permissionless APIS:
1. create_agent
2. create_channel

yrong: https://github.com/Snowfork/snowbridge/pull/940/commits/c834c49190bad783ef1f98be04e703e137b9e6f9

@vgeddes Is your comment `charging a deposit` refers to some guarantee deposit something like the 10 DOT in `asset_hub` to create asset?

contracts/src/Gateway.sol
```diff
@@ -136,45 +137,46 @@ contract Gateway is IGateway, IInitializable {
         // Otherwise malicious relayers can break the bridge by allowing the message handlers below to run out gas and fail silently.
         // In this scenario case, the channel's state would have been updated to accept the message (by virtue of the nonce increment), yet the actual message
         // dispatch would have failed
-        if (gasleft() < DISPATCH_GAS + BUFFER_GAS) {
+        uint256 dispatchGas = (message.dispatchGas > 0 ? message.dispatchGas : DISPATCH_GAS);
```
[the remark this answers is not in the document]
yrong: https://github.com/Snowfork/snowbridge/pull/940/commits/b3522cd98f054643bf45a35f41781325f0b8cf44

contracts/src/Gateway.sol
```diff
@@ -119,8 +119,9 @@ contract Gateway is IGateway, IInitializable {
         // Reward the relayer from the agent contract
         // Expected to revert if the agent for the message origin does not have enough funds to reward the relayer.
         // In that case, the origin should top up the funds of their agent.
-        if (channel.reward > 0) {
-            _transferNativeFromAgent(channel.agent, payable(msg.sender), channel.reward);
+        uint256 reward = (message.reward > 0 ? message.reward : channel.reward);
```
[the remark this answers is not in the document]
yrong: https://github.com/Snowfork/snowbridge/pull/940/commits/b3522cd98f054643bf45a35f41781325f0b8cf44

parachain/pallets/control/src/lib.rs
```diff
@@ -282,13 +285,14 @@ pub mod pallet {
...
 			let location: MultiLocation = T::AgentOrigin::ensure_origin(origin)?;
 
-			let (agent_id, _, _) = Self::convert_location(location)?;
+			let (agent_id, para_id, location) = Self::convert_location(location)?;
 
 			ensure!(Agents::<T>::contains_key(agent_id), Error::<T>::AgentNotExist);
 
 			let message = Message {
-				origin: T::OwnParaId::get(),
+				origin: para_id,
```
yrong, line 293: For `UpdateChannel` and `TransferNativeFromAgent` chage the origin to the original Parachain, not bridgeHub.

parachain/pallets/outbound-queue/src/lib.rs
```diff
@@ -300,6 +402,52 @@ pub mod pallet {
...
+				.ok_or(Error::<T>::LocationToSovereignAccountConversionFailed)?;
+			// charge base fee here before https://github.com/paritytech/polkadot-sdk/pull/1234
+			// There is nuance difference between `xcm-fees-manager` because we charge from the
+			// agent account while they charge from the parachain account
+			let base_fee = Self::estimate_base_fee(ticket).unwrap_or_default();
+			let extra_fee = Self::estimate_extra_fee(ticket).unwrap_or_default();
+			let total_fee = base_fee.saturating_add(extra_fee).saturated_into::<BalanceOf<T>>();
+			T::Token::transfer(
+				&agent_account,
+				&T::LocalPalletId::get().into_account_truncating(),
```
alistair-singh, line 445: What happens to the fees collected in the Pallet account?

parachain/pallets/outbound-queue/src/lib.rs
```diff
@@ -308,14 +456,16 @@ pub mod pallet {
 		id: H256,
 		origin: ParaId,
 		message: BoundedVec<u8, MaxMessageSize>,
+		command: Command,
+		location: MultiLocation,
```
alistair-singh, line 460: See previous comments. A better name would be `agent_location`.

parachain/pallets/outbound-queue/src/lib.rs
```diff
@@ -324,16 +474,23 @@ pub mod pallet {
...
+			let encoded =
+				enqueued_message.encode().try_into().map_err(|_| SubmitError::MessageTooLarge)?;
 
-			let ticket =
-				OutboundQueueTicket { id: message.id, origin: message.origin, message: encoded };
+			let ticket = OutboundQueueTicket {
+				id: message_id,
+				origin: message.origin,
+				message: encoded,
+				location: message.location,
```
alistair-singh, line 491: See previous comments. A better name would be `agent_location`.

parachain/primitives/core/src/outbound.rs
```diff
@@ -129,18 +141,68 @@ impl Command {
...
+			// leave enough space for upgrade with arbitrary initialize logic
+			Command::Upgrade { .. } => 300000,
+		}
+	}
+
+	pub fn base_fee_required(&self) -> bool {
+		match self {
+			Command::CreateAgent { .. } => true,
+			Command::CreateChannel { .. } => true,
+			Command::AgentExecute { .. } => true,
```
alistair-singh, line 173: I agree that `AgentExecute` should require a base fee. But the notion document says otherwise. See [comment](https://www.notion.so/Fees-c15d9159574c4ff1b1f82bdf220d7f94?d=8371b178e0c84f0c96d3ed96658375fe&pvs=4#2e741cda15f747238678134a345cbdf2).

parachain/primitives/core/src/outbound.rs
```diff
@@ -129,18 +141,68 @@ impl Command {
...
+			Command::TransferNativeFromAgent { .. } => true,
+			Command::Upgrade { .. } => false,
+			Command::SetOperatingMode { .. } => false,
+		}
+	}
+
+	pub fn extra_fee_required(&self) -> bool {
+		match self {
+			Command::CreateAgent { .. } => true,
+			Command::CreateChannel { .. } => true,
```
alistair-singh, line 184: Why is an extra fee required for CreateAgent and CreateChannel. I was under the impression that these commands have a fixed cost?

parachain/primitives/router/src/outbound/mod.rs
```diff
@@ -126,18 +128,22 @@ where
...
 
-		log::info!(target: "xcm::ethereum_blob_exporter", "message validated: location = {local_sub_location:?}, agent_id = '{agent_id:?}'");
+		let fees = OutboundQueue::estimate_fee(ticket.clone()).map_err(|err| {
+			log::error!(target: "xcm::ethereum_blob_exporter", "OutboundQueue estimate fee failed. {err:?}");
+			SendError::Fees
+		})?;
+
+		log::info!(target: "xcm::ethereum_blob_exporter", "message validated: location = {local_sub_location:?}, agent_id = '{agent_id:?}', fees = {fees:?}");
 
-		// TODO (SNO-581): Make sure we charge fees for message delivery. Currently this is set to
```
alistair-singh, line 138: +1

web/packages/test/scripts/configure-bridgehub.sh
```diff
@@ -15,26 +14,24 @@ config_beacon_checkpoint()
...
         sleep 3
     done
 }
 
-fund_accounts()
-{
+fund_accounts() {
     echo "Funding substrate accounts"
     transfer_balance $relaychain_ws_url "//Charlie" 1013 1000000000000000 $assethub_sovereign_account
+    transfer_balance $relaychain_ws_url "//Charlie" 1013 1000000000000000 $TEMPLATE_AGENT_ID
```
alistair-singh, line 28: Is this correct?
```suggestion
    transfer_balance $relaychain_ws_url "//Charlie" 1013 1000000000000000 $template_sovereign_account
```

parachain/pallets/outbound-queue/src/lib.rs
```diff
@@ -344,9 +501,22 @@ pub mod pallet {
 				ticket.message.as_bounded_slice(),
 				AggregateMessageOrigin::Parachain(ticket.origin),
 			);
+			Self::charge_fees(&ticket).map_err(|_| SubmitError::ChargeFeeFailed)?;
```
vgeddes, line 504: Just to clarify some design goals which I think are important. I don't think `OutboundQueue` should be charging fees in its message processing pipeline, because of the various exceptions and edge cases.

It is better that the actual charging of fees is done in:
1. The control pallet, for permissionless commands such as `CreateAgent`
2. For `AgentExecute` by the the XCM executor, using https://github.com/paritytech/polkadot-sdk/pull/1234 when it is ready. This means we can't make full progress until that is implemented.
  i. Specify the fee (specifically a static base fee) to charge in our [ExportXcm::validate](https://github.com/Snowfork/snowbridge/blob/4f739fde129de640c667dc10eb0a08b701183bdc/parachain/primitives/router/src/outbound/mod.rs#L140) implementation.
 ii. Then the XCM executor will automatically transfer it to the treasury account.

For charging fees in the control pallet, I suggest this approach:

Define a new trait `ChargeFees`, that is implemented by the `OutboundQueue` pallet:

```
trait ChargeFee {
   fn charge_fee(command: Command, from: AccountId) -> Result<V, E>
}
```

Then the control pallet can use this interface to charge fees for various commands depending on whether they are permissionless or not. We should collect all the various fees in a single treasury account controlled by BridgeHub, for example in:

https://github.com/paritytech/polkadot-sdk/pull/1234/files#diff-984ede31f860995ed9e3720b323fce903524085f550b1a6c5a2a128664aa5846R64

parachain/pallets/outbound-queue/src/lib.rs
```diff
@@ -201,6 +238,33 @@ pub mod pallet {
...
+
+	/// The gas price to calculate the reward in Ether from https://etherscan.io/gastracker
+	/// default 15 gwei
+	#[pallet::type_value]
+	pub fn DefaultGasPrice() -> u128 {
+		15_000_000_000
+	}
+	#[pallet::storage]
+	pub type GasPrice<T: Config> = StorageValue<_, u128, ValueQuery, DefaultGasPrice>;
+
```
vgeddes, line 267: I think we're adding so many fee-related storage items here that we should rather them group them into a single struct. Then we only need a single getter/setter pair.

parachain/pallets/outbound-queue/src/lib.rs
vgeddes: We should actually be collecting the fees in BridgeHub's local treasury account, not the account for the pallet.

https://github.com/paritytech/polkadot-sdk/pull/1234/files#diff-984ede31f860995ed9e3720b323fce903524085f550b1a6c5a2a128664aa5846R64

parachain/pallets/outbound-queue/src/lib.rs
vgeddes, line 460: It shouldn't be necessary to add this location field to the Message type, with the design I suggest in a comment below.

parachain/primitives/core/src/outbound.rs
vgeddes: The meaning of "extra fee" in this context is that the sender needs to pay the full cost of execution on the Ethereum side.

I agree the terminology is confusing and we should settle on better naming :)

parachain/primitives/core/src/outbound.rs
vgeddes: The meaning of "base fee" in this context is that the sender needs to pay a small static fee for the weight involved in processing outbound messages locally.

The "extra fee" is the cost of dispatching on Ethereum, but as you can see from the Notion doc we don't always want to charge this fee on BridgeHub.

parachain/pallets/control/src/lib.rs
```diff
@@ -312,22 +316,21 @@ pub mod pallet {
...
-			// Only allow Parachain as origin location
-			let para_id = match location {
-				MultiLocation { parents: 0, interior: X1(Parachain(index)) } =>
-					Some((index).into()),
+			let para_id = match location.interior.first() {
+				Some(Parachain(index)) => Some((*index).into()),
 				_ => None,
-			};
+			}
+			.ok_or(Error::<T>::LocationToParaIdConversionFailed)?;
```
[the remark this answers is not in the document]
yrong: It will get ParaId from [first Junction](https://github.com/Snowfork/snowbridge/blob/86c3ffc07f6d64d0078b2bd1a0752fcfbe012d3a/parachain/pallets/control/src/lib.rs#L326) of the location, unit tests still work as expected.

parachain/pallets/outbound-queue/src/lib.rs
```diff
@@ -158,12 +187,20 @@ pub mod pallet {
 			/// number of committed messages
 			count: u64,
 		},
+		/// Set base fee
+		BaseFeeSet { amount: u128 },
+		/// Set swap ratio
+		SwapRatioSet { ratio: u128 },
```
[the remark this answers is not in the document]
yrong: Actually this is not a ratio but swap price some integer about 400 https://github.com/Snowfork/snowbridge/blob/86c3ffc07f6d64d0078b2bd1a0752fcfbe012d3a/parachain/pallets/outbound-queue/src/lib.rs#L250-L254

parachain/pallets/control/src/lib.rs
```diff
@@ -146,6 +146,7 @@ pub mod pallet {
 			let message = Message {
 				origin: T::OwnParaId::get(),
 				command: Command::Upgrade { impl_address, impl_code_hash, params },
+				location: MultiLocation::parent(),
```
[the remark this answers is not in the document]
yrong: Not sure explicit add `agent` is better.

parachain/primitives/core/src/outbound.rs
yrong: https://github.com/Snowfork/snowbridge/pull/940#discussion_r1333099420

web/packages/test/scripts/configure-bridgehub.sh
yrong: Yes, we derivate the agent account same way as `agent_id`, for the reason and more detail in https://github.com/Snowfork/snowbridge/pull/940#discussion_r1308520658

parachain/primitives/core/src/outbound.rs
yrong: For `AgentExecute` the reward is paid from the agent which is maintained by parachain themselves.

parachain/pallets/outbound-queue/src/lib.rs
yrong: > charging of fees is done in the control pallet

I still prefer the outbound pallet. The reason is that even we do not charge extra fees for some of the command(e.g. AgentCommand) we still need to provide some estimate api(I may add some runtime rpc for that in outbound pallet) for the source chain to collect fees based upon. So anyway, it's nessessary to take gas cost and reward mechanism into consideration as part of the protocol from the beginning.

> About [#1234](https://github.com/paritytech/polkadot-sdk/pull/1234) the fee manager

Yes, that's the correct way to handle fee. But as commented [here](https://github.com/Snowfork/snowbridge/blob/86c3ffc07f6d64d0078b2bd1a0752fcfbe012d3a/parachain/pallets/outbound-queue/src/lib.rs#L437-L448) this is a workaround before that PR released and I've add `estimate_fee` in https://github.com/Snowfork/snowbridge/blob/86c3ffc07f6d64d0078b2bd1a0752fcfbe012d3a/parachain/primitives/router/src/outbound/mod.rs#L139-L142 for preparation. When that pr got merged we just remove the `charge_fee` logic in outbound pallet.

parachain/pallets/outbound-queue/src/lib.rs
yrong: https://github.com/Snowfork/snowbridge/pull/940#discussion_r1333152954

https://github.com/Snowfork/snowbridge/pull/940#discussion_r1325343050

https://github.com/Snowfork/snowbridge/pull/940#discussion_r1333114054

parachain/pallets/outbound-queue/src/lib.rs
yrong: https://github.com/Snowfork/snowbridge/pull/940/commits/3b784b20f4526bd31023e1279efe85d27613d1b5

parachain/pallets/outbound-queue/src/lib.rs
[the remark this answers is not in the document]
yrong: https://github.com/Snowfork/snowbridge/commit/3b784b20f4526bd31023e1279efe85d27613d1b5

contracts/test/Gateway.t.sol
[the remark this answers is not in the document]
yrong: https://github.com/Snowfork/snowbridge/pull/940/commits/460e10d4f1dc0f4fd49cc937b55b4ab6beefe775

parachain/pallets/outbound-queue/src/lib.rs
vgeddes: > I still prefer done in outbound pallet. The reason is that even we do not charge extra fees for some of the command(e.g. AgentCommand) we still need to provide some estimate api for the source chain to collect fees based upon(We may add some runtime rpc for that in outbound pallet later). Meanwhile control pallet is just one of the upper layer modules to consume the outbound queue and the charge logic should applies all the same.

It is true that the control pallet consumes the outbound queue, but that does not mean we must push all complexity into the lower level. In fact, it should be the other way around, higher layers add more complexity in an incremental fashion. 

One specific problem with your approach is we have provide information about the agent/origin location to the lower level  (the outbound-queue) so that it can charge fees. In my view, this complicates the `Message` type unnecessarily.

Also we are already using the control pallet for regular governance operations, I think it makes sense to implement `set_outbound_fee_config` extrinsic in the control pallet.

> Yes, that's the correct way to handle fee. But as commented here this is a workaround before that PR released

I don't want us to spend effort on temporary workarounds for what it is a critical aspect of our bridge operation.

I would much rather we only make partial progress, which would mean:
1. Charging fees in control pallet for certain commands
2. Specifying the fee for `AgentExecute` command in our implementation of `ExportXcm::validate`.

vgeddes: Furthermore, I also realise now, that if we limit the scope to the audit fix, then we shouldn't need to worry about charging a base fee.

Which means we only need to implement task 1: Charging fees in control pallet for certain commands

yrong: Revamp for charging fees in control pallet for certain commands
https://github.com/Snowfork/snowbridge/pull/940/commits/448963af70d72d568257a2903cd8129a58a021b3

For task2: Specifying the fee for AgentExecute command in our implementation of ExportXcm::validate. actually it's already covered in current [impl](https://github.com/Snowfork/snowbridge/blob/e8308f8f8dadf9705e4299bc05180f9cf3022344/parachain/primitives/router/src/outbound/mod.rs#L138-L143) so I would prefer not to revert it.

smoketest/tests/upgrade_gateway.rs
```diff
@@ -143,20 +141,20 @@ async fn upgrade_gateway() {
...
 		let upgrades = block.events().await.expect("read block events");
 		for upgrade in upgrades.find::<ethereum_control::events::Upgrade>() {
-			let upgrade = upgrade.expect("expect upgrade");
-			assert_eq!(upgrade.impl_address, GATETWAY_UPGRADE_MOCK_CONTRACT.into());
-			assert_eq!(upgrade.impl_code_hash, gateway_upgrade_mock_code_hash.into());
-			assert_eq!(upgrade.params_hash, Some(params_hash.into()));
+			let _upgrade = upgrade.expect("expect upgrade");
+			// assert_eq!(upgrade.impl_address, GATETWAY_UPGRADE_MOCK_CONTRACT.into());
+			// assert_eq!(upgrade.impl_code_hash, gateway_upgrade_mock_code_hash.into());
+			// assert_eq!(upgrade.params_hash, Some(params_hash.into()));
```
claravanstaden, line 147: Can we remove these?

```suggestion
			let _upgrade = upgrade.expect("expect upgrade");
			// assert_eq!(upgrade.impl_address, GATETWAY_UPGRADE_MOCK_CONTRACT.into());
			// assert_eq!(upgrade.impl_code_hash, gateway_upgrade_mock_code_hash.into());
			// assert_eq!(upgrade.params_hash, Some(params_hash.into()));
```

If so, also remove [line 79](https://github.com/Snowfork/snowbridge/pull/940/files#diff-d4bb813697a0a98a0366b54c5ebd604d9572f1b2ff40240eccdadc4455a1918aR79).

web/packages/test/config/launch-config.toml
```diff
@@ -14,14 +14,14 @@ default_args = [
...
-        [relaychain.genesis.runtime.runtime_genesis_config.configuration.config]
-        max_validators_per_core = 2
-        needed_approvals = 1
-        validation_upgrade_cooldown = 2
-        validation_upgrade_delay = 2
+    #     [relaychain.genesis.runtime.runtime_genesis_config.configuration.config]
+    #     max_validators_per_core = 2
+    #     needed_approvals = 1
+    #     validation_upgrade_cooldown = 2
+    #     validation_upgrade_delay = 2
```
claravanstaden, line 24: Did this config cause issues with the smoke tests?

web/packages/test/scripts/configure-base-fee.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
+set -eu
+
+source scripts/set-env.sh
+
+configure_base_fee() {
+    pushd $root_dir/smoketest
+    ./make-bindings.sh
+    cargo test --test configure_base_fee
```
claravanstaden, line 9: Just curious about the decision to use the smoke test here, instead of using polkadot-js like in `xcm-helper.sh`?

web/packages/test/scripts/set-env.sh
```diff
@@ -82,19 +86,21 @@ export RANDAO_COMMIT_DELAY="${ETH_RANDAO_DELAY:-3}"
...
 export REGISTER_NATIVE_TOKEN_FEE="${ETH_REGISTER_NATIVE_TOKEN_FEE:-0}"
 export SEND_NATIVE_TOKEN_FEE="${ETH_SEND_NATIVE_TOKEN_FEE:-0}"
 
 ## Vault
-export BRIDGE_HUB_INITIAL_DEPOSIT="${ETH_BRIDGE_HUB_INITIAL_DEPOSIT:-1000}"
+export BRIDGE_HUB_INITIAL_DEPOSIT="${ETH_BRIDGE_HUB_INITIAL_DEPOSIT:-10000000000000000000}"
 
-address_for()
-{
+## BaseFee to cover the cost of an Ethereum no-op dispatchable in DOT, default 0.1 DOT
```
claravanstaden, line 98: Comments like these are very useful, thank you!

web/packages/test/scripts/set-env.sh
```diff
@@ -82,19 +86,21 @@ export RANDAO_COMMIT_DELAY="${ETH_RANDAO_DELAY:-3}"
...
 
 export CREATE_CALL_INDEX="${ETH_CREATE_CALL_INDEX:-0x3500}"
-export DISPATCH_GAS="${ETH_DISPATCH_GAS:-500000}"
 
 export REGISTER_NATIVE_TOKEN_FEE="${ETH_REGISTER_NATIVE_TOKEN_FEE:-0}"
 export SEND_NATIVE_TOKEN_FEE="${ETH_SEND_NATIVE_TOKEN_FEE:-0}"
 
 ## Vault
-export BRIDGE_HUB_INITIAL_DEPOSIT="${ETH_BRIDGE_HUB_INITIAL_DEPOSIT:-1000}"
+export BRIDGE_HUB_INITIAL_DEPOSIT="${ETH_BRIDGE_HUB_INITIAL_DEPOSIT:-10000000000000000000}"
```
claravanstaden, line 96: Why the big increase? In what currency is this? DOT?

parachain/pallets/control/src/lib.rs
```diff
@@ -303,34 +323,59 @@ pub mod pallet {
...
+
+			let fee_amount: u128 = match fees.get(0) {
+				Some(&MultiAsset { fun: Fungible(amount), .. }) => amount,
+				_ => 0,
+			};
+
+			if fee_amount > 0 {
+				T::Token::transfer(
+					&agent_account,
+					&T::LocalPalletId::get().into_account_truncating(),
```
claravanstaden, line 371: Should the destination account here still be updated to Bridgehub's treasury account? [#940 (comment)](https://github.com/Snowfork/snowbridge/pull/940#discussion_r1332961063)

smoketest/tests/upgrade_gateway.rs
claravanstaden: I see you are still busy with the smoke tests, apologies. 😄

web/packages/test/config/launch-config.toml
yrong: Nope, just for quick smoke tests more agile. I remember the previous change for the `session_length` is for debugging beefy commitment issues which may not be necessary now.

web/packages/test/scripts/configure-base-fee.sh
yrong: I would assume the shell script only applies to some simple scenario, for configuring complex data structures or some business logic required script based on a high-level language like Javascript or Rust is better. As we've already heavily used rust for the smoke test I would prefer the latter.

web/packages/test/scripts/set-env.sh
yrong: It's Ether with the precision of 18, so it's 10 Ether.

parachain/pallets/control/src/lib.rs
yrong: Yes, it's treasury underlying https://github.com/Snowfork/cumulus/pull/63/files#r1338607827

web/packages/test/scripts/set-env.sh
claravanstaden: Cool! A comment stating that would be super helpful.

contracts/test/Gateway.t.sol
```diff
@@ -61,7 +61,7 @@ contract GatewayTest is Test {
     address public account1;
     address public account2;
 
-    uint256 public constant DISPATCH_GAS = 500_000;
+    uint256 public dispatch_gas = 500_000;
```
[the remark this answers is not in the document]
claravanstaden: Is gas not denoted in Gwei though? 🤔

contracts/src/FundAgent.sol
```diff
@@ -0,0 +1,46 @@
...
+    using stdJson for string;
+
+    function setUp() public {}
+
+    function run() public {
+        uint256 privateKey = vm.envUint("PRIVATE_KEY");
+        address deployer = vm.rememberKey(privateKey);
+        vm.startBroadcast(deployer);
+
+        uint256 initialDeposit = vm.envUint("BRIDGE_HUB_INITIAL_DEPOSIT");
```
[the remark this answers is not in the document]
claravanstaden: I see. I just think we need to think about the maintenance overhead that these scripts add. Is it not better to do the necessary funding in the test itself?

web/packages/test/scripts/configure-base-fee.sh
claravanstaden: In principle I agree, but I do worry we are creating dependencies all over the place that are not that obvious. For example, same as my comment here: https://github.com/Snowfork/snowbridge/pull/940#discussion_r1338677303. Now the smoketests depend on some contract scripts and our bash scripts depend on the smoketests. These dependencies aren't always clear and it would be nice if each part of our project is mostly self-contained.

parachain/pallets/control/src/lib.rs
```diff
@@ -303,34 +323,59 @@ pub mod pallet {
...
 				T::OutboundQueue::validate(&message).map_err(|_| Error::<T>::SubmissionFailed)?;
+			Self::charge_fees(&message, &agent_location)?;
 			T::OutboundQueue::submit(ticket).map_err(|_| Error::<T>::SubmissionFailed)?;
 			Ok(())
 		}
 
 		pub fn convert_location(
 			mut location: MultiLocation,
-		) -> Result<(H256, Option<ParaId>, MultiLocation), DispatchError> {
+		) -> Result<(H256, ParaId, MultiLocation), DispatchError> {
```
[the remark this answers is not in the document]
yrong: https://github.com/Snowfork/snowbridge/pull/940/commits/309ce5a9196e0d78f71659f3ee102f4ce33ab598

parachain/primitives/core/src/outbound.rs
```diff
@@ -229,3 +307,159 @@ impl AgentExecuteCommand {
...
+	/// gas cost for each command
+	pub command_gas_map: Option<
+		BoundedBTreeMap<CommandIndex, GasAmount, ConstU32<{ CommandIndex::max_value() as u32 }>>,
+	>,
+	/// gas range applies for all commands
+	pub gas_range: Option<DispatchGasRange>,
+	/// gas price in Wei from https://etherscan.io/gastracker
+	pub gas_price: Option<GasPriceInWei>,
+	/// swap ratio for Ether->DOT from https://www.coingecko.com/en/coins/polkadot/eth with precision difference between Ether->DOT(18->10)
+	pub swap_ratio: Option<FixedU128>,
```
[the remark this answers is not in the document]
yrong: The reason for the `Option` is in some cases we may only want to change partially(e.g. `command_gas_map` for the static gas cost for some commands only), then no need to provide other parameters in the call(though read previous value and feed into the call should be ok but is error prone IMO).

parachain/pallets/outbound-queue/runtime-api/src/lib.rs
```diff
@@ -2,11 +2,17 @@
...
+use xcm::prelude::MultiAssets;
 
 sp_api::decl_runtime_apis! {
 	pub trait OutboundQueueApi
 	{
 		fn prove_message(leaf_index: u64) -> Option<MerkleProof>;
+
+		fn estimate_fee(message: &Message) -> Result<MultiAssets, SubmitError>;
+
+		fn estimate_fee_by_command_index(command_index: u8) -> Result<MultiAssets, SubmitError>;
```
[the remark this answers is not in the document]
yrong: Some revamp accordingly in https://github.com/Snowfork/snowbridge/pull/940/commits/a6649d8fe563434ebaa92dc8692f44992955636c

Yes, though currently only `command` is required for computing fees.  I'm afraid there is a use case that discounts for a particular parachain. So would prefer to the previous `Message` which includes the original ParaId.

And I would assume for most of the cases `estimate_fee_by_command_index` without params should be fine so add that as a shortcut API.

contracts/test/Gateway.t.sol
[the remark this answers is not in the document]
alistair-singh: The price of gas price is in Wei/Gwei/Ether but not gas itself. Gas is just a unit to measure work done in a transaction. 
`Transaction Price (in Wei) = Gas * Gas Price (in Wei)`

contracts/src/FundAgent.sol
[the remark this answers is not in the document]
yrong: Yes, agree. It can be executed directly with `forge script` as we do for deploying contracts and we can add a thin shell wrapper for it. 

But I would prefer to improve these maintenance scripts in a separate PR and make this one more focused. As you mentioned it's already a mammoth PR now😄 🐘.

web/packages/test/scripts/configure-base-fee.sh
yrong: https://github.com/Snowfork/snowbridge/pull/940#discussion_r1339379792

parachain/pallets/control/src/lib.rs
yrong: https://github.com/Snowfork/snowbridge/pull/940/commits/6dfe50247504fbc2c556808496c4a631613d757e for less confusion.

web/packages/test/scripts/set-env.sh
yrong: https://github.com/Snowfork/snowbridge/pull/940/commits/aab0ec007f47982b3cf871e151341a94fd341cbc

parachain/pallets/outbound-queue/src/lib.rs
```diff
@@ -239,6 +206,39 @@ pub mod pallet {
 		) -> DispatchResult {
 			<Self as OwnedBridgeModule<_>>::set_operating_mode(origin, operating_mode)
 		}
+
+		/// Set fee config for outbound message.
+		/// May only be called by root
+		#[pallet::call_index(2)]
+		#[pallet::weight((T::DbWeight::get().reads_writes(2, 3), DispatchClass::Operational))]
+		pub fn set_outbound_fee_config(
```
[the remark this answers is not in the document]
yrong: https://github.com/Snowfork/snowbridge/pull/940/commits/378ca74f33a74e393fd7724cac9b073e14aaefea

contracts/src/FundAgent.sol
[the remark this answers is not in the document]
yrong: https://github.com/Snowfork/snowbridge/pull/940/commits/9a905a4eb3ea09d010d0a95877a4876311cc18d2

claravanstaden:
Hey @yrong! I tried checking out your branch together with https://github.com/Snowfork/cumulus/tree/ron/sno-624 but I couldn't get it to build - I came across some compiler errors in the [bridge hub rococo runtime xcm config](https://github.com/Snowfork/cumulus/blob/ron/sno-624/parachains/runtimes/bridge-hubs/bridge-hub-rococo/src/xcm_config.rs#L36) due to the renaming of `AgentHashedDescription`. Will you let me know when I can checkout your branch and run your branch for a final review? 😄

web/packages/test/scripts/configure-base-fee.sh
claravanstaden: That's okay with me, if we can reduce the inter-dependencies between the E2E script and the smoke tests in a follow-up PR.

parachain/pallets/outbound-queue/src/test.rs
```diff
@@ -228,3 +263,87 @@ fn process_message_fails_on_overweight_message() {
...
+			command: Command::CreateAgent { agent_id: Default::default() },
+		};
+		let fees = OutboundQueue::compute_fee_reward(&message.command).unwrap();
+		assert_eq!(fees.0, 19000000000);
+		assert_eq!(fees.1, 3375000000000000);
+
+		let mut command_gas_map = BoundedBTreeMap::<
+			CommandIndex,
+			GasAmount,
+			ConstU32<{ CommandIndex::max_value() as u32 }>,
```
claravanstaden, line 335:
```suggestion
			ConstU32<{ CommandIndex::MAX as u32 }>,
```

parachain/pallets/outbound-queue/src/lib.rs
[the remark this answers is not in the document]
claravanstaden: The added test only tests updating `command_gas_map`. If we go with this approach of only updating some items like you've explained in https://github.com/Snowfork/snowbridge/pull/940#discussion_r1338804994, we need proper test coverage and cover all of the items, like setting `base_fee`, `gas_range`, etc. The danger if we do not cover all cases is that there could be a typo and the wrong values are set to the wrong config option.

yrong:
> I tried checking out your branch together with https://github.com/Snowfork/cumulus/tree/ron/sno-624 

@claravanstaden Thanks for the review! Actually the cumulus branch is `ron/oak-sno-624` and the companion PR https://github.com/Snowfork/cumulus/pull/63

yrong:
Closed in favor of https://github.com/Snowfork/snowbridge/pull/968

PR #942, opened 2023-08-29
Inbound queue submission no longer fails silently

vgeddes:
Fix for audit issue (3).

Fixes: SNO-626


yrong:
Are we gonna to implement some kind of the refund mechanism as mentioned?

> We recommend implementing an error-handling mechanism in order to return tokens from the Agent to the user if the bridging process does not succeed on the Polkadot side

parachain/pallets/inbound-queue/src/test.rs
```diff
@@ -231,13 +271,40 @@ fn test_submit_happy_path() {
...
+		// Submit message
+		let message = Message {
+			data: OUTBOUND_QUEUE_EVENT_LOG_INVALID_DEST.into(),
+			proof: Proof {
+				block_hash: Default::default(),
+				tx_index: Default::default(),
+				data: Default::default(),
+			},
+		};
+		assert_noop!(
```
claravanstaden, line 301: This is cool, leant something today. 😄

claravanstaden (approved):

yrong (approved):
LGTM!

Though my only concern about this PR is there some irrecoverable error and in that case the bridge will be blocked and there should be some remedial action(from governance) including(not limited):

- Increase the nonce of the inbound channel to bypass that particular error message
- Refund the ERC20 token locked in Agent to the original sender

parachain/pallets/inbound-queue/src/lib.rs
```diff
@@ -208,39 +231,24 @@ pub mod pallet {
...
 				Err(_) => {
-					Self::deposit_event(Event::MessageReceived {
-						dest: envelope.dest,
-						nonce: envelope.nonce,
-						result: MessageDispatchResult::InvalidPayload,
-					});
-					return Ok(())
-				},
+					return Err(Error::<T>::InvalidPayload.into());
+				}
```
claravanstaden, line 239: Nit: You could shorten this:

```
Err(_) => Err(Error::<T>::InvalidPayload.into()),
```

parachain/pallets/inbound-queue/src/lib.rs
```diff
@@ -208,39 +231,24 @@ pub mod pallet {
...
 			};
 
-			// Attempt to convert to XCM
-			let sibling_para =
-				MultiLocation { parents: 1, interior: X1(Parachain(envelope.dest.into())) };
+			// Attempt to send XCM to a dest parachain
+			let dest = MultiLocation { parents: 1, interior: X1(Parachain(envelope.dest.into())) };
+			let (xcm_hash, _) = send_xcm::<T::XcmSender>(dest, xcm).map_err(Error::<T>::from)?;
```
claravanstaden, line 244: Do we still need to do any fee handling here (outside of this issue), using the send xcm price of delivery return parameter?

alistair-singh (approved):
🚀

parachain/pallets/inbound-queue/src/lib.rs
vgeddes: Not really. I did look into it, and the fee returned by `send_xcm` (i.e the fee for sending) seems to be an optional thing that chains can decide to implement.

Ultimately, on the destination chain, some account controlled by the source chain must pay fees anyway.

parachain/pallets/inbound-queue/src/lib.rs
vgeddes: Thanks. Fixed in latest upload.

PR #949, opened 2023-08-31
SNO-632: Submit outbound message with priority control

yrong:
Fix for audit issue (5).
Resolves:  SNO-632


parachain/primitives/core/src/outbound.rs
```diff
@@ -15,6 +15,9 @@ pub trait OutboundQueue {
 
 	/// Submit the message ticket for eventual delivery to Ethereum
 	fn submit(ticket: Self::Ticket) -> Result<MessageHash, SubmitError>;
```
vgeddes, line 17: I was thinking of rather having a single submit method in the trait, which takes a priority parameter:

```suggestion
	fn submit(ticket: Self::Ticket, priority: Priority) -> Result<MessageHash, SubmitError>;
```

The underlying implementation of the trait can still have two different submit method though.

claravanstaden (commented):
I agree with Vincent about introducing a priority parameter rather than a new method, but otherwise it looks good! 👍 

Something that I wanted to ask, is if there are lots of messages in the queue (say a 100 create agent messages), will the priority messages still be processed after the 100 messages? Or will it be moved to the front of the message queue to be processed before the other 100 messages?

EDIT: Nevermind, I see the control messages are processed immediately without being queued. That makes sense! And there isn't a possibility that those messages can be spammed by a malicious user? They are only messages sent by governance?

parachain/primitives/core/src/outbound.rs
yrong: https://github.com/Snowfork/snowbridge/pull/949/commits/ac791fd9855ff1c70e1c5e2861bff9f8977d8057

doubledup (approved):

parachain/primitives/core/src/outbound.rs
```diff
@@ -6,6 +6,13 @@ use sp_std::{borrow::ToOwned, vec, vec::Vec};
 
 pub type MessageHash = H256;
 
+#[derive(Copy, Clone, Encode, Decode, PartialEq, Eq, RuntimeDebug, TypeInfo)]
+pub enum Priority {
+	Normal,
+	High,
+	Emergency,
```
vgeddes, line 13: I'd like us to make conservative, incremental changes. So if we are not actually implementing `Emergency` priority, then I'd rather we not include this variant.

```suggestion
```

parachain/primitives/core/src/outbound.rs
```diff
@@ -6,6 +6,13 @@ use sp_std::{borrow::ToOwned, vec, vec::Vec};
 
 pub type MessageHash = H256;
 
+#[derive(Copy, Clone, Encode, Decode, PartialEq, Eq, RuntimeDebug, TypeInfo)]
+pub enum Priority {
```
vgeddes, line 10: This enum and its variants require doc comments please

parachain/pallets/control/src/lib.rs
```diff
@@ -305,7 +307,8 @@ pub mod pallet {
 		fn submit_outbound(message: Message) -> DispatchResult {
 			let ticket =
 				T::OutboundQueue::validate(&message).map_err(|_| Error::<T>::SubmissionFailed)?;
-			T::OutboundQueue::submit(ticket).map_err(|_| Error::<T>::SubmissionFailed)?;
+			T::OutboundQueue::submit(ticket, Priority::High)
```
vgeddes, line 310: I think everything except the `CreateAgent` command should be high priority. So perhaps make `submit_outbound` accept a priority parameter.

yrong: I would assume there could still be ACL control for `CreateAgent` and it's different from the normal operation from any signed origin(e.g. sending token from source chain).   

So I would prefer all commands in the control pallet with high priority(btw: maybe upgrade/pause should be with more priority like `emergency` though don't support yet).

parachain/pallets/outbound-queue/src/lib.rs
```diff
@@ -337,14 +338,27 @@ pub mod pallet {
...
+				Priority::Normal => {
+					T::MessageQueue::enqueue_message(
+						ticket.message.as_bounded_slice(),
+						AggregateMessageOrigin::Parachain(ticket.origin),
+					);
+					Self::deposit_event(Event::MessageQueued { id: ticket.id });
+				},
+				Priority::High => {
+					ensure!(
+						MessageLeaves::<T>::decode_len().unwrap_or(0) <
```
vgeddes, line 353: So unfortunately, if there is congestion, then this `ensure!` check is always going to fail. The MessageQueue pallet is going to fill up `MessageLeaves` at the start of every block, and there won't be any more space for `High` priority messages.

yrong: Not sure but I would assume the message is queued by [Origin](https://github.com/paritytech/polkadot-sdk/blob/2c35bc3d273f47d45c17078134c048016ef7a57a/substrate/frame/message-queue/src/lib.rs#L766) and the congestion for normal user operations(e.g. sendToken from assetHub) will not block the control commands here origin from bridgeHub.

And I would prefer these low level of the QoS priority control of traffic stuff to be addressed by the internal implementation of `message-queue` pallet.

yrong:
@vgeddes Checked a bit more for the `message-queue`  implementation and now I feel it's not as easy to support priority control as I previously thought. 

Notice the recent commit [Reserve space for high priority messages](https://github.com/Snowfork/snowbridge/pull/949/commits/4cf60f8f21993ac118bbbe1064db3f48cc543ed2) seems more practical so I revamped a bit based on that.

parachain/pallets/outbound-queue/src/lib.rs
yrong: Since it's not easy for the upstream `message-queue` to support priority control natively. After some more thought revamp in https://github.com/Snowfork/snowbridge/pull/949/commits/fe5c8a3d96a7a5ac4f68f0339d0b655eea8287fe
from our side for the feature `swap low priority message with high priority one when hard limit reached`.

parachain/pallets/outbound-queue/src/lib.rs
```diff
@@ -337,14 +349,22 @@ pub mod pallet {
...
+				Priority::Normal => {
+					Self::ensure_not_halted().map_err(|_| SubmitError::BridgeHalted)?;
+					T::MessageQueue::enqueue_message(
+						ticket.message.as_bounded_slice(),
+						AggregateMessageOrigin::Parachain(ticket.origin),
+					);
+					Self::deposit_event(Event::MessageQueued { id: ticket.id });
+				},
+				Priority::High => {
+					Self::do_process_message(&ticket.message.as_bounded_slice(), priority)
```
vgeddes, line 363: This call will fail if `MessageLeaves` is already filled up by the time the code reaches here. As long as the number of queued messages is greater than the maximum length of `MessageLeaves`, this edge case will always be hit.

Lets trace the execution of a block:
1. `on_initialize` is called to reset storage
2. `MessageQueue::on_initialize` delivers messages to the message processor `process_message`
3. `MessageLeaves` gets filled up with low priority messages
4. Governance submits an upgrade message. This fails as there is no more space in `MessageLeaves`.

parachain/pallets/outbound-queue/src/lib.rs
vgeddes: Unfortunately, that is not going to work, and will result in high priority messages being dropped. See my other comment about it.

yrong: Thanks for pointing it out. Yes, I also realize it and revamp in https://github.com/Snowfork/snowbridge/pull/949/commits/42692ab2fd457618d325fb7ce87e89f08bd42c8f

parachain/primitives/core/src/outbound.rs
```diff
@@ -192,6 +206,15 @@ impl Command {
...
+
+	/// The priority of the command
+	pub fn priority(&self) -> Priority {
+		match self {
+			Command::Upgrade { .. } => Priority::High,
+			Command::SetOperatingMode { .. } => Priority::High,
+			_ => Priority::Normal,
+		}
+	}
 }
```
yrong, line 218: The priority of the command.

parachain/pallets/outbound-queue/src/lib.rs
yrong: Thanks for pointing it out. Yes, I also realize it and revamp in https://github.com/Snowfork/snowbridge/commit/42692ab2fd457618d325fb7ce87e89f08bd42c8f

This only concern is about how to get the priority, for me seems not relevant to the parachain(I would assume it's still possible that low priority message from bridge hub or high priority from sibling chains), it's more like an [extended attribute](https://github.com/Snowfork/snowbridge/pull/949/files#r1357042670) of the command.

parachain/primitives/core/src/outbound.rs
vgeddes: I think we should determine the priority using the `AggregateMessageOrigin` origin of the command, because its most efficient.

The implementation of `ProcessMessage::process_message` needs to decode each message to determine the priority. Suppose there are 10000 low priority messages in the queue, followed by 1 high priority message. After delivering 100 low priority messages for processing, the maximum block weight may be reached, and processing stops. And then the situation repeats in the next block.

Ultimately what happens is that the high priority message is never processed because its at the end of the queue.

Instead, since `ProcessMessage::process_message` is also passed an origin parameter, we should use that. Specifically, we should consider the `AggregateMessageOrigin::Parachain(BridgeHub)` origin to be high priority.

parachain/pallets/outbound-queue/src/lib.rs
```diff
@@ -208,6 +212,7 @@ pub mod pallet {
 	{
 		fn on_initialize(_: BlockNumberFor<T>) -> Weight {
 			// Remove storage from previous block
+			PendingHighPriorityMessageCount::<T>::kill();
```
vgeddes, line 215: The idea was to _decrement_ this counter after processing each high priority message.

We can't kill it every block, because there is no guarantee that all pending high priority messages will be processed every block.

parachain/pallets/outbound-queue/src/lib.rs
```diff
@@ -293,6 +301,10 @@ pub mod pallet {
 			MessageLeaves::<T>::append(message_abi_encoded_hash);
 			Nonce::<T>::set(enqueued_message.origin, next_nonce);
 
+			if priority == Priority::High {
+				PendingHighPriorityMessageCount::<T>::mutate(|cnt| *cnt = cnt.saturating_add(1));
```
vgeddes, line 305:
```suggestion
				PendingHighPriorityMessageCount::<T>::mutate(|count| *count = count.saturating_add(1));
```

parachain/primitives/core/src/outbound.rs
```diff
@@ -6,6 +6,16 @@ use sp_std::{borrow::ToOwned, vec, vec::Vec};
 
 pub type MessageHash = H256;
 
+/// Priority for submit the message ticket to OutboundQueue
+#[derive(Copy, Clone, Encode, Decode, PartialEq, Eq, RuntimeDebug, TypeInfo)]
+pub enum Priority {
```
vgeddes, line 11: I've realised that we can't use this `Priority` type any more. See my comment below, for why we need to use message origin as the priority.

parachain/primitives/router/src/outbound/mod.rs
```diff
@@ -147,7 +147,7 @@ where
 				SendError::NotApplicable
 			})?;
 
-		let message_hash = OutboundQueue::submit(ticket).map_err(|_| {
+		let message_hash = OutboundQueue::submit(ticket, Priority::Normal).map_err(|_| {
```
vgeddes, line 150:
```suggestion
		let message_hash = OutboundQueue::submit(ticket).map_err(|_| {
```

parachain/pallets/outbound-queue/src/lib.rs
```diff
@@ -337,9 +349,10 @@ pub mod pallet {
 			Ok(ticket)
 		}
 
-		fn submit(ticket: Self::Ticket) -> Result<MessageHash, SubmitError> {
-			// Make sure the bridge not halted
-			Self::ensure_not_halted().map_err(|_| SubmitError::BridgeHalted)?;
+		fn submit(ticket: Self::Ticket, priority: Priority) -> Result<MessageHash, SubmitError> {
```
vgeddes, line 352:
```suggestion
		fn submit(ticket: Self::Ticket) -> Result<MessageHash, SubmitError> {
```

yrong: https://github.com/Snowfork/snowbridge/pull/949/commits/239f3c37a261c7bfda76fb6dcfbce5805eeecee4

parachain/primitives/core/src/outbound.rs
yrong: https://github.com/Snowfork/snowbridge/pull/949/commits/239f3c37a261c7bfda76fb6dcfbce5805eeecee4

parachain/pallets/outbound-queue/src/lib.rs
yrong: https://github.com/Snowfork/snowbridge/pull/949/commits/239f3c37a261c7bfda76fb6dcfbce5805eeecee4

parachain/primitives/core/src/outbound.rs
yrong: https://github.com/Snowfork/snowbridge/pull/949/commits/239f3c37a261c7bfda76fb6dcfbce5805eeecee4 for the fix.

> And then the situation repeats in the next block. Ultimately what happens is that the high priority message is never processed because its at the end of the queue.

From my previous unit test this is not true, seems the `pallet_message_queue` process message from different queues in a round-robin manner, so after the first 100 low-priority messages get processed in the first block the high-priority message from another parachain will get the chance in the next block.

But anyway, the fix above as your suggestion is better, will get the 1 high-priority message and 99 low-priority messages in the first block, which is exactly what we want.  Unit test related covering in https://github.com/Snowfork/snowbridge/blob/239f3c37a261c7bfda76fb6dcfbce5805eeecee4/parachain/pallets/outbound-queue/src/test.rs#L315

yrong: Though I removed `priority` from the submit interface in the fix above,  still some concerns about always deeming message from bridge hub as high priority, specifically assuming message queue for bridge hub has already been spammed by the `create_agent` message, the `upgrade` will never get the chance to execute.

But consider the fee problem has been addressed in https://github.com/Snowfork/snowbridge/pull/968, maybe we can make it simple in this PR and try to improve it later.

parachain/pallets/outbound-queue/src/lib.rs
```diff
@@ -351,16 +388,14 @@ pub mod pallet {
 
 	impl<T: Config> ProcessMessage for Pallet<T> {
 		type Origin = AggregateMessageOrigin;
+		#[allow(clippy::clone_double_ref)]
 		fn process_message(
 			message: &[u8],
-			_: Self::Origin,
+			_origin: Self::Origin,
```
vgeddes, line 394: This is the origin you need to use to determine the message priority. Not the origin in the encoded message.

parachain/pallets/outbound-queue/src/lib.rs
```diff
@@ -272,6 +279,28 @@ pub mod pallet {
...
+				Command::Upgrade { .. } => false,
+				Command::SetOperatingMode { .. } => false,
+				_ => true,
+			};
+			if halt_check_required {
+				Self::ensure_not_halted().map_err(|_| ProcessMessageError::Yield)?;
+			}
+
+			// Yield low priority message if there was pending high priority message
+			let high_priority = enqueued_message.origin == T::OwnParaId::get();
```
vgeddes, line 292: The problem here is that the message needs to be decoded to determine the origin. This is too expensive.

See my other latest comment about the origin passed to ProcessMessage::process_message. That's the one you should check.

parachain/pallets/outbound-queue/src/lib.rs
```diff
@@ -272,6 +279,23 @@ pub mod pallet {
 			let enqueued_message: EnqueuedMessage =
 				EnqueuedMessage::decode(&mut message).map_err(|_| ProcessMessageError::Corrupt)?;
 
+			// Skp halt check for Upgrade and SetOperatingMode
+			let halt_check_required = !matches!(
```
vgeddes, line 283: We run into the same problem where we need to decode all the pending messages before deciding whether to yield or not.

I want you to try and find a design that relies solely on the origin passed to process_message to decide whether to yield.

Consider using an origin type like this:
Struct 

```rust 
struct AggregateMessageOrigin {
  Self(Priority)
  Parachain(ParaId)
}
```

Then in `process_message`, if origin != Self(Priority::High), when we should yield the message.

yrong: Cool! So does that implicitly mean you prefer  to give `create_agent`  a low Priority to address my previous concern in https://github.com/Snowfork/snowbridge/pull/949#discussion_r1357700866?

> specifically assuming message queue for bridge hub has already been spammed by the `create_agent` message, then the upgrade will never get the chance to execute.

vgeddes: Yes, let's give create-agent a lower priority.

vgeddes: But I don't want the priority to be a parameter passed to OutboundQueue::submit.

Instead OutboundQueue should just check if the command is an upgrade or operating mode change and then prioritise accordingly

yrong: Yes, that's exactly what the current implementation does. We check priority in `validate` phase and set it to the `ticket` accordingly.

parachain/pallets/outbound-queue/src/lib.rs
```diff
@@ -317,42 +336,70 @@ pub mod pallet {
...
+			}
+			T::MessageQueue::enqueue_message(ticket.message.as_bounded_slice(), origin);
 			Self::deposit_event(Event::MessageQueued { id: ticket.id });
 			Ok(ticket.id)
 		}
 	}
 
 	impl<T: Config> ProcessMessage for Pallet<T> {
 		type Origin = AggregateMessageOrigin;
+		#[allow(clippy::clone_double_ref)]
```
vgeddes, line 380: We should fix this. Do you know what the error message is so I can help with it?

yrong: https://github.com/Snowfork/snowbridge/pull/949/commits/cfb1cfba8659f4b459bb68f0d6c1c7959bb06b84

parachain/pallets/outbound-queue/src/lib.rs
```diff
@@ -317,42 +336,70 @@ pub mod pallet {
...
+						*count = count.saturating_add(1)
+					});
+				} else {
+					origin = AggregateMessageOrigin::SelfChain(Priority::Low);
+					Self::ensure_not_halted().map_err(|_| SubmitError::BridgeHalted)?;
+				}
+			} else {
+				origin = AggregateMessageOrigin::Parachain(ticket.origin);
+				Self::ensure_not_halted().map_err(|_| SubmitError::BridgeHalted)?;
+			}
```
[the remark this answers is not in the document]
yrong: https://github.com/Snowfork/snowbridge/pull/949/commits/db7cf6b0b35857b70e88d835c8187564ee1d3bf8

parachain/pallets/outbound-queue/src/lib.rs
```diff
@@ -317,42 +336,70 @@ pub mod pallet {
 			let encoded =
 				enqueued_message.encode().try_into().map_err(|_| SubmitError::MessageTooLarge)?;
 
-			let ticket =
-				OutboundQueueTicket { id: message_id, origin: message.origin, message: encoded };
+			let priority = match message.command {
+				Command::Upgrade { .. } | Command::SetOperatingMode { .. } => Priority::High,
+				_ => Priority::Low,
```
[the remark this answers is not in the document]
yrong: https://github.com/Snowfork/snowbridge/pull/949/commits/7252285dc2776f1b672cebe76aa8960b25772da5

parachain/pallets/outbound-queue/src/lib.rs
```diff
@@ -317,42 +336,70 @@ pub mod pallet {
...
+			};
+
+			let ticket = OutboundQueueTicket {
+				id: message_id,
+				origin: message.origin,
+				message: encoded,
+				priority,
+			};
```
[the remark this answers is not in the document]
yrong: Yes, that make sense. The problem is that message in submit is encoded already so we need to decode to inspect the priority of the command.

parachain/primitives/core/src/outbound.rs
```diff
@@ -355,3 +349,19 @@ pub struct OriginInfo {
...
+	Low,
+	High,
+}
+
+/// Aggregate message origin for the `MessageQueue` pallet.
+#[derive(Encode, Decode, Clone, MaxEncodedLen, Eq, PartialEq, RuntimeDebug, TypeInfo)]
+pub enum AggregateMessageOrigin {
+	#[codec(index = 0)]
+	SelfChain(Priority),
```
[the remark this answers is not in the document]
yrong: `Self` is a keyword in rust.

parachain/primitives/core/src/outbound.rs
```diff
@@ -355,3 +349,19 @@ pub struct OriginInfo {
 	/// The deterministic ID of the agent for this origin
 	pub agent_id: H256,
 }
+
+/// Priority
+#[derive(Encode, Decode, Clone, MaxEncodedLen, Eq, PartialEq, RuntimeDebug, TypeInfo)]
+pub enum Priority {
+	Low,
```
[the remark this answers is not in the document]
yrong: https://github.com/Snowfork/snowbridge/pull/949/commits/cc47ac5fd0a4e5530590e61419a8ea23b0edbfe0

parachain/primitives/core/src/outbound.rs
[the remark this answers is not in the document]
vgeddes: ahh. Good catch.

Then lets rather use `BridgeHub`.

parachain/primitives/core/src/outbound.rs
```diff
@@ -355,3 +349,26 @@ pub struct OriginInfo {
 	/// The deterministic ID of the agent for this origin
 	pub agent_id: H256,
 }
+
+/// Priority
```
vgeddes, line 353:
```suggestion
/// Priority for an outbound message
```

parachain/pallets/outbound-queue/src/lib.rs
```diff
@@ -144,6 +152,9 @@ pub mod pallet {
 	#[pallet::unbounded]
 	pub(super) type Messages<T: Config> = StorageValue<_, Vec<PreparedMessage>, ValueQuery>;
 
+	#[pallet::storage]
```
vgeddes, line 155:
```suggestion
	/// Number of high priority messages that are waiting to be processed
	#[pallet::storage]
```

parachain/pallets/outbound-queue/src/lib.rs
```diff
@@ -99,6 +105,8 @@ pub mod pallet {
 		#[pallet::constant]
 		type DeliveryReward: Get<u128>;
 
+		type HighPriorityCommands: Contains<Command>;
```
vgeddes, line 108:
```suggestion
		/// Determines whether a command is high priority
		type HighPriorityCommands: Contains<Command>;
```

parachain/pallets/outbound-queue/src/lib.rs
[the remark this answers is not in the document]
vgeddes: Ah makes sense. OK, lets keep the existing code then.

parachain/primitives/core/src/outbound.rs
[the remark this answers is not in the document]
yrong: https://github.com/Snowfork/snowbridge/pull/949/commits/783762c4305a6f28ca738c7c73940b540eba7f9d

vgeddes:
I've changed my mind about one aspect of how we are dealing with priorities.

When we launch initially in production, the `CreateAgent` and `CreateChannel` commands will be disabled by default, by ensuring that a call filter in the runtime prevents them from being executed.

So there is no need to make sure that `Upgrade` and `SetOperatingMode` are given priority over the above disabled commands. When we eventually enable them, we can tackle the problem at that time.

Accordingly, I would like to change the definition of our `AggregateMessageOrigin` to the following:

```rust
#[derive(Encode, Decode, Clone, MaxEncodedLen, Eq, PartialEq, RuntimeDebug, TypeInfo)]
pub enum ExportOrigin {
	Here,
	Sibling(ParaId),
}

/// Aggregate message origin for the `MessageQueue` pallet.
#[derive(Encode, Decode, Clone, MaxEncodedLen, Eq, PartialEq, RuntimeDebug, TypeInfo)]
pub enum AggregateMessageOrigin {
	/// Message is to be exported via a bridge
	Export(ExportOrigin)
}
```

We have singe `Export` variant in the aggregate origin reserved for our bridge. This is because other variants will be reserved for other message processors. See for example this PR which extends AggregateMessageOrigin: https://github.com/paritytech/polkadot-sdk/pull/1246/files#diff-bae5ed28e8fcfedf07c29cd6c3834e147b1723c4ceddb42076e25318668bf148R83

yrong:
> We have singe Export variant in the aggregate origin reserved for our bridge. This is because other variants will be reserved for other message processors.

So in the end are we gonna to merge into the [cumulus primitive](https://github.com/paritytech/polkadot-sdk/blob/a2dd345b1964eeb967f3fe0b07da3ac9b94f81c7/cumulus/primitives/core/src/lib.rs#L83-L94) as a seperated entry? Some revamp accordingly in  https://github.com/Snowfork/snowbridge/pull/949/commits/a6317f0e357d6e4cbbb48b2b2bc02747ee8403b4

vgeddes:
> So in the end are we gonna to merge into the [cumulus enum](https://github.com/paritytech/polkadot-sdk/blob/a2dd345b1964eeb967f3fe0b07da3ac9b94f81c7/cumulus/primitives/core/src/lib.rs#L83-L94) as a seperate entry?

Yeh, exactly 👍

vgeddes (commented):
Added some more comments:
https://github.com/Snowfork/snowbridge/commit/a6317f0e357d6e4cbbb48b2b2bc02747ee8403b4#comments

vgeddes (approved):

PR #951, opened 2023-08-31
Inbound queue nonce update should use saturating_add

claravanstaden:
To prevent `u64` overflow.

Resolves: [SNO-647](https://linear.app/snowfork/issue/SNO-647)


parachain/pallets/inbound-queue/src/lib.rs
```diff
@@ -190,10 +190,10 @@ pub mod pallet {
 
 			// Verify message nonce
 			<Nonce<T>>::try_mutate(envelope.dest, |nonce| -> DispatchResult {
-				if envelope.nonce != *nonce + 1 {
+				if envelope.nonce != nonce.saturating_add(1) {
```
doubledup, line 193: Why is this better than overflow? With overflow, attackers can replay messages in order starting from the first message sent. With saturation, they can send arbitrary messages provided `envelope.nonce == u64::MAX`.

claravanstaden: Good point! Let me know if the change I made looks sufficient.

doubledup: Looks good!

PR #952, opened 2023-09-01
SNO-625: Change owner of the asset from gateway to agent

yrong:
Resolves: SNO-625


vgeddes (commented):
> TBC: Or just use agent as owner?

Yeah we should use the agent as the owner, not `msg.sender`.

vgeddes:
If the owner is the agent, its makes handling accounting for the 10 DOT deposit much more simpler. Since the `extraFee` is already being deposited into the agent contract.

claravanstaden (approved):

contracts/src/Assets.sol
```diff
@@ -21,7 +21,7 @@ library Assets {
     event TokenSent(
         address indexed token, address indexed sender, ParaID destinationChain, bytes destinationAddress, uint128 amount
     );
-    event TokenRegistrationSent(address token);
+    event TokenRegistrationSent(address indexed token, address indexed owner, uint256 deposit);
```
doubledup, line 24: Looks like the 3rd field here is set to a fee:
```suggestion
    event TokenRegistrationSent(address indexed token, address indexed owner, uint256 registerTokenFee);
```

smoketest/tests/register_token.rs
```diff
@@ -1,28 +1,29 @@
...
-	providers::{Http, Provider},
-	signers::{LocalWallet, Signer},
-	utils::parse_units,
-	utils::rlp::Encodable,
+    core::types::Address,
+    middleware::SignerMiddleware,
+    providers::{Http, Provider},
+    signers::{LocalWallet, Signer},
+    utils::parse_units,
+    utils::rlp::Encodable,
```
doubledup, line 8: Let's remove the conversion from tabs to spaces and stick with tabs for consistency.

alistair-singh (approved):

parachain/primitives/router/src/inbound/mod.rs
```diff
@@ -225,6 +224,13 @@ impl Command {
 			),
 		}
 	}
+
+	fn convert_to_substrate_address(chain_id: u64, address: H160) -> [u8; 32] {
+		GlobalConsensusEthereumAccountConvertsFor::<[u8; 32]>::from_params(
+			&chain_id,
+			address.as_fixed_bytes(),
+		)
+	}
```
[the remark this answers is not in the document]
yrong: Reverted.

smoketest/tests/register_token.rs
yrong: Seems it's wrongly formatted by my IDE, to resolve it move `rustfmt.toml` to top level sharing the same rules.

parachain/primitives/router/src/inbound/mod.rs
```diff
@@ -37,6 +37,8 @@ pub enum Command {
 	RegisterToken {
 		/// The address of the gateway
 		gateway: H160,
+		/// The address of the owner
+		owner: H160,
```
vgeddes, line 41: Let's use more specific naming. All tokens are really always owned by a single agent
```suggestion
		agent: H160,
```

vgeddes (commented):
Seems we still need to update the runtime configuration for AssetHub to work with the new asset multilocation:

https://github.com/Snowfork/cumulus/blob/38c2e26a4ead087d0837b5abdc3cdc3e0e6e0d8e/parachains/runtimes/assets/asset-hub-kusama/src/xcm_config.rs#L607

Also raises a new concern: Do we need to include the address of the agent for AssetHub in static runtime config.

Can we simply the runtime configuration by leaving the gateway address in the asset multilocation, like so:

```
/AccountKey20(Gateway)/AccountKey20(Agent)/AccountKey20(Token)
```

parachain/primitives/router/src/inbound/mod.rs
```diff
@@ -37,6 +37,8 @@ pub enum Command {
 	RegisterToken {
 		/// The address of the gateway
 		gateway: H160,
```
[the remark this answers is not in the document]
yrong: Nope, it's required for `pallet_bridge_transfer` as the reserved prefix and should not change.

vgeddes:
I'm inclined to not go forward with this for the time being.

I like the simplicity of having the gateway as the asset owner.

@yrong can you close this PR, but keep the branch around just in case.

PR #953, opened 2023-09-01
Updates of the Gateway address in inbound-queue pallet make the bridge stuck

alistair-singh:
Removing the `set_gateway` extrinsic, making the config static and immutable.

Resolves: SNO-640
Cumulus Companion: https://github.com/Snowfork/cumulus/pull/61

### DIFF
--- a/parachain/pallets/inbound-queue/src/lib.rs
+++ b/parachain/pallets/inbound-queue/src/lib.rs
@@ -89,6 +89,7 @@ pub mod pallet {
 		type WeightInfo: WeightInfo;
 
 		// Gateway contract address
+		#[pallet::constant]
 		type GatewayAddress: Get<H160>;
 
 		#[cfg(feature = "runtime-benchmarks")]
--- a/parachain/pallets/inbound-queue/src/lib.rs
+++ b/parachain/pallets/inbound-queue/src/lib.rs
@@ -19,11 +19,8 @@ mod test;
 
 use codec::{Decode, DecodeAll, Encode};
 use frame_support::{
-	traits::{
-		fungible::{Inspect, Mutate},
-		GenesisBuild,
-	},
-	CloneNoBound, DefaultNoBound, EqNoBound, PartialEqNoBound,
+	traits::fungible::{Inspect, Mutate},
+	CloneNoBound, EqNoBound, PartialEqNoBound,
 };
 use frame_system::ensure_signed;
 use scale_info::TypeInfo;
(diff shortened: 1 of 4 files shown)


parachain/pallets/inbound-queue/src/lib.rs
```diff
@@ -91,6 +88,9 @@ pub mod pallet {
 
 		type WeightInfo: WeightInfo;
 
+		// Gateway contract address
+		type GatewayAddress: Get<H160>;
```
vgeddes, line 92:
```suggestion
		#[pallet::constant]
		type GatewayAddress: Get<H160>;
```

alistair-singh: Resolved in [6693a52](https://github.com/Snowfork/snowbridge/pull/953/commits/6693a52a450f830726e8cd6b129b0702517e0cfb)

yrong:
Then would suggest deploying `GatewayProxy` with [create2](https://eips.ethereum.org/EIPS/eip-1014) in a more invariable and predictable way.

claravanstaden (approved):
So updating the Gateway contract on Ethereum would require an upgrade on the Polkadot side as well?

yrong:
> So updating the Gateway contract on Ethereum would require an upgrade on the Polkadot side as well?

Yes, it's hard coded in asset_hub [here](https://github.com/Snowfork/cumulus/blob/38c2e26a4ead087d0837b5abdc3cdc3e0e6e0d8e/parachains/runtimes/assets/asset-hub-kusama/src/xcm_config.rs#L547)

alistair-singh:
> So updating the Gateway contract on Ethereum would require an upgrade on the Polkadot side as well?

Yes it would. But because we use the proxy upgrade pattern this will only ever need to happen once.

alistair-singh:
> Then would suggest deploying `GatewayProxy` with [create2](https://eips.ethereum.org/EIPS/eip-1014) in a more invariable and predictable way.

We will first deploy with a zero address and then runtime upgrade to update the address.

yrong:
> We will first deploy with a zero address and then runtime upgrade to update the address.

With `create2` the address is fully predictable(as the proxy code is pretty simple and will not change) so there is no need for a runtime upgrade.

vgeddes:
> > We will first deploy with a zero address and then runtime upgrade to update the address.
> 
> With `create2` the address is fully predictable(as the proxy code is pretty simple and unlikely to change) so there is no need for a runtime upgrade. Another benefit it will be the same address across multiple networks(e.g. goerli and mainnet).

The `create2` option does come with various drawbacks that complicate operations. For example, the deployed bytecode needs to be the same across all networks. And when we deploy the gateway proxy, we also need to deploy an implementation contract, and need to upgrade it even though the BridgeHubs don't exist yet.

So I am not too keen on using `create2` for these reasons.

vgeddes (approved):

yrong (approved):

PR #955, opened 2023-09-02
Get code and generate code hash in upgrade test

alistair-singh:
Test fails every time we change the Gateway contracts interface because of the hard-coded code hash. This makes the test dynamically generate the code_hash.


claravanstaden (approved):
This is cool! I think it would be useful to extract getting the contract address to the helper module and do this for all the contracts. But we could do that later too.

PR #956, opened 2023-09-05
Dynamic signature sampling

doubledup:
This counts the number of times each validator's signature is used in the current session. It then uses the recommendation from W3F to adjust the required number of signatures for a commitment to be accepted.

@vgeddes For now, I've stored these counts by validator address only (not session id) as I'm not yet sure of a use for the counts from past sessions. If we need them, I'll extend the key to include session id.

If we can store them by validator index, there's a potential optimisation where we can pack multiple counts into a single word, eg. keep a `uint256[]` where each count is a `uint16` packed into an entry in that array - similar to how we work with bitfields. We'll have at most `2**16-1 = 65 535` uses of each validator's signature in `submitInitial` calls. For reference, with 6s block time & 24h sessions, we'll see `24*60*60/6 = 14 400` blocks each session.

Resolves: SNO-671, SNO-672, SNO-673, SNO-674, SNO-634


yrong:
@doubledup I remember the initial suggestion from Web3 team is `Limiting validators reuse per relayer`(also mentioned in the Oak audit report), right? Just curious what's the benefit of this solution over the previous one or why they abandon that one?

There is a draft PR https://github.com/Snowfork/snowbridge/pull/879 for that though I closed later. I'm not sure but my intuition will the current implementation cost much more gas(considering an invariant length array `validatorsUsedThisSession` requires deletion in each session and some logarithmic arithmetic introduced)?

vgeddes:
> @doubledup I remember the initial suggestion from Web3 team is `Limiting validators reuse per relayer`(also mentioned in the Oak audit report), right? Just curious what's the benefit of this solution over the previous one or why they abandon that one?
> 
> There is a draft PR #879 for that though I closed later. I'm not sure but my intuition will the current implementation cost much more gas(considering an invariant length array `validatorsUsedThisSession` requires deletion in each session and some logarithmic arithmetic introduced)?

@Ron after a call with Web3 (I can't remember if you were on the call), we decided to go with this [solution](https://hackmd.io/wsVcL0tZQA-Ks3b5KJilFQ?view#Solution-3-Increasing-checked-signatures-with-the-number-of-times-the-validator’s-signature-is-used), where the number of checked signatures dynamically increases. The benefit of this approach is that relayers don't need to put up deposits.

yrong:
Thanks for letting me know.  Then another question for the comment
> The benefit of this approach is that relayers don't need to put up deposits.

Is the deposit required for option 1? I guess it's for later slashing right? Then is it possible to only force the check(Limiting validators reuse per relayer) without slashing?

doubledup:
@yrong For gas costs, we can potentially reduce them by packing multiple counts into an array like `uint256[]`. This requires a stable ordering of the validators within a session though - something I'm not yet sure we can count on.

The log arithmetic is already fairly performant as we're rounding to an integer rather than calculating a fraction: https://github.com/Snowfork/snowbridge/pull/956/files#diff-473fead6cd500a4a05f06d0cea4d9ef166c9bdbe2a2e1a06a33d2e1aedcc428eR427-R433

yrong:
> For now, I've stored these counts by validator address only (not session id) as I'm not yet sure of a use for the counts from past sessions. If we need them, I'll extend the key to include session id.

Yes, I prefer to extend the key to include session id and would suspect we can merge the two storages `signatureCounters` and `validatorsUsedThisSession` into one.  Moreover, doubt if it's still necessary for the auxiliary storage `validatorsUsedThisSession` only used for cleanup purposes.  Just curious Is there still a reward with gas refund for freeing up storage space? (vaguely recall it's canceled in some EIP but I can't find it now).

vgeddes (commented):
I suggest a different data structure for the usage counters, one that has efficient space complexity and can be deleted in constant time.

Since we know we need to keep counters for both the current and next session, we can update the `ValidatorSet` struct with a `counters` field as follows:

```
    struct ValidatorSet {
        uint128 id;
        uint128 length;
        bytes32 root;
        uint256[] counters;
    }
```

This `counters` field is a packed array of `uint16` usage counters for each validator. 

This means we can track whether a validator is used 0, 1, 2, 3 or 4 and more times. This level of quantization/compression may be sufficient to satisfy the algorithm proposed by W3F.

Updating the packed array can be performed using bitfield math. Deleting it is a simple `delete currentValidatorSet.counters` 1-liner.

vgeddes:
> I suggest a different data structure for the usage counters, one that has efficient space complexity and can be deleted in constant time.

Oops I see that you suggested mostly the same thing in the PR description :+1:

However, if we make sure that the packed array is scoped to a single session, then it is possible to track validators by their leaf index, as they will remain static during a session.

As I said previously we need to have a packed array for both the current and next session.

On session handover, we also need to copy over the counters from the next session to the current session, i.e:
```
copyCounters(from=nextValidatorSet.counters, to=currentValidatorSet.counters);
```

and then we reset the counters for the next session to zero and to match the size of the new next validator set.

I did some gas tests with this approach and it doesn't look too bad honestly:
```
| test/PackedArrayGasTest.t.sol:ArrayThing contract |                 |        |        |        |         |
|---------------------------------------------------|-----------------|--------|--------|--------|---------|
| Deployment Cost                                   | Deployment Size |        |        |        |         |
| 298544                                            | 1519            |        |        |        |         |
| Function Name                                     | min             | avg    | median | max    | # calls |
| copyCounters                                      | 57593           | 57593  | 57593  | 57593  | 1       |
| deleteCountersA                                   | 7920            | 7920   | 7920   | 7920   | 1       |
| initCountersA                                     | 168104          | 168104 | 168104 | 168104 | 1       |
| initCountersB                                     | 168047          | 168047 | 168047 | 168047 | 1       |

```

Test Script (assuming 1000 validators):
```solidity
// SPDX-License-Identifier: Apache-2.0
pragma solidity ^0.8.20;

import "forge-std/Test.sol";
import "forge-std/console.sol";

import {Agent} from "../src/Agent.sol";

contract ArrayThing {

    uint256 constant NUM_VALIDATORS = 1024;
    uint256 constant ELEMENT_SIZE = 2;
    uint256 constant CONTAINER_SIZE = 32;
    uint256 constant NUM_CONTAINERS = (NUM_VALIDATORS * ELEMENT_SIZE) / CONTAINER_SIZE;

    uint256[] public countersA;
    uint256[] public countersB;

    function initCountersA() public {
        countersA = new uint256[](NUM_CONTAINERS);
    }

    function initCountersB() public {
        countersB = new uint256[](NUM_CONTAINERS);
    }

    function deleteCountersA() public {
        delete countersA;
    }

    function copyCounters() public {
        for (uint i = 0; i < countersB.length; i++) {
            countersA[i] = countersB[i];
        }
    }
}

contract PackedArrayGasTest is Test {

    ArrayThing public thing;

    function setUp() public {
        thing = new ArrayThing();
    }

    function testCounters() public {
        thing.initCountersA();
        thing.initCountersB();
        thing.copyCounters();
        thing.deleteCountersA();
    }
}
```

yrong:
> On session handover, we also need to copy over the counters from the next session to the current session, i.e:

```copyCounters(from=nextValidatorSet.counters, to=currentValidatorSet.counters);```

Not sure fully understand the point here for the copy operation, Is the [validate logic](https://github.com/Snowfork/snowbridge/blob/6b08d309db4a7f5a7ec20ae490173644b6ea47ff/contracts/src/BeefyClient.sol#L569-L571) checks commitment can only be updated incrementally?  

So when the handover happens, the counters in the previous session will not be in use anymore? Please correct me if I'm wrong.

File: contracts/src/BeefyClient.sol [hunk 1]
```diff
@@ -321,6 +348,10 @@ contract BeefyClient {
             nextValidatorSet.id = leaf.nextAuthoritySetID;
             nextValidatorSet.length = leaf.nextAuthoritySetLen;
             nextValidatorSet.root = leaf.nextAuthoritySetRoot;
+
+            for (uint256 i = 0; i < validatorsUsedThisSession.length; i++) {
```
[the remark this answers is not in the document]
doubledup: The packing optimisation in the description should address this.

contracts/src/BeefyClient.sol
```diff
@@ -145,6 +148,12 @@ contract BeefyClient {
     // Currently pending tickets for commitment submission
     mapping(bytes32 => Ticket) public tickets;
 
+    // The addresses of the validators used for submitInitial calls in the current session. Used to reset
+    // signatureCounters on session handover.
+    address[] public validatorsUsedThisSession;
+    // The number of times each validator signature has been used in the current session
+    mapping(address validatorAccount => uint256 count) public signatureCounters;
```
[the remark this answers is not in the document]
doubledup: Right, the `submitInitial` calls for the next session form part of that session's signature uses, not the current session's.

yrong:
@doubledup I've updated Readme with required steps to regenerate beefy test fixtures. Hope that's helpful.

contracts/src/BeefyClient.sol
```diff
@@ -390,49 +414,24 @@ contract BeefyClient {
...
+            return BASE_SIGNATURE_COUNT;
+        }
+
+        // log_2 rounded down is the index of the highest bit
+        uint256 logSignatureUse = Bits.highestBitSet(signatureUseCount);
+        // round up by adding 1 when there is more than 1 bit set in signatureUseCount
+        if (signatureUseCount & logSignatureUse != uint256(0)) {
+            ++logSignatureUse;
+        }
+        uint256 dynamicSignatures = 1 + 2 * logSignatureUse;
```
alistair-singh, line 432: This math produces signature counts that sometimes jump to a lower dynamic signature count. Is this expected? I made `minSignatures` external and ran the following test:
```js

    function testMinSignatures() public {
        // Test using signature counts 8 to 13
        assertEq(beefyClient.minSignatures(100, 8), 31);
        assertEq(beefyClient.minSignatures(100, 9), 33);
        assertEq(beefyClient.minSignatures(100, 10), 33);
        assertEq(beefyClient.minSignatures(100, 11), 33);
        assertEq(beefyClient.minSignatures(100, 12), 31); // At 12 it jumps back down to 31 again.
        assertEq(beefyClient.minSignatures(100, 13), 33);

        // Test using signature counts 70 to 75
        assertEq(beefyClient.minSignatures(100, 70), 39);
        assertEq(beefyClient.minSignatures(100, 71), 39);
        assertEq(beefyClient.minSignatures(100, 72), 37); // Again at 72
        assertEq(beefyClient.minSignatures(100, 73), 37); // And here at 73
        assertEq(beefyClient.minSignatures(100, 74), 39);
        assertEq(beefyClient.minSignatures(100, 75), 39);
    }
```
Either way, we need unit tests for this.

alistair-singh (commented):
Looks good so far. Left some comments.

contracts/src/BeefyClient.sol
```diff
@@ -243,8 +253,23 @@ contract BeefyClient {
...
-        tickets[createTicketID(msg.sender, commitmentHash)] =
-            Ticket(msg.sender, uint64(block.number), uint32(vset.length), 0, keccak256(abi.encodePacked(bitfield)));
+        // Increment the counter and store the previous value in the ticket
+        uint16 signatureCount = Counter.get(vset.counters, proof.index);
+        Counter.set(vset.counters, proof.index, signatureCount + 1);
+        if (commitment.validatorSetID == currentValidatorSet.id) {
+            currentValidatorSet.counters = vset.counters;
+        } else if (commitment.validatorSetID == nextValidatorSet.id) {
+            nextValidatorSet.counters = vset.counters;
+        }
```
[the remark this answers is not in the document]
yrong: https://github.com/Snowfork/snowbridge/pull/956/commits/3a4c5d540b68eae86fea0b805e4cc57f34d5a1e6

contracts/src/DeployScript.sol
```diff
@@ -33,10 +34,21 @@ contract DeployScript is Script {
...
+            uint128(beefyCheckpointRaw.readUint(".current.id")),
+            uint128(beefyCheckpointRaw.readUint(".current.length")),
+            beefyCheckpointRaw.readBytes32(".current.root"),
+            counters
+        );
+        BeefyClient.ValidatorSet memory next = BeefyClient.ValidatorSet(
+            uint128(beefyCheckpointRaw.readUint(".next.id")),
+            uint128(beefyCheckpointRaw.readUint(".next.length")),
+            beefyCheckpointRaw.readBytes32(".next.root"),
+            counters
```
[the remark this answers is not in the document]
yrong: Nice catch! Fixed in https://github.com/Snowfork/snowbridge/pull/956/commits/aafe25bf008826e94b301ddc1c3f410f8e5015a0

contracts/src/BeefyClient.sol
yrong: https://github.com/Snowfork/snowbridge/pull/956/commits/57bfb94054b4206794dc7937ccce80efa90106f3

alistair-singh: Just a quick check on the math without the round-up across the whole `uint16` range.
The table below is to be read as when the signature count for a single validator is within the range in column 1 then it will choose the corresponding dynamic signature choice in column 2. In the worst case of a validator signature being used `65535` times, then the number of dynamic signatures needed is `55`.

Sig Count Values|Dynamic Sigs Choice
-|-
0 | 24
1 | 25
2 - 3 | 27
4 - 7 | 29
8 - 15 | 31
16 - 31 | 33
32 - 63 | 35
64 - 127 | 37
128 - 255 | 39
256 - 511 | 41
512 - 1023 | 43
1024 - 2047 | 45
2048 - 4095 | 47
4096 - 8191 | 49
8192 - 16383 | 51
16384 - 32767 | 53
32768 - 65535 | 55

yrong: > M is the minimum number of signatures (24 in our code base).

My only concern is the gas cost.  Vaguely recall some benchmarks before shows gas cost doubled. So I'm wondering if **24**(compared with the previous 10) is absolutely required for the initial launch or we can make it configurable through a governance call?

alistair-singh: @vgeddes How did we land on 24 as the minimum number of signatures?

contracts/src/utils/Counter.sol
```diff
@@ -0,0 +1,24 @@
...
+        return uint16((self[element] >> (16 * inside)) & 0xFFFF);
+    }
+
+    function set(uint256[] storage self, uint256 index, uint16 value) internal {
+        uint256 element = index >> 4;
+        uint8 inside = uint8(index) & 0x0F;
+        uint256 zero = uint256(0xFFFF) << (16 * inside);
+        uint256 shiftedValue = uint256(value) << (16 * inside);
+        self[element] = self[element] & zero | shiftedValue;
+    }
```
yrong, line 23: Seems nuance that makes `value = get(index)` not always the same value as set by `set(index, value)`, Is that as expected?

alistair-singh: This shouldn't be the case. I have added more unit tests for counters:
https://github.com/Snowfork/snowbridge/pull/956/commits/c631af0355189f8414064e4d9a8f32fea317f302

What inputs did you use to produce the behavior? I can add it as a test case.

yrong:
```
function testCounterGetAndSetNotMatch() public {
        counters = Counter.createCounter(300);
        uint256 index = 0;
        uint16 value = 11;
        counters.set(index, value);
        uint16 new_value = counters.get(index);
        console.log("round1:index at %d set %d and get %d", index, value, new_value);
        assertEq(value, new_value);
        value = value + 1;
        counters.set(index, value);
        new_value = counters.get(index);
        console.log("round2:index at %d set %d and get %d", index, value, new_value);
        assertEq(value, new_value);
}
```

Test failed at round2 assert.

alistair-singh: Nice catch! Addressed in [0e0978d](https://github.com/Snowfork/snowbridge/pull/956/commits/0e0978d8bef1aac762889f22b5c8f1d53e2fa286).

contracts/src/BeefyClient.sol
yrong: Seems from https://hackmd.io/9OedC7icR5m-in_moUZ_WQ based on total DOT supply.

yrong: ### Before
| test/mocks/BeefyClientMock.sol:BeefyClientMock contract |                 |       |        |        |         |
|---------------------------------------------------------|-----------------|-------|--------|--------|---------|
| Deployment Cost                                         | Deployment Size |       |        |        |         |
| 2197297                                                 | 11304           |       |        |        |         |
| Function Name                                           | min             | avg   | median | max    | # calls |
| commitPrevRandao                                        | 763             | 17921 | 21097  | 21097  | 19      |
| createFinalBitfield                                     | 1134            | 9966  | 10597  | 10597  | 15      |
| createInitialBitfield                                   | 544             | 52080 | 52804  | 70348  | 48      |
| encodeCommitment_public                                 | 5556            | 5556  | 5556   | 5556   | 1       |
| initialize_public                                       | 2259            | 84250 | 92059  | 92059  | 23      |
| latestBeefyBlock                                        | 488             | 488   | 488    | 488    | 3       |
| minimumSignatureThreshold_public                        | 267             | 267   | 267    | 267    | 1       |
| setLatestBeefyBlock                                     | 20364           | 20364 | 20364  | 20364  | 1       |
| submitFinal                                             | 4829            | 40169 | 6704   | 112156 | 15      |
| submitInitial                                           | 14308           | 58987 | 61586  | 61853  | 20      |

### After 
| test/mocks/BeefyClientMock.sol:BeefyClientMock contract |                 |        |        |        |         |
|---------------------------------------------------------|-----------------|--------|--------|--------|---------|
| Deployment Cost                                         | Deployment Size |        |        |        |         |
| 2687425                                                 | 14246           |        |        |        |         |
| Function Name                                           | min             | avg    | median | max    | # calls |
| BASE_SIGNATURE_COUNT                                    | 269             | 269    | 269    | 269    | 1       |
| commitPrevRandao                                        | 763             | 17926  | 21097  | 21097  | 19      |
| createFinalBitfield                                     | 1153            | 23646  | 25159  | 26481  | 15      |
| createInitialBitfield                                   | 566             | 52131  | 52826  | 70370  | 50      |
| encodeCommitment_public                                 | 5578            | 5578   | 5578   | 5578   | 1       |
| initialize_public                                       | 7903            | 168210 | 183478 | 183478 | 23      |
| latestBeefyBlock                                        | 510             | 510    | 510    | 510    | 3       |
| minSignatures_public                                    | 247             | 2179   | 2225   | 2455   | 47      |
| setLatestBeefyBlock                                     | 20364           | 20364  | 20364  | 20364  | 1       |
| submitFinal                                             | 4851            | 81266  | 6726   | 239927 | 15      |
| submitInitial                                           | 39473           | 104226 | 108951 | 109218 | 20      |

contracts/src/DeployScript.sol
```diff
@@ -33,10 +34,22 @@ contract DeployScript is Script {
...
+            beefyCheckpointRaw.readBytes32(".current.root"),
+            currCounters
+        );
+        uint256[] memory nextCounters = Counter.createCounter(beefyCheckpointRaw.readUint(".next.length"));
+        BeefyClient.ValidatorSet memory next = BeefyClient.ValidatorSet(
+            uint128(beefyCheckpointRaw.readUint(".next.id")),
+            uint128(beefyCheckpointRaw.readUint(".next.length")),
+            beefyCheckpointRaw.readBytes32(".next.root"),
+            nextCounters
+        );
```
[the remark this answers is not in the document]
vgeddes: There is another issue in that we're now overloading the the usage of the `ValidatorSet` type for two purposes:
* Providing initial state to the BeefyClient (validatorSetLen, etc)
* Storing work state that is internal to the BeefyClient  (the counters)

I suggest refactoring and creating a new type `InitialValidatorSet` to accommodate the different concern.

contracts/src/BeefyClient.sol
```diff
@@ -390,49 +411,27 @@ contract BeefyClient {
...
+
+        // We need to get the value of log2 rounded up. To do this we can compare if signatureUseCount is
+        // greater than the signatureUseCount is larger than its most significant bit. If it is larger then
+        // we need to round up by incrementing the log2SignatureUse.
+        if (signatureUseCount > (1 << log2SignatureUse)) {
+            ++log2SignatureUse;
+        }
+        uint256 dynamicSignatures = 1 + 2 * log2SignatureUse;
+
+        return BASE_SIGNATURE_COUNT + dynamicSignatures;
```
vgeddes, line 434: I think this code would read better if we drop all the unnecessary intermediate variable names `log2Signatures`, `dynamicSignatures`, and just accumulate the mathematical operations into a `count` variable.

contracts/src/BeefyClient.sol
```diff
@@ -390,49 +411,27 @@ contract BeefyClient {
...
+        // log2 rounded down is the index of the highest bit
+        uint256 log2SignatureUse = Bits.highestBitSet(signatureUseCount);
+
+        // We need to get the value of log2 rounded up. To do this we can compare if signatureUseCount is
+        // greater than the signatureUseCount is larger than its most significant bit. If it is larger then
+        // we need to round up by incrementing the log2SignatureUse.
+        if (signatureUseCount > (1 << log2SignatureUse)) {
+            ++log2SignatureUse;
+        }
+        uint256 dynamicSignatures = 1 + 2 * log2SignatureUse;
```
vgeddes, line 432: what do the magic `1` and `2` constants refer to? Should at least add a comment

contracts/src/BeefyClient.sol
```diff
@@ -390,49 +411,27 @@ contract BeefyClient {
...
+
+        if (signatureUseCount == 0) {
+            return BASE_SIGNATURE_COUNT;
+        }
+
+        // log2 rounded down is the index of the highest bit
+        uint256 log2SignatureUse = Bits.highestBitSet(signatureUseCount);
+
+        // We need to get the value of log2 rounded up. To do this we can compare if signatureUseCount is
+        // greater than the signatureUseCount is larger than its most significant bit. If it is larger then
```
vgeddes, line 427: This comment looks like it needs to be reformatted. for example, there is an unnecessary newline after `then`.

contracts/src/BeefyClient.sol
```diff
@@ -321,6 +343,7 @@ contract BeefyClient {
             nextValidatorSet.id = leaf.nextAuthoritySetID;
```
vgeddes, line 343: Does ` currentValidatorSet = nextValidatorSet` actually copy the counters over too? Is that confirmed in unit tests?

As that is definitely needed.

vgeddes (commented):
Looks good, added some minor comments.

contracts/src/BeefyClient.sol
vgeddes: > vgeddes How did we land on 24 as the minimum number of signatures?

Hey, ah, I'm not sure. Was a minimal number decided upon in the meeting with W3F last week?

claravanstaden: I think so. Looks like `29`, although Alistair Stewart said `27` should be OK too. [Number of Signatures needed for Sampling Light client of BEEFY](https://hackmd.io/9OedC7icR5m-in_moUZ_WQ) cc @alistair-singh correct me if I'm wrong. 😄

contracts/src/utils/Counter.sol
```diff
@@ -0,0 +1,24 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2023 Snowfork <hello@snowfork.com>
+pragma solidity 0.8.20;
+
+library Counter {
```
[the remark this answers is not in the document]
vgeddes: Perhaps an example ASCII diagram of the layout would also help in making sense of all the bit operations being performed in this library. Just an idea.

contracts/src/utils/Counter.sol
```diff
@@ -0,0 +1,24 @@
...
+    function get(uint256[] storage self, uint256 index) internal view returns (uint16) {
+        uint256 element = index >> 4;
+        uint8 inside = uint8(index) & 0x0F;
+        return uint16((self[element] >> (16 * inside)) & 0xFFFF);
+    }
+
+    function set(uint256[] storage self, uint256 index, uint16 value) internal {
+        uint256 element = index >> 4;
+        uint8 inside = uint8(index) & 0x0F;
+        uint256 zero = ~(uint256(0xFFFF) << (16 * inside));
```
vgeddes, line 20: Do you have an intuitive sense what these operations and magic constants mean? If not, lets try make sense of it, and then add documentation/comments.

vgeddes (commented):
One more thing @alistair-singh, we also need to update the documentation at https://docs.snowbridge.network/architecture/verification/polkadot to describe the design of the dynamic signature sampling algorithm.

Can either edit the markdown in this PR, or via the online gitbook editor. Either way let's make sure this task is tracked.

yrong:
Another change we may need is for relay logic to not always pick from the first validator
https://github.com/Snowfork/snowbridge/blob/ba4e80140f3b39921f9c32c3174a069bb53e063d/relayer/relays/beefy/ethereum-writer.go#L156-L158
Instead, it makes more sense to pick from the validator set randomly.

contracts/src/BeefyClient.sol
```diff
@@ -153,6 +159,11 @@ contract BeefyClient {
      */
     bytes2 public constant MMR_ROOT_ID = bytes2("mh");
 
+    /**
+     * @dev Minimum number of signatures required to validate a new commitment.
+     */
+    uint256 public constant BASE_SIGNATURE_COUNT = 24;
```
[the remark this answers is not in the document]
alistair-singh: Resolved in https://github.com/Snowfork/snowbridge/pull/956/commits/6a5b02cb9de4f76b527443068e51a79736a9fdde

contracts/src/utils/Counter.sol
```diff
@@ -0,0 +1,24 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2023 Snowfork <hello@snowfork.com>
+pragma solidity 0.8.20;
+
+library Counter {
+    function createCounter(uint256 length) internal pure returns (uint256[] memory counter) {
+        uint256 counterLength = length / 16 + (length % 16 == 0 ? 0 : 1);
+        counter = new uint256[](counterLength);
```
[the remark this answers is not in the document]
alistair-singh: Resolved in https://github.com/Snowfork/snowbridge/pull/956/commits/dc5f1d2d416e80572aedaa9d054ee9ba52feb123

contracts/src/BeefyClient.sol
alistair-singh: Resolved in https://github.com/Snowfork/snowbridge/pull/956/commits/dc5f1d2d416e80572aedaa9d054ee9ba52feb123

contracts/src/BeefyClient.sol
alistair-singh: I have refactored the counters into separate variables. This [test](https://github.com/Snowfork/snowbridge/pull/956/files#diff-2984b4351f0684d24f6eb93d006264cb1442612c4f6a9869de336506cd85f01aR25-R32) tests the overwriting of counters with new counters making sure previous values are zeroed.

contracts/src/utils/Counter.sol
```diff
@@ -0,0 +1,24 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2023 Snowfork <hello@snowfork.com>
+pragma solidity 0.8.20;
+
+library Counter {
+    function createCounter(uint256 length) internal pure returns (uint256[] memory counter) {
+        uint256 counterLength = length / 16 + (length % 16 == 0 ? 0 : 1);
```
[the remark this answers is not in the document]
alistair-singh: Test added in https://github.com/Snowfork/snowbridge/pull/956/commits/52ae45b0f7909379cd95168e55b36c68aded4e74

alistair-singh:
> Instead, it makes more sense to pick from the validator set randomly.

Addressed in https://github.com/Snowfork/snowbridge/pull/956/commits/249447164a84b9916382797da68f08bb37c00675

contracts/src/BeefyClient.sol
alistair-singh: Final version of the signatureSampling is added in https://github.com/Snowfork/snowbridge/pull/956/commits/572ee397ac140d56a1d7eee2bb6dfeb93163450a. Need to regenerate test fixtures.

contracts/src/BeefyClient.sol
alistair-singh: addressed in https://github.com/Snowfork/snowbridge/pull/956/commits/572ee397ac140d56a1d7eee2bb6dfeb93163450a

contracts/src/BeefyClient.sol
alistair-singh: Addressed in https://github.com/Snowfork/snowbridge/pull/956/commits/5eda996bc29e88d237f19b78b93f2d2a168379fd

contracts/src/BeefyClient.sol
alistair-singh: Dropped minimum to 16 based on this comment about slash rate in [element](https://matrix.to/#/!xvbzTAMdhwhpLCmmLe:parity.io/$S8RBXphyc_2iqg0By85l9vKleW4Xpx-li75RhyQVGaI?via=parity.io&via=matrix.org&via=web3.foundation).
In change [e5c161c](https://github.com/Snowfork/snowbridge/pull/956/commits/e5c161c97fde55aac2e3bf8ace3a717776adf90a)

contracts/src/utils/Counter.sol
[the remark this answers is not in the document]
alistair-singh: Addressed in https://github.com/Snowfork/snowbridge/pull/956/commits/af16c9598bd6914ec1e93f7454e204a69ad49594.

contracts/src/utils/Counter.sol
alistair-singh: Addressed in https://github.com/Snowfork/snowbridge/pull/956/commits/af16c9598bd6914ec1e93f7454e204a69ad49594

yrong (approved):
Awesome!

contracts/src/BeefyClient.sol
```diff
@@ -390,49 +417,57 @@ contract BeefyClient {
...
+
+        // We must substrate minimumSignatures from the number of validators or we might end up
+        // requiring more signatures than there are validators.
+        uint256 extraValidators = validatorSetLen - minimumSignatureSamples;
+        // Using highest bit set to yield floor(log2(x))
+        uint256 validatorsLog2 = Bits.highestBitSet(extraValidators);
+        // We require ceil(log2(x)) so we round up if there were any extra bits set.
+        if (extraValidators > (1 << validatorsLog2)) {
+            validatorsLog2++;
+        }
```
vgeddes, line 444: Lets factor this out into a `ceil` function in a `Math.sol` utility library

contracts/src/BeefyClient.sol
```diff
@@ -390,49 +417,57 @@ contract BeefyClient {
...
+        // https://hackmd.io/wsVcL0tZQA-Ks3b5KJilFQ?view#Solution-2-Signature-Checks-dynamically-depend-on-the-No-of-initial-Claims-per-session
+        // It must be harder for a mallicious relayer to spam submitInitial to bias the RANDAO.
+        // If we detect that a signature is used many times (spam), we increase the number of signature samples required on submitFinal.
+
+        // Using highest bit set to yield floor(log2(x))
+        uint256 log2SignatureUse = Bits.highestBitSet(signatureUseCount);
+        // We require ceil(log2(x)) so we round up if there were any extra bits set.
+        if (signatureUseCount > (1 << log2SignatureUse)) {
+            ++log2SignatureUse;
+        }
```
vgeddes, line 462: Ditto, please factor out `ceil` into a utility library.

contracts/src/BeefyClient.sol
```diff
@@ -390,49 +417,57 @@ contract BeefyClient {
...
+        // Based on formula provided here: https://hackmd.io/9OedC7icR5m-in_moUZ_WQ
+        samples += 1 + 2 * log2SignatureUse;
+
+        // Make sure that samples are bounded by the validatorSetLength
+        if (samples > validatorSetLen) {
+            samples = validatorSetLen;
+        }
```
vgeddes, line 469: Also known as `Math.min(x, y)`.

contracts/src/BeefyClient.sol
```diff
@@ -390,49 +417,57 @@ contract BeefyClient {
...
-     *
-     * In any case, an order of magnitude increase in validator set sizes will likely require a re-architecture
-     * of Polkadot that would make this contract obsolete well before the assumption becomes a problem.
-     *
-     * Constants generated with the help of scripts/minsigs.py
-     */
-    function minimumSignatureThreshold(uint256 validatorSetLen) internal pure returns (uint256) {
-        if (validatorSetLen <= 10) {
+    // Calculates the number of signature samples required by validator set length and the number of times a validator
+    // signature was used.
```
vgeddes, line 422: Should clarify what "number of times a validator signature was used" actually means in relation to `signatureUseCount`. Which signature is being used multiple times where?

contracts/src/BeefyClient.sol
```diff
@@ -142,6 +147,9 @@ contract BeefyClient {
     ValidatorSet public currentValidatorSet;
     ValidatorSet public nextValidatorSet;
 
+    uint256[] currentValidatorSetCounters;
+    uint256[] nextValidatorSetCounters;
```
vgeddes, line 151: Needs doc comments please.

contracts/src/BeefyClient.sol
```diff
@@ -243,8 +265,14 @@ contract BeefyClient {
...
-        tickets[createTicketID(msg.sender, commitmentHash)] =
-            Ticket(msg.sender, uint64(block.number), uint32(vset.length), 0, keccak256(abi.encodePacked(bitfield)));
+        tickets[createTicketID(msg.sender, commitmentHash)] = Ticket(
+            msg.sender,
+            uint64(block.number),
+            uint32(vset.length),
+            signatureSamples(vset.length, signatureCount),
+            0,
+            keccak256(abi.encodePacked(bitfield))
+        );
```
vgeddes, line 275: I think there is an alternative syntax for this initializer which is more readable as it includes field names.

```solidity
Ticket ticket = Ticket({foo=1, foo=2})
```

contracts/src/BeefyClient.sol
```diff
@@ -390,49 +418,57 @@ contract BeefyClient {
...
         }
+
+        // Start with the minimum number of signatures.
+        uint256 samples = minimumSignatureSamples;
+
+        // We must substrate minimumSignatures from the number of validators or we might end up
+        // requiring more signatures than there are validators.
+        uint256 extraValidators = validatorSetLen - minimumSignatureSamples;
+        // Using highest bit set to yield floor(log2(x))
+        uint256 validatorsLog2 = Bits.highestBitSet(extraValidators);
```
vgeddes, line 440: Please extract `log2` into a Math utility library

contracts/src/BeefyClient.sol
```diff
@@ -390,49 +418,57 @@ contract BeefyClient {
...
+            return samples;
+        }
+
+        // To address the concurrency issue specified in the link below:
+        // https://hackmd.io/wsVcL0tZQA-Ks3b5KJilFQ?view#Solution-2-Signature-Checks-dynamically-depend-on-the-No-of-initial-Claims-per-session
+        // It must be harder for a mallicious relayer to spam submitInitial to bias the RANDAO.
+        // If we detect that a signature is used many times (spam), we increase the number of signature samples required on submitFinal.
+
+        // Using highest bit set to yield floor(log2(x))
+        uint256 log2SignatureUse = Bits.highestBitSet(signatureUseCount);
```
vgeddes, line 458: Ditto, extract log2 into a math utility library.

vgeddes, line 458: The variable names `log2SignatureUse` and `validatorsLog2` are inconsistent. rather pick one style, and preferably drop the `2`.

contracts/src/BeefyClient.sol
alistair-singh: Addressed in https://github.com/Snowfork/snowbridge/pull/956/commits/51746e56272d3a94dab96546e0422abbd0b33826

contracts/src/BeefyClient.sol
alistair-singh: These variables are removed in https://github.com/Snowfork/snowbridge/pull/956/commits/2a9d732f2d84c0585ae3dee3ef191206b9eb6fb0 as the Math util functions means we do not have to keep them around.

alistair-singh: Addressed in https://github.com/Snowfork/snowbridge/pull/956/commits/2a9d732f2d84c0585ae3dee3ef191206b9eb6fb0.

contracts/src/BeefyClient.sol
alistair-singh: Addressed in https://github.com/Snowfork/snowbridge/pull/956/commits/2a9d732f2d84c0585ae3dee3ef191206b9eb6fb0.

contracts/src/BeefyClient.sol
alistair-singh: Addressed in https://github.com/Snowfork/snowbridge/pull/956/commits/2a9d732f2d84c0585ae3dee3ef191206b9eb6fb0.

contracts/src/BeefyClient.sol
alistair-singh: Addressed in https://github.com/Snowfork/snowbridge/pull/956/commits/2a9d732f2d84c0585ae3dee3ef191206b9eb6fb0.

contracts/src/BeefyClient.sol
alistair-singh: Addressed in https://github.com/Snowfork/snowbridge/pull/956/commits/4fb659255abf429d2157f413c4e00bf3dbb9b750

contracts/src/BeefyClient.sol
alistair-singh: Addressed in https://github.com/Snowfork/snowbridge/pull/956/commits/17d468f0427d9e040b1d8438d9202045834b6fe5.

contracts/src/BeefyClient.sol
alistair-singh: Addressed in https://github.com/Snowfork/snowbridge/pull/956/commits/17d468f0427d9e040b1d8438d9202045834b6fe5.

contracts/src/BeefyClient.sol
```diff
@@ -390,49 +419,35 @@ contract BeefyClient {
...
-     * so we are well within those limits.
-     *
-     * In any case, an order of magnitude increase in validator set sizes will likely require a re-architecture
-     * of Polkadot that would make this contract obsolete well before the assumption becomes a problem.
-     *
-     * Constants generated with the help of scripts/minsigs.py
-     */
-    function minimumSignatureThreshold(uint256 validatorSetLen) internal pure returns (uint256) {
-        if (validatorSetLen <= 10) {
+    // Calculates the number of signature samples required by validator set length and the number of times a validator
```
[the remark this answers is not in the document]
alistair-singh: Addressed in https://github.com/Snowfork/snowbridge/pull/956/commits/62f65cb9523a45b1e098ba183cd8dc357ae1cd06.

contracts/src/BeefyClient.sol
```diff
@@ -390,49 +419,35 @@ contract BeefyClient {
...
+    // signature was with a call to `submitInitial`.
```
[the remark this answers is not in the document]
alistair-singh: Addressed in https://github.com/Snowfork/snowbridge/pull/956/commits/62f65cb9523a45b1e098ba183cd8dc357ae1cd06.

contracts/src/BeefyClient.sol
```diff
@@ -390,49 +419,35 @@ contract BeefyClient {
...
+        // Start with the minimum number of signatures.
+        uint256 samples = minimumSignatureSamples;
+
+        // We must substrate minimumSignatures from the number of validators or we might end up
+        // requiring more signatures than there are validators.
+        samples += Math.ceilingOfLog2(validatorSetLen - minimumSignatureSamples);
+
+        // To address the concurrency issue specified in the link below:
+        // https://hackmd.io/wsVcL0tZQA-Ks3b5KJilFQ?view#Solution-2-Signature-Checks-dynamically-depend-on-the-No-of-initial-Claims-per-session
+        // It must be harder for a mallicious relayer to spam submitInitial to bias the RANDAO.
```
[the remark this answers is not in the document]
alistair-singh: Addressed in https://github.com/Snowfork/snowbridge/pull/956/commits/62f65cb9523a45b1e098ba183cd8dc357ae1cd06.

contracts/src/utils/Counter.sol
```diff
@@ -0,0 +1,70 @@
...
+        uint256 element = index >> 4;
+        // Mask out the first 4 bytes of the logical index to give us the bit-index.
+        uint8 inside = uint8(index) & 0x0F;
+        // find the element in the array, shift until its bit index and mask to only take the first 16 bits.
+        return uint16((self[element] >> (16 * inside)) & 0xFFFF);
+    }
+
+    /// @dev Sets the counter at the logical index.
+    /// @param self The counter array.
+    /// @param index The logic index.
```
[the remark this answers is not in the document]
alistair-singh: Addressed in https://github.com/Snowfork/snowbridge/pull/956/commits/62f65cb9523a45b1e098ba183cd8dc357ae1cd06.

contracts/test/BeefyClient.t.sol
```diff
@@ -493,6 +659,78 @@ contract BeefyClientTest is Test {
...
+
+        assertEq(29, beefyClient.signatureSamples_public(30, 10), "dynamicSignatures incorrect.");
+        assertEq(30, beefyClient.signatureSamples_public(30, 17), "dynamicSignatures incorrect.");
+    }
+
+    function testSignatureSamplingRanges() public {
+        assertEq(25, beefyClient.signatureSamples_public(setSize, 0), "dynamicSignatures incorrect.");
+        assertEq(26, beefyClient.signatureSamples_public(setSize, 1), "dynamicSignatures incorrect.");
+        assertEq(28, beefyClient.signatureSamples_public(setSize, 2), "dynamicSignatures incorrect.");
+        assertEq(30, beefyClient.signatureSamples_public(setSize, 3), "dynamicSignatures incorrect.");
```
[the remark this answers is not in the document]
alistair-singh: Addressed in https://github.com/Snowfork/snowbridge/pull/956/commits/7d3d49b0d034b6221e320865401dea53b5ec2f25.

contracts/test/BeefyClient.t.sol
```diff
@@ -493,6 +659,78 @@ contract BeefyClientTest is Test {
...
+        regenerateBitField(bitFieldFile3SignatureCount, samples);
+    }
+
+    function testSignatureSampleWithUseMajorityIfVsetIsSmallerThanMinSignatures() public {
+        assertEq(7, beefyClient.signatureSamples_public(9, 0), "dynamicSignatures incorrect.");
+        assertEq(11, beefyClient.signatureSamples_public(16, 0), "dynamicSignatures incorrect.");
+        assertEq(16, beefyClient.signatureSamples_public(17, 0), "dynamicSignatures incorrect.");
+    }
+
+    function testSignatureSampleNeverYieldsASampleGreaterThanTheValidatorSetLength() public {
```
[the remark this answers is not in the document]
alistair-singh: Addressed in https://github.com/Snowfork/snowbridge/pull/956/commits/7d3d49b0d034b6221e320865401dea53b5ec2f25.

contracts/test/BeefyClient.t.sol
```diff
@@ -493,6 +659,78 @@ contract BeefyClientTest is Test {
...
+        assertEq(18, beefyClient.signatureSamples_public(19, 0), "dynamicSignatures incorrect.");
+        assertEq(18, beefyClient.signatureSamples_public(20, 0), "dynamicSignatures incorrect.");
+        assertEq(19, beefyClient.signatureSamples_public(21, 0), "dynamicSignatures incorrect.");
+        assertEq(20, beefyClient.signatureSamples_public(30, 0), "dynamicSignatures incorrect.");
+
+        assertEq(29, beefyClient.signatureSamples_public(30, 10), "dynamicSignatures incorrect.");
+        assertEq(30, beefyClient.signatureSamples_public(30, 17), "dynamicSignatures incorrect.");
+    }
+
+    function testSignatureSamplingRanges() public {
```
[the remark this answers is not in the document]
alistair-singh: Addressed in https://github.com/Snowfork/snowbridge/pull/956/commits/7d3d49b0d034b6221e320865401dea53b5ec2f25.

contracts/src/utils/Math.sol
```diff
@@ -0,0 +1,41 @@
...
+     * @dev Returns the ceiling of log2 of a number using bitwise arithmetic.
+     */
+    function ceilingOfLog2(uint256 x) internal pure returns (uint256) {
+        uint256 result = floorOfLog2(x);
+        // We round up if there were any extra bits set.
+        if (x > (1 << result)) {
+            result++;
+        }
+        return result;
+    }
```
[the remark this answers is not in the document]
alistair-singh: Resolved in https://github.com/Snowfork/snowbridge/pull/956/commits/f532b7230322bed1ad168a4893f6c6df1eebc8f4.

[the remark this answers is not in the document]
alistair-singh: Tests for log2 added in https://github.com/Snowfork/snowbridge/pull/956/commits/0e7716dbb619bc2b69986a0147ae8b335de86df7

contracts/src/BeefyClient.sol
```diff
@@ -391,48 +420,37 @@ contract BeefyClient {
...
-     * In any case, an order of magnitude increase in validator set sizes will likely require a re-architecture
-     * of Polkadot that would make this contract obsolete well before the assumption becomes a problem.
-     *
-     * Constants generated with the help of scripts/minsigs.py
+     * See https://hackmd.io/9OedC7icR5m-in_moUZ_WQ for full analysis.
      */
-    function minimumSignatureThreshold(uint256 validatorSetLen) internal pure returns (uint256) {
-        if (validatorSetLen <= 10) {
+    function signatureSamples(uint256 validatorSetLen, uint256 signatureUseCount) internal view returns (uint256) {
+        // There are less validators than the minimum signatures so validate 2/3 majority.
```
vgeddes, line 432: This case was originally a special case for supporting local testnets and rococo using the old algorithm, so relayers would not have to provide all signatures.

However, its important that we test the dynamic signature algorithm on Rococo, and even locally if possible. This case effectively disables the algorithm.

Lets remove the special case

contracts/src/BeefyClient.sol
```diff
@@ -212,9 +230,14 @@ contract BeefyClient {
         external
     {
         ValidatorSet memory vset;
+        uint16 signatureCount;
         if (commitment.validatorSetID == currentValidatorSet.id) {
+            signatureCount = currentValidatorSetCounters.get(proof.index);
+            currentValidatorSetCounters.set(proof.index, signatureCount + 1);
```
vgeddes, line 236: I kinda think we should use saturating arithmetic here. Otherwise the submission will fail on overflow. Obviously this is unlikely to happen to act in practice, but we should be paranoid just in case.

```suggestion
            currentValidatorSetCounters.set(proof.index, signatureCount.saturatingAdd(1));
```

Will need to add a `saturatingAdd(uint16 value)` helper to `Math.sol`.

contracts/src/BeefyClient.sol
```diff
@@ -91,12 +94,14 @@ contract BeefyClient {
...
      * @param blockNumber the block number for this commitment
      * @param validatorSetLen the length of the validator set for this commitment
+     * @param signatureCountRequired the number of signatures required
      * @param bitfield a bitfield signalling which validators they claim have signed
      */
     struct Ticket {
         address account;
         uint64 blockNumber;
         uint32 validatorSetLen;
+        uint256 signatureCountRequired;
```
vgeddes, line 104: This field name sounds like its a boolean rather than a number.

```suggestion
        uint256 numRequiredSignatures;
```

contracts/src/BeefyClient.sol
```diff
@@ -391,48 +420,37 @@ contract BeefyClient {
...
-     *
-     * In any case, an order of magnitude increase in validator set sizes will likely require a re-architecture
-     * of Polkadot that would make this contract obsolete well before the assumption becomes a problem.
-     *
-     * Constants generated with the help of scripts/minsigs.py
+     * See https://hackmd.io/9OedC7icR5m-in_moUZ_WQ for full analysis.
      */
-    function minimumSignatureThreshold(uint256 validatorSetLen) internal pure returns (uint256) {
-        if (validatorSetLen <= 10) {
+    function signatureSamples(uint256 validatorSetLen, uint256 signatureUseCount) internal view returns (uint256) {
```
vgeddes, line 431: I think we should be clearer in our terminology. "Sampling" is the process of selecting the individuals in a subset. This function does not do that. Rather it calculates the size of the subset. So lets rename the function accordingly:

```suggestion
    function computeNumRequiredSignatures(uint256 validatorSetLen, uint256 signatureUsageCount) internal view returns (uint256) {
```

contracts/src/BeefyClient.sol
```diff
@@ -391,48 +420,37 @@ contract BeefyClient {
...
-        } else if (validatorSetLen < 10923) {
-            return 15;
-        } else if (validatorSetLen < 21846) {
-            return 16;
-        } else {
-            return 17;
         }
+
+        // Start with the minimum number of signatures.
+        uint256 samples = minimumSignatureSamples;
```
vgeddes, line 438:
```suggestion
        uint256 numRequiredSignatures = minNumRequiredSignatures;
```

contracts/src/BeefyClient.sol
```diff
@@ -391,48 +420,37 @@ contract BeefyClient {
...
+
+        // We must substrate minimumSignatures from the number of validators or we might end up
+        // requiring more signatures than there are validators.
+        samples += Math.log2(validatorSetLen - minimumSignatureSamples, Math.Rounding.Ceil);
```
vgeddes, line 442:
```suggestion
        numRequiredSignatures += Math.log2(validatorSetLen - minNumRequiredSignatures, Math.Rounding.Ceil);
```

contracts/src/BeefyClient.sol
```diff
@@ -391,48 +420,37 @@ contract BeefyClient {
...
+
+        // To address the concurrency issue specified in the link below:
+        // https://hackmd.io/wsVcL0tZQA-Ks3b5KJilFQ?view#Solution-2-Signature-Checks-dynamically-depend-on-the-No-of-initial-Claims-per-session
+        // It must be harder for a malicious relayer to spam submitInitial to bias the RANDAO.
+        // If we detect that a signature is used many times (spam), we increase the number of signature samples required on submitFinal.
+        if (signatureUseCount > 0) {
```
vgeddes, line 448: UsageCount sounds better than UseCount

```suggestion
        if (signatureUsageCount > 0) {
```

contracts/src/BeefyClient.sol
```diff
@@ -447,7 +465,7 @@ contract BeefyClient {
     ) internal view {
         Ticket storage ticket = tickets[ticketID];
         // Verify that enough signature proofs have been supplied
-        uint256 signatureCount = minimumSignatureThreshold(vset.length);
+        uint256 signatureCount = ticket.signatureCountRequired;
```
vgeddes, line 468:
```suggestion
        uint256 numRequiredSignatures = ticket.signatureCountRequired;
```

contracts/.envrc-example
```diff
@@ -5,6 +5,7 @@ source_up_if_exists
 ## BeefyClient
 export RANDAO_COMMIT_DELAY=3
 export RANDAO_COMMIT_EXP=8
+export BEEFY_MINIMUM_SIGNATURE_SAMPLES=16
```
vgeddes, line 8:
```suggestion
export MINIMUM_REQUIRED_SIGNATURES=16
```

alistair-singh: Resolved in https://github.com/Snowfork/snowbridge/pull/956/commits/88f2b0749ba0bfde1a78ba2e9ff5e58e77b73207

contracts/src/BeefyClient.sol
alistair-singh: Addressed in https://github.com/Snowfork/snowbridge/pull/956/commits/46250d8ff06a500a194a2c57acd59397e343af46

contracts/src/BeefyClient.sol
alistair-singh: Addressed in https://github.com/Snowfork/snowbridge/pull/956/commits/46250d8ff06a500a194a2c57acd59397e343af46

contracts/src/BeefyClient.sol
alistair-singh: Addressed in https://github.com/Snowfork/snowbridge/pull/956/commits/46250d8ff06a500a194a2c57acd59397e343af46

contracts/src/BeefyClient.sol
alistair-singh: Addressed in https://github.com/Snowfork/snowbridge/pull/956/commits/46250d8ff06a500a194a2c57acd59397e343af46

contracts/src/BeefyClient.sol
alistair-singh: Addressed in https://github.com/Snowfork/snowbridge/pull/956/commits/46250d8ff06a500a194a2c57acd59397e343af46

contracts/src/BeefyClient.sol
alistair-singh: Addressed in https://github.com/Snowfork/snowbridge/pull/956/commits/46250d8ff06a500a194a2c57acd59397e343af46

contracts/src/BeefyClient.sol
alistair-singh: Addressed in https://github.com/Snowfork/snowbridge/pull/956/commits/b9b7127027e342adde23c43af16d46dea6d92db2

contracts/src/BeefyClient.sol
alistair-singh: Removed this case in https://github.com/Snowfork/snowbridge/pull/956/commits/54da625f2e636fa34bd4b0563969756028589f79

contracts/src/BeefyClient.sol
```diff
@@ -455,7 +455,9 @@ contract BeefyClient {
             numRequiredSignatures += 1 + 2 * Math.log2(signatureUsageCount, Math.Rounding.Ceil);
         }
 
-        return Math.min(numRequiredSignatures, validatorSetLen);
+        // Never require more signatures than 2/3 majority.
+        uint256 validatorSetMajority = validatorSetLen - (validatorSetLen - 1) / 3;
+        return Math.min(numRequiredSignatures, validatorSetMajority);
```
claravanstaden, line 463: Are there unit tests that test these conditions?

alistair-singh: This cases:
https://github.com/Snowfork/snowbridge/pull/956/commits/90923476fb11fe7016dea487b4123d06331bc171#diff-87ab70972cd5c5d114ecf33d954f8a1880f72747fbbebb53db2141bd93203737R686-R688

But I have added a few more to the same test.

contracts/src/BeefyClient.sol
```diff
@@ -391,48 +425,42 @@ contract BeefyClient {
...
+        // https://hackmd.io/wsVcL0tZQA-Ks3b5KJilFQ?view#Solution-2-Signature-Checks-dynamically-depend-on-the-No-of-initial-Claims-per-session
+        // It must be harder for a malicious relayer to spam submitInitial to bias the RANDAO.
+        // If we detect that a signature is used many times (spam), we increase the number of signature samples required on submitFinal.
+        if (signatureUsageCount > 0) {
+            // Based on formula provided here: https://hackmd.io/9OedC7icR5m-in_moUZ_WQ
+            numRequiredSignatures += 1 + 2 * Math.log2(signatureUsageCount, Math.Rounding.Ceil);
+        }
+
+        // Never require more signatures than 2/3 majority.
+        uint256 validatorSetMajority = validatorSetLen - (validatorSetLen - 1) / 3;
```
vgeddes, line 462: I believe the correct term is [quorum](https://en.wikipedia.org/wiki/Quorum_(distributed_computing)) instead of `majority`.

```suggestion
        uint256 validatorSetQuorum = validatorSetLen - (validatorSetLen - 1) / 3;
```

Also can we add unit test to ensure that `validatorSetLen - (validatorSetLen - 1) / 3` is always at least 2/3 of `validatorSetLen`. I don't think I tested that math previously.

vgeddes: To clarify, I was wanting a test that covers `validatorSetLen - (validatorSetLen - 1) / 3` on its own.

Also that expression needs to be factored out into its function, as it is duplicated in `submitInitial`, where there is a check that the relayer provided a claims bitfield with at least 2/3 bits set.

contracts/src/BeefyClient.sol
```diff
@@ -212,9 +233,14 @@ contract BeefyClient {
         external
     {
         ValidatorSet memory vset;
+        uint16 numRequiredSignatures;
```
vgeddes, line 236: Looks we renamed this incorrectly.
```suggestion
        uint16 signatureUsageCount;
```

contracts/src/BeefyClient.sol
```diff
@@ -212,9 +233,14 @@ contract BeefyClient {
...
         if (commitment.validatorSetID == currentValidatorSet.id) {
+            numRequiredSignatures = currentValidatorSetCounters.get(proof.index);
```
vgeddes, line 238: ditto

```suggestion
            signatureUsageCount = currentValidatorSetCounters.get(proof.index);
```

contracts/src/BeefyClient.sol
```diff
@@ -212,9 +233,14 @@ contract BeefyClient {
...
+            currentValidatorSetCounters.set(proof.index, numRequiredSignatures.saturatingAdd(1));
             vset = currentValidatorSet;
         } else if (commitment.validatorSetID == nextValidatorSet.id) {
+            numRequiredSignatures = nextValidatorSetCounters.get(proof.index);
```
vgeddes, line 242:
```suggestion
            signatureUsageCount = nextValidatorSetCounters.get(proof.index);
```

contracts/src/BeefyClient.sol
```diff
@@ -243,8 +269,16 @@ contract BeefyClient {
...
         }
 
-        tickets[createTicketID(msg.sender, commitmentHash)] =
-            Ticket(msg.sender, uint64(block.number), uint32(vset.length), 0, keccak256(abi.encodePacked(bitfield)));
+        tickets[createTicketID(msg.sender, commitmentHash)] = Ticket({
+            account: msg.sender,
+            blockNumber: uint64(block.number),
+            validatorSetLen: uint32(vset.length),
+            numRequiredSignatures: computeNumRequiredSignatures(
+                vset.length, numRequiredSignatures, minNumRequiredSignatures
```
vgeddes, line 277:
```suggestion
                vset.length, signatureUsageCount, minNumRequiredSignatures
```

contracts/src/BeefyClient.sol
```diff
@@ -391,48 +425,42 @@ contract BeefyClient {
...
-            return 13;
-        } else if (validatorSetLen < 5462) {
-            return 14;
-        } else if (validatorSetLen < 10923) {
-            return 15;
-        } else if (validatorSetLen < 21846) {
-            return 16;
-        } else {
-            return 17;
+    function computeNumRequiredSignatures(
```
vgeddes, line 437: I think this function is an absolutely excellent candidate for [fuzz testing](https://book.getfoundry.sh/forge/fuzz-testing) using foundry, to make sure that for all inputs it doesn't error out or produce undefined results.

It will also help us determine whether we need to add checks to bound the inputs in some way.

contracts/src/BeefyClient.sol
```diff
@@ -91,12 +97,14 @@ contract BeefyClient {
...
      * @param blockNumber the block number for this commitment
      * @param validatorSetLen the length of the validator set for this commitment
+     * @param numRequiredSignatures the number of signatures required
      * @param bitfield a bitfield signalling which validators they claim have signed
      */
     struct Ticket {
         address account;
         uint64 blockNumber;
         uint32 validatorSetLen;
+        uint256 numRequiredSignatures;
```
vgeddes, line 107: As this ticket struct is persisted in storage, we should try and take care to pack the individual fields in as few bits as possible. This optimizes gas costs when writing and reading from storage: https://medium.com/coinmonks/gas-optimization-in-solidity-part-i-variables-9d5775e43dde

The `account`, `blockNumber`, and `validatorSetLen` fields all fit into a single 32-byte slot. 

However, we can optimize further by using these types:
```
uint16 validatorSetLen;
uint16 numRequiredSignatures;
``` 

Now the first four fields all fit in a 32-byte slot.

Since there are currently only 1000 validators on Kusama, I don't see the number of validators increasing by 65x without a major re-architecture of Polkadot consensus. And we can always deploy a new BeefyClient via governance if that happens.

vgeddes: Actually it turns out that `ticket.account` isn't actually being used anywhere (unless I'm mistaken), so we can remove that field. Then we can leave `validatorSetLen` alone, and the only optimization we need to make is `uint32 numRequiredSignatures`.

contracts/src/BeefyClient.sol
alistair-singh: Addressed in https://github.com/Snowfork/snowbridge/pull/956/commits/3da2bef5c6b3b13b02a5f4cd45a63e375c8c39e9

contracts/src/BeefyClient.sol
alistair-singh: Addressed in https://github.com/Snowfork/snowbridge/pull/956/commits/3da2bef5c6b3b13b02a5f4cd45a63e375c8c39e9

contracts/src/BeefyClient.sol
alistair-singh: Addressed in https://github.com/Snowfork/snowbridge/pull/956/commits/3da2bef5c6b3b13b02a5f4cd45a63e375c8c39e9

contracts/src/BeefyClient.sol
alistair-singh: Addressed in https://github.com/Snowfork/snowbridge/pull/956/commits/3da2bef5c6b3b13b02a5f4cd45a63e375c8c39e9

contracts/src/BeefyClient.sol
alistair-singh: Addressed in https://github.com/Snowfork/snowbridge/pull/956/commits/d8144f3ca33f55bce887b03f0aea30ed736a5d8e

contracts/src/BeefyClient.sol
alistair-singh: Switched to using fuzz tests for `computeQuorum`, `computeNumRequiredSignatures` and `utils/Math.sol` functions.

contracts/src/BeefyClient.sol
alistair-singh: Addressed in https://github.com/Snowfork/snowbridge/pull/956/commits/5488978e8d91994bf7b027a5f82d4a41f30ff6c0.

vgeddes: Hey Alistair, its dangerous to just cast the result of `computeNumSignatures` to uint32, as it could result in undefined output.

While I know that practically the numSignatures value will never ever require more than 32 bits, auditors are still going to point this out.

Instead, I would add a `Math.min(type(uint32).max, ...)` as the last expression in `computeNumSignatures`, to clamp the maximum value to `type(uint32).max`.

Then the cast is safe.

vgeddes (approved):
Lets merge it!

PR #960, opened 2023-09-07
Allow Polkadot governance to withdraw funds from any agent

vgeddes:
This is necessary so that the sovereign accounts / agents of system parachains can be rebalanced.

The AssetHub parachain will have an agent on Ethereum which collects fees from users. In order for AssetHub parachain to withdraw these funds and rebalance, it needs to call the `transfer_native_from_agent` extrinsic on BridgeHub using XCM.

However there is no code on AssetHub to initiate this XCM call, which leaves us with a problem.

The solution is to add `force_transfer_native_from_agent` which allows the governance of the Polkadot relay chain to withdraw funds from any funds from any agent.

### DIFF
--- a/parachain/pallets/control/src/lib.rs
+++ b/parachain/pallets/control/src/lib.rs
@@ -109,6 +109,7 @@ pub mod pallet {
 		AgentNotExist,
 		ChannelAlreadyCreated,
 		ChannelNotExist,
+		Unsupported,
 	}
 
 	#[pallet::storage]
@@ -270,9 +271,12 @@ pub mod pallet {
 			Ok(())
 		}
 
-		/// Sends a message to the Gateway contract to transfer asset from agent
+		/// Sends a message to the Gateway contract to transfer ether from an agent
+		/// controlled by `origin`.
 		///
 		/// - `origin`: Must be `MultiLocation`
+		/// - `recipient`: Recipient of funds
+		/// - `amount`: Amount to transfer
 		#[pallet::call_index(5)]
 		#[pallet::weight(T::WeightInfo::transfer_native_from_agent())]
 		pub fn transfer_native_from_agent(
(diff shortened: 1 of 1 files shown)


yrong:
If the only difference is to allow sudo as another kind of origin then I would prefer to revamp a bit in https://github.com/Snowfork/cumulus/blob/f5180878ca95afd9f8b9525d54daa5ed4a2b9505/parachains/runtimes/bridge-hubs/bridge-hub-rococo/src/lib.rs#L689 without adding a new extrinsic.

alistair-singh:
I do not think we need a new extrinsic as we can do this via XCM with the existing `transfer_native_from_agent` extrinsic. Governance can impersonate any agent with `DescendOrigin`. So the following XCM can be used to achieve the same effect:
```rs
DescendOrigin(1000),
Transact(OriginKind::Xcm, transfer_native_from_agent),
```

yrong:
@vgeddes Can this PR be safely closed considering the comment above in https://github.com/Snowfork/snowbridge/pull/960#issuecomment-1710923597?

vgeddes:
> If the only difference is to allow sudo as another kind of origin then I would prefer to revamp a bit in https://github.com/Snowfork/cumulus/blob/f5180878ca95afd9f8b9525d54daa5ed4a2b9505/parachains/runtimes/bridge-hubs/bridge-hub-rococo/src/lib.rs#L689 without adding a new extrinsic.
> 
> Something like `type AgentOrigin = EitherOf<EnsureRoot<Self::AccountId>,EnsureXcm<Everything>>;`

This won't work, since the root account doesn't have an xcm location associated with it and so we don't know which agent to withdraw from.

vgeddes:
> I do not think we need a new extrinsic as we can do this via XCM with the existing `transfer_native_from_agent` extrinsic. Governance can impersonate any agent with `DescendOrigin`. So the following XCM can be used to achieve the same effect:
> 
> ```rust
> DescendOrigin(1000),   // Impersonate Asset hub.
> Transact(OriginKind::Xcm, transfer_native_from_agent),
> ```

This won't really work either since it  will require custom code on the relay chain to issue such an xcm message, and that's a no go.

vgeddes:
I've implemented these `force_*` methods in this PR rather: https://github.com/Snowfork/snowbridge/pull/968

PR #961, opened 2023-09-08
Remove call index from the cross-chain interface

yrong:
Resolves: https://linear.app/snowfork/issue/SNO-657

Requires: https://github.com/Snowfork/cumulus/pull/66


parachain/primitives/router/src/inbound/mod.rs
```diff
@@ -123,6 +122,8 @@ impl Command {
 				let asset_id = Self::convert_token_address(network, gateway, token);
 
 				let mut instructions = create_instructions(origin_location);
+
+				let create_call_index: [u8; 2] = [53, 0];
```
yrong, line 126: As Oak mentioned, we can make it configurable load from some storage.
But I would suspect even this is not necessary as there is also some strict check flow for polkadot/cumulus runtime upgrade to not override an existing call so the call index is unlike to change. So just hardcode here should be fine.

vgeddes: That's a good point Ron, which I never realised. The call index should ideally remain the same for all AssetHub runtimes.

If you can confirm that, I'm happy for you to proceed with this PR.

yrong: > The call index should ideally remain the same for all AssetHub runtimes.

Yes, it's the same for all asset-hub runtimes(e.g. asset-hub-polkado
https://github.com/paritytech/polkadot-sdk/blob/f204e3264f945c33b4cea18a49f7232c180b07c5/cumulus/parachains/runtimes/assets/asset-hub-polkadot/src/lib.rs#L772)

alistair-singh (commented):
Looks good so far!

alistair-singh (approved):

parachain/primitives/router/src/inbound/mod.rs
```diff
@@ -103,7 +103,8 @@ impl Command {
 								Parent,
 								GlobalConsensus(network),
 								origin_location,
-							).into(),
+							)
+								.into(),
```
claravanstaden, line 107: Is the formatting right? It looks a little weird. 🤔

claravanstaden (approved):
Nice!

contracts/src/SubstrateTypes.sol
```diff
@@ -56,18 +56,13 @@ library SubstrateTypes {
      * `NativeTokensMessage::Create`
      */
     // solhint-disable-next-line func-name-mixedcase
-    function RegisterToken(address gateway, address token, bytes2 createCallIndex)
-        internal
-        view
-        returns (bytes memory)
-    {
```
vgeddes, line 63: Please don't make any formatting/whitespace changes. If necessary, we should disable any auto-formatters until the audit is complete.

This will help the auditors.

contracts/src/Assets.sol
```diff
@@ -95,17 +95,14 @@ library Assets {
 
     /// @dev Enqueues a create native token message to substrate.
     /// @param token The ERC20 token address.
-    function registerToken(address token, bytes2 createTokenCallID)
-        external
-        returns (bytes memory payload, uint256 extraFee)
-    {
```
vgeddes, line 101: Please don't make any formatting/whitespace changes. If necessary, we should disable any auto-formatters until the audit is complete.

This will help the auditors.

parachain/primitives/router/src/inbound/mod.rs
```diff
@@ -17,30 +17,30 @@ const MINIMUM_DEPOSIT: u128 = 1;
 /// we may want to evolve the protocol so that the ethereum side sends XCM messages directly.
 /// Instead having BridgeHub transcode the messages into XCM.
 #[derive(Clone, Encode, Decode, RuntimeDebug)]
-pub enum VersionedMessage {
-	V1(MessageV1),
+pub enum VersionedMessage<G> {
```
vgeddes, line 20: I'm uncomfortable adding generic types to `VersionedMessage` which are only going to be used for conversion purposes. Its perhaps a sign that conversion is becoming complicated enough to justify some refactoring.

So instead, perhaps we should introduce a separate converter that implements `sp_runtime::traits::Convert<VersionedMessage, Xcm<()>>`. The converter can be parameterised with all the conversion config it requires.

parachain/primitives/router/src/inbound/mod.rs
```diff
@@ -70,13 +70,13 @@ pub enum Destination {
 	ForeignAccountId20 { para_id: u32, id: [u8; 20] },
 }
 
-impl From<MessageV1> for Xcm<()> {
-	fn from(val: MessageV1) -> Self {
+impl<G: Get<[u8; 2]>> From<MessageV1<G>> for Xcm<()> {
+	fn from(val: MessageV1<G>) -> Self {
```
vgeddes, line 74: See man comment above, I think we should introduce a separate converter type, ie `Converter::convert(msgv1) -> Xcm`

parachain/primitives/router/src/inbound/mod.rs
yrong: https://github.com/Snowfork/snowbridge/pull/961/commits/2bcd6bec5df6861c402a0bd19135ec329f27965c

parachain/primitives/router/src/inbound/mod.rs
yrong: Yes, it's formatted by the [rule](https://github.com/Snowfork/snowbridge/blob/main/parachain/rustfmt.toml). Notice sometimes we've format inconsistent which makes cooperating development/code review a bit challenge. 

So would suggest as a team to always share the same formatting rule. Btw: config the ide with [auto format ](https://github.com/rust-lang/rust-analyzer/issues/3916#issuecomment-611617573) could be helpful.

contracts/src/SubstrateTypes.sol
yrong: This is a relevent change for remove `createCallIndex` which shorten code line and triggers the format. 

I would assume this is revelant to intention of the fix and keep it, or we may need to disable the rule [here](https://github.com/Snowfork/snowbridge/blob/4f739fde129de640c667dc10eb0a08b701183bdc/hooks/pre-commit#L23).

contracts/src/Assets.sol
yrong: https://github.com/Snowfork/snowbridge/pull/961#discussion_r1334933344

parachain/pallets/inbound-queue/src/lib.rs
```diff
@@ -214,20 +223,23 @@ pub mod pallet {
 
 			// Decode message into XCM
 			let xcm = match inbound::VersionedMessage::decode_all(&mut envelope.payload.as_ref()) {
-				Ok(inbound::VersionedMessage::V1(message_v1)) => message_v1.into(),
+				Ok(message) => inbound::VersionedMessageToXcmConverter::<
```
vgeddes, line 226: A couple of issues with this implementation:

1. The inbound queue does not need to know about  the `RegisterCallIndex` type. That is an implementation detail of the converter. I would remove it from the pallet config.

2. Instead, the converter itself should be provided to the pallet config. Example:

```rust
type MessageConverter: ConvertMessage
```

where `ConvertMessage` is a trait implemented by the converter:

```
pub trait ConvertMessage {
  fn convert(message: VersionedMessage) -> Result<Xcm<()>, ConvertMessageError>;
}

pub enum ConvertMessageError {
  UnsupportedVersion,
}
```

parachain/primitives/router/src/inbound/mod.rs
```diff
@@ -70,135 +68,172 @@ pub enum Destination {
...
-								GlobalConsensus(network),
-								origin_location,
-							).into(),
-						},
-					]
-					.into(),
-				),
-			]
-		};
+impl<G: Get<[u8; 2]>, F: XcmFeeGetter> sp_runtime::traits::TryConvert<VersionedMessage, Xcm<()>>
```
vgeddes, line 98: Please use a type alias for `[u8; 2]`, to increase descriptiveness.

```suggestion
impl<G: Get<CallIndex>, F: XcmFeeGetter> sp_runtime::traits::TryConvert<VersionedMessage, Xcm<()>>
```

parachain/primitives/router/src/inbound/mod.rs
```diff
@@ -70,135 +68,172 @@ pub enum Destination {
...
+	for VersionedMessageToXcmConverter<G, F>
+{
+	fn try_convert(message: VersionedMessage) -> Result<Xcm<()>, VersionedMessage> {
```
vgeddes, line 101: The problem with `TryConvert` is that is doesn't allow us to return a real error, such as `UnsupportedVersion`.

parachain/primitives/router/src/inbound/mod.rs
```diff
@@ -70,135 +68,172 @@ pub enum Destination {
...
 				};
-				instructions.append(&mut fragment);
-				instructions.into()
+				Ok(xcm)
 			},
 		}
 	}
+}
 
+impl<G: Get<[u8; 2]>, F: XcmFeeGetter> VersionedMessageToXcmConverter<G, F> {
```
vgeddes, line 236:
```suggestion
impl<G: Get<CallIndex>, F: XcmFeeGetter> VersionedMessageToXcmConverter<G, F> {
```

parachain/pallets/inbound-queue/src/lib.rs
yrong: Thanks for the suggestion! Some revamp in https://github.com/Snowfork/snowbridge/pull/961/commits/18a90eb78006ebbd441184d89dcb9fdc9535e3a2#diff-7d758f1d757c44217a6d64358b4e11fa6fadaf9d160c655ee9b076257dd1be86

parachain/primitives/router/src/inbound/mod.rs
```diff
@@ -70,135 +69,158 @@ pub enum Destination {
...
-	pub fn convert(self, chain_id: u64) -> Xcm<()> {
-		let network = NetworkId::Ethereum { chain_id };
-		// TODO (SNO-582): The fees need to be made configurable and must match the weight
-		// required by the generated XCM script when executed on the foreign chain.
-		let buy_execution_fee_amount = 2_000_000_000;
-		let buy_execution_fee = MultiAsset {
-			id: Concrete(MultiLocation::parent()),
-			fun: Fungible(buy_execution_fee_amount),
-		};
+#[derive(TypeInfo, PalletError, Encode, Decode)]
```
vgeddes, line 76: Should also derive `Copy` and `RuntimeDebug`.

parachain/primitives/router/src/inbound/mod.rs
```diff
@@ -70,135 +69,158 @@ pub enum Destination {
 	ForeignAccountId20 { para_id: u32, id: [u8; 20] },
 }
 
-impl From<MessageV1> for Xcm<()> {
-	fn from(val: MessageV1) -> Self {
-		val.message.convert(val.chain_id)
-	}
+pub struct VersionedMessageToXcmConverter<G> {
```
vgeddes, line 72: This struct name is a bit too verbose. Let's shorten it. And rename the generic parameter `G` to `CreateAssetCall`.

```suggestion
pub struct MessageToXcm<CreateAssetCall> {
```

parachain/primitives/router/src/inbound/mod.rs
```diff
@@ -70,135 +69,158 @@ pub enum Destination {
...
-					Destination::ForeignAccountId20 { para_id, id } => (
-						Some(para_id),
-						MultiLocation {
-							parents: 0,
-							interior: X1(AccountKey20 { network: None, key: id }),
-						},
-					),
+pub type CallIndex = [u8; 2];
+
+impl<G: Get<CallIndex>> ConvertMessage for VersionedMessageToXcmConverter<G> {
```
vgeddes, line 87:
```suggestion
impl<CreateAssetCall> ConvertMessage for MessageToXcm<CreateAssetCall>
where
    CreateAssetCall: Get<CallIndex>
{
```

parachain/primitives/router/src/inbound/mod.rs
```diff
@@ -70,135 +69,158 @@ pub enum Destination {
...
+	fn convert(message: VersionedMessage) -> Result<Xcm<()>, ConvertMessageError> {
+		match message {
+			VersionedMessage::V1(val) => {
```
vgeddes, line 90: The name `val` isn't great. I think let's destructure it:

```suggestion
			VersionedMessage::V1(MessageV1 { chain_id, message }) => {
```

parachain/primitives/router/src/inbound/mod.rs
```diff
@@ -70,135 +69,158 @@ pub enum Destination {
...
 				};
-				instructions.append(&mut fragment);
-				instructions.into()
+				Ok(xcm)
 			},
 		}
 	}
+}
 
+impl<G: Get<CallIndex>> VersionedMessageToXcmConverter<G> {
```
vgeddes, line 223:
```suggestion
impl<CreateAssetCall: Get<CallIndex>> MessageToXcm<CreateAssetCall> {
```

parachain/primitives/router/src/inbound/mod.rs
```diff
@@ -70,135 +69,158 @@ pub enum Destination {
...
-		// TODO (SNO-582): The fees need to be made configurable and must match the weight
-		// required by the generated XCM script when executed on the foreign chain.
-		let buy_execution_fee_amount = 2_000_000_000;
-		let buy_execution_fee = MultiAsset {
-			id: Concrete(MultiLocation::parent()),
-			fun: Fungible(buy_execution_fee_amount),
-		};
+#[derive(TypeInfo, PalletError, Encode, Decode)]
+pub enum ConvertMessageError {
+	UnsupportedVersion,
```
vgeddes, line 78:
```suggestion
        /// The message version is not supported for conversion.
	UnsupportedVersion,
```

parachain/primitives/router/src/inbound/mod.rs
```diff
@@ -70,135 +69,158 @@ pub enum Destination {
...
-		let network = NetworkId::Ethereum { chain_id };
-		// TODO (SNO-582): The fees need to be made configurable and must match the weight
-		// required by the generated XCM script when executed on the foreign chain.
-		let buy_execution_fee_amount = 2_000_000_000;
-		let buy_execution_fee = MultiAsset {
-			id: Concrete(MultiLocation::parent()),
-			fun: Fungible(buy_execution_fee_amount),
-		};
+#[derive(TypeInfo, PalletError, Encode, Decode)]
+pub enum ConvertMessageError {
```
vgeddes, line 77:
```suggestion
/// Reason why a message conversion failed.
pub enum ConvertMessageError {
```

parachain/pallets/inbound-queue/src/test.rs
```diff
@@ -136,6 +136,7 @@ const GATEWAY_ADDRESS: [u8; 20] = hex!["eda338e4dc46038493b885327842fd3e301cab39
 parameter_types! {
 	pub const EthereumNetwork: xcm::v3::NetworkId = xcm::v3::NetworkId::Ethereum { chain_id: 15 };
 	pub const GatewayAddress: H160 = H160(GATEWAY_ADDRESS);
+	pub const RegisterCallIndex: [u8;2] = [53, 0];
```
vgeddes, line 139:
```suggestion
	pub const CreateAssetCall: CallIndex = [53, 0];
```

parachain/pallets/inbound-queue/src/lib.rs
```diff
@@ -124,6 +130,8 @@ pub mod pallet {
 		Send(SendError),
 		/// Operational mode errors
 		OperationalMode(bp_runtime::OwnedBridgeModuleError),
+		/// Convert error
+		ConvertError(ConvertMessageError),
```
vgeddes, line 134:
```suggestion
		/// Message conversion error
		ConvertMessage(ConvertMessageError),
```

parachain/pallets/inbound-queue/src/lib.rs
```diff
@@ -85,6 +89,8 @@ pub mod pallet {
 		#[pallet::constant]
 		type GatewayAddress: Get<H160>;
 
+		type MessageConverter: ConvertMessage;
```
vgeddes, line 92: Needs a documentation comment.

```suggestion
		/// Convert inbound message to XCM
		type MessageConverter: ConvertMessage;
```

alistair-singh (approved):

parachain/primitives/core/src/outbound.rs
```diff
@@ -51,6 +51,10 @@ pub enum SubmitError {
 	MessageTooLarge,
 	/// The bridge has been halted for maintenance
 	BridgeHalted,
+	/// Gas config invalid
+	InvalidGas(u128),
+	/// Estimate fee failed
+	EstimateFeeFailed,
```
vgeddes: Obsolete code?

vgeddes (approved):
Lets merge this in 🚀

PR #962, opened 2023-09-12
Configurable dispatch gas

yrong:
Resolves: SNO-650


vgeddes:
Looks good so far 👌

vgeddes:
As discussed with @musnit we may also want to include relayer reward in the message to the Gateway, to make the messaging protocol more declarative (also known as _intents_ in DeFi terminology).

But let's confirm this first in our meeting tomorrow.

yrong:
> we may also want to include relayer reward in the message to the Gateway, to make the messaging protocol more declarative

Agree. The reward should not be some config per channel as we do now. It seems more like the config for each kind of the message(i.e. incentivize relayer with more reward for a heavy dispatch).

yrong:
Closed in favor of https://github.com/Snowfork/snowbridge/pull/940

PR #963, opened 2023-09-13
Updates lodestar

claravanstaden:
Tested with node `v18.16.1`. E2E stack works.


yrong:
From this comment https://github.com/ChainSafe/lodestar/issues/5851#issue-1837932118 seems we also need to upgrade node release to >=18.17.0?

alistair-singh:
> From this comment [ChainSafe/lodestar#5851 (comment)](https://github.com/ChainSafe/lodestar/issues/5851#issue-1837932118) seems we also need to upgrade node release to >=18.17.0?

We only need to see if the testnet starts up and works correctly with the version of the 1.10 version of lodestar. In prod we will deploy using the container vended by Chainsafe which will have the correct version of node. I think the version of node 20.5.0 is not available from nix.

claravanstaden:
Closing for now, until we've done more tests and need to upgrade.

PR #966, opened 2023-09-21
Update for ark integration

yrong:
The initial intention is for migrating bls algorithm from milagro to [ark curves](https://github.com/arkworks-rs/curves). But for G1 point can't be scale encoded and cached into storage so the benchmark is worse than milagro. 

It's still in investigation and I'll come back when finished the oak issues.

### DIFF
--- a/parachain/primitives/beacon/src/ark_bls.rs
+++ b/parachain/primitives/beacon/src/ark_bls.rs
@@ -1,0 +1,180 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2023 Snowfork <hello@snowfork.com>
+
+// core
+use core::{borrow::Borrow, ops::Neg};
+// crates.io
+use ark_bls12_381::{
+	g2::Config as G2Config, Bls12_381, G1Affine, G1Projective, G2Affine, G2Projective,
+};
+use ark_ec::{
+	hashing::{
+		curve_maps::wb::WBMap, map_to_curve_hasher::MapToCurveBasedHasher, HashToCurve,
+		HashToCurveError,
+	},
+	models::short_weierstrass::Projective,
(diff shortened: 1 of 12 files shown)


parachain/primitives/beacon/src/ark_bls.rs
```diff
@@ -0,0 +1,180 @@
...
+}
+
+pub fn fast_aggregate_verify(
+	pub_keys: Vec<Vec<u8>>,
+	message: Vec<u8>,
+	signature: Vec<u8>,
+) -> Result<bool, BlsError> {
+	let sig = Signature::from_bytes(&signature).map_err(|_| BlsError::InvalidSignature)?;
+	let pub_keys: Result<Vec<PublicKey>, _> =
+		pub_keys.into_iter().map(|pk| PublicKey::from_bytes(&pk)).collect();
```
yrong, line 95: The decompressed pubkey should be read from cached storage as we do for Milagro.

claravanstaden:
Cool investigation! Why can't G1 point be scale encoded?

parachain/primitives/beacon/src/ark_bls.rs
vgeddes, line 1: Hey, curious where you got this code. 

Did you see this comment from Seun by the way? It's another BLS lib based on Arkworks.

https://matrix.to/#/!dOiGCcBecwSGXtifov:web3.foundation/$169358770621189BwbgH:matrix.org?via=web3.foundation&via=matrix.org&via=matrix.parity.io

yrong: No, it's also from other projects. I've checked the implementation a bit and seems the logic/library used underying almost the same. And I would assume what we need is only the primitives for `fast_aggregate_verify` so a thin wrapper of the Arkworks library should be enough.

PR #968, opened 2023-10-04
Improved fees and rewards logic for messages to Ethereum

vgeddes:
Changes:
* Messages to ethereum now include the maximum gas required to dispatch a message payload
* Relayers to Ethereum are refunded gas cost, and are given a static reward on top of that.
* Added remaining required APIs to control pallet in BridgeHub, such as `force_transfer_native`, `set_operating_mode`, etc.
* the `create_agent` and `create_channel` extrinsics in the control pallet now charge an upfront deposit.

Cumulus companion: https://github.com/Snowfork/cumulus/pull/67

Fixes: SNO-625, SNO-650


contracts/src/Gateway.sol
```diff
@@ -136,53 +134,81 @@ contract Gateway is IGateway, IInitializable {
...
             catch {
                 success = false;
             }
         }
 
+        // Calculate the remaining funds in the channel agent contract
+        uint256 agentBalance = channel.agent.balance;
+        if (channel.agent.balance <= MINIMUM_THRESHOLD_GAS * tx.gasprice) {
+            agentBalance = 0;
+        }
```
yrong, line 186: Just curious why add this check or what is the bad effect without it?

contracts/src/Gateway.sol
```diff
@@ -136,53 +134,81 @@ contract Gateway is IGateway, IInitializable {
...
+        uint256 gasUsed = startGas - gasleft() + BASE_GAS_USED;
+        uint256 refund = gasUsed * tx.gasprice;
+
+        // Add the reward to the refund amount. If the sum is more than the funds available
+        // in the channel agent, then reduce the total amount
+        uint256 amount = _min(refund + message.reward, agentBalance);
+
+        // Do the payment if there funds available in the agent
+        if (amount > 0) {
+            _transferNativeFromAgent(channel.agent, payable(msg.sender), amount);
```
yrong, line 198: Also not sure fully get the point of `BASE_GAS_USED` here.  Is it absolutely required or just increase the `message.reward` in the bridgehub config a bit more to cover the extra reward?

parachain/pallets/control/src/lib.rs
```diff
@@ -161,15 +186,10 @@ pub mod pallet {
 		pub fn create_agent(origin: OriginFor<T>) -> DispatchResult {
 			let origin_location: MultiLocation = T::AgentOrigin::ensure_origin(origin)?;
 
-			let (agent_id, _, location) = Self::convert_location(origin_location)?;
+			Self::charge_fee(&origin_location)?;
```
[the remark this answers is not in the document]
yrong: Btw, fix in https://github.com/Snowfork/snowbridge/pull/968/commits/7c6f69a231a127eb05cb2cf098e8d11c4d36faea
to make smoke tests like `create_agent|create_channel` back to work.

parachain/pallets/control/src/mock.rs
```diff
@@ -74,135 +171,88 @@ parameter_types! {
...
+pub type LocationToAccountId = (
+	// The parent (Relay-chain) origin converts to the parent `AccountId`.
+	ParentIsPreset<AccountId>,
+	// Sibling parachain origins convert to AccountId via the `ParaId::into`.
+	SiblingParachainConvertsVia<Sibling, AccountId>,
+	// Straight up local `AccountId32` origins just alias directly to `AccountId`.
+	AccountId32Aliases<RelayNetwork, AccountId>,
+	// Other nested consensus systems on sibling parachains or relay chain.
+	HashedDescription<AccountId, DescribeFamily<DescribeAllTerminal>>,
+);
```
yrong, line 205: Same config as in bridge_hub and https://github.com/Snowfork/cumulus/pull/67/files#r1346854601

contracts/src/Gateway.sol
vgeddes: If there is only a tiny amount of ether (known as "dust") in the agent contract, then the gas cost of transferring that dust will be more than the value of the dust. So in that case it makes sense to skip the transfer.

Either way it's not a huge issue. Just a small optimisation in the relayers favour.

contracts/src/Gateway.sol
vgeddes: `BASE_GAS_USED` actually refers to all the gas used outside the code that is being metered. This includes the fixed cost of 21,000 that every relayer pays for a transaction, before the code in the transaction even starts executing.

vgeddes: I would prefer to be explicit and keep it in.

parachain/pallets/control/src/lib.rs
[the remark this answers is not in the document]
vgeddes: Thanks Ron, much appreciated.

yrong (approved):

parachain/pallets/control/src/lib.rs
```diff
@@ -166,41 +200,47 @@ pub mod pallet {
...
 		pub fn create_agent(origin: OriginFor<T>) -> DispatchResult {
-			let origin_location: MultiLocation = T::AgentOrigin::ensure_origin(origin)?;
-
-			Self::charge_fee(&origin_location)?;
+			let origin_location: MultiLocation = T::AgentOwnerOrigin::ensure_origin(origin)?;
 
-			let OriginInfo { reanchored_location, agent_id, .. } =
-				Self::process_origin_location(&origin_location)?;
+			// Ensure that origin location is some consensus system on a sibling parachain
+			let (para_id, agent_id, ancestry) = ensure_sibling_or_sibling_child::<T>(&origin_location)?;
```
yrong, line 229: Nice to remove the reanchor logic.

yrong:
Nice refactor! Just some minor fix for clippy.
@vgeddes Seems we remove relayers for template node in https://github.com/Snowfork/snowbridge/pull/968/commits/2a9633ba2f688021a7bb59f95d7d48fbf20c302f
It's required for the smoke test `cargo test --test transfer_native_from_agent` so revert to make it work.

parachain/pallets/control/src/lib.rs
vgeddes: yeah we don't really need to reanchor anymore. It seems that was only required in order to support local locations within BridgeHub. However we only need to support locations on sibling parachains.

web/packages/test/config/launch-config.toml
```diff
@@ -14,14 +14,14 @@ default_args = [
     "--pruning", "archive",
 ]
 
-    [relaychain.genesis.runtime]
-    session_length_in_blocks = 20
+    # [relaychain.genesis.runtime]
```
alistair-singh, line 17: The commented-out config should be deleted.

parachain/primitives/core/src/outbound.rs
```diff
@@ -20,44 +32,51 @@ pub trait OutboundQueue {
...
+#[derive(Copy, Clone, Encode, Decode, PartialEq, Eq, RuntimeDebug, TypeInfo)]
 pub enum SubmitError {
 	/// Message is too large to be safely executed on Ethereum
 	MessageTooLarge,
 	/// The bridge has been halted for maintenance
 	BridgeHalted,
+	/// Gas config invalid
+	InvalidGas(u128),
+	/// Estimate fee failed
+	EstimateFeeFailed,
```
alistair-singh, line 56: Cannot find any usage of these errors in the code.

hooks/pre-commit
```diff
@@ -23,7 +23,7 @@ chronic typos .
 (cd relayer && chronic mage lint && chronic go fmt ./...)
 
 # format rust
-(cd parachain && chronic cargo +nightly-"$SNOWBRIDGE_RUST_NIGHTLY" fmt)
+# (cd parachain && chronic cargo +nightly-"$SNOWBRIDGE_RUST_NIGHTLY" fmt)
```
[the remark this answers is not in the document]
yrong: I would expect It's a temp disable and we'll enable it after the audit.
@vgeddes Just curious are you using the auto-format feature provided by the IDE or some other format rule?  would suggest we follow the same rule, related comments in inhttps://github.com/Snowfork/snowbridge/pull/961#discussion_r1334932702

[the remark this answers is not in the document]
vgeddes: Ah oops. I temporarily disabled that as rust nightly wasn't installed in my env. I now see that `scripts/init.sh` installs it. So I will uncomment these lines.

parachain/pallets/outbound-queue/src/lib.rs
```diff
@@ -293,7 +320,8 @@ pub mod pallet {
 
 			let ticket =
 				OutboundQueueTicket { id: message_id, origin: message.origin, message: encoded };
-			Ok((ticket, T::Fee::get()))
+
+			Ok((ticket, delivery_fee))
```
yrong, line 324: should it be `T::Fee::get() + delivery_fee`?  I would assume the `T::Fee::get()` to cover the cost in bridgeHub(previously I named as `base_fee` not only for the cost of submit message but also the cost for constructing the merkle tree when messages are committed).

hooks/pre-commit
```diff
@@ -16,14 +16,14 @@ check_duplicate_versions() {
...
 
-# lint and format for relayer codes
-(cd relayer && chronic mage lint && chronic go fmt ./...)
+# check format for relayer codes
+(cd relayer && chronic gofmt -l .)
 
-# format rust
-# (cd parachain && chronic cargo +nightly-"$SNOWBRIDGE_RUST_NIGHTLY" fmt)
+# check format of rust
+(cd parachain && chronic cargo +$RUST_NIGHTLY_VERSION fmt)
```
yrong, line 26: Would prefer also add ```--check``` here.

parachain/pallets/outbound-queue/src/lib.rs
vgeddes: It just covers the cost on the Ethereum side for now. Can always do a follow up PR to improve, but its not necessary for the audit.

hooks/pre-commit
vgeddes: done

vgeddes: done

hooks/pre-commit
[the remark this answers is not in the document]
vgeddes: I've changed the pre-commit to only _check_ whether formatting is invalid. I don't quite like it when git hooks start rewriting code automatically.

Instead, I'd like for everyone to enable auto-formatting in their IDEs. Example configuration for VS code:

```json
{
    "solidity.formatter": "forge",
    "[solidity]": {
        "editor.defaultFormatter": "JuanBlanco.solidity",
        "editor.formatOnSave": true
    },
    "[rust]": {
        "editor.defaultFormatter": "rust-lang.rust-analyzer",
        "editor.formatOnSave": true
    },
    "go.lintTool": "revive",
    "go.lintOnSave": "package",
    "go.lintFlags": [
        "-config=${workspaceFolder}/relayer/revive.toml"
    ],
    "[go]": {
        "editor.defaultFormatter": "golang.go",
        "editor.formatOnSave": true,
    }
}
```

PR #970, opened 2023-10-09
Update constant fee for InboundMessage

yrong:
Resolves: SNO-692, SNO-660
Requires: https://github.com/Snowfork/cumulus/pull/69


parachain/primitives/router/src/inbound/mod.rs
```diff
@@ -225,7 +247,9 @@ where
 	}
 }
 
-impl<CreateAssetCall: Get<CallIndex>> MessageToXcm<CreateAssetCall> {
+impl<CreateAssetCall: Get<CallIndex>, FeeGetter: XcmFeeGetter>
+	MessageToXcm<CreateAssetCall, FeeGetter>
+{
```
alistair-singh, line 252: I think using `where` is a little bit easier to read.
```suggestion
impl<CreateAssetCall, FeeGetter> MessageToXcm<CreateAssetCall, FeeGetter>
where  
    CreateAssetCall: Get<CallIndex>,
    FeeGetter: XcmFeeGetter,
{
```

parachain/primitives/router/src/inbound/mod.rs
```diff
@@ -87,15 +91,33 @@ pub trait ConvertMessage {
...
+
+/// A meter that assigns a constant amount of gas for the execution of a command
+pub struct ConstantFeeForInboundMessage;
+
+impl XcmFeeGetter for ConstantFeeForInboundMessage {
+	fn fee(command: &Command) -> u128 {
+		match command {
+			Command::RegisterToken { .. } => 2_000_000_000,
+			Command::SendToken { .. } => 1_000_000_000,
+		}
```
alistair-singh, line 107: We did lose a little bit of coverage with these new lines though I am not sure if we need tests for such a simple implementation.

alistair-singh (approved):
Looks good!

parachain/primitives/router/src/inbound/mod.rs
yrong: https://github.com/Snowfork/snowbridge/pull/970/commits/f8fe653d239e445e3ab44da4a590e71ace49bde7

parachain/primitives/router/src/inbound/mod.rs
```diff
@@ -87,15 +91,33 @@ pub trait ConvertMessage {
...
 	CreateAssetCall: Get<CallIndex>,
+	FeeGetter: XcmFeeGetter,
 {
 	fn convert(message: VersionedMessage) -> Result<Xcm<()>, ConvertMessageError> {
 		match message {
-			VersionedMessage::V1(MessageV1 { chain_id, message }) => {
+			VersionedMessage::V1(MessageV1 { chain_id, message: command }) => {
 				let network = Ethereum { chain_id };
-				let buy_execution_fee_amount = 2_000_000_000;
+				let buy_execution_fee_amount = FeeGetter::fee(&command);
```
vgeddes, line 120: I am not too sure about this design. For one, we may need to construct an xcm message that buys execution fees multiple times, for example, on AssetHub, and then the final destination parachain.

So I think we should start simple and rather rely on the following generic parameters:

```rust
CreateAssetExecutionFee: Get<u128>
SendTokenExecutionFee: Get<u128>
```

Then if we need more complexity, we can revisit as necessary

yrong: Yean make sense. Could be some double map storage like `(Multilocation,Command)->Fee` but let's make it simple now. Some revamp in
https://github.com/Snowfork/snowbridge/pull/970/commits/31e3019f3c9d1b5ed2951e2eeb2460824cd43ad4

alistair-singh (approved):

parachain/primitives/router/src/inbound/mod.rs
```diff
@@ -87,15 +92,21 @@ pub trait ConvertMessage {
...
+	for MessageToXcm<CreateAssetCall, CreateAssetExecutionFee, SendTokenExecutionFee>
 where
 	CreateAssetCall: Get<CallIndex>,
+	CreateAssetExecutionFee: Get<u128>,
+	SendTokenExecutionFee: Get<u128>,
 {
 	fn convert(message: VersionedMessage) -> Result<Xcm<()>, ConvertMessageError> {
 		match message {
-			VersionedMessage::V1(MessageV1 { chain_id, message }) => {
+			VersionedMessage::V1(MessageV1 { chain_id, message: command }) => {
```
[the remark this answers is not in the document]
yrong: https://github.com/Snowfork/snowbridge/pull/970/commits/c51c633a507d9237ee39a08b797001ea7d3f026f

vgeddes (approved):

PR #971, opened 2023-10-17
Reward relayer from asset hub sovereign account that matches asset hub

claravanstaden:
Updates the way we get sovereign account IDs from parachain IDs. Use the SiblingId struct instead of Id, since that is what AssetHub does. The sovereign account we use should be the same as the account ID used on AssetHub.

### DIFF
--- a/.github/workflows/parachain.yml
+++ b/.github/workflows/parachain.yml
@@ -80,7 +80,6 @@ jobs:
           --manifest-path parachain/Cargo.toml
           --workspace
           --features runtime-benchmarks
-          --exclude snowbridge-query-events
           --exclude snowbridge-ethereum-beacon-client
       # Run tests for beacon light client, excluding benchmark tests
       - name: Tests for beacon light client excluding benchmark tests
--- a/parachain/pallets/inbound-queue/src/lib.rs
+++ b/parachain/pallets/inbound-queue/src/lib.rs
@@ -35,7 +35,7 @@ use xcm::v3::{
 use envelope::Envelope;
 use snowbridge_core::{
 	inbound::{Message, Verifier},
-	ParaId,
+	ParaId, SiblingParaId
 };
 use snowbridge_router_primitives::{
 	inbound,
@@ -218,9 +218,12 @@ pub mod pallet {
 				}
 			})?;
 
+			// Convert to a sibling parachain ID (uses bytes "sibl" instead of "para" to derive
+			// the sovereign account ID, to match how Assethub derives the account ID).
+			let sibling_para_id: SiblingParaId = envelope.dest.into();
 			// Reward relayer from the sovereign account of the destination parachain
 			// Expected to fail if sovereign account has no funds
-			let sovereign_account = envelope.dest.into_account_truncating();
+			let sovereign_account = sibling_para_id.into_account_truncating();
 			T::Token::transfer(&sovereign_account, &who, T::Reward::get(), Preservation::Preserve)?;
 
 			// Decode message into XCM
--- a/parachain/pallets/inbound-queue/src/lib.rs
+++ b/parachain/pallets/inbound-queue/src/lib.rs
@@ -25,7 +25,6 @@ use frame_support::{
 use frame_system::ensure_signed;
 use scale_info::TypeInfo;
 use sp_core::H160;
-use sp_runtime::traits::AccountIdConversion;
 use sp_std::convert::TryFrom;
 use xcm::v3::{
 	send_xcm, Junction::*, Junctions::*, MultiLocation, SendError as XcmpSendError, SendXcm,
@@ -35,7 +34,7 @@ use xcm::v3::{
 use envelope::Envelope;
 use snowbridge_core::{
 	inbound::{Message, Verifier},
-	ParaId, SiblingParaId
+	ParaId, sibling_sovereign_account
 };
 use snowbridge_router_primitives::{
 	inbound,
@@ -218,12 +217,9 @@ pub mod pallet {
 				}
 			})?;
 
-			// Convert to a sibling parachain ID (uses bytes "sibl" instead of "para" to derive
-			// the sovereign account ID, to match how Assethub derives the account ID).
-			let sibling_para_id: SiblingParaId = envelope.dest.into();
 			// Reward relayer from the sovereign account of the destination parachain
 			// Expected to fail if sovereign account has no funds
-			let sovereign_account = sibling_para_id.into_account_truncating();
+			let sovereign_account = sibling_sovereign_account::<T>(envelope.dest);
 			T::Token::transfer(&sovereign_account, &who, T::Reward::get(), Preservation::Preserve)?;
 
 			// Decode message into XCM
--- a/web/packages/test/scripts/set-env.sh
+++ b/web/packages/test/scripts/set-env.sh
@@ -61,10 +61,10 @@ skip_relayer="${SKIP_RELAYER:-false}"
 
 ## Important accounts
 
-# Account for assethub (1000 sibling parachain 5Eg2fntNprdN3FgH4sfEaaZhYtddZQSQUqvYJ1f2mLtinVhV in testnet)
+# Account for assethub (Sibling parachain 1000 5Eg2fntNprdN3FgH4sfEaaZhYtddZQSQUqvYJ1f2mLtinVhV in testnet)
 assethub_sovereign_account="${ASSETHUB_SOVEREIGN_ACCOUNT:-0x7369626ce8030000000000000000000000000000000000000000000000000000}"
-# Account for template (Para 1001)
-template_sovereign_account="${TEMPLATE_SOVEREIGN_ACCOUNT:-0x70617261e9030000000000000000000000000000000000000000000000000000}"
+# Account for template (Sibling parachain 1001 5Eg2fntP2UgYZNc5tW8xmmCmeXJ3hmNXaZ9MvcpbMdvh1bBJ in testnet)
+template_sovereign_account="${TEMPLATE_SOVEREIGN_ACCOUNT:-0x7369626ce9030000000000000000000000000000000000000000000000000000}"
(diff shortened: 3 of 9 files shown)


yrong:
Nice catch! Then do we also need to change the derived `AccountId` in control pallet [here](https://github.com/Snowfork/snowbridge/blob/7ad8ab8013dc0cd3694891a871676725ec43370c/parachain/pallets/control/src/lib.rs#L229)?

claravanstaden:
> Nice catch! Then do we also need to change the derived `AccountId` in control pallet [here](https://github.com/Snowfork/snowbridge/blob/7ad8ab8013dc0cd3694891a871676725ec43370c/parachain/pallets/control/src/lib.rs#L229)?

@yrong perhaps!

 @alistair-singh will [this](https://github.com/Snowfork/snowbridge/pull/971/files#diff-7d758f1d757c44217a6d64358b4e11fa6fadaf9d160c655ee9b076257dd1be86R216) always be asset hub's parachain ID? For send and register token it would be, right? We shouldn't rely on this always being assethub, I know, but just thinking of the downstream effects of the sovereign account ID being different to what one would expect.

vgeddes:
Clara, can you describe where you saw `SiblingParaId::into_account_truncating` being used to create sovereign accounts? There are indeed seem to be many different ways to derive sovereign accounts, and it has been quite confusing for me.

I've been using `ParaId::into_account_truncating`, because that's what the XCM executor does when it converts an xcm sibling origin into a local sovereign account: https://github.com/paritytech/polkadot-sdk/blob/8b3905d2a5847ff6bfc552fb30009bdefe788170/polkadot/xcm/xcm-builder/src/location_conversion.rs#L323

yrong:
>  because that's what the XCM executor does when it converts an xcm sibling origin into a local sovereign account

Actually from runtime config here https://github.com/Snowfork/cumulus/blob/9feea846bae186d7da395f791f207d043ef435eb/parachains/runtimes/bridge-hubs/bridge-hub-rococo/src/xcm_config.rs#L101C31-L101C31 the `ID` used is *sibling*.

claravanstaden:
> Clara, can you describe where you saw `SiblingParaId::into_account_truncating` being used to create sovereign accounts? There are indeed seem to be many different ways to derive sovereign accounts, and it has been quite confusing for me.

@vgeddes in https://github.com/paritytech/polkadot-sdk/pull/1672 xcm execution fees are paid from the asset hub sovereign account. The parachain ID is converted to a sovereign account ID using this trait: https://github.com/acatangiu/polkadot-sdk/blob/pallet-xcm-fully-support-reserve-transfers/cumulus/parachains/runtimes/assets/asset-hub-kusama/src/xcm_config.rs#L78

This uses bytes "sibl" + the para ID to derive the account ID, instead of bytes "para" + para ID, like we do. So then there is a mismatch between what we see as the parachain soveign account to what asset hub sees it to be.

I am not completely sure how this affects the rest of the places where we use ParaId, and if they should be converted to using `Sibling` instead of `Id`. I am marking this PR draft until I've looked at it further.

vgeddes:
Ah I see, yes you're both correct.

I was confused, because when I read the code [here](https://github.com/paritytech/polkadot-sdk/blob/8b3905d2a5847ff6bfc552fb30009bdefe788170/polkadot/xcm/xcm-builder/src/location_conversion.rs#L323), I thought `ParaId` was an alias for the `Id` type, when in fact it is a generic parameter, which can be substituted with `Sibling`.

parachain/pallets/inbound-queue/src/lib.rs
```diff
@@ -211,9 +211,12 @@ pub mod pallet {
 				}
 			})?;
 
+			// Convert to a sibling parachain ID (uses bytes "sibl" instead of "para" to derive
+			// the sovereign account ID, to match how Assethub derives the account ID).
+			let sibling_para_id: SiblingParaId = envelope.dest.into();
 			// Reward relayer from the sovereign account of the destination parachain
 			// Expected to fail if sovereign account has no funds
-			let sovereign_account = envelope.dest.into_account_truncating();
+			let sovereign_account = sibling_para_id.into_account_truncating();
```
vgeddes, line 219: We need to do the same conversion in this pallet as well as the control pallet, so lets make the converter a pluggable type.

```suggestion
			let sovereign_account = T::SiblingSovereignAccountOf::convert(para_id)
```

where `T::SiblingSovereignAccountOf` is a type that implements [Convert<ParaId, T::AccountId>](https://paritytech.github.io/polkadot-sdk/master/sp_runtime/traits/trait.Convert.html#tymethod.convert)

vgeddes: Actually, wait the above is totally overkill. Instead we can simplify by adding a free function in snowbridge core:

```rust
fn sibling_sovereign_account<T>(para_id: ParaId) -> T::AccountId
where
   T: frame_system::Config
{
    SiblingParaId::from(para_id).into_account_truncating()
}
```

vgeddes (commented):
Looks good. Though also need to update the control pallet which also does similar conversions: https://github.com/Snowfork/snowbridge/blob/6d3766947cb2a266c392579a5b890e33e031f338/parachain/pallets/control/src/lib.rs#L263

And to simplify the conversions and abstract the conversion logic away from the pallets, we can make use of a helper function I describe here:
https://github.com/Snowfork/snowbridge/pull/971/files#r1363609484

alistair-singh:
> @alistair-singh will [this](https://github.com/Snowfork/snowbridge/pull/971/files#diff-7d758f1d757c44217a6d64358b4e11fa6fadaf9d160c655ee9b076257dd1be86R216) always be asset hub's parachain ID? For send and register token it would be, right? We shouldn't rely on this always being assethub, I know, but just thinking of the downstream effects of the sovereign account ID being different to what one would expect.

@yrong is correct. Everywhere on bridge hub where we do fee-related things with sovereign accounts we need to use sibling.

.github/workflows/parachain.yml
```diff
@@ -80,7 +80,6 @@ jobs:
           --manifest-path parachain/Cargo.toml
           --workspace
           --features runtime-benchmarks
-          --exclude snowbridge-query-events
```
yrong, line 83: Since we're clean the workflow here, would suggest merge `Tests for everything except the beacon light client` with `Benchmark tests for beacon light client`(Something simple as `cargo test --features runtime-benchmarks` should be fine).

web/packages/test/scripts/set-env.sh
```diff
@@ -61,8 +61,8 @@ skip_relayer="${SKIP_RELAYER:-false}"
 
 ## Important accounts
 
-# Account for assethub (1000 5Ec4AhPZk8STuex8Wsi9TwDtJQxKqzPJRCH7348Xtcs9vZLJ in testnet)
-assethub_sovereign_account="${ASSETHUB_SOVEREIGN_ACCOUNT:-0x70617261e8030000000000000000000000000000000000000000000000000000}"
+# Account for assethub (1000 sibling parachain 5Eg2fntNprdN3FgH4sfEaaZhYtddZQSQUqvYJ1f2mLtinVhV in testnet)
+assethub_sovereign_account="${ASSETHUB_SOVEREIGN_ACCOUNT:-0x7369626ce8030000000000000000000000000000000000000000000000000000}"
 # Account for template (Para 1001)
 template_sovereign_account="${TEMPLATE_SOVEREIGN_ACCOUNT:-0x70617261e9030000000000000000000000000000000000000000000000000000}"
```
yrong, line 67: This account needs to change accordingly.

.github/workflows/parachain.yml
claravanstaden: @yrong can you remember why we split the tests up in the first place?

web/packages/test/scripts/set-env.sh
claravanstaden: Good catch! Addressed in https://github.com/Snowfork/snowbridge/pull/971/commits/2386f85ec08d79dd12f26774e8a473144e85b2ed.

.github/workflows/parachain.yml
yrong: Can't remember correctly. But I would assume since feature `beacon-spec-mainnet` is already covered in `runtime-benchmarks` as in https://github.com/Snowfork/snowbridge/blob/79dba80857f06c550560cbcf818152e0a97137b3/parachain/pallets/ethereum-beacon-client/Cargo.toml#L75, a split in CI is unnessessary.

vgeddes: I would rather we focus this PR on the sovereign account issue, and leave CI/devops stuff for a separate PR.

vgeddes (approved):
Looks good!

.github/workflows/parachain.yml
vgeddes: The 1-liner change to remove the obsolete `snowbridge-query-events` thingy is reasonable though. 👍

vgeddes: > yrong can you remember why we split the tests up in the first place?

I was the one that did that. Because of the following constraints with the beacon client pallet:

* Benchmark tests require the `beacon-spec-mainnet` feature: https://github.com/Snowfork/snowbridge/blob/167b2502d7ae83059cadd78fce0f9b22ae127805/parachain/pallets/ethereum-beacon-client/Cargo.toml#L75

* However, when the `beacon-spec-mainnet` feature is enabled, normal tests are disabled, because they are designed to only work with the minimal spec: https://github.com/Snowfork/snowbridge/blob/167b2502d7ae83059cadd78fce0f9b22ae127805/parachain/pallets/ethereum-beacon-client/src/lib.rs#L15

This is why we need to run the tests separately.

^ @yrong

alistair-singh (approved):
Looks good 🚀

.github/workflows/parachain.yml
yrong: @vgeddes Yes, it's reasonable for the seperation between  `mainnet-spec` and `minimal-spec` for package `snowbridge-ethereum-beacon-client`.  

But the seperation I refer to above is different for running benchmark tests between `ethereum-beacon-client` and other packages which seems somehow unnessessary.

But anyway, we can leave CI/devops stuff for a separate PR.

PR #972, opened 2023-10-18
Polkadot SDK Update V2

claravanstaden:
Companion: https://github.com/Snowfork/polkadot-sdk/pull/3

Fixes: SNO-707


claravanstaden:
Smoke tests work

- [x] Create agent
- [x] Create channel
- [ ] Transfer native from agent (needs to be tested)
- [ ] Upgrade gateway (needs fee)
- [x] Register token
- [ ] Send token (allowed reserve location xcm config to be fixed)
- [ ] Transfer token (depends on https://github.com/paritytech/polkadot-sdk/pull/1672)

smoketest/tests/transfer_token.rs
```diff
@@ -75,11 +80,26 @@ async fn transfer_token() {
...
+            Junction::AccountKey20 {
+                network: None,
+                key: DESTINATION_ADDRESS.into(),
+            },
         ),
     });
 
-    let bridge_transfer_call = TransactionApi.transfer_asset_via_bridge(assets, destination);
+    let bridge_transfer_call =
+        TransactionApi.reserve_transfer_assets(destination, beneficiary, assets, 0);
```
yrong, line 102: Seems not correct.  Actually the `0` is the index of fee asset which should be `DOT`, so the assets should be an array which contains both the transfer asset and fee asset.

claravanstaden: Yep, `transfer_token` [doesn't work yet](https://github.com/Snowfork/snowbridge/pull/972#issuecomment-1778921269). @vgeddes mentioned we can fix it after merge, since it is dependent on https://github.com/paritytech/polkadot-sdk/pull/1672 too.

parachain/pallets/inbound-queue/src/test.rs
```diff
@@ -164,7 +157,7 @@ impl SendXcm for MockXcmSender {
 		match dest {
 			Some(MultiLocation { interior, .. }) => {
 				if let X1(Parachain(1001)) = interior {
-					return Err(XcmpSendError::NotApplicable)
+					return Err(XcmpSendError::NotApplicable);
```
vgeddes, line 160: Are you sure your formatting settings are correct?

Both of my formatter's are not making this change:
`cargo +$RUST_NIGHTLY_VERSION fmt`
```"rust-analyzer.rustfmt.overrideCommand": [
        "rustup",
        "run",
        "nightly-2023-05-23",
        "--",
        "rustfmt",
        "--"
    ],
```

parachain/pallets/outbound-queue/src/lib.rs
```diff
@@ -242,7 +242,7 @@ pub mod pallet {
 		pub(crate) fn commit_messages() {
 			let count = MessageLeaves::<T>::decode_len().unwrap_or_default() as u64;
 			if count == 0 {
-				return
+				return;
```
vgeddes, line 245: Ditto, I suspect your formatter may be misconfigured.

parachain/primitives/router/src/outbound/mod.rs
```diff
@@ -283,17 +288,17 @@ impl<'a, Call> XcmConverter<'a, Call> {
...
+					return Err(AssetResolutionFailed);
 				}
 				if token_network.is_some() && token_network != &Some(*self.ethereum_network) {
-					return Err(AssetResolutionFailed)
+					return Err(AssetResolutionFailed);
 				}
 				(token_address.into(), *amount)
 			} else {
-				return Err(AssetResolutionFailed)
+				return Err(AssetResolutionFailed);
```
vgeddes, line 301: Ditto, your formatter may be misconfigured.

smoketest/tests/register_token.rs
```diff
@@ -1,27 +1,27 @@
...
-		runtime_types::xcm::v3::{
-			junction::{
-				Junction::{AccountKey20, GlobalConsensus},
-				NetworkId,
-			},
-			junctions::Junctions::X3,
-			multilocation::MultiLocation,
-		},
-	},
+    contracts::{i_gateway, weth9},
```
vgeddes, line 13: We tabs instead of spaces for indent.

I think we need to copy the `rustfmt.toml` from the parachains dir to this smoketests crate.

claravanstaden: Good point! https://github.com/Snowfork/snowbridge/pull/972/commits/cc9b1fc2aa6f50764287be80db7eb4e8666d4fe7

parachain/primitives/router/src/outbound/mod.rs
claravanstaden: Fixed in https://github.com/Snowfork/snowbridge/pull/972/commits/cc9b1fc2aa6f50764287be80db7eb4e8666d4fe7.

parachain/pallets/outbound-queue/src/lib.rs
claravanstaden: Fixed in https://github.com/Snowfork/snowbridge/pull/972/commits/cc9b1fc2aa6f50764287be80db7eb4e8666d4fe7.

parachain/pallets/inbound-queue/src/test.rs
claravanstaden: Apologies, I had trouble with my local setup where the formatter didn't work in the `polkadot-sdk` submodule, so I turned it off temporarily but forgot to turn it back in. Fixed in Fixed in https://github.com/Snowfork/snowbridge/pull/972/commits/cc9b1fc2aa6f50764287be80db7eb4e8666d4fe7..

smoketest/tests/register_token.rs
yrong: One single  `rustfmt.toml` in root dir(snowbridge) should applies to all sub modules. And just checked our rules are exactly the same as Parity's(which in polkadot-sdk).

parachain/primitives/router/src/outbound/mod.rs
```diff
@@ -134,11 +135,11 @@ where
 		// convert fee to MultiAsset
 		let fee = MultiAsset::from((MultiLocation::parent(), fee)).into();
 
-		Ok((ticket.encode(), fee))
+		Ok(((ticket.encode(), XcmHash::default()), fee))
```
yrong, line 138: Should be `message_id` for the ticket not the `XcmHash::default()`.

smoketest/tests/register_token.rs
claravanstaden: What do you run to format it then? If I run 

`chronic cargo +$RUST_NIGHTLY_VERSION fmt`

from the snowbridge dir I get ``cargo metadata` exited with an error: error: could not find `Cargo.toml` in `snowbridge` or any parent directory`

yrong: Just run in parachain dir, it will respect the config(rustfmt.toml or .rustfmt.toml) in any other parent directory. Or place the rule file in [global config directory](https://docs.rs/dirs/4.0.0/dirs/fn.config_dir.html) (e.g. .config/rustfmt/) will also do the job.

parachain/pallets/inbound-queue/src/test.rs
vgeddes: Looks like your formatter is acting weird again, for example in in `pallets/ethereum-beacon-client/src/lib.rs` 😄

parachain/primitives/router/src/outbound/mod.rs
vgeddes: I think we can fix this in a follow up PR, as we need to make some other related changes to message IDs anyway.

vgeddes (approved):
Nice work! Lets merge it in... 🚀

PR #975, opened 2023-10-20
Delivery fee should cover cost of local processing

yrong:
We need to take base fee(fee in bridgehub) into consideration, more context in https://github.com/Snowfork/snowbridge/pull/968#discussion_r1357725585


vgeddes:
This is a great start Ron! I like how you are using `WeightToFee` for accurate calculation.

Lets rather make this work its own standalone PR, as I really want to get `ron/sno-632` merged into today.

We need to add a couple of more things:

First, its important that the local "base" fee is collected into BridgeHub's treasury account. The delivery fee however must somehow end up into the sovereign account of the origin parachain. So that the origin can balance its sovereign account on Ethereum.

The problem is that https://github.com/paritytech/polkadot-sdk/pull/1234 is going to put the total "base fee + delivery fee" into BridgeHub's treasury account.

There is a solution however. In our [ExportXcm::deliver](https://github.com/Snowfork/snowbridge/blob/c3052038bd2ea56c466963c68b09974d6dc71614/parachain/primitives/router/src/outbound/mod.rs#L140) function, we assign to the origin parachain an amount of fungible "credits" equal to the delivery fee.

Then the origin parachain can call an extrinsic `EthereumControl::redeem` to redeem those credits for actual DOT from the BridgeHub treasury.

vgeddes:
> This is a great start Ron! I like how you are using `WeightToFee` for accurate calculation.
> 
> Lets rather make this work its own standalone PR, as I really want to get `ron/sno-632` merged into today.
> 
> We need to add a couple of more things:
> 
> First, its important that the local "base" fee is collected into BridgeHub's treasury account. The delivery fee however must somehow end up into the sovereign account of the origin parachain. So that the origin can balance its sovereign account on Ethereum.
> 
> The problem is that [paritytech/polkadot-sdk#1234](https://github.com/paritytech/polkadot-sdk/pull/1234) is going to put the total "base fee + delivery fee" into BridgeHub's treasury account.
> 
> There is a solution however. In our [ExportXcm::deliver](https://github.com/Snowfork/snowbridge/blob/c3052038bd2ea56c466963c68b09974d6dc71614/parachain/primitives/router/src/outbound/mod.rs#L140) function, we assign to the origin parachain an amount of fungible "credits" equal to the delivery fee.
> 
> Then the origin parachain can call an extrinsic `EthereumControl::redeem` to redeem those credits for actual DOT from the BridgeHub treasury.

I also mentioned this idea in https://matrix.to/#/!gxqZwOyvhLstCgPJHO:matrix.parity.io/$URHJPeCFmUb2byOeqCD0LTdobb7oUyNGOerpApgaMPc?via=matrix.parity.io&via=parity.io&via=matrix.org

parachain/pallets/control/src/lib.rs
```diff
@@ -390,22 +397,47 @@ pub mod pallet {
...
 			let message = Message { origin, command };
-			let (ticket, delivery_fee) =
+			let (ticket, fee) =
 				T::OutboundQueue::validate(&message).map_err(|_| Error::<T>::SubmissionFailed)?;
 
-			if let PaysFee::Yes(sovereign_account) = pays_fee {
+			let payment = match pays_fee {
+				PaysFee::Yes(account) =>
+					Some((account, fee.base_fee.saturating_add(fee.delivery_fee))),
+				PaysFee::Partial(account) => Some((account, fee.base_fee)),
```
yrong, line 432: Only charge `base_fee` for the partial case(e.g. transfer_native_from_agent).

parachain/pallets/outbound-queue/src/lib.rs
```diff
@@ -354,18 +364,31 @@ pub mod pallet {
 				payload.len() < T::MaxMessagePayloadSize::get() as usize,
 				SubmitError::MessageTooLarge
 			);
+
+			let base_fee = T::WeightToFee::weight_to_fee(
+				&T::WeightInfo::do_process_message()
+					.saturating_add(T::WeightInfo::do_commit_one_message()),
+			);
```
yrong, line 371: Use `WeightToFee` to estimate the `base_fee`, benchmark shows that `T::WeightInfo::do_commit_one_message()>T::WeightInfo::do_commit_messages()/T::MaxMessagesPerBlock::get()`, so we charge based upon the former.

vgeddes (commented):
Hey Ron, as per our meeting today, we should introduce a new pallet `refunds`, which has responsibility for:
* Adding vouchers
* Tracking vouchers for each sibling parachain
* Redeeming vouchers for a refund

The other pallets can communicate with the `refunds` pallet via a trait.

yrong:
> we should introduce a new pallet refunds

Hey Vincent, TBO I'm a bit hesitate to introduce a new pallet just for this which seems a bit overkill, since all fee related const/logic for estimation are already in outbound queue, I would prefer to maintain the vouchers also in outbound queue.

Or can we make things even more simple to not introduce the vouchers storage from the beginning? Is that possible just return the `base_fee` for `XcmFeeManager` to charge, https://github.com/Snowfork/snowbridge/blob/f76a4f7512552317b71299350c42948d7cb27b36/parachain/primitives/router/src/outbound/mod.rs#L137

We can still add a runtime api to return both the `base_fee` and `delivery_fee` for the source parachain to estimate the total fee cost. So the source chain can charge enough from the end user and use that to rebalance it's sovereign account on Ethereum by itself.

vgeddes:
Good point Ron, maybe this voucher/refund system is too complicated.

Alternatively, we could also refund the sovereign account of the origin parachain directly in `ExportXcm::deliver`.

Looking at the [implementation](https://github.com/paritytech/polkadot-sdk/blob/d0fd2660031b92a3896db3ea47e65f6f8af9a9ae/polkadot/xcm/xcm-executor/src/lib.rs#L847-L855) of `ExportMessage`, the executor takes the fees, and then it calls `ExportXcm::deliver`.

So it should be safe to transfer DOT from the treasury to the sovereign account in `deliver`.

parachain/primitives/router/src/outbound/mod.rs
```diff
@@ -132,7 +132,7 @@ where
 		})?;
 
 		// convert fee to MultiAsset
-		let fee = MultiAsset::from((MultiLocation::parent(), fee)).into();
+		let fee = MultiAsset::from((MultiLocation::parent(), fee.base_fee)).into();
```
yrong, line 135: We should return `base_fee` instead of the `delivery_fee` for the  `XcmManager` to charge.

yrong:
> maybe this voucher/refund system is too complicated.

Yes, and as mentioned above, I don't think all the voucher/refund stuff are nessessary at all. The only fix required is here https://github.com/Snowfork/snowbridge/pull/975#discussion_r1369346636

parachain/pallets/outbound-queue/src/lib.rs
```diff
@@ -354,18 +360,31 @@ pub mod pallet {
 				payload.len() < T::MaxMessagePayloadSize::get() as usize,
 				SubmitError::MessageTooLarge
 			);
+
+			let base_fee = T::WeightToFee::weight_to_fee(
+				&T::WeightInfo::do_submit_message()
+					.saturating_add(T::WeightInfo::do_process_message())
+					.saturating_add(T::WeightInfo::do_commit_one_message()),
+			);
```
yrong, line 368: So the base_fee has three components:

1. cost for submit message
2. cost for process message
3. cost for commit message

parachain/primitives/router/src/outbound/mod.rs
vgeddes: We do need to make sure that the computed origin is charged the delivery fee. Otherwise the parachain that is the parent of the computed origin needs to estimate the fees themselves - this approach I am skeptical of.

In either case, lets wait until this PR is ready: https://github.com/paritytech/polkadot-sdk/pull/2021/files

It allows to flexibly decide how to handle the fee using a `HandleFee` trait.

yrong: I do think it's the responsibility of source chain(i.e asset_hub or other sibling chain) to ensure all costs have been charged from their end user(signed origin).

More detail as we discussed before in https://snowfork.slack.com/archives/G01BGMGLAC9/p1696933144716239?thread_ts=1696931181.004119&cid=G01BGMGLAC9, the fee charged will be collected into treasury account of source chain and later use to rebalance their sovereign account on Ethereum.

But anyway, I'll stop here for the upstream PRs ready and we agree on the implementation.

vgeddes: I think we can offer parachains the option to charge fees from their users themselves. For example, in future, the `create_channel` and `update_channel` extrinsics can give parachains flexibility in choosing the model they want.

However, by default, I want an easy way for parachains to integrate with our bridge, and charging fees on BridgeHub makes that possible.

vgeddes:
Lets limit the scope of this PR, as it does solve a few important problems:
* Benchmarks for processing a message and committing a single message.
* Charging base_fee + delivery_fee (See my comment in https://github.com/Snowfork/snowbridge/pull/975#discussion_r1372100507 why I want this to be the default)
* Remove all voucher/redeem and estimate_fee runtime API.

parachain/pallets/control/src/tests.rs
```diff
@@ -506,3 +508,34 @@ fn check_sibling_sovereign_account() {
 		);
 	});
 }
+
+#[test]
+fn charge_fee() {
```
alistair-singh, line 513: All the happy case path tests such as the [create_agent test](https://github.com/Snowfork/snowbridge/pull/975/files#diff-494b86672bfad8881d114f1bb5765ab5ef9def9f532a51aa13b80b13f0b06714R9) should assert that the fee was charged or not charged as opposed to one test for all cases if possible. If not possible maybe we should have separate asserts for each case in this test.

alistair-singh (commented):
Looks good!

parachain/pallets/control/src/tests.rs
yrong: https://github.com/Snowfork/snowbridge/pull/975/commits/9037a8ce0bcb74a2b231d07c67f5cdb9a8350098

parachain/primitives/router/src/outbound/mod.rs
yrong: https://github.com/Snowfork/snowbridge/pull/975/commits/828063441d9ae5a2b0fbbdd331e608c646e09b78

yrong:
> Remove all voucher/redeem and estimate_fee runtime API.

Hey, Vincent.  Actually voucher/redeem has been removed in previous commit, but for `estimate_fee` I prefer to keep as it could be helpful for chains to estimate fee in advance before a real transfer. Also the return of the api include both `base_fee` and `delivery_fee` which is not obvious in current impl.

vgeddes:
> > Remove all voucher/redeem and estimate_fee runtime API.
> 
> Hey, Vincent. Actually voucher/redeem has been removed in previous commit, but for `estimate_fee` I prefer to keep as it could be helpful for chains to estimate fee in advance before a real transfer. Also the return of the runtime api includes both `base_fee` and `delivery_fee` which is not obvious in current impl.

OK, lets keep `estimate_fee`.

parachain/pallets/control/src/lib.rs
```diff
@@ -82,22 +88,18 @@ where
...
 /// Whether a fee should be withdrawn to an account for sending an outbound message
 #[derive(Clone, PartialEq, RuntimeDebug)]
-enum PaysFee<T>
+pub enum PaysFee<T>
 where
 	T: Config,
 {
 	Yes(AccountIdOf<T>),
+	Partial(AccountIdOf<T>),
 	No,
```
vgeddes, line 97: Needs doc comments please

parachain/pallets/control/src/mock.rs
```diff
@@ -146,7 +149,45 @@ impl pallet_xcm_origin::Config for Test {
 }
 
 parameter_types! {
+	pub const HeapSize: u32 = 32 * 1024;
+	pub const MaxStale: u32 = 32;
+	pub static ServiceWeight: Option<Weight> = Some(Weight::from_parts(100, 100));
+}
+
+impl pallet_message_queue::Config for Test {
```
vgeddes, line 157: Since we no longer need to test voucher/refund stuff, do we really still need to use a real MessageQueue in the tests?

parachain/pallets/outbound-queue/runtime-api/src/lib.rs
```diff
@@ -2,11 +2,14 @@
...
 
+use snowbridge_core::outbound::{Message, OutboundFee};
 use snowbridge_outbound_queue_merkle_tree::MerkleProof;
 
 sp_api::decl_runtime_apis! {
 	pub trait OutboundQueueApi
 	{
 		fn prove_message(leaf_index: u64) -> Option<MerkleProof>;
+
+		fn estimate_fee(message: Message) -> Option<OutboundFee<u128>>;
```
vgeddes, line 13: Prefer this name instead: `calculate_fee`.

parachain/pallets/outbound-queue/src/api.rs
```diff
@@ -19,3 +21,14 @@ where
 	);
 	Some(proof)
 }
+
+pub fn estimate_fee<Runtime>(message: Message) -> Option<OutboundFee<u128>>
```
vgeddes, line 25: prefer this name: `calculate_fee`.

parachain/primitives/core/src/outbound.rs
```diff
@@ -16,34 +16,31 @@ pub type FeeAmount = u128;
 pub type GasAmount = u128;
 pub type GasPriceInWei = u128;
 
+/// OutboundFee which covers the cost of execution on both BridgeHub and Ethereum
+#[derive(Copy, Clone, Default, Encode, Decode, PartialEq, Eq, RuntimeDebug, TypeInfo)]
+pub struct OutboundFee<Balance: BalanceT> {
+	/// Cover the cost on BridgeHub
```
vgeddes, line 22:
```suggestion
	/// Fee for processing the message locally
```

parachain/primitives/core/src/outbound.rs
```diff
@@ -16,34 +16,31 @@ pub type FeeAmount = u128;
...
+	pub base_fee: Balance,
```
vgeddes, line 23: Lets rename `base_fee` to `acceptance`, as `base` is way too generic.

parachain/primitives/core/src/outbound.rs
```diff
@@ -16,34 +16,31 @@ pub type FeeAmount = u128;
...
+	/// Cover the cost on Ethereum
```
vgeddes, line 24:
```suggestion
	/// Fee for delivering the message to Ethereum for execution
```

parachain/primitives/router/src/outbound/mod.rs
```diff
@@ -132,7 +132,11 @@ where
 		})?;
 
 		// convert fee to MultiAsset
-		let fee = MultiAsset::from((MultiLocation::parent(), fee)).into();
+		let fee = MultiAsset::from((
+			MultiLocation::parent(),
+			fee.base_fee.saturating_add(fee.delivery_fee),
```
vgeddes, line 137: I would rather we add a helper function `Fees::total`.

```suggestion
			fees.total(),
```

parachain/primitives/core/src/outbound.rs
```diff
@@ -16,34 +16,31 @@ pub type FeeAmount = u128;
 pub type GasAmount = u128;
 pub type GasPriceInWei = u128;
 
+/// OutboundFee which covers the cost of execution on both BridgeHub and Ethereum
+#[derive(Copy, Clone, Default, Encode, Decode, PartialEq, Eq, RuntimeDebug, TypeInfo)]
+pub struct OutboundFee<Balance: BalanceT> {
```
vgeddes, line 21: Lets rename this to `Fees`.

This struct is already in an `outbound` namespace, so no need to have `Outbound` in the name. Also there multiple sub-fees, so lets make the name plural.

parachain/primitives/core/src/outbound.rs
```diff
@@ -16,34 +16,31 @@ pub type FeeAmount = u128;
...
+	/// Cover the cost on BridgeHub
+	pub base_fee: Balance,
+	/// Cover the cost on Ethereum
+	pub delivery_fee: Balance,
```
vgeddes, line 25: The struct already has the word `Fees` in it.

```suggestion
	pub delivery: Balance,
```

parachain/pallets/outbound-queue/src/lib.rs
vgeddes: Hmm, actually, I don't think the base fee should include the cost for submitting a message.

The extrinsics in the control pallet are all benchmarked, and so the weight of those extrinsics include the weight of submitting a message. So if we include the cost again, we're essentially double-billing users.

Its only process_message and commit message which we need to benchmark

parachain/pallets/outbound-queue/src/lib.rs
```diff
@@ -239,7 +243,7 @@ pub mod pallet {
 
 	impl<T: Config> Pallet<T> {
 		/// Generate a messages commitment and insert it into the header digest
-		pub(crate) fn commit_messages() {
+		pub(crate) fn do_commit_messages() {
```
vgeddes, line 246: This name change is unnecessary.

I only named `do_process_message` that way so as not to clash with the existing `process_message`.

parachain/pallets/outbound-queue/src/lib.rs
yrong: [Remove benchmark](https://github.com/Snowfork/snowbridge/pull/975/commits/3f685eae097d6b69b2295c78022000cf549040fd)

parachain/pallets/control/src/lib.rs
yrong: [Add comments](https://github.com/Snowfork/snowbridge/pull/975/commits/95d673a11ab13f0b59a6be72be049555b6cacc81)

parachain/pallets/control/src/mock.rs
yrong: It's required for the new tests added in https://github.com/Snowfork/snowbridge/commit/9037a8ce0bcb74a2b231d07c67f5cdb9a8350098

parachain/pallets/outbound-queue/runtime-api/src/lib.rs
yrong: [Rename to calculate_fee](https://github.com/Snowfork/snowbridge/pull/975/commits/8d6ec9d2774444a4ae356dbb9803ca67eb186475)

parachain/primitives/core/src/outbound.rs
yrong: [Comments](https://github.com/Snowfork/snowbridge/pull/975/commits/ae1cbd5b0326461fab4de40f701b3b803280fc23)

parachain/primitives/router/src/outbound/mod.rs
yrong: [Add method for total fee](https://github.com/Snowfork/snowbridge/pull/975/commits/727288d67a7a073294a1b10122cad6ce02e9f45a)

parachain/primitives/core/src/outbound.rs
yrong: [Short name](https://github.com/Snowfork/snowbridge/pull/975/commits/f864192282433e928493e657ac2a23800f6af305)

parachain/pallets/outbound-queue/src/lib.rs
yrong: [Rename to commit_messages](https://github.com/Snowfork/snowbridge/pull/975/commits/6691413ab99ce893189b3409249541084141c42f)

parachain/pallets/outbound-queue/runtime-api/src/lib.rs
```diff
@@ -2,11 +2,14 @@
...
 
+use snowbridge_core::outbound::{Fees, Message};
 use snowbridge_outbound_queue_merkle_tree::MerkleProof;
 
 sp_api::decl_runtime_apis! {
 	pub trait OutboundQueueApi
 	{
 		fn prove_message(leaf_index: u64) -> Option<MerkleProof>;
+
+		fn calculate_fee(message: Message) -> Option<Fees<u128>>;
```
vgeddes, line 13: Should make the fee type a generic parameter, as the balance type is configured in the runtime crate.

```suggestion
		fn calculate_fee(message: Message) -> Option<Fees<Balance>>;
```

parachain/pallets/outbound-queue/runtime-api/src/lib.rs
```diff
@@ -2,11 +2,14 @@
 // SPDX-FileCopyrightText: 2023 Snowfork <hello@snowfork.com>
 #![cfg_attr(not(feature = "std"), no_std)]
 
+use snowbridge_core::outbound::{Fees, Message};
 use snowbridge_outbound_queue_merkle_tree::MerkleProof;
 
 sp_api::decl_runtime_apis! {
 	pub trait OutboundQueueApi
```
vgeddes, line 9:
```suggestion
	pub trait OutboundQueueApi<Balance> where Balance: BalanceT {
```

parachain/pallets/outbound-queue/src/api.rs
```diff
@@ -19,3 +21,14 @@ where
 	);
 	Some(proof)
 }
+
+pub fn calculate_fee<Runtime>(message: Message) -> Option<Fees<u128>>
```
vgeddes, line 25: Need to use `Balance` generic parameter instead of `u128`.

parachain/pallets/outbound-queue/runtime-api/src/lib.rs
yrong: [Use Balance generic for the api](https://github.com/Snowfork/snowbridge/pull/975/commits/61d1df5e6b8a208e50998d19fde9a026d6df8afa)

parachain/pallets/outbound-queue/runtime-api/src/lib.rs
```diff
@@ -2,11 +2,15 @@
...
+use snowbridge_core::outbound::{Fees, Message, SubmitError};
 use snowbridge_outbound_queue_merkle_tree::MerkleProof;
 
 sp_api::decl_runtime_apis! {
-	pub trait OutboundQueueApi
+	pub trait OutboundQueueApi<Balance> where Balance: BalanceT
 	{
 		fn prove_message(leaf_index: u64) -> Option<MerkleProof>;
+
+		fn calculate_fee(message: Message) -> Result<Fees<Balance>,SubmitError>;
```
vgeddes, line 14: Missing a space:

```suggestion
		fn calculate_fee(message: Message) -> Result<Fees<Balance>, SubmitError>;
```

parachain/pallets/outbound-queue/src/api.rs
```diff
@@ -19,3 +20,12 @@ where
 	);
 	Some(proof)
 }
+
+pub fn calculate_fee<Runtime>(
+	message: Message,
+) -> Result<Fees<<Runtime as Config>::Balance>, SubmitError>
```
vgeddes, line 26: Hmm, actually, should rename `Runtime` to `T`. They are the same thing really. And I think its not necessary to cast Runtime into Config.

```suggestion
pub fn calculate_fee<T>(
	message: Message,
) -> Result<Fees<T::Balance>, SubmitError>
```

parachain/pallets/outbound-queue/src/api.rs
```diff
@@ -4,6 +4,7 @@
 
 use crate::{Config, MessageLeaves};
 use frame_support::storage::StorageStreamIter;
+use snowbridge_core::outbound::{Fees, Message, OutboundQueue, SubmitError};
 use snowbridge_outbound_queue_merkle_tree::{merkle_proof, MerkleProof};
 
 pub fn prove_message<Runtime>(leaf_index: u64) -> Option<MerkleProof>
```
vgeddes, line 10: ditto, lets fix this up too.

```suggestion
pub fn prove_message<T>(leaf_index: u64) -> Option<MerkleProof>
```

parachain/pallets/outbound-queue/src/lib.rs
```diff
@@ -354,18 +358,30 @@ pub mod pallet {
...
 			let encoded =
 				enqueued_message.encode().try_into().map_err(|_| SubmitError::MessageTooLarge)?;
 
-			let ticket =
-				OutboundQueueTicket { id: message_id, origin: message.origin, message: encoded };
+			let ticket = OutboundQueueTicket {
+				id: message_id,
+				origin: message.origin,
+				message: encoded,
+				fee,
```
vgeddes, line 381: I'm not seeing the `OutboundQueueTicket.fee` field being used anywhere. Can it removed?

parachain/primitives/router/src/outbound/mod.rs
```diff
@@ -132,7 +132,7 @@ where
 		})?;
 
 		// convert fee to MultiAsset
-		let fee = MultiAsset::from((MultiLocation::parent(), fee)).into();
+		let fee = MultiAsset::from((MultiLocation::parent(), fee.total())).into();
```
vgeddes, line 135:
```suggestion
		let fee = MultiAsset::from((MultiLocation::parent(), fees.total())).into();
```

vgeddes (commented):
Last round of minor comments

parachain/primitives/router/src/outbound/mod.rs
yrong: [Rename to fees](https://github.com/Snowfork/snowbridge/pull/975/commits/356d17c9b8c57a14e6d411185d1473c942b4579d)

parachain/pallets/outbound-queue/src/lib.rs
yrong: [Remove fee from OutboundQueueTicket](https://github.com/Snowfork/snowbridge/pull/975/commits/8fa7dacdbe5ea9c260b3da7af7706a8955139dec)

parachain/pallets/outbound-queue/src/api.rs
yrong: [Rename generic Runtime to T](https://github.com/Snowfork/snowbridge/pull/975/commits/5548e949c2affcb7bd70fc5c912793013346e1d5)

vgeddes (approved):

PR #977, opened 2023-10-25
Improve gas metering

vgeddes:
* Give more gas to `AgentExecuteCommand::TokenTransfer` as `IERC.transferFrom` can consume more than 40000 gas under worst case assumptions
* Make the `testAgentExecution` unit test consume gas more realistically in `IERC.transferFrom`.


contracts/src/Gateway.sol
```diff
@@ -124,12 +128,10 @@ contract Gateway is IGateway, IInitializable {
...
-        // Otherwise malicious relayers can break the bridge by allowing the message handlers below to run out gas and fail silently.
-        // In this scenario case, the channel's state would have been updated to accept the message (by virtue of the nonce increment), yet the actual message
-        // dispatch would have failed
+        // Make sure relayers provide enough gas so that message dispatch
+        // does not run out of gas.
         uint256 maxDispatchGas = message.maxDispatchGas;
-        if (gasleft() < maxDispatchGas + BUFFER_GAS) {
+        if (gasleft() < maxDispatchGas + DISPATCH_OVERHEAD_GAS) {
             revert NotEnoughGas();
         }
```
yrong, line 136: It's a bit weird with the name `maxDispatchGas` and check  `gasleft()>max****`, would prefer to rename as `dispatchGasRequired`.

There could be some `maxDispatchGas` when arbitrary transact(i.e. arbitray dispatchGas) introduced and check `dispatchGasRequired<maxDispatchGas<gasLimtOfRelayer` or the bridge will get stuck.

vgeddes: The better name would actually be `maxRequiredDispatchGas`, but I don't think this is a priority for us now.

The problem with the `dispatchGasRequired`, is that it says that the dispatch requires a _minimum_ amount. However, the field in reality actually says that the dispatch requires a _maximum_ amount.

Anyway, this naming is orthogonal to the issue I'm trying to fix, so resolving it.

PR #981, opened 2023-10-30
Support Dencun consensus upgrade

yrong:
Resolves: SNO-406, SNO-775

- [x] Upgrade [geth](https://github.com/ethereum/go-ethereum/releases/tag/v1.13.5) and [lodestar](https://github.com/ChainSafe/lodestar/releases/tag/v1.12.0) to recent release
- [x] Add `VersionedExecutionPayloadHeader` to support multiple versions of the beacon updates
- [x] Cleanup for unused codes/packages(e.g. package `ethashproof` based on the deprecated POW consensus which can be removed).
- [x] Test e2e work as expected 
- [x] Test the switchover process


vgeddes:
Nice work Ron! 🙏

yrong:
@vgeddes I would prefer to split https://linear.app/snowfork/issue/SNO-406 into multiple PRs, in this one we only focus on upgrading packages and add versioned data support in next one.

vgeddes (changes requested):
I've just noticed that there are a lot of unrelated changes to the E2E scripts, like making processes run in parallel etc. I vaguely recall previously reviewing such a change and saying that it makes troubleshooting harder if the E2E tests break.

The other reason for not wanting such changes is that roughly 80% of the initialization time is spent in compiling rust code. And the parallelization changes you're introducing here don't improve that. So I think the net gain is minimal.

So I'd like for you to please focus this PR on only the code relevant to the Dencun upgrades.

yrong:
> there are a lot of unrelated changes to the E2E scripts, like making processes run in parallel etc.

[Remove unrelated changes](https://github.com/Snowfork/snowbridge/pull/981/commits/28f23ddaa3d7be7234922898d8ededdf5712d24e)

relayer/go.mod
```diff
@@ -1,82 +1,124 @@
...
+	gopkg.in/ini.v1 v1.67.0 // indirect
 	gopkg.in/natefinch/npipe.v2 v2.0.0-20160621034901-c1b8fa8bdcce // indirect
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
+	rsc.io/tmplfunc v0.0.3 // indirect
 )
+
+replace github.com/tranvictor/ethutils => github.com/Snowfork/ethutils v0.0.0-20231121055358-90d4f9fa98fd
+
+replace github.com/snowfork/ethashproof => github.com/yrong/ethashproof v0.0.0-20231121055841-7e61e22d5099
```
vgeddes, line 124: This is an obsolete dependency that was used to generate proofs of POW for our previous ethereum light client. We should remove any code from the relayer that uses it.

.gitmodules
```diff
@@ -13,3 +13,9 @@
 [submodule "contracts/lib/canonical-weth"]
 	path = contracts/lib/canonical-weth
 	url = https://github.com/Snowfork/canonical-weth
+[submodule "lodestar"]
+    path = lodestar
+    url = https://github.com/ChainSafe/lodestar
+[submodule "go-ethereum"]
+    path = go-ethereum
+    url = https://github.com/ethereum/go-ethereum
```
vgeddes, line 21: The convention is that we use Nix where possible for local E2E dependencies. I believe we've discussed this before.

Nix has support for a very recent go-ethereum: https://search.nixos.org/packages?channel=unstable&show=go-ethereum&from=0&size=50&sort=relevance&type=packages&query=go-ethereum

.gitmodules
```diff
@@ -13,3 +13,9 @@
 [submodule "contracts/lib/canonical-weth"]
 	path = contracts/lib/canonical-weth
 	url = https://github.com/Snowfork/canonical-weth
+[submodule "lodestar"]
```
vgeddes, line 16: Why are we making this a submodule?

relayer/go.mod
claravanstaden: There's actually quite a bit of cleanup that we can do in the relayer to remove unused code, like here: https://github.com/Snowfork/snowbridge/blob/main/relayer/chain/ethereum/header.go#L89

.gitmodules
yrong: Last time when I upgrade lodestar to v1.11 met an issue which breaks the api of dumping beacon state we rely on to generate the ancestry proof,  made a fix in https://github.com/Snowfork/lodestar/pull/1 to resolve it. 

Since it's already fixed in v1.12 and just tested that e2e smoke tests work as expected so yes we do not need the fork anymore.

But that's somehow explains the reason why I prefer to fork the upstream repos like geth and lodestar, it allows us to test with the most recent version and make any fix or hack as we want, like the [hack](https://github.com/Snowfork/snowbridge/blob/cce771451a53260523bc5da6efc46c60ebb51f10/web/packages/test/scripts/build-binary.sh#L41-L49) for a fast block time to speedup the e2e test.

relayer/go.mod
yrong: Yea! I'll try to remove the unused and make some cleanup.

.gitmodules
yrong: What I'm testing with in this PR is 1.13.5 https://github.com/Snowfork/snowbridge/blob/cce771451a53260523bc5da6efc46c60ebb51f10/relayer/go.mod#L7
but I think 1.13.4 as the recent nix package in the search result should also be fine.

Some other reason why I prefer to fork the upstream repos in https://github.com/Snowfork/snowbridge/pull/981#discussion_r1401401017

relayer/go.mod
yrong: Some cleanup in https://github.com/Snowfork/snowbridge/pull/981/commits/75e3342faa0a749aca7b9aeec42721798e6102d5

relayer/relays/beacon/header/syncer/syncer.go
```diff
@@ -394,15 +394,17 @@ func (s *Syncer) GetHeaderUpdate(blockRoot common.Hash, checkpoint *cache.Proof)
...
 		return scale.HeaderUpdatePayload{}, err
 	}
 
-	executionHeaderBranch, err := s.getExecutionHeaderBranch(block)
+	// Todo: Upgrade to Deneb based on fork epoch
+	executionPayloadScale, err := api.CapellaExecutionPayloadToScale(block.GetExecutionPayload(), s.activeSpec)
 	if err != nil {
 		return scale.HeaderUpdatePayload{}, err
 	}
+	versionedExecutionPayloadHeader := scale.VersionedExecutionPayloadHeader{Capella: &executionPayloadScale}
```
yrong, line 407: @vgeddes as you mentioned in this PR we just introduced `VersionedExecutionPayloadHeader` and made `Capella` the only item for now and left `Deneb` support to future PRs.

vgeddes (commented):
Looks good. Lets hold off on merging though. We have an audit starting on the 4th Dec, and I need to ask the auditors whether they will have time to audit these changes.

parachain/Cargo.toml
```diff
@@ -14,7 +14,6 @@ members = [
     "pallets/system/runtime-api",
     "runtime/runtime-common",
     "runtime/rococo-common",
-    "runtime/tests",
```
yrong, line 17: Seems we add runtime tests back in previous PR, would suggest removing it or will take a long time to run `cargo` commands only for our pallets/codes.

web/packages/test/scripts/deploy-ethereum.sh
```diff
@@ -6,7 +6,26 @@ source scripts/set-env.sh
...
+            else
+                timestamp=$(date -d'+1000second' +%s)
+            fi
+        fi
+        jq \
+            --argjson timestamp "$timestamp" \
+            '
+            .config.CancunTime = $timestamp
+            ' \
+            config/genesis.json >$output_dir/genesis.json
```
yrong, line 28: The hack here is to test the switchover process. The point here is that we start with a `Shapella` setup and after 180 seconds switch to `Decun`, then we check that the beacon relayer works just as expected.

relayer/relays/beacon/header/syncer/scale/beacon_scale.go
```diff
@@ -294,3 +294,20 @@ type BeaconState struct {
...
+
+func (v VersionedExecutionPayloadHeader) Encode(encoder scale.Encoder) error {
+	var err error
+	if v.Capella != nil {
+		encoder.PushByte(0)
+		err = encoder.Encode(v.Capella)
+	} else if v.Deneb != nil {
+		encoder.PushByte(1)
+		err = encoder.Encode(v.Deneb)
+	}
```
alistair-singh, line 311: We should have an else here that sets `err` if no branch is hit.

parachain/runtime/tests/Cargo.toml
```diff
@@ -195,36 +197,19 @@ runtime-benchmarks = [
...
+snowbridge-system = { path = "../../pallets/system" }
+snowbridge-system-runtime-api = { path = "../../pallets/system/runtime-api" }
+snowbridge-core = { path = "../../primitives/core" }
+snowbridge-ethereum-beacon-client = { path = "../../pallets/ethereum-beacon-client" }
+snowbridge-inbound-queue = { path = "../../pallets/inbound-queue" }
+snowbridge-outbound-queue = { path = "../../pallets/outbound-queue" }
+snowbridge-outbound-queue-runtime-api = { path = "../../pallets/outbound-queue/runtime-api" }
+snowbridge-router-primitives = { path = "../../primitives/router" }
+snowbridge-runtime-common = { path = "../../runtime/runtime-common" }
+snowbridge-rococo-common = { path = "../../runtime/rococo-common" }
```
alistair-singh, line 215: Nit: ws

parachain/runtime/tests/Cargo.toml
```diff
@@ -4,6 +4,8 @@ version = "0.1.0"
 authors = ["Snowfork <contact@snowfork.com>"]
 edition = "2021"
 
+[workspace]
```
alistair-singh, line 7: What does this line do?

relayer/relays/beacon/header/syncer/scale/beacon_scale.go
yrong: https://github.com/Snowfork/snowbridge/pull/981/commits/9673b21dfa790eb9a4fc6d3ea2ea50e02e2e9196

web/packages/test/scripts/build-binary.sh
```diff
@@ -14,8 +14,8 @@ build_binaries() {
     check_local_changes "polkadot"
     check_local_changes "substrate"
 
-  # Check that all 3 binaries are available and no changes made in the polkadot and substrate dirs
-    if [[ ! -e "target/release/polkadot" || ! -e "target/release/polkadot-execute-worker" || ! -e "target/release/polkadot-prepare-worker" || "$changes_detected" -eq 1 ]]; then
```
yrong, line 18: Remove the change detect for building the polkadot binary, IMO in most cases we only change the cumulus codes, and building polkadot is unnecessary.  In case of a major substrate version upgrade, we can build Polkadot manually.

parachain/runtime/tests/Cargo.toml
yrong: Removed.

relayer/crypto/sr25519/sr25519.go
```diff
@@ -27,7 +27,7 @@ func GenerateKeypair(network uint8) (*Keypair, error) {
 	return NewKeypairFromSeed("//"+hexutil.Encode(data), network)
 }
 
-func NewKeypairFromSeed(seed string, network uint8) (*Keypair, error) {
+func NewKeypairFromSeed(seed string, network uint16) (*Keypair, error) {
```
vgeddes, line 30: Ditto, seems unrelated.

parachain/pallets/ethereum-beacon-client/src/lib.rs
```diff
@@ -549,10 +549,12 @@ pub mod pallet {
 			);
 
 			// Checks that we don't skip execution headers, they need to be imported sequentially.
+			let compact_execution_header: CompactExecutionHeader =
```
vgeddes, line 552: The changes you made here in `process_execution_header_update` seem unnecessary. I'd like to reduce the amount of changes if at all possible

parachain/primitives/beacon/src/lib.rs
```diff
@@ -5,9 +5,11 @@
 pub mod bits;
 pub mod bls;
 pub mod config;
+pub mod deneb;
```
vgeddes, line 8: I think the module layout can be improved. Its confusing to add a `deneb` module at the top level. It should be a child of `types`.

For example:
```rust
use crate::types::capella::ExecutionPayloadHeader
use crate::types::deneb::ExecutionPayloadHeader
```

parachain/primitives/beacon/src/lib.rs
```diff
@@ -5,9 +5,11 @@
...
 pub mod merkle_proof;
 pub mod receipt;
 pub mod ssz;
+pub mod ssz_deneb;
```
vgeddes, line 12: Ditto, this module layout can be improved.

For example:
```rust
use crate::ssz::capella::SSZExecutionPayloadHeader
use crate::types::deneb::SSZExecutionPayloadHeader
```

parachain/primitives/beacon/src/types.rs
```diff
@@ -386,6 +391,55 @@ pub struct CompactBeaconState {
...
+#[cfg_attr(
+	feature = "std",
+	derive(Serialize, Deserialize),
+	serde(deny_unknown_fields, bound(serialize = ""), bound(deserialize = ""))
+)]
+#[codec(mel_bound())]
+pub enum VersionedExecutionPayloadHeader {
+	Capella(ExecutionPayloadHeader),
+	Deneb(ExecutionPayloadHeaderDeneb),
+}
```
vgeddes, line 405: Rather use modules as a namespace, rather than adding `Deneb` as a suffix.

```suggestion
pub enum VersionedExecutionPayloadHeader {
	Capella(capella::ExecutionPayloadHeader),
	Deneb(deneb::ExecutionPayloadHeader),
}
```

go.work
```diff
@@ -1,3 +1,6 @@
 go 1.20
 
-use ./relayer
+use (
+	./go-ethereum
```
vgeddes, line 4: Just curious why this is change necessary? I am not too familiar with `go.work` files.

parachain/pallets/ethereum-beacon-client/src/mock.rs
```diff
@@ -86,11 +86,40 @@ pub mod minimal {
 				version: [3, 0, 0, 1], // 0x03000001
 				epoch: 0,
 			},
+			deneb: Fork {
+				version: [4, 0, 0, 1], // 0x04000001
+				epoch: 4294967295,
+			},
+		};
+		pub const DenebChainForkVersions: ForkVersions = ForkVersions{
```
vgeddes, line 94:
```suggestion
		pub const DenebChainForkVersions: ForkVersions = ForkVersions {
```

parachain/pallets/ethereum-beacon-client/src/mock.rs
```diff
@@ -86,11 +86,40 @@ pub mod minimal {
...
+			deneb: Fork {
+				version: [4, 0, 0, 1], // 0x04000001
+				epoch: 0,
+			},
 		};
 	}
 
 	impl ethereum_beacon_client::Config for Test {
 		type RuntimeEvent = RuntimeEvent;
+		#[cfg(feature = "beacon-spec-minimal-deneb")]
```
vgeddes, line 120: I don't see these features defined in `Cargo.toml`. Is it obsolete code?

Either way, I would prefer not to add these features. Perhaps there is another solution.

relayer/crypto/sr25519/sr25519.go
```diff
@@ -18,7 +18,7 @@ type Keypair struct {
 	keyringPair *signature.KeyringPair
 }
 
-func GenerateKeypair(network uint8) (*Keypair, error) {
+func GenerateKeypair(network uint16) (*Keypair, error) {
```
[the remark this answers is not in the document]
yrong: Removed in https://github.com/Snowfork/snowbridge/pull/981/commits/d1a5b8648f7f4b44041f3b9260667989a290cba2

relayer/crypto/sr25519/sr25519.go
yrong: Removed in https://github.com/Snowfork/snowbridge/pull/981/commits/d1a5b8648f7f4b44041f3b9260667989a290cba2

parachain/pallets/ethereum-beacon-client/src/lib.rs
yrong: Clean in https://github.com/Snowfork/snowbridge/pull/981/commits/3580c4de7d6400880f3a4df171ffb42b5000737b

parachain/primitives/beacon/src/lib.rs
yrong: Reorganized in https://github.com/Snowfork/snowbridge/pull/981/commits/3580c4de7d6400880f3a4df171ffb42b5000737b

parachain/primitives/beacon/src/lib.rs
yrong: https://github.com/Snowfork/snowbridge/pull/981/commits/3580c4de7d6400880f3a4df171ffb42b5000737b

Though `use crate::ssz::deneb::SSZExecutionPayloadHeader` instead because there are some conflict imports between `ssz_rs` and `sp_core` which causes a bit messy to merge.

parachain/primitives/beacon/src/types.rs
yrong: Reorganized in https://github.com/Snowfork/snowbridge/commit/3580c4de7d6400880f3a4df171ffb42b5000737b

go.work
yrong: Not necessary just removed.

parachain/pallets/ethereum-beacon-client/src/mock.rs
yrong: Yeah, obsolete and removed.
https://github.com/Snowfork/snowbridge/pull/981/commits/53e4fa68f9f3408e6f7c1147b28f8625566f687e

yrong: Seems Decun will be activated on testnet soon(end of this Month?). I'll regenerate all the test fixtures and do some more tests then, though it does not block merging this PR.

vgeddes (approved):
+1.

relayer/relays/beacon/header/syncer/scale/beacon_scale.go
```diff
@@ -294,3 +295,26 @@ type BeaconState struct {
...
+	ErrEncodeVersionedExecutionPayloadHeader = errors.New("error scale encode VersionedExecutionPayloadHeader")
+)
+
+func (v VersionedExecutionPayloadHeader) Encode(encoder scale.Encoder) error {
+	var err error
+	if v.Capella != nil {
+		encoder.PushByte(0)
+		err = encoder.Encode(v.Capella)
+	} else if v.Deneb != nil {
+		encoder.PushByte(1)
```
claravanstaden, line 314: Why is this byte prefix needed?

relayer/relays/beacon/header/syncer/util/util.go
```diff
@@ -48,6 +48,18 @@ func ToUint64(stringVal string) (uint64, error) {
 	return intVal, err
 }
 
+func ToUint64AllowEmpty(stringVal string) (uint64, error) {
```
claravanstaden, line 51: Is this used somewhere?

claravanstaden (commented):
Hey @yrong! Great job, this is awesome!

I made a few comments. I would have preferred the relayer cleanup (removing the Ethereum PoW code) in a separate PR, but what's done is done.

I am not convinced adding geth and lodestar as submodules is necessary. Did we make any changes at all that warrants using forked versions? Seems like unnecessary maintenance and I would prefer we use the original projects.

web/packages/test/scripts/deploy-ethereum.sh
```diff
@@ -60,22 +81,13 @@ start_lodestar() {
             --params.ALTAIR_FORK_EPOCH 0 \
             --params.BELLATRIX_FORK_EPOCH 0 \
             --params.CAPELLA_FORK_EPOCH 0 \
-            --params.DEPOSIT_CHAIN_ID 11155111 \
-            --params.DEPOSIT_NETWORK_ID 11155111 \
```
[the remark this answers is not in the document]
yrong: I tested the local setup without these and it just works fine so I think it's not required.

web/packages/test/scripts/deploy-ethereum.sh
```diff
@@ -60,22 +81,13 @@ start_lodestar() {
...
+            --params.DENEB_FORK_EPOCH 20 \
             --eth1=true \
             --rest.namespace="*" \
-            --jwt-secret config/jwtsecret \
+            --jwt-secret $config_dir/jwtsecret \
+            --chain.archiveStateEpochFrequency 1 \
```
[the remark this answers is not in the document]
yrong: Without `chain.archiveStateEpochFrequency=1` the debug beacon state API won't work, more details in https://github.com/ChainSafe/lodestar/issues/5846#issuecomment-1729847506

.gitmodules
```diff
@@ -16,3 +16,9 @@
 [submodule "contracts/lib/prb-math"]
 	path = contracts/lib/prb-math
 	url = https://github.com/PaulRBerg/prb-math
+[submodule "lodestar"]
+	path = lodestar
+	url = https://github.com/Snowfork/lodestar
+[submodule "go-ethereum"]
+	path = go-ethereum
+	url = https://github.com/Snowfork/go-ethereum
```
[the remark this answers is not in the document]
yrong: I prefer to maintain a forked version for the reason https://github.com/Snowfork/snowbridge/pull/981#discussion_r1401401017

Btw: There is one commit in https://github.com/Snowfork/lodestar/commit/24851a96b0dfdb19b22f1ccc92a88b53cabef6ae to make beacon node work in mock mode without requiring the execution node(geth).

parachain/pallets/ethereum-client/src/mock.rs
```diff
@@ -263,6 +267,10 @@ pub mod mainnet {
 				version: [144, 0, 0, 114], // 0x90000072
 				epoch: 56832,
 			},
+			deneb: Fork {
+				version: [4, 0, 16, 32], // 0x04001020
+				epoch: 4294967295,
```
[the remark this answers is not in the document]
yrong: Yeah, just a placeholder and I'll update it when the date of deneb upgrading get finalized.

relayer/go.mod
```diff
@@ -4,79 +4,109 @@ go 1.20
...
+	golang.org/x/net v0.20.0 // indirect
+	golang.org/x/sys v0.16.0 // indirect
+	golang.org/x/text v0.14.0 // indirect
+	golang.org/x/tools v0.17.0 // indirect
+	google.golang.org/protobuf v1.32.0 // indirect
+	gopkg.in/ini.v1 v1.67.0 // indirect
 	gopkg.in/natefinch/npipe.v2 v2.0.0-20160621034901-c1b8fa8bdcce // indirect
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
+	rsc.io/tmplfunc v0.0.3 // indirect
```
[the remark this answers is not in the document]
yrong: Guess it's because I upgraded the `github.com/ethereum/go-ethereum` library and all transitive dependencies.

relayer/relays/beacon/header/syncer/scale/beacon_scale.go
yrong: It's required to match the codec encoding of `VersionedExecutionPayloadHeader` in the beacon light client.

relayer/relays/beacon/header/syncer/util/util.go
yrong: Yeah, used [here](https://github.com/Snowfork/snowbridge/blob/61dadc99ac650d5c424c033121df7169effd6f61/relayer/relays/beacon/header/syncer/api/api_response.go#L652) for backward compatible.

.gitmodules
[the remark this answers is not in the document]
claravanstaden: We need to use geth for mock mode to deploy contracts, etc, so is this even a benefit?

Also, adding these dependencies as submodules makes our already large project even larger. Let's maybe discuss in our meeting today.

relayer/chain/ethereum/trie.go
```diff
@@ -12,7 +14,7 @@ import (
 
 func MakeTrie(items types.Receipts) (*trie.Trie, error) {
 
-	trie := new(trie.Trie)
+	trie := trie.NewEmpty(trie.NewDatabase(rawdb.NewMemoryDatabase(), nil))
```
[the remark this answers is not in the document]
claravanstaden: I guess because of the go-ethereum upgrade?

relayer/relays/beacon/header/syncer/scale/beacon_scale.go
claravanstaden: Can't we just use the fork version to determine which version of the header to use?

web/packages/test/scripts/deploy-ethereum.sh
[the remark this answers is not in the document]
claravanstaden: Ok nice! 👍🏻

.gitmodules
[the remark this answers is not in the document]
yrong: Another reason is that last time when I upgraded lodestar to v1.12 the previous way to start the beacon node [here](https://github.com/Snowfork/snowbridge/blob/d77201bc370ab2a4da15de701871b4a643bd5b5c/web/packages/test/scripts/deploy-ethereum.sh#L49) did not even work(errors can't remember), forces me switch to [build from source](https://github.com/Snowfork/snowbridge/blob/61dadc99ac650d5c424c033121df7169effd6f61/web/packages/test/scripts/build-binary.sh#L76) which is also the recommend way for the [installation](https://chainsafe.github.io/lodestar/getting-started/installation/)

relayer/relays/beacon/header/syncer/scale/beacon_scale.go
yrong: Yes, we can. But the point here is to use a single extrinsic to provide multiple version support.

.gitmodules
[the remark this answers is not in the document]
yrong: As discussed to use submodules for now, we may switch to another beacon client implementation(e.g. Prysm) in the future.

relayer/chain/ethereum/trie.go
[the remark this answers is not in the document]
yrong: Exactly.

.gitmodules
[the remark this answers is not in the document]
claravanstaden: Yep, for lodestar. I believe we said in the call go-ethereum can stay in nixpkgs.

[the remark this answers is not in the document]
yrong: I would prefer to build geth from source as well because the versions of Nixpkgs always fall behind the recent release.

e.g. I upgraded both modules to the most recent release in [commit](https://github.com/Snowfork/snowbridge/pull/981/commits/328d541c2a06c25d75e5ce4137f9d3c1f601407c) to test the Deneb hard fork on Goerli which requires geth to https://github.com/ethereum/go-ethereum/releases/tag/v1.13.10 and there is no nixpkgs available yet.

relayer/relays/beacon/state/beacon.go
```diff
@@ -196,7 +196,8 @@ type BlockRootsContainer interface {
 type BeaconBlock interface {
 	UnmarshalSSZ(buf []byte) error
 	GetBeaconSlot() uint64
-	GetExecutionPayload() *ExecutionPayloadCapella
+	CapallaExecutionPayload() *ExecutionPayloadCapella
```
vgeddes, line 199: Naming should be consistent

```suggestion
	ExecutionPayloadCapella() *ExecutionPayloadCapella
```

relayer/relays/beacon/state/beacon.go
```diff
@@ -196,7 +196,8 @@ type BlockRootsContainer interface {
...
+	DenebExecutionPayload() *ExecutionPayloadDeneb
```
vgeddes, line 200:
```suggestion
	ExecutionPayloadDeneb() *ExecutionPayloadDeneb
```

relayer/relays/beacon/state/beacon.go
yrong: https://github.com/Snowfork/snowbridge/pull/981/commits/d4bc62bc6f4ce11f379441d7f7f913dbea7fc773

relayer/relays/beacon/state/beacon.go
yrong: https://github.com/Snowfork/snowbridge/pull/981/commits/d4bc62bc6f4ce11f379441d7f7f913dbea7fc773

scripts/init.sh
```diff
@@ -20,3 +20,9 @@ cargo install cargo-fuzz
 
 echo "Installing web packages"
 (cd web && pnpm install)
+
+echo "Download geth to replace the nix version"
+geth_package=geth-$(uname | tr '[:upper:]' '[:lower:]')-amd64-1.13.10-bc0be1b1
```
vgeddes, line 25: I think its simpler to hardcode the source filenames based on the output of:

```shell
nix eval --impure --raw --expr 'builtins.currentSystem'
```

We need to support macos too, not just linux/amd64

yrong: It seems the [download](https://geth.ethereum.org/downloads) page only provides the amd64 versions, and I tested the `darwin-amd64` did work in my M1.

vgeddes: I think you need to look that download page a bit more - there are links to downloads for many architectures:

Screenshot 2024-01-18 at 15 20 55

So depending on the value of `builtins.currentSystem`, which should be one of the architectures listed in `flake.nix`, download the appropriate binary.

yrong: https://github.com/Snowfork/snowbridge/pull/981/commits/fb5366f196f78b7f950763a5fbc087385d1dcb62

Btw: `nix eval --impure --raw --expr 'builtins.currentSystem'` return `aarch64-darwin` in my laptop which does not match the `arm64-darwin` in the download link so switch to `uname -m` instead.

alistair-singh (approved):

claravanstaden (approved):

vgeddes (approved):

PR #982, opened 2023-10-30
Refactor outbound queue pallet

vgeddes:
The main change is improve the calculation of delivery fees. It turns out that the pallet does need to know the `ETH/DOT` exchange rate in order to calculate a fee that covers the additional delivery reward (on top of the refund). The previous fee calculation was incorrect and did not do this.

The fee config for the pallet can now be set by governance using the `set_fee_config` extrinsic.

Other Changes:
* Support the new `QueuePausedQuery` feature of the `MessageQueue` pallet improve implementation of priority message handling.
* Moved some code around to improve organisation and readability
* Improved tests
* Fee calculation code is a lot cleaner
* Improved documentation for all pallets.
* Outbound messages are versioned so that we can handle forkless upgrades without marking older messages emitted from MessageQueue as corrupted.

Fixes: SNO-635

### DIFF
--- a/parachain/pallets/control/src/lib.rs
+++ b/parachain/pallets/control/src/lib.rs
@@ -2,14 +2,39 @@
 // SPDX-FileCopyrightText: 2023 Snowfork <hello@snowfork.com>
 //! Governance API for controlling the Ethereum side of the bridge
 //!
-//! * upgrade: Upgrade the gateway contract (permissioned)
-//! * set_operating_mode: Update the operating mode of the gateway contract (permissioned)
-//! * create_agent: Create agent for a sibling (permissionless)
-//! * create_channel: Create channel for a sibling (permissionless)
-//! * update_channel: Update a channel for a sibling (permissionless)
-//! * force_update_channel: Allow root to update a channel for a sibling (permissioned)
-//! * transfer_native_from_agent: Withdraw ether from an agent (permissionless)
-//! * force_transfer_native_from_agent: Allow root to withdraw ether from an agent (permissionless)
+//! # Extrinsics
+//!
+//! ## Agents
+//!
+//! Agents are smart contracts on Ethereum that act as proxies for consensus systems on Polkadot
+//! networks.
+//!
+//! * [`Call::create_agent`]: Create agent for a sibling parachain
+//! * [`Call::transfer_native_from_agent`]: Withdraw ether from an agent
+//!
+//! The `create_agent` extrinsic should be called via an XCM `Transact` instruction from the sibling
+//! parachain.
+//!
+//! ## Channels
+//!
+//! Each sibling parachain has its own dedicated messaging channel for sending and receiving
+//! messages. As a prerequisite to creating a channel, the sibling should have already created
+//! an agent using the `create_agent` extrinsic.
+//!
+//! * [`Call::create_channel`]: Create channel for a sibling
+//! * [`Call::update_channel`]: Update a channel for a sibling
+//!
+//! ## Governance
+//!
+//! Only Polkadot governance itself can call these extrinsics. Delivery fees are waived.
+//!
+//! * [`Call::upgrade`]`: Upgrade the gateway contract
+//! * [`Call::set_operating_mode`]: Update the operating mode of the gateway contract
+//! * [`Call::force_update_channel`]: Allow root to update a channel for a sibling
+//! * [`Call::force_transfer_native_from_agent`]: Allow root to withdraw ether from an agent
+//!
+//! Typically, Polkadot governance will use the `force_transfer_native_from_agent` and
+//! `force_update_channel` and extrinsics to manage agents and channels for system parachains.
 #![cfg_attr(not(feature = "std"), no_std)]
 
 pub use pallet::*;
--- a/parachain/pallets/control/src/lib.rs
+++ b/parachain/pallets/control/src/lib.rs
@@ -43,11 +43,8 @@ use frame_support::{
 };
 use frame_system::pallet_prelude::*;
 use snowbridge_core::{
-	outbound::{
-		Command, Initializer, Message, OperatingMode, OutboundQueue as OutboundQueueTrait,
-		SendError,
-	},
-	sibling_sovereign_account, AgentId, ParaId,
+	outbound::{Command, Initializer, Message, OperatingMode, ParaId, SendError, SendMessage},
+	sibling_sovereign_account, AgentId,
 };
 
 #[cfg(feature = "runtime-benchmarks")]
@@ -92,9 +89,9 @@ pub enum PaysFee<T>
 where
 	T: Config,
 {
-	/// Fully charge includes (base_fee + delivery_fee)
+	/// Fully charge includes (local + remote fee)
 	Yes(AccountIdOf<T>),
-	/// Partially charge includes base_fee only
+	/// Partially charge includes local only
 	Partial(AccountIdOf<T>),
 	/// No charge
 	No,
@@ -115,7 +112,7 @@ pub mod pallet {
 		type MessageHasher: Hash<Output = H256>;
 
 		/// Send messages to Ethereum
-		type OutboundQueue: OutboundQueueTrait<Balance = BalanceOf<Self>>;
+		type OutboundQueue: SendMessage<Balance = BalanceOf<Self>>;
 
 		/// The ID of this parachain
 		type OwnParaId: Get<ParaId>;
@@ -413,7 +410,7 @@ pub mod pallet {
 
 			let payment = match pays_fee {
 				PaysFee::Yes(account) => Some((account, fee.total())),
-				PaysFee::Partial(account) => Some((account, fee.base)),
+				PaysFee::Partial(account) => Some((account, fee.local)),
 				PaysFee::No => None,
 			};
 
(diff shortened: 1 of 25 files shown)


parachain/pallets/outbound-queue/src/lib.rs
```diff
@@ -314,94 +397,46 @@ pub mod pallet {
...
-			ensure!(
-				payload.len() < T::MaxMessagePayloadSize::get() as usize,
-				SendError::MessageTooLarge
+		/// Calculate fee in native currency for delivering a message.
+		pub(crate) fn calculate_fee(command: &Command) -> Option<Fee<T::Balance>> {
+			let max_gas = Self::maximum_overall_required_gas(command);
+			let remote_fee = Self::calculate_remote_fee(
+				max_gas,
+				Self::fee_config().fee_per_gas,
+				Self::fee_config().reward,
```
yrong, line 406: Since we need to config `fee_per_gas` and `reward` anyway, can we make them `DOT` based then no need to introduce the `swap_ratio` config?

yrong: If we add `exchange_rate` into the calculation, we also need to take token precision into consideration, DOT is 10 while Ether 18.

parachain/pallets/outbound-queue/src/lib.rs
```diff
@@ -314,94 +397,46 @@ pub mod pallet {
...
-			let ticket =
-				OutboundQueueTicket { id: message_id, origin: message.origin, message: encoded };
-
-			Ok((ticket, fee))
+			let remote_fee = FixedU128::from(remote_fee)
+				.checked_div(&Self::fee_config().exchange_rate)?
+				.into_inner()
+				.checked_div(FixedU128::accuracy())
+				.expect("accuracy is not zero; qed")
+				.into();
```
yrong, line 413: If we add `exchange_rate` into the calculation, we also need to take token precision into consideration, DOT is 10 while Ether 18.

parachain/pallets/outbound-queue/src/types.rs
```diff
@@ -0,0 +1,101 @@
...
+	pub fee_per_gas: u128,
+	/// Ether reward for delivering message
+	pub reward: u128,
+}
+
+impl Default for FeeConfigRecord {
+	fn default() -> Self {
+		Self { exchange_rate: FixedU128::saturating_from_rational(1, 1), fee_per_gas: 1, reward: 0 }
+	}
+}
```
yrong, line 83: would suggest the default value to be more realistic, some [config](https://github.com/Snowfork/snowbridge/blob/584ba5171a96f37426c08746f0e329f894a912c2/parachain/primitives/core/src/outbound.rs#L380-L408) from the previous impl to be referenced, with the [default value](https://github.com/Snowfork/snowbridge/blob/584ba5171a96f37426c08746f0e329f894a912c2/parachain/pallets/outbound-queue/src/lib.rs#L162-L170) there is no need to add a genesis config.

parachain/pallets/outbound-queue/src/mock.rs
```diff
@@ -0,0 +1,177 @@
...
+}
+
+impl crate::Config for Test {
+	type RuntimeEvent = RuntimeEvent;
+	type Hashing = Keccak256;
+	type MessageQueue = MessageQueue;
+	type MaxMessagePayloadSize = ConstU32<1024>;
+	type MaxMessagesPerBlock = ConstU32<20>;
+	type OwnParaId = OwnParaId;
+	type GasMeter = ();
```
yrong, line 89: Prefer to use `ConstantGasMeter` rather than the unit struct as close to the production config as possible.

parachain/pallets/outbound-queue/src/queue_paused_query_impl.rs
```diff
@@ -0,0 +1,28 @@
+//! Implementation for [`frame_support::traits::QueuePausedQuery`]
+use super::*;
+use frame_support::traits::QueuePausedQuery;
+use snowbridge_core::outbound::{AggregateMessageOrigin, ExportOrigin};
+
+impl<T> QueuePausedQuery<AggregateMessageOrigin> for Pallet<T>
```
yrong, line 6: Cool.

parachain/pallets/outbound-queue/src/lib.rs
yrong: Notice that reward in Ether is required anyway pay for the relayer so `exchange_rate`  seems inevitable?

parachain/pallets/outbound-queue/src/lib.rs
vgeddes: Nice catch! This means we should divide the result again by `1 * 10*^-(18 - 10)` right?

parachain/pallets/outbound-queue/src/lib.rs
vgeddes: Yes, exactly. I missed that in the previous PRs.

claravanstaden (approved):
Very nice, super clean. ✨ The extra docs are also great. 

I see you have a [companion branch](https://github.com/Snowfork/polkadot-sdk/tree/refactor-outbound-queue) on polkadot-sdk, which you can also open a PR for.

parachain/pallets/control/src/lib.rs
```diff
@@ -2,14 +2,39 @@
...
-//! * transfer_native_from_agent: Withdraw ether from an agent (permissionless)
-//! * force_transfer_native_from_agent: Allow root to withdraw ether from an agent (permissionless)
+//! # Extrinsics
+//!
+//! ## Agents
+//!
+//! Agents are smart contracts on Ethereum that act as proxies for consensus systems on Polkadot
+//! networks.
+//!
+//! * [`Call::create_agent`]: Create agent for a sibling parachain
```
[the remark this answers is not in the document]
yrong: I think the agent itself is not limited to parachain, the `AgentIdOf` with the internally convert [DescribeAllTerminal](https://github.com/paritytech/polkadot-sdk/blob/9faea380dce6db9aabba29cb01328df229764863/polkadot/xcm/xcm-builder/src/location_conversion.rs#L113-L120) applies to different cases.

But for the current impl the sovereign account of the agent is only for the specific parachain. To @vgeddes just wondering is your previous concern in https://github.com/Snowfork/snowbridge/pull/940#discussion_r1308520658 still valid?

parachain/pallets/outbound-queue/src/types.rs
vgeddes: Nice, never knew about `type_value`. It will definitely simplify the code.

parachain/pallets/outbound-queue/src/lib.rs
```diff
@@ -25,56 +43,92 @@
...
+//!
+//! Messages are weighed to determine the maximum amount of gas they could
+//! consume on Ethereum. Using this upper bound, a final fee can be calculated.
+//!
+//! The fee calculation also requires the following parameters:
+//! * ETH/DOT exchange rate
+//! * Ether fee per unit of gas
+//!
+//! By design, it is expected that governance should manually update these
+//! parameters every few weeks using the [`Call::set_fee_config`] extrinsic.
```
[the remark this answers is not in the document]
vgeddes: Yeh exactly, there will be some buffer in the exchange rate and fee_per_gas to accommodate most spikes.

parachain/pallets/outbound-queue/src/lib.rs
vgeddes: Fixed in https://github.com/Snowfork/snowbridge/pull/982/commits/a0d7916624012f2b32c4c6ce846e8ba092c6d5bb

parachain/pallets/outbound-queue/src/types.rs
vgeddes: Fixed in https://github.com/Snowfork/snowbridge/pull/982/commits/a0d7916624012f2b32c4c6ce846e8ba092c6d5bb

parachain/pallets/outbound-queue/src/mock.rs
vgeddes: Fixed in https://github.com/Snowfork/snowbridge/pull/982/commits/a0d7916624012f2b32c4c6ce846e8ba092c6d5bb

parachain/pallets/outbound-queue/src/lib.rs
```diff
@@ -314,94 +397,56 @@ pub mod pallet {
...
-			let delivery_fee = Self::delivery_fee(&message.command);
-
-			let command = message.command.clone();
-			let fee = Fees { base: base_fee, delivery: delivery_fee };
+			let remote = FixedU128::from(remote)
+				.checked_div(&Self::fee_config().exchange_rate)
+				.expect("exchange rate is not zero; qed")
+				.into_inner()
+				.checked_div(FixedU128::accuracy())
+				.expect("accuracy is not zero; qed");
```
yrong, line 414: Use `FixedU128::from_inner(remote)` then the last step `checked_div(FixedU128::accuracy())` can be ignored.

yrong (approved):

parachain/pallets/outbound-queue/src/lib.rs
vgeddes: While merging the exchange rate and decimal conversion into a single factor may simplify the code, it increases complexity for governance, who now need to calculate and review the math manually. This complexity increases the chance of errors.

PR #988, opened 2023-11-02
Improve xcm integration test with CI

yrong:
Require https://github.com/Snowfork/polkadot-sdk/pull/20


parachain/pallets/inbound-queue/src/lib.rs
```diff
@@ -130,8 +130,8 @@ pub mod pallet {
 	#[pallet::event]
 	#[pallet::generate_deposit(pub(super) fn deposit_event)]
 	pub enum Event<T> {
-		/// A message was received from Ethereum
-		MessageReceived {
+		/// Xcm message sent to dest chain
+		MessageSent {
```
vgeddes, line 134: This doesn't need to be renamed. The inbound queue really did receive a message.

The XCM subsystem should already emit an `XcmSent` event or something like that.

yrong: Or can we rename as `MessageProcessed`?

vgeddes: Look at it this way, the bridge has several layered protocols:

1. The lowest layer is the point-to-point link between the Gateway and in the inbound queue pallet, powered by light clients.
2. The layer above is XCM, which provides a link from the Gateway to any parachain on Polkadot
3. Finally, there is a the application logic built on top of XCM, be it token transfers, or Transact

Logically, this `MessageReceived` is the correct naming for layer (1). A messages was sent by the gateway, and received by the inbound queue pallet.

To track the execution of messages in layer 2, we should use whatever events are emitted by the various XCM pallets, along with the Xcm Hash Ids.

alistair-singh (approved):
Nice work!

parachain/pallets/inbound-queue/src/lib.rs
vgeddes, line 1: I'm currently touching this part of the code, and these non-essential refactorings will get in the way. Can we reduce the scope of this PR to just adding the integration tests?

vgeddes (approved):
Looks great!

PR #989, opened 2023-11-02
Pull in latest Reserve Transfer code

alistair-singh:
Polkadot-Sdk companion: https://github.com/Snowfork/polkadot-sdk/pull/14

### DIFF
--- a/parachain/pallets/control/src/lib.rs
+++ b/parachain/pallets/control/src/lib.rs
@@ -64,7 +64,7 @@ use xcm_executor::traits::ConvertLocation;
 
 use frame_support::{
 	pallet_prelude::*,
-	traits::{tokens::Preservation, Contains, EnsureOrigin},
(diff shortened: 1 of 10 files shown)


claravanstaden (approved):
Nice catch with the duplicate trait!

parachain/pallets/control/src/lib.rs
```diff
@@ -122,6 +122,13 @@ where
 	No,
 }
 
+pub struct AllowSiblingsOnly;
```
vgeddes, line 125: I'd prefer that this goes into our core primitives crate, as its very general.

alistair-singh: Addressed in https://github.com/Snowfork/snowbridge/pull/989/commits/e3e6c8d915c8451c3c583eb5ae8396e246f23d43

PR #994, opened 2023-11-06
Track registered native tokens

vgeddes:
This is useful for implementing the "Polkadot-native assets on Ethereum" story post-launch. It allows the gateway to know whether a particular ERC20 token has been registered or not.

### DIFF
--- a/contracts/src/Assets.sol
+++ b/contracts/src/Assets.sol
@@ -41,6 +41,10 @@ library Assets {
     ) external returns (bytes memory payload, uint256 extraFee) {
         AssetsStorage.Layout storage $ = AssetsStorage.layout();
 
+        if (!$.registeredNativeTokens[token]) {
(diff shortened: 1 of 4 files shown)


claravanstaden (approved):

alistair-singh (approved):

yrong (approved):

contracts/src/Assets.sol
```diff
@@ -96,6 +104,11 @@ library Assets {
             revert InvalidToken();
         }
 
+        if ($.registeredNativeTokens[token]) {
```
alistair-singh, line 107: Just a note here, if token creation fails on the substrate side mapping will be populated and stop a retry by re-submitting the registerToken. Someone would need to fix it manually on the substrate end.

vgeddes: good point! will remove this check.

PR #996, opened 2023-11-07
Fix asset hub fee

claravanstaden:
The decimal precision in the tests do not match the decimals in the [rococo runtime config](https://github.com/Snowfork/polkadot-sdk/blob/snowbridge/cumulus/parachains/runtimes/bridge-hubs/bridge-hub-rococo/src/lib.rs#L538). Using the `calculate_fee` test to generate the user fee value for transfer token back to Ethereum, it generates a value that is too low.

Polkadot-sdk companion: https://github.com/Snowfork/polkadot-sdk/pull/21


alistair-singh (approved):

vgeddes (approved):
Nice catch! I should have made 12 decimals the default, for compat with Rococo. Its only Polkadot mainnet that uses 10 decimals.

PR #1000, opened 2023-11-09
Include all fees in XCM message sent to AssetHub

alistair-singh:
Resolves: https://linear.app/snowfork/issue/SNO-740

Transfer fees for xcm execution with the xcm program.
1. Embed fees into the XCM program using teleport.
2. Burn fees from AssetHub sovereign on bridgehub.
3. On the destination, if there are any fees left over after xcm execution move them to the treasury pallet.


parachain/pallets/inbound-queue/src/lib.rs
```diff
@@ -276,15 +275,32 @@ pub mod pallet {
...
+			T::XcmExecutor::prepare_and_execute(
+				origin,
+				message,
+				&mut xcm_hash,
+				Weight::zero(),
+				Weight::zero(),
+			)
+			.ensure_complete()
+			.map_err(Error::<T>::from)?;
+
```
vgeddes, line 303: I don't think its necessary to run XCM locally. Seems like this adds unnecessary complexity.

Instead burn the fee amount from AssetHub's sovereign account, using [T::Token::burn_from](https://github.com/Snowfork/polkadot-sdk/blob/f7fbba753247de67d36af870b2c69b8f2efc629a/substrate/frame/support/src/traits/tokens/fungible/regular.rs#L252) and then include a [ReceiveTeleportedAsset](https://github.com/paritytech/polkadot/blob/962bc21352f5f80a580db5a28d05154ede4a9f86/xcm/src/v3/mod.rs#L406) instruction in the message you send to AssetHub. That will send the fee amount to the holding there.

parachain/primitives/router/src/inbound/mod.rs
```diff
@@ -83,141 +100,187 @@ pub enum ConvertMessageError {
...
+		let total: MultiAsset = (MultiLocation::parent(), total_amount).into();
+
+		let bridge_location: MultiLocation = (Parent, Parent, GlobalConsensus(network)).into();
+		// TODO(alistair): Get real fee refund locaton
+		let fee_refund_location: MultiLocation = (Parent, Parent, GlobalConsensus(network)).into();
+		let owner = GlobalConsensusEthereumConvertsFor::<[u8; 32]>::from_chain_id(&chain_id);
+		let asset_id = Self::convert_token_address(network, token);
+		let create_call_index: [u8; 2] = CreateAssetCall::get();
+
+		let xcm: Xcm<()> = vec![
```
vgeddes, line 179: Nice! Is looking much cleaner.

vgeddes (commented):
Looks good Alistair! I have gone ahead and made a few changes:

* On AssetHub and the final dest, we should not transfer remaining assets in holding to a treasury account, simply because we can't assume any such account exists. Indeed AssetHub does not have a treasury account, and so remaining execution weight is actually refunded to an account that is used to reward parachain collators. I'm happy for any excess weight refunds / assets to be trapped and claimable, which is default behavior.

* When forwarding funds to the final dest para, the `DepositReserveAsset` instruction actually inserts a `ReserveAssetDeposited` instruction into the sent message. So no need to insert `ReceivedTeleportedAssets` instruction. In any case, `ReceivedTeleportedAssets` implies that AssetHub and the dest para both mutually trust each other as trusted teleporters, which is not always the case. So we need to use reserve-backed deposits.
* I removed some unit tests which I don't think are particularly useful relative. We can use the emulated simular tests to verify that the message packets are correct.

PR #1001, opened 2023-11-09
Improve runtime tests

yrong:


vgeddes (commented):
Nice! 🚀

vgeddes (approved):
Excellent!

yrong:
@claravanstaden I've moved the tests in https://github.com/Snowfork/polkadot-sdk/pull/17 to our own repo and made some fix. To seperate from cumulus it's more convenient to maintain/extend tests as required.

parachain/runtime-tests/Cargo.toml
vgeddes, line 1: Can we make this small change. Move this crate to directory `parachain/runtime/tests/Cargo.toml`.

Since @alistair-singh is also introducing a `runtime-common` crate, we should put all the runtime specific crates under `parachain/runtime`.

yrong: https://github.com/Snowfork/snowbridge/pull/1001/commits/b1ae6a3840bce122960a4590cd4e085c87bb7eb9

vgeddes (approved):

.github/workflows/parachain.yml
```diff
@@ -160,11 +171,31 @@ jobs:
...
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+      - name: setup rust toolchain
+        run: rustup show
+      - name: snowbridge integration tests
+        run: >
+          RUST_LOG=xcm=trace cargo test 
           --manifest-path polkadot-sdk/Cargo.toml
-          -p bridge-hub-rococo-integration-tests
+          -p bridge-hub-rococo-integration-tests snowbridge
```
vgeddes, line 197: What package is named `snowbridge`? I don't recall any package like that?

.github/workflows/parachain.yml
```diff
@@ -84,6 +91,7 @@ jobs:
           --workspace
           --features runtime-benchmarks
           --exclude snowbridge-ethereum-beacon-client
+          --exclude snowbridge-runtime-tests
```
vgeddes, line 94: Do we also now need to run `cargo test --exclude snowbridge-runtime-tests` on the command line too?

.github/workflows/parachain.yml
yrong: Actually it's not a package name, the mod name to filter snowbrige tests only.

parachain/Cargo.toml
```diff
@@ -12,6 +12,6 @@ members = [
     "pallets/ethereum-beacon-client",
     "pallets/control",
     "pallets/control/runtime-api",
+    "runtime/tests",
```
vgeddes, line 15: The tests crate is only used by the runtimes in `polkadot-sdk/`, so I don't think we should add it to the workspace.

Also because its going to bloat our local compile times.

PR #1003, opened 2023-11-09
Pass structured event log to ethereumInboundQueue.submit

vgeddes:
We were making life difficult for ourselves by passing the message data (event log) as raw bytes. Makes troubleshooting harder, and also increases complexity of on-chain unnecessarily, as it has to decode the bytes. This also brings with it a small security risk, if the decoding can be exploited somehow.

I have also updated the `register_token` smoketest to pretty-print the event log (thanks @yrong for initiating that).

```rust
// Generated from smoketests:
//   cd smoketests
//   ./make-bindings
//   cargo test --test register_token -- --nocapture
fn mock_event_log() -> Log {
	Log {
		// gateway address
		address: hex!("eda338e4dc46038493b885327842fd3e301cab39").into(),
		topics: vec![
			hex!("5066fbba677e15936860e04088ca4cad3acd4c19706962196a5346f1457f7169").into(),
			// destination parachain id
			hex!("00000000000000000000000000000000000000000000000000000000000003e8").into(),
			// message id
			hex!("afad3c9777134532ae230b4fad334eef2e0dacbb965920412a7eaa59b07d640f").into(),
		],
		// Nonce + Payload
		data: hex!("00000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000040000000000000000000000000000000000000000000000000000000000000001e000f000000000000000087d1f7fdfee7f651fabc8bfcb6e086c278b77a7d0000").into(),
	}
}
```

The only issue is that our codebase now has two `Log` types:
1. `snowbridge_core::inbound::Log`
2.  `snowbridge_ethereum::log::Log`

I didn't standardize on `snowbridge_ethereum::log::Log`, as that crate uses various ethereum dependencies that are kinda deprecated and unmaintained. Rust folks are now using this library for ethereum types: https://github.com/alloy-rs/core. Anyway, this older `Log` version is still used by the beacon light client.

So I created `snowbridge_core::inbound::Log`, which is only used by the inbound-queue crate, and converted to `snowbridge_ethereum::log::Log` when needed.


parachain/primitives/core/src/inbound.rs
```diff
@@ -5,23 +5,23 @@
...
 use scale_info::TypeInfo;
-use sp_core::H256;
+use sp_core::{H160, H256};
 use sp_runtime::RuntimeDebug;
 use sp_std::vec::Vec;
 
 /// A trait for verifying inbound messages from Ethereum.
 pub trait Verifier {
-	fn verify(message: &Message) -> Result<(), VerificationError>;
+	fn verify(event: &Log, proof: &Proof) -> Result<(), VerificationError>;
```
yrong, line 14: Maybe not direct relevant to change in this PR, I would assume event log is not required for the verification, with the proof and `header.receipts_root` we can generate the `receipt` which internally contains the log event.

yrong: Also we have duplicated definition for `Log` structure in 

https://github.com/Snowfork/snowbridge/blob/6b592235b907e00e9fe6ab4446ff48564bd41474/parachain/pallets/inbound-queue/src/envelope.rs#L12C1-L17

and 

https://github.com/Snowfork/snowbridge/blob/6b592235b907e00e9fe6ab4446ff48564bd41474/parachain/primitives/ethereum/src/log.rs#L8-L13

Maybe we can make some furthur refactoring.

parachain/primitives/core/src/inbound.rs
```diff
@@ -32,12 +32,36 @@ pub type MessageNonce = u64;
...
+	TooManyTopics,
+}
+
+/// Event log
+#[derive(Clone, Encode, Decode, PartialEq, RuntimeDebug, TypeInfo)]
+pub struct Log {
+	pub address: H160,
+	pub topics: Vec<H256>,
+	pub data: Vec<u8>,
+}
```
yrong, line 54: Also we have duplicated definition for the `Log` structure in 

https://github.com/Snowfork/snowbridge/blob/6b592235b907e00e9fe6ab4446ff48564bd41474/parachain/primitives/ethereum/src/log.rs#L8-L13

So maybe we can make some furthur refactoring.

parachain/primitives/core/src/inbound.rs
vgeddes: Yeah, that's a good observation. I saw that as well while doing this PR. 

However I think its easier to troubleshoot problems if the `Log` is submitted as well. Otherwise if there are issues, we would have to decode the `Proof` manually to extract the Log. Which brings us full circle to the problem this PR is solving - improving debuggability.

PR #1004, opened 2023-11-09
make tests, integration-tests, and coverage run in parallel

vgeddes:
To speed up results.

### DIFF
--- a/.github/workflows/parachain.yml
+++ b/.github/workflows/parachain.yml
@@ -130,6 +130,7 @@ jobs:
           flags: rust
 
   check-cumulus-bridgehub:
+    needs: check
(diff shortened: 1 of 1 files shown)


.github/workflows/parachain.yml
```diff
@@ -130,6 +130,7 @@ jobs:
           flags: rust
 
   check-cumulus-bridgehub:
+    needs: check
```
yrong, line 133: To speed up we may not need to add this.

yrong:
@vgeddes I've integrate the change here to https://github.com/Snowfork/snowbridge/pull/1001 and made some more improvement, maybe we can just review/comment in that PR?

PR #1009, opened 2023-11-15
Improve Gateway API for sending tokens to Polkadot

vgeddes:
The previous API was messy, and had duplicate `sentToken` functions in order handle different recipient address types.

I've changed my mind on this design, and have introduced `MultiAddress` type, which encapsulates different substrate address types.


contracts/src/interfaces/IGateway.sol
```diff
@@ -65,7 +65,11 @@ interface IGateway {
 
     /// @dev Emitted once the funds are locked and an outbound message is successfully queued.
     event TokenSent(
-        address indexed token, address indexed sender, ParaID destinationChain, bytes destinationAddress, uint128 amount
+        address indexed token,
+        address indexed sender,
+        ParaID destinationChain,
+        MultiAddress destinationAddress,
```
alistair-singh, line 71: We may want to re-run e2e before final commit to check if the assets still work.

alistair-singh (approved):

contracts/src/MultiAddress.sol
```diff
@@ -0,0 +1,53 @@
...
+using {isIndex, asIndex, isAddress32, asAddress32, isAddress20, asAddress20} for MultiAddress global;
+
+/// @dev An address for an on-chain account
+struct MultiAddress {
+    Kind kind;
+    bytes data;
+}
+
+enum Kind {
+    Index,
```
claravanstaden, line 14: What is this `Index` kind? A parachain ID?

contracts/src/Assets.sol
```diff
@@ -58,34 +50,35 @@ library Assets {
...
+                payload = SubstrateTypes.SendTokenToAddress32(
+                    token, destinationChain, destinationAddress.asAddress32(), amount
+                );
+            } else if (destinationAddress.isAddress20()) {
+                payload = SubstrateTypes.SendTokenToAddress20(
+                    token, destinationChain, destinationAddress.asAddress20(), amount
+                );
+            } else {
+                revert Unsupported();
+            }
```
claravanstaden, line 77: What if the `destinationAddress` is `Index`?

claravanstaden (approved):
Nice consolidation!

contracts/src/MultiAddress.sol
vgeddes: If you install [pallet_indices](https://paritytech.github.io/polkadot-sdk/master/pallet_indices/index.html) on a substrate chain, then its possible for the chain to map a `u32` to a `AccountId` under the hood.

So its basically a "short" id for accountid.

In fact Substrate allows all sorts of mappings:
https://paritytech.github.io/polkadot-sdk/master/sp_runtime/enum.MultiAddress.html

contracts/src/Assets.sol
vgeddes: That's not supported right now, requires more plumbing on the BridgeHub side, to convert the message into the right XCM.

Not many parachains use `pallet_indices` anyway.

yrong (approved):
Cool!

contracts/src/Assets.sol
```diff
@@ -58,34 +50,35 @@ library Assets {
...
+        } else {
+            if (destinationAddress.isAddress32()) {
+                payload = SubstrateTypes.SendTokenToAddress32(
+                    token, destinationChain, destinationAddress.asAddress32(), amount
+                );
+            } else if (destinationAddress.isAddress20()) {
+                payload = SubstrateTypes.SendTokenToAddress20(
```
musnit, line 72: ooh will this make it easier for native Moonbeam support?

musnit: ah i also see we had it already, so just a refactor

PR #1011, opened 2023-11-16
Fix deprecated inbound fixtures

yrong:
Resolves: SNO-749

Btw, met a compiling error https://github.com/rust-lang/rust/issues/112637 several times in my local setup, as suggested in the thread upgrade rust to 1.71.


parachain/primitives/router/src/inbound/mod.rs
```diff
@@ -99,6 +99,7 @@ pub struct MessageToXcm<
 
 /// Reason why a message conversion failed.
 #[derive(Copy, Clone, TypeInfo, PalletError, Encode, Decode, RuntimeDebug)]
+#[cfg_attr(feature = "std", derive(PartialEq))]
```
vgeddes, line 102: We can be more specific here
```suggestion
#[cfg_attr(any(test, feature = "runtime-benchmarks"), derive(PartialEq))]
```

parachain/pallets/inbound-queue/src/lib.rs
```diff
@@ -311,5 +313,20 @@ pub mod pallet {
...
+		pub fn validate(envelope: &Envelope) -> Result<Channel, Error<T>> {
+			// Verify that the message was submitted from the known Gateway contract
+			ensure!(T::GatewayAddress::get() == envelope.gateway, Error::<T>::InvalidGateway);
+
+			// Retrieve the registered channel for this message
+			let channel: Channel =
+				T::ChannelLookup::lookup(envelope.channel_id).ok_or(Error::<T>::InvalidChannel)?;
+			Ok(channel)
+		}
 	}
```
vgeddes, line 331: I don't think this is an improvement. The previous code was easier to follow, even if the `submit` function was rather long. Its mostly straight-line code without any branching.

parachain/pallets/inbound-queue/src/test.rs
```diff
@@ -330,53 +41,20 @@ fn test_submit_happy_path() {
...
+		let envelope = InboundQueue::decode(log).unwrap();
+		let err = InboundQueue::validate(&envelope).unwrap_err();
+		assert_eq!(err, Error::<Test>::InvalidChannel);
 	});
 }
 
 #[test]
 fn test_submit_with_invalid_gateway() {
 	new_tester().execute_with(|| {
-		let relayer: AccountId = Keyring::Bob.into();
```
vgeddes, line 360: What was wrong with this approach to testing? I felt it more complete because it was calling `submit` which is what the user will call.

Now its calling `decode` and `validate` instead, and I can't the benefits to this approach.

parachain/pallets/inbound-queue/src/test.rs
```diff
@@ -2,325 +2,36 @@
...
 		// Submit message
-		let message = Message {
-			event_log: mock_event_log(),
-			proof: Proof {
-				block_hash: Default::default(),
-				tx_index: Default::default(),
-				data: Default::default(),
-			},
-		};
+		let message = make_create_message().message;
```
vgeddes, line 26: I feel like this new `make_create_message` fixture makes our tests inconsistent.

Previously, all tests were using `mock_event_log_*` fixtures, and now this test is using a different fixture, which returns different data types.

vgeddes (commented):
Ron, I am not convinced this makes our tests any better.

parachain/pallets/inbound-queue/src/test.rs
```diff
@@ -2,325 +2,36 @@
...
-use crate::{self as inbound_queue, Error, Event as InboundQueueEvent};
-
-type Block = frame_system::mocking::MockBlock<Test>;
-
-frame_support::construct_runtime!(
-	pub enum Test
-	{
-		System: frame_system::{Pallet, Call, Storage, Event<T>},
-		Balances: pallet_balances::{Pallet, Call, Storage, Config<T>, Event<T>},
-		EthereumBeaconClient: snowbridge_ethereum_beacon_client::{Pallet, Call, Storage, Event<T>},
```
vgeddes, line 32: Also, I'm not sure how the beacon client pallet ended up in the mock runtime, sometime in the past. We don't really need as it, since there is a `MockVerifier`.

And the `MockVerifier` will verifyy any dummy proof, which is good enough to let us test the inbound queue.

yrong:
> I am not convinced this makes our tests any better.

@vgeddes  Actually the previous [fixture](https://github.com/Snowfork/snowbridge/blob/main/parachain/pallets/inbound-queue/src/benchmarking/fixtures.rs) is deprecated and does not work any more so we need to update it.

```
2023-11-16 20:51:37.451  INFO main ethereum-beacon-client: 💫 Verifying message with block hash 0x5f46…0363
2023-11-16 20:51:37.451 ERROR main ethereum-beacon-client: 💫 Verification of receipt inclusion failed for block 0x5f46…0363: <wasm:stripped>
```

Another issue is that in unit test we ignore the verification with a mock impl https://github.com/Snowfork/snowbridge/blob/ae1358fe138d9b14cf075fe52fb8995baf570c3d/parachain/pallets/inbound-queue/src/test.rs#L186

I would like to reuse the fixture for both benchmark and unit test so we have test to cover the complete flow.

parachain/pallets/inbound-queue/src/test.rs
claravanstaden: It is because I needed to implement the trait `BenchmarkHelper` https://github.com/Snowfork/snowbridge/blob/main/parachain/pallets/inbound-queue/src/test.rs#L140 that I added so that the inbound queue benchmarks would include the log verification in the inbound queue pallet.

vgeddes:
> Actually the previous [fixture](https://github.com/Snowfork/snowbridge/blob/main/parachain/pallets/inbound-queue/src/benchmarking/fixtures.rs) is deprecated and does not work any more so we need to update it.

Thanks for catching this - I've cut SNO-749 to track it.

> Since we've already introduced `EthereumBeaconClient` for the benchmark and have fixture data for it. So I would prefer to not use the mock and reuse the fixture for both benchmark and the unit test.

We can't really reuse the the benchmarking fixture for all the negative tests - the ones that test invalid messages. So we need mocks for those anyway. So if we're going to use mocks, lets be consistent with how we use them.

There are already a bunch of tests in the beacon-client pallet for testing verification of receipts and inclusion of logs in receipts. I don't want to duplicate those tests in the inbound-queue pallet.

The inbound-queue tests should treat `T::Verifier` as a black box and not be concerned whether it works correctly or not, because it is not the focus of the tests.

For accurate benchmarking on the other hand, it is different matter, and the benchmarking code needs to initialize a beacon client. But this is still separate from unit tests.

rust-toolchain.toml
```diff
@@ -3,7 +3,7 @@
 # https://github.com/rust-lang/rustup/issues/2686
 # The auto-installation behaviour in rustup will likely be removed:
 # https://github.com/rust-lang/rustup/issues/1397
-channel = "1.70.0"
+channel = "1.71.0"
```
vgeddes, line 6: Polkadot is already using 1.73.0: https://github.com/paritytech/polkadot-sdk/releases/tag/polkadot-v1.3.0

vgeddes (approved):
Looks much better and focused! Though consider upgrading to Rust 1.73.0

rust-toolchain.toml
yrong: https://github.com/Snowfork/snowbridge/pull/1011/commits/f70142dcbd837e0f16a7df21ab06c7cea3f6f75c

vgeddes (approved):

PR #1012, opened 2023-11-16
Asset transfer to penpal

alistair-singh:
Resolves: SNO-708
Polkadot-sdk companion: https://github.com/Snowfork/polkadot-sdk/pull/34

Configuration required for PenPal (or an integrating parachain).
1. Have a foreign assets pallet instance configured in your runtime.
2. Configure xcm executor to allow ethereum assets reserves from asset hub.

## TODO
- [x] Add foreign assets pallet instance and configure it appropriately to get ethereum assets from asset hub.
- [x] Add emulated integration test using penpal.
- [x] Add e2e test to test that tests the same scenario.


vgeddes (approved):
Looks great! 🚀

smoketest/make-bindings.sh
```diff
@@ -23,5 +23,6 @@ fi
 # Fetch metadata from BridgeHub and generate client
 subxt codegen --url ws://localhost:11144 > src/parachains/bridgehub.rs
 subxt codegen --url ws://localhost:12144 > src/parachains/assethub.rs
-subxt codegen --url ws://localhost:9944  > src/parachains/relaychain.rs
 subxt codegen --url ws://localhost:13144 > src/parachains/template.rs
```
yrong, line 26: Is template chain still in use or just safe to remove?

yrong (approved):
LGTM!

smoketest/make-bindings.sh
alistair-singh: Not safe yet, we will still need to add an extrinsic to penpal to call create_agent/create_channel. Although penpal has `sudo` pallet so we actually dont need to add anything and can just use a sudo call with pallet_xcm.send to call bridgehub.

PR #1013, opened 2023-11-20
Improve runtime tests

yrong:
The added unit test focus on the flow of submit inbound message(including proof verification logic) and check `MessageReceived` event with specified `xcm_hash` as expected.

Together with the previous emulated [test](https://github.com/Snowfork/polkadot-sdk/blob/52bcb07b166fa12b1a1446fb655d3ae42fe84170/cumulus/parachains/integration-tests/emulated/tests/bridges/bridge-hub-rococo/src/tests/snowbridge.rs#L230) then we have a basic ensurance that inbound flow always work as expected without runing e2e smoke test which is a bit too slow.


vgeddes (commented):
I like the goals of the PR, however, I think the implementation introduces a big problem - tight coupling between the runtime benchmark code and the runtime test code. In other words, the modularity & composability of the codebase has been compromised.

The main symptom is that the runtime tests now depend on the `runtime-benchmarks` feature, which is not good.

This problem can be solved by some of the following approaches:
1. Adding a `runtime-tests` feature to the inbound-queue crate
2. Move the fixture data to `pallets/inbound-queue/src/fixtures.rs`, and conditionally declare it as a module using `#[cfg(any(feature = "runtime-tests", feature = "runtime-benchmarks))]`
3. Conditionally compile `Config::Helper` also using  `#[cfg(any(feature = "runtime-tests", feature = "runtime-benchmarks))]`

PR #1015, opened 2023-11-22
Fee for `Gateway.sendToken` should cover all costs

vgeddes:
For token transfers from Ethereum, if the final destination of the transfer is not AssetHub, we need to charge double the sendToken fee to cover cost of the extra XCM message to the final destination.

I've also improved the API that allows users to query fees for registering and sending tokens. The APIs give more detailed fee info, and in the case of SendToken, accounts for special case above.

requires: https://github.com/Snowfork/polkadot-sdk/pull/49

Fixes: SNO-761


yrong:
> For token transfers from Ethereum, if the final destination of the transfer is not AssetHub, we need to charge double the sendToken fee to cover cost of the extra XCM message to the final destination.

The estimation to double the cost may not be accurate, the xcm fee totally depends on respective config/impl for a specific parachain and may not be same as(or even differs a lot from) assetHub.

E.g. Like for acala the [trader](https://github.com/AcalaNetwork/Acala/blob/d2998be5bf1617f8aeaf740e53f634bfef944142/runtime/acala/src/xcm_config.rs#L93-L103) they use to calculate the fee are totally different from [trader](https://github.com/Snowfork/polkadot-sdk/blob/a0c2f3a7e39b89a926c12f5048b7f41d7dc5b172/cumulus/parachains/runtimes/assets/asset-hub-rococo/src/xcm_config.rs#L580C17-L580C17) in assetHub and they don't even use the asset pallet for MultiCurrency support.

contracts/src/Assets.sol
```diff
@@ -75,12 +87,19 @@ library Assets {
             } else {
                 revert Unsupported();
             }
+            // If the final destination chain is not AssetHub, then the fee needs to additionally
+            // include the cost of executing an XCM on the final destination parachain.
+            extraFee = 2 * $.sendTokenFee;
```
alistair-singh, line 92: Suggestion: Why not call `sendTokenFee(ParaID assetHubParaID, ParaID destinationChain)` here? Is it because we are worried about gas costs because its `external`? I think we should have an `internal` implementation and use it for both here and in `sendTokenFee(ParaID assetHubParaID, ParaID destinationChain)` so that the fee can be tuned in one place.

alistair-singh (approved):
Minor suggestion. But looks good!

vgeddes:
That's a good point Ron. And we will have the same issue for Xcm::Transact. I understand now that we need to implement something like this:
1.  The `sendToken` function takes an `destinationChainFee` parameter, which is the amount of DOT it will cost to execute the XCM.
2.  When called, the `destinationChainFee` is converted to ETH using a `DOT/ETH` exchange rate
3.  This converted amount is transferred from the user to the agent contract for AssetHub
4.  We need to include `destinationChainFee` amount in the message sent to BridgeHub, so that the XCM messages can have the correct `BuyExecution` instructions.

I think we can reasonably expect that parachains should accept DOT as the fee currency. This simplifies things. No need for custom fee tokens.

There are other changes to make if we go with this approach. For example, [Channel.outboundFee](https://github.com/Snowfork/snowbridge/blob/eb13b900de69e9e63c4c1e49ad56d352d0b309f8/contracts/src/Types.sol#L49) should be in DOT, not ETH. Then we convert to ETH using the  exchange rate when `sendToken` or `registerToken` is called.

Potential issues: The currency conversion with a lagging & mostly static exchange rate would lead to either of the following scenarios:

Scenario 1:
1. Suppose actual XR is 1ETH/100DOT, however Gateway is configured with an XR of 1ETH/50DOT.
2. Then user calls `sendToken`, paying a fee of 1 ETH for a XCM fee of 50 DOT.
3. User ends up overpaying ETH to AssetHub's agent, as DOT is worth less in reality.
4. Bridge security & operations are maintained, however users are potentially unhappy.

Scenario 2:
1. Suppose actual XR is 1ETH/100DOT, however Gateway is configured with an XR of 1ETH/200DOT.
2. Then user calls `sendToken`, paying a fee of 1 ETH for a XCM fee of 200 DOT.
3. User ends up underpaying ETH to AssetHub's agent, as DOT is worth more in reality.
4. Bridge security & operations are maintained. however, in the longer run, AssetHub may end up short of funds to pay for rewards on BridgeHub.

contracts/src/Assets.sol
vgeddes: yeah mostly worried about gas costs. 

Gonna have to redo this code though, see my latest comment further below.

yrong:
> The sendToken function takes a destinationChainFee parameter, which is the amount of DOT it will cost to execute the XCM.

Or could be loaded from some map storage something like 
(Multilocation(Acala)->0.01DOT),(MultiLocation(Moonbeam)->0.02DOT)
That storage can be configured from the control pallet.

> When called, the destinationChainFee is converted to ETH using a DOT/ETH exchange rate

Since we've already introduced the `exchange rate` configurable in https://github.com/Snowfork/snowbridge/blob/eb13b900de69e9e63c4c1e49ad56d352d0b309f8/parachain/pallets/outbound-queue/src/lib.rs#L242,  so maybe we can reuse it on Ethereum side with another control command set both.

As you mentioned the main issue is the outdated price. So we may need to introduce some price oracle like [Chainlink Data Feeds ](https://data.chain.link/) or remember as @musnit mentioned before with an on-chain swap.

vgeddes:
>> The sendToken function takes a destinationChainFee parameter, which is the amount of DOT it will cost to execute the XCM.

> Or could be loaded from some map storage something like (Multilocation(Acala)->0.01DOT),(MultiLocation(Moonbeam)->0.02DOT), That storage can be configured from the control pallet.

Well the idea is that indeed there need to be some mapping of parachain to fees. But its much simpler for clients to maintain that mapping offchain, and pass the required `destinationChainFee` to our API.

>> When called, the destinationChainFee is converted to ETH using a DOT/ETH exchange rate

> Since we've already introduced the `exchange rate` configurable in https://github.com/Snowfork/snowbridge/blob/eb13b900de69e9e63c4c1e49ad56d352d0b309f8/parachain/pallets/outbound-queue/src/lib.rs#L242,  so maybe we can reuse it on Ethereum side with another control command set both at the same time.

Agree :+1:

I'm implementing that in this PR.

> As you mentioned the main issue is the outdated price. So we may need to introduce some price oracle like [Chainlink Data Feeds ](https://data.chain.link/) or remember as @musnit mentioned before with an on-chain swap.

I think a outdated exchange rate is OK for our bridge, for various reasons:
1. The exchange rate is only used to calculate fees, which are generally much less than the actual value of the tokens being transferred.
2. At least for the past few months the ETH/DOT rate has been quite stable around 1/400.
3. We can always upgrade the bridge to implement a better mechanism (such as on-chain swaps)

The issue with all Chainlink services is that they are all centralized. They are not actually decentralised and trustless, since they are governed and maintained by a single company.

yrong:
https://github.com/Snowfork/snowbridge/blob/2fcc4159610b4ed7f718eca424a2cf25483c6de9/parachain/pallets/control/src/lib.rs#L339

I would assume this param can be removed now, should be exactly the same as the cost in

https://github.com/Snowfork/snowbridge/blob/2fcc4159610b4ed7f718eca424a2cf25483c6de9/parachain/pallets/inbound-queue/src/lib.rs#L256

alistair-singh:
> Scenario 1:
> 
> 1. Suppose actual XR is 1ETH/100DOT, however Gateway is configured with an XR of 1ETH/50DOT.
> 2. Then user calls `sendToken`, paying a fee of 1 ETH for a XCM fee of 50 DOT.
> 3. User ends up overpaying ETH to AssetHub's agent, as DOT is worth less in reality.
> 4. Bridge security & operations are maintained, however users are potentially unhappy.
> 
> Scenario 2:
> 
> 1. Suppose actual XR is 1ETH/100DOT, however Gateway is configured with an XR of 1ETH/200DOT.
> 2. Then user calls `sendToken`, paying a fee of 1 ETH for a XCM fee of 200 DOT.
> 3. User ends up underpaying ETH to AssetHub's agent, as DOT is worth more in reality.
> 4. Bridge security & operations are maintained. however, in the longer run, AssetHub may end up short of funds to pay for rewards on BridgeHub.

Both these scenarios apply to our conversion rate on the substrate side as well. So we may want one extrinsic so that governance can upgrade the conversion rates on both sides of the bridge, maybe from the control pallet?

contracts/src/Assets.sol
```diff
@@ -44,54 +36,99 @@ library Assets {
...
+    /// @param token The ERC20 token address.
+    function registerToken(address token) external returns (Ticket memory ticket) {
         if (!token.isContract()) {
             revert InvalidToken();
         }
 
-        payload = SubstrateTypes.RegisterToken(token);
-        extraFee = $.registerTokenFee;
+        ticket.costs = _registerTokenCosts();
+        ticket.payload = SubstrateTypes.RegisterToken(token);
```
vgeddes, line 131: I realize that we should include `$.assetHubCreateAssetFee` in the message payload to BridgeHub, to be consistent with sendToken

```suggestion
        ticket.payload = SubstrateTypes.RegisterToken(token, $.assetHubCreateAssetFee);
```

vgeddes: Seems we also need to have a `$.assetHubCreateAssetDeposit` to match the XCM conversion code in

https://github.com/Snowfork/snowbridge/blob/510880c717bf34941c75aff2ed4c63b272964fa3/parachain/primitives/router/src/inbound/mod.rs#L165

yrong: Addressed in https://github.com/Snowfork/snowbridge/pull/1015/commits/edf418d7890a2d8d40e816cff1c0923e64840772

> currently assetHubCreateAssetFee is fee + deposit

Actually it's for different purposes, `assetHubCreateAssetFee` is only paying for xcm execution while  `CreateAssetDeposit` is for the required deposit(0.1 DOT).
https://github.com/Snowfork/snowbridge/blob/510880c717bf34941c75aff2ed4c63b272964fa3/parachain/primitives/router/src/inbound/mod.rs#L177-L180

Ideally the `registerTokenFee`(in Ether) should match the `CreateAssetDeposit`(in DOT), but  I would suggest we keep it seperated, we can charge more than the required deposit to avoid spamming.

vgeddes: sounds fine 👍

yrong (approved):

PR #1016, opened 2023-11-22
LOW-1: Inconsistency between implementation and documentation regarding the minimum number of signatures

alistair-singh:
Resolves: SNO-751

See my review comments for more details about the change. The old fuzz tests still pass without modification. Please review carefully.


contracts/src/BeefyClient.sol
```diff
@@ -448,22 +448,10 @@ contract BeefyClient {
     ) internal pure returns (uint256) {
         // Start with the minimum number of signatures.
         uint256 numRequiredSignatures = minRequiredSignatures;
-
-        // We must subtract minimumSignatures from the number of validators or we might end up
-        // requiring more signatures than there are validators.
-        uint256 extraValidatorsLen = validatorSetLen.saturatingSub(minRequiredSignatures);
```
alistair-singh, line 454: We no longer need this `extraValidatorsLen` because we clamp the output to be less than 2/3 quorum.

contracts/src/BeefyClient.sol
```diff
@@ -448,22 +448,10 @@ contract BeefyClient {
...
-        if (extraValidatorsLen > 1) {
-            numRequiredSignatures += Math.log2(extraValidatorsLen, Math.Rounding.Ceil);
```
alistair-singh, line 456: We do not need to check `extraValidatorsLen` as `Math.log2(extraValidatorsLen, Math.Rounding.Ceil)` returns `0` with `0` as input.

contracts/src/BeefyClient.sol
```diff
@@ -448,22 +448,10 @@ contract BeefyClient {
...
-        } else if (extraValidatorsLen == 1) {
-            // log2 returns 0 when there is one validator so when this case is hit the single
-            // validator needs to be verified.
-            numRequiredSignatures += 1;
-        }
-
-        if (signatureUsageCount > 0) {
```
alistair-singh, line 463: We do not need to check `signatureUsageCount ` as `Math.log2(signatureUsageCount, Math.Rounding.Ceil)` returns `0` with `0` as input.

scripts/beefy_signature_sampling.py
```diff
@@ -15,8 +15,7 @@ def samples(ratio_per_validator, validators_length, slash_rate, randao_commit_ex
 
     result += ceil(log2(validators_length));
 
-    if(signature_use_count > 0):
-        result += 1 + 2 * ceil(log2(signature_use_count))
+    result += 1 + (2 * ceil(log2(signature_use_count)) if signature_use_count > 1 else 0)
```
alistair-singh, line 18: This reflects the new math. The main difference is that there is one extra signature required when signature count is 0. This is because both the python and solidity code only added 1 if the signature count was greater than 0.

Old math:

Signature Use Count|Total|Static|Dynamic
-|-|-|-
0|27|18|9
1|28|18|10
2|30|18|12
3|32|18|14
4|32|18|14
5|34|18|16
6|34|18|16
7|34|18|16
8|34|18|16
9|36|18|18
10|36|18|18
11|36|18|18
12|36|18|18
13|36|18|18
14|36|18|18
15|36|18|18
16|36|18|18
17|38|18|20
18|38|18|20
19|38|18|20

Signature Use Count|Total|Static|Dynamic
-|-|-|-
0|28|18|10
1|28|18|10
2|30|18|12
3|32|18|14
4|32|18|14
5|34|18|16
6|34|18|16
7|34|18|16
8|34|18|16
9|36|18|18
10|36|18|18
11|36|18|18
12|36|18|18
13|36|18|18
14|36|18|18
15|36|18|18
16|36|18|18
17|38|18|20
18|38|18|20
19|38|18|20

contracts/test/BeefyClient.t.sol
```diff
@@ -242,6 +242,9 @@ contract BeefyClientTest is Test {
         // Signature count is now 1 after a second submitInitial.
         beefyClient.submitInitial(commitment, bitfield, finalValidatorProofs[0]);
 
+        // Signature count is now 2 after a second submitInitial.
+        beefyClient.submitInitial(commitment, bitfield, finalValidatorProofs[0]);
```
alistair-singh, line 246: To make these tests pass we need another submitInitial to increment the signature count. See comment below on the new math.

yrong: Guess to trigger the error [here](https://github.com/Snowfork/snowbridge/blob/13b406832cedee0066d9823d4704e6fcab70d258/contracts/src/BeefyClient.sol#L480-L482)?

Just curious how to make the invalid proof in https://github.com/Snowfork/snowbridge/blob/13b406832cedee0066d9823d4704e6fcab70d258/contracts/test/BeefyClient.t.sol#L256 to trigger that error.

yrong: Maybe not relevant to the change here, but I saw multiple `InvalidValidatorProof` errors in different places in function [verifyCommitment](https://github.com/Snowfork/snowbridge/blob/13b406832cedee0066d9823d4704e6fcab70d258/contracts/src/BeefyClient.sol#L470), would suggest we name the error properly with different names.

alistair-singh: Addressed in https://github.com/Snowfork/snowbridge/pull/1016/commits/cd0034bf59fca50ef38fa8ab9f38fdb66140a35b. I just added an extra error called `InvalidValidatorProofLength` for this case and modified the affected cases.

yrong (approved):

vgeddes (approved):
Looks much cleaner 🚀

PR #1017, opened 2023-11-22
LOW-2: The functions of the Uint16Array library can access and modify out of (the real) bounds positions

alistair-singh:
Resolves: SNO-752


vgeddes (approved):
Looks good.

PR #1022, opened 2023-11-23
Derive AgentId for BridgeHub in a consistent manner

yrong:
Resolves: SNO-760


vgeddes:
Looks good so far. Of course, if the junction is not `Here`, we should pass it to `xcm_builder::DescribeFamily`.

vgeddes:
And of course then we would no longer need to hardcode the AgentID here: https://github.com/Snowfork/snowbridge/blob/d97e3b780953edfb5296e83bcbb850c66caafe57/parachain/primitives/core/src/lib.rs#L139. Can delete it.

alistair-singh (approved):
🚀

yrong:
Merge though integration tests failed,  will handle that in other PRs more related.

PR #1025, opened 2023-11-25
LOW-3: Missing input validation

alistair-singh:
Resolves: SNO-753

Extra merkle proof validations from the audit issue are not addressed.

### DIFF
--- a/contracts/src/BeefyClient.sol
+++ b/contracts/src/BeefyClient.sol
@@ -241,6 +241,10 @@ contract BeefyClient {
     function submitInitial(Commitment calldata commitment, uint256[] calldata bitfield, ValidatorProof calldata proof)
         external
     {
+        if (commitment.blockNumber <= latestBeefyBlock) {
+            revert StaleCommitment();
+        }
+
         ValidatorSetState storage vset;
         uint16 signatureUsageCount;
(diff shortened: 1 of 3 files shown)


contracts/src/BeefyClient.sol
```diff
@@ -215,6 +216,9 @@ contract BeefyClient {
         ValidatorSet memory _initialValidatorSet,
         ValidatorSet memory _nextValidatorSet
     ) {
+        if (_nextValidatorSet.id != _initialValidatorSet.id + 1) {
+            revert InvalidValidatorSetData();
```
vgeddes, line 220: Can just `revert()` here without raising a custom exception. I don't want to add a custom exception `InvalidValidatorSetData` which may never ever be used.

alistair-singh: Addressed in https://github.com/Snowfork/snowbridge/pull/1025/commits/9fd119560d062741cc7513e102778fa7d21e5850.

vgeddes: Nit, would prefer a string that looks more robotic: `invalid-constructor-params`

vgeddes (approved):
+1

PR #1029, opened 2023-11-28
Remove workspace flag from build script

claravanstaden:
I added the `--workspace` flag to the build polkadot and build polkadot-parachain commands because it prevented recompilation if the source files did not change. However, building the workspace included compiling some dev features, like fuzzing, which lead to a compilation error (for which I added a temp fix here: https://github.com/Snowfork/polkadot-sdk/blob/snowbridge/substrate/frame/nomination-pools/src/lib.rs#L3409). 

This PR removes the `--workspace` flag. Unfortunately the binaries rebuild on every start-services run now, at least partially. Will optimize later, but for now this fixes the compilation error and I can revert the change in https://github.com/Snowfork/polkadot-sdk/blob/snowbridge/substrate/frame/nomination-pools/src/lib.rs#L3409.

### DIFF
--- a/web/packages/test/scripts/build-binary.sh
+++ b/web/packages/test/scripts/build-binary.sh
@@ -12,14 +12,16 @@ build_binaries() {
     fi
 
     echo "Building polkadot binary"
-    cargo build --release --workspace --locked --bin polkadot --bin polkadot-execute-worker --bin polkadot-prepare-worker
+    cargo build --release --locked --bin polkadot --bin polkadot-execute-worker --bin polkadot-prepare-worker
     cp target/release/polkadot $output_bin_dir/polkadot
     cp target/release/polkadot-execute-worker $output_bin_dir/polkadot-execute-worker
     cp target/release/polkadot-prepare-worker $output_bin_dir/polkadot-prepare-worker
 
     echo "Building polkadot-parachain binary"
-    cargo build --release --workspace --locked --bin polkadot-parachain $features
-    cp target/release/polkadot-parachain $output_bin_dir/polkadot-parachain
+    cd cumulus/polkadot-parachain
+    cargo build --release --locked --bin polkadot-parachain $features
+    cp ../../target/release/polkadot-parachain $output_bin_dir/polkadot-parachain
+    cd -
 
     popd
 }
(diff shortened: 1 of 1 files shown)


vgeddes (approved):
+1

web/packages/test/scripts/build-binary.sh
```diff
@@ -12,14 +12,16 @@ build_binaries() {
...
     cp target/release/polkadot-execute-worker $output_bin_dir/polkadot-execute-worker
     cp target/release/polkadot-prepare-worker $output_bin_dir/polkadot-prepare-worker
 
     echo "Building polkadot-parachain binary"
-    cargo build --release --workspace --locked --bin polkadot-parachain $features
-    cp target/release/polkadot-parachain $output_bin_dir/polkadot-parachain
+    cd cumulus/polkadot-parachain
+    cargo build --release --locked --bin polkadot-parachain $features
+    cp ../../target/release/polkadot-parachain $output_bin_dir/polkadot-parachain
+    cd -
```
yrong, line 24: Maybe no need cd to the cumulus directory, build from the top level with command
```cargo build --release --locked -p polkadot-parachain-bin --bin polkadot-parachain $features``` should be fine.

claravanstaden: Good point. 😄 Addressed in https://github.com/Snowfork/snowbridge/pull/1029/commits/056b9f49f119621a295e65944da4a5542da79db4.

yrong: > Unfortunately the binaries rebuild on every start-services run now, at least partially.

I would assume it's because we build two binaries(i.e. relaychain&parachain)? Considering that usually we do not change the code of relay chain, maybe skipping the build of relay chain(in case binary exists) will help.

vgeddes: Good point, seems unnecessary to rebuild the parachain every time

claravanstaden: I added this quickly: https://github.com/Snowfork/snowbridge/pull/1029/commits/395b5752327e8c269e0828fb9398d2ced475673a

PR #1032, opened 2023-11-30
Track registered tokens

vgeddes:
New implementation which supersedes https://github.com/Snowfork/snowbridge/pull/994.

Also improved the `Assets.sol` library, so that `Gateway.sol` doesn't have to be concerned with AssetHub and asset stuff.


alistair-singh (approved):
🚀

PR #1033, opened 2023-12-01
Use pallet instance in inbound queue origin

claravanstaden:
Change to accomodate this requested change: https://github.com/paritytech/polkadot-sdk/pull/2522#discussion_r1407622955 The EthereumInboundQueue pallet ID should form part of the origin to AssetHub.

Companion PR: https://github.com/Snowfork/polkadot-sdk/pull/51


parachain/runtime/rococo-common/src/lib.rs
```diff
@@ -8,6 +8,8 @@
 use frame_support::parameter_types;
 use xcm::opaque::lts::NetworkId;
 
+pub const INBOUND_QUEUE_MESSAGES_PALLET_INDEX: u8 = 80;
```
vgeddes, line 11: Looks like there is an API for retrieving a pallet's index: https://substrate.stackexchange.com/a/794

parachain/primitives/router/src/inbound/mod.rs
```diff
@@ -165,6 +185,7 @@ where
 			BuyExecution { fees: xcm_fee, weight_limit: Unlimited },
 			// Fund the snowbridge sovereign with the required deposit for creation.
 			DepositAsset { assets: Definite(deposit.into()), beneficiary: bridge_location },
+			DescendOrigin(X1(PalletInstance(inbound_queue_pallet_index))),
```
vgeddes, line 188: Shouldn't this `DescendOrigin` be placed at the start of message, before `BuyExecution`?

As otherwise, we have three origins in this message, which makes it more complicated.

cc @alistair-singh

parachain/runtime/rococo-common/src/lib.rs
vgeddes: Ah ok makes sense 👌

parachain/primitives/router/src/inbound/mod.rs
yrong: @claravanstaden  I agree with @vgeddes  here, better to move `DescendOrigin` to first, revamp accordingly in 
https://github.com/Snowfork/snowbridge/pull/1034/files#diff-9218e396410f01117c1c305b9cf006f6498bb0b3c0aedb8952a8450b4b1c5715

claravanstaden: Thanks @yrong, I merged it.

PR #1042, opened 2023-12-04
Set Chain ID to 5

claravanstaden:


parachain/pallets/inbound-queue/src/mock.rs
```diff
@@ -130,7 +130,7 @@ impl Verifier for MockVerifier {
 const GATEWAY_ADDRESS: [u8; 20] = hex!["eda338e4dc46038493b885327842fd3e301cab39"];
 
 parameter_types! {
-	pub const EthereumNetwork: xcm::v3::NetworkId = xcm::v3::NetworkId::Ethereum { chain_id: 15 };
+	pub const EthereumNetwork: xcm::v3::NetworkId = xcm::v3::NetworkId::Ethereum { chain_id: 5 };
```
claravanstaden, line 133: Is it ok to hardcode the chain ID in tests? Do we want to use the constant here? I am not sure about adding the `rococo-common` crate here as a dependency just to use the chain ID.

parachain/primitives/router/src/inbound/tests.rs
```diff
@@ -6,11 +6,11 @@ use sp_core::crypto::Ss58Codec;
 use xcm::v3::prelude::*;
 use xcm_executor::traits::ConvertLocation;
 
-const NETWORK: NetworkId = Ethereum { chain_id: 15 };
+const NETWORK: NetworkId = Ethereum { chain_id: 5 };
```
claravanstaden, line 9: Same as https://github.com/Snowfork/snowbridge/pull/1042/files#r1415273254.

web/packages/test/scripts/deploy-ethereum.sh
```diff
@@ -60,6 +60,8 @@ start_lodestar() {
             --params.ALTAIR_FORK_EPOCH 0 \
             --params.BELLATRIX_FORK_EPOCH 0 \
             --params.CAPELLA_FORK_EPOCH 0 \
+            --params.DEPOSIT_CHAIN_ID 11155111 \
+            --params.DEPOSIT_NETWORK_ID 11155111 \
```
yrong, line 64: Curious about what are these two parameters for or required?

yrong (approved):

web/packages/test/scripts/deploy-ethereum.sh
claravanstaden: They are the only Lodestar config I could find related to the chain ID. I will double check if they are required. 😄

PR #1044, opened 2023-12-05
Add scan-beefy command to relayer

alistair-singh:
Adds a debug command that scans for beefy commitments using the same logic as the beefy relayer.

In the relayer folder:
```cmd
go run snowbridge-relay scan-beefy \
   --polkadot-url wss://rococo-rpc.polkadot.io \
   --beefy-block $BLOCK_FROM_LIGHT_CLIENT \
   --validator-set-id $VALIDATOR_FROM_LIGHT_CLIENT
```


yrong (approved):
LGTM!

PR #1047, opened 2023-12-11
Lock down token transfer to AssetHub only

alistair-singh:
Resolves: SNO-771
Polkadot-Sdk: https://github.com/Snowfork/polkadot-sdk/pull/62

### DIFF
--- a/parachain/primitives/router/src/outbound/mod.rs
+++ b/parachain/primitives/router/src/outbound/mod.rs
@@ -9,7 +9,10 @@ use core::slice::Iter;
 
 use codec::{Decode, Encode};
 
-use frame_support::{ensure, traits::Get};
+use frame_support::{
+	ensure,
+	traits::{ContainsPair, Get},
+};
 use snowbridge_core::{
 	outbound::{AgentExecuteCommand, Command, Message, SendMessage},
 	ChannelId, ParaId,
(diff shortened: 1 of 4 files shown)


parachain/primitives/router/src/outbound/mod.rs
```diff
@@ -111,6 +130,12 @@ where
 			command: Command::AgentExecute { agent_id, command: agent_execute_command },
 		};
 
+		// filter out message
+		if !ExportFilter::contains(&local_sub, &outbound_message) {
```
yrong, line 134: Notice that `ExportFilter` is declared as `EverythingBut<NonAssetHubTokenTransfers>` in runtime, so could it be simplified as `NonAssetHubTokenTransfers` and remove the not operation here.

alistair-singh: Yeah, I was also a bit unsure on which way to go with this, so I looked at other "Filter" constraints in the Polkadot codebase to see what is the standard way, and it seems that if the filter returns true, the call should be allowed. This allows you to use the traits `Everything` to allow all, and `Nothing` to block all.

Examples:
* [CallFilter](https://github.com/Snowfork/polkadot-sdk/blob/alistair/lock-down-token-transfer/substrate/frame/contracts/src/wasm/runtime.rs#L406)
* [XcmTeleportFilter](https://github.com/Snowfork/polkadot-sdk/blob/alistair/lock-down-token-transfer/polkadot/xcm/pallet-xcm/src/lib.rs#L1422)
* [XcmReserveTransferFilter](https://github.com/Snowfork/polkadot-sdk/blob/alistair/lock-down-token-transfer/polkadot/xcm/pallet-xcm/src/lib.rs#L1343)
* [XcmExecuteFilter](https://github.com/Snowfork/polkadot-sdk/blob/alistair/lock-down-token-transfer/polkadot/xcm/pallet-xcm/src/lib.rs#L297)
* [SafeCallFilter](https://github.com/Snowfork/polkadot-sdk/blob/alistair/lock-down-token-transfer/polkadot/xcm/xcm-executor/src/lib.rs#L558)

yrong: IMHO the filters above are more of a whitelist whereas we're blocking a blacklist item here. So will it make more sense to rename it as `ExportInterceptor` and make the simplification?

claravanstaden (approved):

PR #1051, opened 2023-12-14
Audit issue 7: Remove obsolete `outbound_fee` parameter

vgeddes:
Outbound fees on ethereum are now calculated dynamically using an exchange rate, and so this `outbound_fee` parameter is obsolete and should have been previously removed.

> In parachain/pallets/system/src/lib.rs:441-465 and in parachain/pallets/system/src/lib.rs:479-495, the update_channel and force_update_channel extrinsics permit to modify channels by specifying the mode and the outbound_fee.
> However, the outbound_fee is currently disregarded and not processed either in the pallet or within the Gateway contract; it is solely emitted in the event.
> Despite being defined in the comment as the “Fee charged to users for sending outbound messages to Polkadot”, this specific fee per channel is not currently implemented.


alistair-singh (approved):

yrong (approved):
LGTM!

claravanstaden (approved):

PR #1052, opened 2023-12-14
Audit issue 9: Check if max nonce reached in outbound-queue pallet

vgeddes:
> In parachain/pallets/outbound-queue/src/lib.rs:313, the nonce is correctly incremented using saturated addition, safeguarding against overflow.
> 
> However, when the nonce reaches its maximum value, messages with that one can still be sent across the bridge. Only the initial message with the maximum nonce will succeed; subsequent ones will be rejected.


parachain/pallets/outbound-queue/src/lib.rs
```diff
@@ -198,6 +198,8 @@ pub mod pallet {
 		InvalidFeeConfig,
 		/// Invalid Channel
 		InvalidChannel,
+		/// Maximum Nonce reached
+		MaxNonceReached,
```
alistair-singh, line 202: This error looks to be unused.

parachain/pallets/outbound-queue/src/lib.rs
```diff
@@ -308,10 +310,19 @@ pub mod pallet {
...
-
-			let next_nonce = Nonce::<T>::get(queued_message.channel_id).saturating_add(1);
+			// Obtain next nonce
+			let nonce = <Nonce<T>>::try_mutate(
+				queued_message.channel_id,
+				|nonce| -> Result<u64, ProcessMessageError> {
+					if *nonce == u64::MAX {
+						return Err(Unsupported)
+					}
+					*nonce = nonce.saturating_add(1);
```
yrong, line 320: Could be simplified with something like `checked_add(1).ok_or(Error::<T>::Overflow)?`

PR #1053, opened 2023-12-14
Add CustomDigestItem type

vgeddes:
The outbound queue pallet inserts a merkle root as a custom digest item into the parachain header.

To play nice with other bridges, this type allows these bridges to insert their own custom digest items, while making sure they don't conflict with each other.


alistair-singh (approved):

yrong:
Comment in
https://github.com/Snowfork/polkadot-sdk/pull/65/files#r1427503626

contracts/src/Verification.sol
```diff
@@ -133,8 +136,17 @@ library Verification {
     {
         for (uint256 i = 0; i < header.digestItems.length; i++) {
             DigestItem memory item = header.digestItems[i];
-            if (item.kind == DIGEST_ITEM_OTHER && item.data.length == 32 && commitment == bytes32(item.data)) {
-                return true;
+            if (item.kind == DIGEST_ITEM_OTHER && item.data.length == 33 && item.data[0] == DIGEST_ITEM_PREFIX) {
+                bytes memory commitment_in_header = new bytes(32);
+                for (uint256 j = 1; j < 33;) {
```
vgeddes, line 141: Can't we use array slices here? https://docs.soliditylang.org/en/v0.6.0/types.html#array-slices

yrong: > As of now, array slices are only implemented for calldata arrays. 

As noted on the above page, or error is raised as follows:
`Error (1227): Index range access is only supported for dynamic calldata arrays.`

vgeddes: Right, so isn't `header.digestItems[i].data` exactly that?

vgeddes: If we remove this line, and reference the calldata directly, then it should compile?
```
 DigestItem memory item = header.digestItems[i];
```

yrong: https://github.com/Snowfork/snowbridge/pull/1053/commits/7ba6c0ec0a3cd78d038578db1849bf9f604ab06e

yrong (approved):
All smoke tests work as expected, should be fine now.

claravanstaden (approved):
Cool!

contracts/src/Verification.sol
```diff
@@ -76,6 +76,9 @@ library Verification {
     uint256 public constant DIGEST_ITEM_PRERUNTIME = 6;
     uint256 public constant DIGEST_ITEM_RUNTIME_ENVIRONMENT_UPDATED = 8;
 
+    /// @dev Prefix of enum variants of DigestItem(0 is reserved for snowbridge)
+    bytes1 public constant DIGEST_ITEM_PREFIX = 0;
```
vgeddes, line 80: Prefer this naming:
```suggestion
    bytes1 public constant DIGEST_ITEM_OTHER_SNOWBRIDGE = 0x00;
```

web/packages/test/scripts/configure-assethub.sh
vgeddes, line 1: hmm, this file does not seem relevant to this PR?

contracts/src/Verification.sol
yrong: [d0bb6c9](https://github.com/Snowfork/snowbridge/pull/1053/commits/d0bb6c91d135c6be705ce85f43e166d29b6960a0)

web/packages/test/scripts/configure-assethub.sh
yrong: It's required for `reserve_transfer` but is already covered in another PR, so removed here.

PR #1056, opened 2023-12-15
Fix for smoke tests

yrong:
- Seems `Utility::batch_all` through xcm is not allowed in bridgeHub which we depend on for some of the smoke tests, some revamp accordingly
- Xcm version of Ethereum should be configured in assetHub for `reserve_transfer`


smoketest/src/helper.rs
```diff
@@ -286,7 +286,7 @@ pub async fn construct_transfer_native_from_agent_call(
 	Ok(call)
 }
 
-pub async fn governance_bridgehub_call_from_relay_chain(
+pub async fn governance_bridgehub_batch_call_from_relay_chain(
```
claravanstaden, line 289: I guess we can remove this if it is not used anymore?

claravanstaden (approved):
Nice! 👏🏻

vgeddes (approved):
+1

vgeddes (approved):
+1

PR #1060, opened 2023-12-16
Make base fee configurable from storage.

alistair-singh:
Polkadot-sdk: https://github.com/Snowfork/polkadot-sdk/pull/68

### DIFF
--- a/parachain/runtime/tests/src/lib.rs
+++ b/parachain/runtime/tests/src/lib.rs
@@ -5,7 +5,7 @@
 
 mod test_cases;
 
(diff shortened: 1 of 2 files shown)


parachain/runtime/tests/src/lib.rs
```diff
@@ -5,7 +5,7 @@
 
 mod test_cases;
 
-use asset_hub_rococo_runtime::xcm_config::bridging::to_ethereum::BridgeHubEthereumBaseFeeInROC;
+use asset_hub_rococo_runtime::xcm_config::bridging::to_ethereum::DefaultBridgeHubEthereumBaseFeeInROC;
```
alistair-singh, line 8: Using the const version in the tests as the storage version causes it to panic due to there is no underlying storage provider.

PR #1063, opened 2023-12-17
Init agents and channels with migration

alistair-singh:
Polkadot-sdk: https://github.com/Snowfork/polkadot-sdk/pull/70

Adds a migration script that will initialize the agents and channels if not initialized by the genesis build.
Adds a new extrinsic which can be used to force initialization.

Tested the extrinsic manually and via unit tests. Tested the migration by doing a manual runtime upgrade and by using the `try-runtime` tool from Parity.

```
$ try-runtime  --runtime  gen.raw.wasm  on-runtime-upgrade  live --uri wss://rococo-bridge-hub-rpc.dwellir.com:443 2>&1 | grep ethereum_system
[2023-12-18T15:55:55Z INFO  ethereum_system::migration] Agents and channels not initialized. Initialization will run.
[2023-12-18T15:55:55Z INFO  ethereum_system::migration] Ethereum system initialized.
[2023-12-18T15:55:56Z INFO  ethereum_system::migration] Ethereum system initialized.
[2023-12-18T15:55:56Z INFO  ethereum_system::migration] Agents and channels are initialized. Initialization will not run.
[2023-12-18T15:55:56Z INFO  ethereum_system::migration] Ethereum system already initialized. Skipping.
```

### DIFF
--- a/parachain/pallets/system/src/lib.rs
+++ b/parachain/pallets/system/src/lib.rs
@@ -622,7 +622,10 @@ pub mod pallet {
 		}
 
 		/// Initializes agents and channels.
-		pub fn initialize(para_id: ParaId, asset_hub_para_id: ParaId) -> Result<(), DispatchError> {
+		pub(crate) fn initialize(
+			para_id: ParaId,
+			asset_hub_para_id: ParaId,
+		) -> Result<(), DispatchError> {
 			let bridge_hub_agent_id = agent_id_of::<T>(&MultiLocation::here())?;
 			// Agent for BridgeHub
 			Agents::<T>::insert(bridge_hub_agent_id, ());
--- a/parachain/pallets/system/src/lib.rs
+++ b/parachain/pallets/system/src/lib.rs
@@ -47,6 +47,7 @@ mod tests;
 
 #[cfg(feature = "runtime-benchmarks")]
 mod benchmarking;
+pub mod migration;
 
 pub mod api;
 pub mod weights;
@@ -133,7 +134,10 @@ pub mod pallet {
 
 	use super::*;
 
+	const STORAGE_VERSION: StorageVersion = StorageVersion::new(1);
+
 	#[pallet::pallet]
+	#[pallet::storage_version(STORAGE_VERSION)]
 	pub struct Pallet<T>(_);
 
 	#[pallet::config]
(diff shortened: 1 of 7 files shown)


parachain/pallets/system/src/lib.rs
```diff
@@ -256,34 +257,7 @@ pub mod pallet {
...
-				ParentThen(X1(Parachain(self.asset_hub_para_id.into()))).into();
-			let asset_hub_agent_id =
-				agent_id_of::<T>(&asset_hub_location).expect("infallible; qed");
-			let asset_hub_channel_id: ChannelId = self.asset_hub_para_id.into();
-			Agents::<T>::insert(asset_hub_agent_id, ());
-			Channels::<T>::insert(
-				asset_hub_channel_id,
-				Channel { agent_id: asset_hub_agent_id, para_id: self.asset_hub_para_id },
-			);
+			Pallet::<T>::initialize(self.para_id, self.asset_hub_para_id).expect("infallible; qed");
```
yrong, line 260: Since the initialize call was introduced, maybe we can just remove the GenesisConfig here?

parachain/pallets/system/src/lib.rs
```diff
@@ -594,6 +568,29 @@ pub mod pallet {
...
+		///
+		/// - `origin`: Must be root
+		/// - `own_para_id`: The parachain id of the parachain housing this pallet. Most likely
+		///   BridgeHub.
+		/// - `asset_hub_para_id`: The parachain id of AssetHub.
+		#[pallet::call_index(10)]
+		#[pallet::weight((T::WeightInfo::force_initialize(), DispatchClass::Operational))]
+		pub fn force_initialize(
+			origin: OriginFor<T>,
+			own_para_id: ParaId,
```
yrong, line 587: `own_para_id` seems unnecessary which can be retrieved from [runtime](https://github.com/Snowfork/polkadot-sdk/blob/00e8eb2e04375260d728f143352e21cbf7a4436c/cumulus/parachains/runtimes/bridge-hubs/bridge-hub-rococo/src/lib.rs#L337).

vgeddes: Disagree, actually, the runtime is configured with the para_id using the genesis config for the ParachainInfo pallet.

vgeddes (commented):
As per offline discussions, lets deliver this is as migration script, and remove the governance extrinsic.

parachain/pallets/system/src/lib.rs
alistair-singh: Removed the initialize call.

alistair-singh:
> As per offline discussions, lets deliver this is as migration script, and remove the governance extrinsic.

@vgeddes @yrong I have removed the extrinsic and this is ready for final review.

parachain/pallets/system/src/lib.rs
```diff
@@ -643,6 +617,49 @@ pub mod pallet {
 			});
 			Ok(())
 		}
+
+		/// Checks if the pallet has been initialized.
+		pub(crate) fn is_initialized() -> bool {
```
vgeddes, line 622: I think this can be moved to the migration script

vgeddes: Actually, ignore this.

vgeddes:
We also need to initialize the storage for the runtime tests, 

for example in https://github.com/Snowfork/snowbridge/blob/afc668ef634deb0ed6a8e44f7ee75c0c22f80ed1/parachain/runtime/tests/src/test_cases.rs#L37

To ensure that channels have been created before sending tokens back to Ethereum

vgeddes (approved):

PR #1064, opened 2023-12-17
Fix penpal smoketest

alistair-singh:
Add fee to fix test.

### DIFF
--- a/smoketest/tests/send_token.rs
+++ b/smoketest/tests/send_token.rs
@@ -51,9 +51,10 @@ async fn send_token() {
 		.unwrap();
 	assert_eq!(receipt.status.unwrap().as_u64(), 1u64);
 
+	let fee = gateway.quote_register_token_fee().call().await.unwrap();
+
 	// Lock tokens into vault
 	let amount: u128 = U256::from(value).low_u128();
-	let fee: u128 = 30_000_000_000_000_000;
 	let receipt = gateway
 		.send_token(
 			weth.address(),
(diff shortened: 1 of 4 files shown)


claravanstaden (approved):

smoketest/tests/send_token_to_penpal.rs
```diff
@@ -63,6 +63,7 @@ async fn send_token_to_penpal() {
 
 	// Lock tokens into vault
 	let amount: u128 = U256::from(value).low_u128();
+	let fee: u128 = 30_000_000_000_000_000;
```
vgeddes, line 66: There is an API `IGateway.quoteSendTokenFee` to retrieve the fee. I think I prefer to use that because then the test would verify that it returns a fee large enough.

alistair-singh: addressed in https://github.com/Snowfork/snowbridge/pull/1064/commits/dd7cb06727d41448bfaca30f1c5a1720127788c3

smoketest/tests/send_token.rs
```diff
@@ -51,9 +51,10 @@ async fn send_token() {
 		.unwrap();
 	assert_eq!(receipt.status.unwrap().as_u64(), 1u64);
 
+	let fee = gateway.quote_register_token_fee().call().await.unwrap();
```
vgeddes, line 54: Wrong API, need to call `quoteSendTokenFee`.

smoketest/tests/send_token_to_penpal.rs
```diff
@@ -61,6 +61,8 @@ async fn send_token_to_penpal() {
 		.unwrap();
 	assert_eq!(receipt.status.unwrap().as_u64(), 1u64);
 
+	let fee = gateway.quote_register_token_fee().call().await.unwrap();
```
vgeddes, line 64: ditto. need to call `quoteSendTokenFee`.

vgeddes (approved):
+1

PR #1065, opened 2023-12-17
Take LengthToFee into account

yrong:


parachain/pallets/inbound-queue/src/lib.rs
```diff
@@ -322,9 +327,14 @@ pub mod pallet {
...
-			let refund: BalanceOf<T> = T::WeightToFee::weight_to_fee(&T::WeightInfo::submit());
-			refund.saturating_add(pricing_parameters.rewards.local)
+			let weight_fee = T::WeightToFee::weight_to_fee(&T::WeightInfo::submit());
+			let length_fee = T::LengthToFee::weight_to_fee(&Weight::from_parts(
+				T::MaxMessagePayloadSize::get() as u64,
+				0,
+			));
+			let refund = weight_fee.saturating_add(length_fee);
+			let reward = T::PricingParameters::get().rewards.local;
+			refund.saturating_add(reward)
```
vgeddes, line 337: I think we should break this out into a function named for example `Pallet::calculate_delivery_cost` which takes a `length` parameter.

Then we can compute the worst-case cost and actual cost.

Note that you also need to use this function when rewarding the relayer.

parachain/pallets/inbound-queue/src/lib.rs
```diff
@@ -322,9 +327,14 @@ pub mod pallet {
 	/// API for accessing the delivery cost of a message
 	impl<T: Config> Get<BalanceOf<T>> for Pallet<T> {
 		fn get() -> BalanceOf<T> {
-			let pricing_parameters = T::PricingParameters::get();
-			let refund: BalanceOf<T> = T::WeightToFee::weight_to_fee(&T::WeightInfo::submit());
-			refund.saturating_add(pricing_parameters.rewards.local)
+			let weight_fee = T::WeightToFee::weight_to_fee(&T::WeightInfo::submit());
+			let length_fee = T::LengthToFee::weight_to_fee(&Weight::from_parts(
+				T::MaxMessagePayloadSize::get() as u64,
```
vgeddes, line 332: I looked at this problem earlier today. We need to consider the maximum possible length of the `message` parameter passed to submit. According to my rough estimates of receipt proof length, and transaction receipt size, this is going to be around 2KiB.

parachain/pallets/inbound-queue/src/mock.rs
```diff
@@ -219,6 +219,8 @@ impl inbound_queue::Config for Test {
 	#[cfg(feature = "runtime-benchmarks")]
 	type Helper = Test;
 	type WeightToFee = IdentityFee<u128>;
+	type LengthToFee = IdentityFee<u128>;
+	type MaxMessagePayloadSize = ConstU32<1024>;
```
vgeddes, line 223: The size is the payload as well as the proof, and according to my guess its going to be around 2KIB,

```suggestion
	type MaxMessageSize = ConstU32<2048>;
```

parachain/pallets/inbound-queue/src/lib.rs
yrong: https://github.com/Snowfork/snowbridge/pull/1065/commits/0eaa53fb013fff61bb5cbd65a41e3b7adcd91d69

parachain/pallets/inbound-queue/src/mock.rs
yrong: This is the mock and yeah 2048 is the value in runtime.

vgeddes: Rename to `MaxMessageSize` please

yrong: https://github.com/Snowfork/snowbridge/pull/1065/commits/1033090342e61c1d5a86a893187794c971e553f1

parachain/pallets/inbound-queue/src/lib.rs
```diff
@@ -253,8 +258,15 @@ pub mod pallet {
 			// Reward relayer from the sovereign account of the destination parachain
 			// Expected to fail if sovereign account has no funds
 			let sovereign_account = sibling_sovereign_account::<T>(channel.para_id);
-			let delivery_cost = Self::get();
-			T::Token::transfer(&sovereign_account, &who, delivery_cost, Preservation::Preserve)?;
+			let delivery_cost = Self::calculate_delivery_cost(message.encode().len() as u32);
+			if delivery_cost > T::Token::minimum_balance() {
```
vgeddes, line 262: This seems unnecessary. In practice `delivery_cost` is always going to be greater than the minimum balance.

parachain/pallets/inbound-queue/src/envelope.rs
```diff
@@ -30,10 +30,10 @@ pub struct Envelope {
 #[derive(Copy, Clone, RuntimeDebug)]
 pub struct EnvelopeDecodeError;
 
-impl TryFrom<Log> for Envelope {
+impl TryFrom<&Log> for Envelope {
```
vgeddes, line 33: This change seems unnecessary and unrelated :)

yrong: Or we need an extra clone here. https://github.com/Snowfork/snowbridge/blob/1033090342e61c1d5a86a893187794c971e553f1/parachain/pallets/inbound-queue/src/lib.rs#L236

parachain/pallets/inbound-queue/src/lib.rs
yrong: So will that always be true in other networks? Notice the `ED` in Polkadot is pretty high.
https://github.com/polkadot-fellows/runtimes/blob/30e0dbfdcb78722ed61325c0ebf1efdcdb6033ba/relay/polkadot/constants/src/lib.rs#L28

parachain/pallets/inbound-queue/src/envelope.rs
vgeddes: Ok, lets keep your change them

parachain/pallets/inbound-queue/src/lib.rs
vgeddes: Still seems unnecessary. The relayer would have already have had to fund their account with an amount greater than `ED` in order to submit the message.

vgeddes (approved):
+1

alistair-singh (approved):
Looks good!

PR #1067, opened 2023-12-18
Verify that channel exists when validating outbound message

vgeddes:
Polkadot-sdk: https://github.com/Snowfork/polkadot-sdk/pull/73
This is a simpler version of https://github.com/Snowfork/snowbridge/pull/1047

### DIFF
--- a/parachain/runtime/tests/src/test_cases.rs
+++ b/parachain/runtime/tests/src/test_cases.rs
@@ -4,6 +4,7 @@
 //! Module contains predefined test-case scenarios for `Runtime` with bridging capabilities.
 
 use asset_hub_rococo_runtime::xcm_config::bridging::to_ethereum::DefaultBridgeHubEthereumBaseFee;
+use bridge_hub_rococo_runtime::EthereumSystem;
 use codec::Encode;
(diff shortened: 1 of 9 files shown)


yrong (approved):
LGTM!

parachain/runtime/tests/src/test_cases.rs
```diff
@@ -279,46 +287,6 @@ pub fn send_transfer_token_message_fee_not_enough<Runtime, XcmConfig>(
 				fee_amount,
 			);
 			// check err is NotHoldingFees
-			assert_err!(outcome.ensure_complete(), NotHoldingFees);
-		});
-}
-
-pub fn send_transfer_token_message_insufficient_fund<Runtime, XcmConfig>(
```
claravanstaden, line 286: Why do we remove this test?

alistair-singh: This is a test utility that's called from multiple tests, not a test itself. This was duplicated for each test case and I refactored it into a single [test utility that](https://github.com/Snowfork/snowbridge/pull/1067/files#diff-6b7b2955e408c054f0caa88f257fec221c05670262882c78a2794f54b3705ee8R247-L250).

PR #1077, opened 2023-12-20
Fix sibling location

yrong:
Address for the comment https://github.com/paritytech/polkadot-sdk/pull/2522#discussion_r1425275414
Resolves: SNO-803


smoketest/src/constants.rs
```diff
@@ -30,7 +30,7 @@ pub const BRIDGE_HUB_AGENT_ID: [u8; 32] =
 	hex!("03170a2e7597b7b7e3d84c05391d139a62b157e78786d8c082f29dcf4c111314");
 // Agent for asset hub parachain 1000
 pub const ASSET_HUB_AGENT_ID: [u8; 32] =
-	hex!("72456f48efed08af20e5b317abf8648ac66e86bb90a411d9b0b713f7364b75b4");
+	hex!("81c5ab2571199e3188135178f3c2c8e2d268be1313d029b30f534fa579b69b79");
```
claravanstaden, line 33: So does the ID change because we now consistently refer derive the agent ID using `parents: 1`?

claravanstaden (approved):

alistair-singh (approved):

parachain/primitives/router/src/outbound/tests.rs
```diff
@@ -1,11 +1,14 @@
...
+use snowbridge_core::{
+	outbound::{Fee, SendError, SendMessageFeeProvider},
+	DescribeHere,
+};
 
 use xcm::v3::prelude::SendError as XcmSendError;
 use xcm_builder::{DescribeAllTerminal, DescribeFamily, HashedDescription};
 
-pub type AgentIdOf = HashedDescription<H256, DescribeFamily<DescribeAllTerminal>>;
+pub type AgentIdOf = HashedDescription<H256, (DescribeHere, DescribeFamily<DescribeAllTerminal>)>;
```
vgeddes, line 11: We're also repeating this line in the BridgeHub `xcm_config.rs` configuration, can you check?

In which case we should factor out both definitions into our [parachain/runtime/runtime-common](https://github.com/Snowfork/snowbridge/blob/main/parachain/runtime/runtime-common/src/lib.rs) crate.

yrong: Yeah, several duplications and refactoring with https://github.com/Snowfork/snowbridge/pull/1077/commits/8c1d1d89dd5df205fbe90c4cf256252bc771494c

Btw, prefer to move to primitives as it's not only used in runtime tests.

vgeddes (approved):
+1

PR #1083, opened 2023-12-21
Update milagro

yrong:
Requires: https://github.com/Snowfork/milagro_bls/pull/3 (remove unnecessary formatting there for audit)


parachain/Cargo.toml
```diff
@@ -29,3 +29,6 @@ snowbridge-outbound-queue-runtime-api = { path = "./pallets/outbound-queue/runti
 snowbridge-router-primitives = { path = "./primitives/router" }
 snowbridge-runtime-common = { path = "./runtime/runtime-common" }
 snowbridge-rococo-common = { path = "./runtime/rococo-common" }
+
+[workspace.dependencies]
+milagro_bls = { git = "https://github.com/snowfork/milagro_bls", default-features = false, rev = "b29806d51686f36589bdb0c759ca7d63165e1054" }
```
claravanstaden, line 34: Why do we do it this way, instead of just adding it in the beacon primitive crate?

yrong: It's for reuse purposes, declare the dependency in only one place then reference as

```
milagro_bls = { workspace = true }
```

claravanstaden: But do we not only use it in one place, in the primitives crate?

yrong: https://github.com/Snowfork/snowbridge/pull/1083/commits/dd84492ceda3e1cf210f5bb99d8a88d05d950390

yrong:
Closed in favor of https://github.com/Snowfork/snowbridge/pull/1124

PR #1090, opened 2023-12-22
Ensure commitPrevRandao called after submitInitial

vgeddes:
Make sure to explicitly check that the ticket has been initialized before capturing PREVRANDAO.

In practice, we were protected because of this check, however its still good to be explicit:
https://github.com/Snowfork/snowbridge/blob/cdd922433ff509d1108ec2541946e5b230697fff/contracts/src/BeefyClient.sol#L316


alistair-singh (approved):
LGTM!

PR #1096, opened 2024-01-03
Fix refund surplus

yrong:
Avoid the `Asset Trapped` event log in AH when register/send token.


contracts/src/Gateway.sol
```diff
@@ -206,7 +206,7 @@ contract Gateway is IGateway, IInitializable {
         // Calculate a gas refund, capped to protect against huge spikes in `tx.gasprice`
         // that could drain funds unnecessarily. During these spikes, relayers should back off.
         uint256 gasUsed = _transactionBaseGas() + (startGas - gasleft());
-        uint256 refund = gasUsed * Math.min(tx.gasprice, message.maxFeePerGas);
+        uint256 refund = Math.min(maxDispatchGas * message.maxFeePerGas, gasUsed * tx.gasprice);
```
yrong, line 209: Actually `maxDispatchGas * message.maxFeePerGas` is what we charged at most from the end user.

parachain/primitives/router/src/inbound/mod.rs
```diff
@@ -198,10 +224,17 @@ where
...
-			// are claimable by the physical origin (BridgeHub)
-			ClearOrigin,
+			SetAppendix(Xcm(vec![
+				RefundSurplus,
+				ClearOrigin,
+				DepositAsset {
+					assets: Wild(AllCounted(1u32)),
+					beneficiary: MultiLocation {
+						parents: 1,
+						interior: X1(Junction::Parachain(SelfParaId::get().into())),
```
yrong, line 234: It's for refunding the surplus fee to sovereign BH in AH.

parachain/primitives/router/src/inbound/mod.rs
```diff
@@ -268,17 +302,17 @@ where
 						xcm: vec![
 							// Buy execution on target.
 							BuyExecution { fees: dest_para_fee_asset, weight_limit: Unlimited },
-							// Deposit asset to beneficiary.
-							DepositAsset { assets: Definite(asset.into()), beneficiary },
+							// Deposit asset and fee to beneficiary.
+							DepositAsset { assets: Wild(AllCounted(2u32)), beneficiary },
```
yrong, line 306: It's for refunding the surplus fee to the beneficiary.

contracts/src/Gateway.sol
yrong, line 209: `message.maxDispatchGas` is what we set in BH for estimating the gas cost and charge from the user, assuming it's somehow underestimated and far less than real `gasUsed` here, we may refund too much, so IMO better to use that to calculate the minimal refund.

parachain/primitives/router/src/inbound/mod.rs
```diff
@@ -198,10 +224,17 @@ where
...
+				RefundSurplus,
+				ClearOrigin,
+				DepositAsset {
+					assets: Wild(AllCounted(1u32)),
+					beneficiary: MultiLocation {
+						parents: 1,
+						interior: X1(Junction::Parachain(SelfParaId::get().into())),
+					},
+				},
+			])),
```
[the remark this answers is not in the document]
yrong: Yes, the surplus fee asset will get trapped in AH.

contracts/src/Gateway.sol
vgeddes: Lets rather consider this separately. It is unrelated to refunding surplus weight on AssetHub.

parachain/primitives/router/src/inbound/mod.rs
```diff
@@ -268,17 +301,17 @@ where
 						xcm: vec![
 							// Buy execution on target.
 							BuyExecution { fees: dest_para_fee_asset, weight_limit: Unlimited },
-							// Deposit asset to beneficiary.
-							DepositAsset { assets: Definite(asset.into()), beneficiary },
+							// Deposit asset and fee to beneficiary.
+							DepositAsset { assets: Wild(AllCounted(2u32)), beneficiary },
```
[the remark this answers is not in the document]
yrong: Yes, It's what is left in the holding registry and exactly the token asset plus the fee asset left.

contracts/src/Gateway.sol
yrong: Reverted in https://github.com/Snowfork/snowbridge/pull/1096/commits/f426fcc931fa23d330383d988d800f5c1a447c35

@vgeddes Do you think the fix here makes sense? If so I'll add another PR for it.

parachain/primitives/router/src/inbound/mod.rs
```diff
@@ -198,10 +224,17 @@ where
 					.encode()
 					.into(),
 			},
-			RefundSurplus,
-			// Clear the origin so that remaining assets in holding
-			// are claimable by the physical origin (BridgeHub)
-			ClearOrigin,
+			SetAppendix(Xcm(vec![
```
alistair-singh, line 227: So I am not sure what we gain from using SetAppendix at the end of an xcm message? SetAppendix works like the `finally` of a `try/catch/finally` block. Setting it at the end means it won't run if there are any errors in the instructions before it. The `Transact` always returns success, you have to call `ExpectTransactStatus` instruction to make it trigger xcm error handlers. So I would remove the SetAppendix completely and just add the DepositAsset after the ClearOrigin, or move the SetAppendix closer to the top so if it errors anywhere then the left over fees go to Bridgehub sovereign.

yrong: Nice catch, I revert in

parachain/primitives/router/src/inbound/mod.rs
```diff
@@ -185,6 +211,18 @@ where
 			DescendOrigin(X1(PalletInstance(inbound_queue_pallet_index))),
 			// Change origin to the bridge.
 			UniversalOrigin(GlobalConsensus(network)),
+			// SetAppendix to refund the surplus fee asset to BH sovereign in AH
```
vgeddes, line 214:
```suggestion
			// Refund the surplus execution weight to BH sovereign
```

parachain/primitives/router/src/inbound/mod.rs
alistair-singh: On second thought I think we should rather move the SetAppendix after the BuyExecution but before the DepositAsset. That way any error after buying execution will result in the funds being returned to the bridgehub sovereign.

parachain/primitives/router/src/inbound/mod.rs
```diff
@@ -185,6 +211,18 @@ where
 			DescendOrigin(X1(PalletInstance(inbound_queue_pallet_index))),
 			// Change origin to the bridge.
 			UniversalOrigin(GlobalConsensus(network)),
+			// Refund the surplus execution weight to BH sovereign
+			SetAppendix(Xcm(vec![
+				RefundSurplus,
+				ClearOrigin,
+				DepositAsset {
+					assets: Wild(AllCounted(1u32)),
+					beneficiary: MultiLocation {
```
vgeddes, line 220: The origin in this context should actually be `GlobalConsensus(Ethereum)`.

Also it better to be more specific, since other bridges are also using the BH sovereign account

yrong: https://github.com/Snowfork/snowbridge/pull/1096/commits/ec4ce325c1b7bf49d5ea530863375c9c6827ee8a

parachain/primitives/router/src/inbound/mod.rs
yrong: https://github.com/Snowfork/snowbridge/pull/1096/commits/ec4ce325c1b7bf49d5ea530863375c9c6827ee8a

alistair-singh (approved):
🚀

vgeddes (commented):
This one can wait until after launch so that we can minimize changes for the upcoming audit. It fixes a very minor problem.

parachain/primitives/router/src/inbound/mod.rs
vgeddes: Sorry, I actually meant `recipient` instead of `origin`. The recipient of the surplus funds should be `GlobalConsensus(Ethereum)`. That way, some other bridge on BridgeHub can't come over and claim our surplus fees.

PR #1097, opened 2024-01-04
Test fixture for sepolia

yrong:
- [x] Test fixture from Sepolia testnet
- [x] Revamp conditional compile with `all-features` enabled and improve CI to cover the checks.


relayer/relays/beacon/header/syncer/syncer.go
```diff
@@ -117,16 +117,18 @@ func (s *Syncer) GetSyncCommitteePeriodUpdate(from uint64) (scale.Update, error)
...
 	if err != nil {
 		return scale.Update{}, fmt.Errorf("beacon header hash tree root: %w", err)
 	}
 
+	blockRootsProof, err := s.GetBlockRoots(uint64(finalizedHeader.Slot))
+	// It's possible that beaconState is not available when GetSyncCommitteePeriodUpdate, in this case
+	// there is no blockRootsProof and just ignore it in beacon client
+	if err != nil && err != ErrBeaconStateAvailableYet {
+		return scale.Update{}, fmt.Errorf("fetch block roots: %w", err)
+	}
```
yrong, line 130: Allow SyncCommiteeUpdate without blockRootsProof

parachain/pallets/ethereum-beacon-client/src/lib.rs
```diff
@@ -418,17 +418,18 @@ pub mod pallet {
...
+					verify_merkle_branch(
+						update.block_roots_root,
+						&update.block_roots_branch,
+						config::BLOCK_ROOTS_SUBTREE_INDEX,
+						config::BLOCK_ROOTS_DEPTH,
+						update.finalized_header.state_root
+					),
+					Error::<T>::InvalidBlockRootsRootMerkleProof
+				);
+			}
```
yrong, line 432: For the reason mentioned in https://github.com/Snowfork/snowbridge/pull/1097#discussion_r1441432101

yrong:
@claravanstaden 
In this PR I've also revamped the condition compile a bit to enable all features which is what Parity does for `polkadot-sdk` in their CI, the main difference with https://github.com/paritytech/polkadot-sdk/pull/2792 is that when running test with the two conflicting features `fast-runtime` and `runtime-benchmarks` enabled at the same time, the minimal spec will get applied and benchmark for beacon client will be ignored.  IMO it will make the conditional compile code a bit cleaner.

Please let me know what you think and if anything important missing.

parachain/pallets/ethereum-beacon-client/src/lib.rs
claravanstaden: Hmmm but how can this merkle proof be conditional? Could it not lead to unexpected behaviour? As an example, this check allows finalized header verification without providing the block roots proof. What could happen is we keep processing these beacon headers without the necessary ancestry proofs (needed for the execution header imports). Then, our execution header imports could fail because there are no ancestry proofs to verify the execution header against a beacon header.

parachain/pallets/ethereum-beacon-client/src/lib.rs
```diff
@@ -518,7 +519,9 @@ pub mod pallet {
 				});
 			};
 
-			if update.finalized_header.slot > latest_finalized_state.slot {
+			if update.finalized_header.slot > latest_finalized_state.slot &&
+				!update.block_roots_branch.is_empty()
+			{
```
claravanstaden, line 524: If we don't store the finalized header if the `block_roots_branch` is empty, does it makes sense for the relayer to send this update in the first place? Would it not be a pointless update then?

parachain/pallets/ethereum-beacon-client/src/mock.rs
```diff
@@ -1,22 +1,127 @@
+use hex_literal::hex;
+use std::{fs::File, path::PathBuf};
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: 2023 Snowfork <hello@snowfork.com>
```
claravanstaden, line 4:
```suggestion
// SPDX-License-Identifier: Apache-2.0
// SPDX-FileCopyrightText: 2023 Snowfork <hello@snowfork.com>
use hex_literal::hex;
use std::{fs::File, path::PathBuf};
```

License should be the first lines in the file.

relayer/cmd/generate_beacon_data.go
```diff
@@ -215,8 +186,17 @@ func generateBeaconData(cmd *cobra.Command, _ []string) error {
...
+		if err != nil {
+			return fmt.Errorf("get sync committee update: %w", err)
+		}
+		syncCommitteeUpdate := syncCommitteeUpdateScale.Payload.ToJSON()
+		writeJSONToFile(syncCommitteeUpdate, fmt.Sprintf("sync-committee-update.%s.json", activeSpec.ToString()))
+		log.Info("created sync committee update file")
+
 		// generate FinalizedUpdate for next epoch
-		log.Info("waiting for a new finalized_update in next epoch and in current sync period,several seconds required...")
+		log.Info("waiting finalized_update in next epoch(6.4 minutes later), be patient and drink a cup of tea...")
```
claravanstaden, line 199: 🍵 🙃

relayer/cmd/generate_beacon_data.go
```diff
@@ -266,32 +237,7 @@ func generateBeaconData(cmd *cobra.Command, _ []string) error {
...
-			nextSyncCommitteeUpdateScale, err := s.GetSyncCommitteePeriodUpdate(initialSyncPeriod + 1)
-			if err != nil {
-				return fmt.Errorf("get sync committee update: %w", err)
-			}
-			nextSyncCommitteeUpdate := nextSyncCommitteeUpdateScale.Payload.ToJSON()
-			writeJSONToFile(nextSyncCommitteeUpdate, fmt.Sprintf("next-sync-committee-update.%s.json", activeSpec.ToString()))
-			log.Info("created next sync committee update file")
-		}
-
+		// Generate benchmark fixture for mainnet spec only
```
claravanstaden, line 240: How would we regenerate the minimal spec test fixture data? 🤔

relayer/relays/beacon/header/syncer/syncer.go
claravanstaden: I'm not sure this change is a good one. Does it even make sense to send the finalized header update without an ancestry proof? That's the main reason we send beacon header updates in the first place.

web/packages/test/scripts/configure-bridgehub.sh
```diff
@@ -6,9 +6,10 @@ source scripts/xcm-helper.sh
 
 config_beacon_checkpoint() {
     pushd $root_dir
-    local check_point_call=$($relay_bin generate-beacon-checkpoint --spec $active_spec --url $beacon_endpoint_http)
+    local check_point_hex=$($relay_bin generate-beacon-checkpoint --spec $active_spec --url $beacon_endpoint_http)
     popd
-    send_governance_transact_from_relaychain $BRIDGE_HUB_PARAID "$check_point_call" 180000000000 900000
+    local transact_call="0x5200"$check_point_hex
```
claravanstaden, line 11: Nice, it is better to keep the call index just here.

claravanstaden (commented):
Hey @yrong! Thanks for these changes. I think resolving the all-features problem in https://github.com/paritytech/polkadot-sdk/pull/2792 in the PR is better than the fixes I attempted. I'm not sure we will be able to convince Parity of the conditional `add_benchmarks` in the Rococo runtime, we'll have to see.

I am not comfortable with allowing beacon header updates without ancestry proofs. But perhaps I am missing something in my understanding, please let me know.

relayer/relays/beacon/header/syncer/syncer.go
yrong: It's an edge case that I run `generateBeaconTestFixture` for the `SyncCommitteePeriodUpdate` from sepolia and beaconState is not available at that moment. Code in https://github.com/Snowfork/snowbridge/blob/faacc1e1cb43f5e04b9dae059453d3da717c8cc2/relayer/cmd/generate_beacon_data.go#L189-L193

But I think it's possible to switch to another time to generate the test fixture and as you mentioned it could lead to some unexpected behavior to relax the check so just revert it.

parachain/pallets/ethereum-beacon-client/src/lib.rs
yrong: https://github.com/Snowfork/snowbridge/pull/1097#discussion_r1444244154

relayer/cmd/generate_beacon_data.go
yrong: Yes, I think it's fine to also generate benchmark fixtures for minimal spec, then we can avoid the conditional add_benchmarks hack in the Rococo runtime.

https://github.com/Snowfork/snowbridge/pull/1097/commits/63a1bcdb9a432abaa0d25d564375694242f1d26b

alistair-singh (approved):
LGTM!

relayer/relays/beacon/header/syncer/syncer.go
```diff
@@ -117,14 +117,14 @@ func (s *Syncer) GetSyncCommitteePeriodUpdate(from uint64) (scale.Update, error)
...
 	if err != nil {
-		return scale.Update{}, fmt.Errorf("fetch block roots: %w", err)
+		return scale.Update{}, fmt.Errorf("beacon header hash tree root: %w", err)
 	}
 
-	finalizedHeaderBlockRoot, err := finalizedHeader.ToSSZ().HashTreeRoot()
+	blockRootsProof, err := s.GetBlockRoots(uint64(finalizedHeader.Slot))
 	if err != nil {
-		return scale.Update{}, fmt.Errorf("beacon header hash tree root: %w", err)
+		return scale.Update{}, fmt.Errorf("fetch block roots: %w", err)
```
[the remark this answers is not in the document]
yrong: Not necessary and reverted in https://github.com/Snowfork/snowbridge/pull/1097/commits/87ba289125ee400a3658e20682bd148ea0377cfc

claravanstaden (approved):

vgeddes (changes requested):
Looks great Ron.

However the conditional `add_benchmarks` in the Rococo runtime is problematic. I'm sure Parity will reject it.

As you suggest, lets also generate minimal fixtures for benchmarks.

vgeddes (approved):
Ignore my previous, seems we no longer need to conditionally compile `add_benchmarks`.

Approved!

PR #1098, opened 2024-01-09
Rate limit beefy updates

yrong:
Resolves: SNO-806

In this PR we introduce `beefy-update-period`, a static config param to decide how often to sync with the consensus updates. 

Though it's possible to make it self-adaptive/dynamic to be determined by how far away from the next pending user message, it did increase complexity to the current workflow (we may need to subscribe to recent message commitment of each parachain/channel) so I'm a bit hesitate of that.

 Only focus on the static config in this PR and we can improve it later in another PR if necessary.


relayer/relays/beefy/polkadot-listener.go
```diff
@@ -118,18 +119,24 @@ func (li *PolkadotListener) scanCommitments(
 				case requests <- task:
 					logEntry.Info("New commitment with handover added to channel")
 					currentValidatorSet++
+					lastSyncedBeefyBlock = committedBeefyBlock
 				}
 			} else if validatorSetID == currentValidatorSet {
 				if result.Depth > li.config.FastForwardDepth {
```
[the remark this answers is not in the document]
yrong: So are you referring to reusing the `fast-forward-depth` for the update period? If so I'm afraid not agree.

IMO `fast-forward-depth` is the parameter that describes how far the sync falls behind the relay chain, if far enough then just ignore the non-handover updates. 

Whereas the `update-period` is the sampling config for how often to sync with the relay chain, e.g:  assuming beefy commitment in relaychain updates every 8 blocks and we set 32 here, so only 1/4 of the non-handover updates are sampled.

So they are different and can't be reused.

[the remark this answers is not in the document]
vgeddes: Yeah, that's correct - they are used for different purposes.

vgeddes:
> Though it's possible to make it self-adaptive/dynamic to be determined by how far away from the next pending user message, it did increase complexity to the current workflow (we may need to subscribe to recent message commitment of each parachain/channel) so I'm a bit hesitate of that.

Agree, the simpler approach is fine for now. 👍

web/packages/test/scripts/build-binary.sh
```diff
@@ -17,7 +17,8 @@ build_binaries() {
     # Check that all 3 binaries are available and no changes made in the polkadot and substrate dirs
     if [[ ! -e "target/release/polkadot" || ! -e "target/release/polkadot-execute-worker" || ! -e "target/release/polkadot-prepare-worker" || "$changes_detected" -eq 1 ]]; then
         echo "Building polkadot binary, due to changes detected in polkadot or substrate, or binaries not found"
-        cargo build --release --locked --bin polkadot --bin polkadot-execute-worker --bin polkadot-prepare-worker
+        # Increase session length to 6 mins
+        ROCOCO_EPOCH_DURATION=60 cargo build --release --locked --bin polkadot --bin polkadot-execute-worker --bin polkadot-prepare-worker
```
vgeddes, line 21: 6 minutes is a long time to wait to test handover cases. I would shorten the session to 20 blocks.

relayer/relays/beefy/polkadot-listener.go
```diff
@@ -118,18 +119,24 @@ func (li *PolkadotListener) scanCommitments(
...
 					logEntry.Warn("Discarded commitment with depth not fast forward")
 					continue
 				}
+				if committedBeefyBlock < lastSyncedBeefyBlock+li.config.UpdatePeriod {
+					logEntry.Warn("Discarded commitment with update period too frequently")
```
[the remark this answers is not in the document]
yrong: That's correct! Prefer `Info` though.

web/packages/test/scripts/build-binary.sh
yrong: https://github.com/Snowfork/snowbridge/pull/1098/commits/95f024c9b4c233e3b6a7d041a45953318efc12ea

relayer/relays/beefy/config.go
```diff
@@ -12,6 +12,7 @@ type Config struct {
 type SourceConfig struct {
 	Polkadot         config.PolkadotConfig `mapstructure:"polkadot"`
 	FastForwardDepth uint64                `mapstructure:"fast-forward-depth"`
+	UpdatePeriod     uint64                `mapstructure:"update-period"`
```
vgeddes, line 15: Should add a comment that this quantity is in number of blocks rather than seconds

yrong: https://github.com/Snowfork/snowbridge/pull/1098/commits/92458230cd25e527f362571ae12ceeff6caba099

relayer/relays/beefy/config.go
```diff
@@ -10,9 +10,11 @@ type Config struct {
 }
 
 type SourceConfig struct {
-	Polkadot         config.PolkadotConfig `mapstructure:"polkadot"`
-	FastForwardDepth uint64                `mapstructure:"fast-forward-depth"`
-	UpdatePeriod     uint64                `mapstructure:"update-period"`
+	Polkadot config.PolkadotConfig `mapstructure:"polkadot"`
+	// Depth to ignore the beefy updates too far away(in number of blocks)
+	FastForwardDepth uint64 `mapstructure:"fast-forward-depth"`
+	// Period to sample the beefy updates(in number of blocks)
```
vgeddes, line 16:
```suggestion
	// Period to sample the beefy updates (in number of blocks)
```

relayer/relays/beefy/config.go
```diff
@@ -10,9 +10,11 @@ type Config struct {
 }
 
 type SourceConfig struct {
-	Polkadot         config.PolkadotConfig `mapstructure:"polkadot"`
-	FastForwardDepth uint64                `mapstructure:"fast-forward-depth"`
-	UpdatePeriod     uint64                `mapstructure:"update-period"`
+	Polkadot config.PolkadotConfig `mapstructure:"polkadot"`
+	// Depth to ignore the beefy updates too far away(in number of blocks)
```
vgeddes, line 14:
```suggestion
	// Depth to ignore the beefy updates too far away (in number of blocks)
```

vgeddes (approved):
+1

PR #1103, opened 2024-01-09
Upgrade to Solidity 0.8.23

vgeddes:
Fixes: SNO-759

Also updates Foundry version in Nix environment. Make sure to *reload* your envs


yrong:
Seems the title of the PR is not correct(0.9.23)? 
Btw, except for upgrading the version, seems no incompatible solidity change is required, right?

yrong (approved):
+1

PR #1107, opened 2024-01-10
Handle fee with error instead of silently ignoring

yrong:
Resolves: SNO-820
Requires: https://github.com/Snowfork/polkadot-sdk/pull/95

In edge cases like `BridgeHubEthereumBaseFee` insufficient or price params not correctly initialized, prefer to handle the fee with the error thrown instead of silently ignore.


yrong:
Closes with https://github.com/Snowfork/snowbridge/pull/1109

PR #1109, opened 2024-01-11
Refactoring XcmExportFeeToSibling and more tests for outbound fee

yrong:
Resolves: SNO-820, SNO-812

This is the sibling PR for https://github.com/Snowfork/snowbridge/pull/1107 without the change required for the xcm primitives in https://github.com/Snowfork/polkadot-sdk/pull/95. 

In this PR we focus on 
- adding more sanity checks for the params and more tests accordingly
- improving the error log for `XcmExportFeeToSibling`


parachain/runtime/runtime-common/src/lib.rs
```diff
@@ -98,32 +112,46 @@ impl<Balance, AccountId, FeeAssetLocation, EthereumNetwork, AssetTransactor, Fee
...
+		let result = AssetTransactor::deposit_asset(
+			&MultiAsset { id: Concrete(token_location), fun: Fungible(remote_fee.into()) },
+			&MultiLocation { parents: 1, interior: X1(Parachain(para_id)) },
+			context,
 		);
-		fees
+		if result.is_err() {
+			log::error!(
+				target: LOG_TARGET,
+				"transact fee asset failed with err {:?}",
```
alistair-singh, line 144:
```suggestion
				"transact fee asset failed: {:?}",
```

parachain/runtime/runtime-common/src/lib.rs
```diff
@@ -98,32 +112,46 @@ impl<Balance, AccountId, FeeAssetLocation, EthereumNetwork, AssetTransactor, Fee
...
-		log::info!(
-			target: "xcm::fees",
-			"XcmExportFeeToSibling skipped: {fees:?}, context: {context:?}, reason: {reason:?}",
+		let (fee_index, total_fee) = maybe_total_supplied_fee.unwrap();
+		let local_fee = FeeProvider::local_fee();
+		let remote_fee = total_fee.saturating_sub(local_fee);
+		if local_fee == Balance::zero() || remote_fee == Balance::zero() {
+			log::error!(
+				target: LOG_TARGET,
+				"fees to refund not correct with local_fee {:?} and remote_fee {:?}",
```
alistair-singh, line 129:
```suggestion
				"calculated refund incorrect with local_fee: {:?} and remote_fee: {:?}",
```

alistair-singh (approved):
Looks good. Added minor comments.

parachain/primitives/core/src/pricing.rs
```diff
@@ -23,18 +24,20 @@ pub struct Rewards<Balance> {
...
 
 impl<Balance> PricingParameters<Balance>
 where
 	Balance: BaseArithmetic + Unsigned + Copy,
 {
 	pub fn validate(&self) -> Result<(), InvalidPricingParameters> {
-		if self.exchange_rate == FixedU128::zero() {
+		if self.exchange_rate < FixedU128::from_rational(1, 10000) ||
+			self.exchange_rate > FixedU128::from_rational(1, 1)
+		{
```
vgeddes, line 37: I prefer not to change the validation logic. I also thought about this approach, but ultimately decided it wasn't worth it. 

1. If the exchange rate does actually mistakenly get set to `1/9999` (slightly less than the limit), then that practically still ends up blocking the bridge anyway.
2. The upper limit is arbitrary, and if the real-world exchange rate breaches the limit (quite possible), then a runtime upgrade would be required to fix the validation code.

parachain/primitives/core/src/pricing.rs
```diff
@@ -23,18 +24,20 @@ pub struct Rewards<Balance> {
...
 			return Err(InvalidPricingParameters)
 		}
-		if self.fee_per_gas == U256::zero() {
+		if self.fee_per_gas < gwei(1) || self.fee_per_gas > gwei(1000) {
```
vgeddes, line 40: ditto, see comment above. Not in favour of changing the validation logic.

parachain/runtime/runtime-common/src/tests.rs
```diff
@@ -0,0 +1,246 @@
...
+
+impl SendMessageFeeProvider for MockErrOutboundQueue {
+	type Balance = u128;
+
+	fn local_fee() -> Self::Balance {
+		1
+	}
+}
+
+pub struct SuccessfulTransactor;
```
vgeddes, line 60: Prefer this name:

```suggestion
pub struct MockAssetTransactor;
```

parachain/runtime/runtime-common/src/tests.rs
```diff
@@ -0,0 +1,246 @@
...
+		TokenLocation,
+		EthereumNetwork,
+		SuccessfulTransactor,
+		MockOkOutboundQueue,
+	>::handle_fee(fee.clone(), Some(&ctx), reason);
+	assert_eq!(result, fee)
+}
+
+#[test]
+fn handle_fee_fail_for_fees_not_met() {
```
vgeddes, line 207: Needs a comment explaining what condition is being tested.

parachain/runtime/runtime-common/src/tests.rs
```diff
@@ -0,0 +1,246 @@
...
+		TokenLocation,
+		EthereumNetwork,
+		SuccessfulTransactor,
+		MockOkOutboundQueue,
+	>::handle_fee(fee.clone(), Some(&ctx), reason);
+	assert_eq!(result, fee)
+}
+
+#[test]
+fn handle_fee_fail_for_transact() {
```
vgeddes, line 228: Needs a comment explaining what condition is actually being tested.

parachain/runtime/runtime-common/src/tests.rs
```diff
@@ -0,0 +1,246 @@
...
+		_what: &MultiAsset,
+		_from: &MultiLocation,
+		_to: &MultiLocation,
+		_context: &XcmContext,
+	) -> Result<Assets, XcmError> {
+		Ok(Assets::default())
+	}
+}
+
+pub struct NotFoundTransactor;
```
vgeddes, line 104: Instead of adding `NotFoundTransactor`, just parameterize the return value:

```rust
pub struct MockAssetTransactor<Get<R1>, Get<R2>>;

impl TransactAsset for MockAssetTransactor {
	fn can_check_in(
		_origin: &MultiLocation,
		_what: &MultiAsset,
		_context: &XcmContext,
	) -> XcmResult {
		R1::get()
	}

	fn withdraw_asset(
		_what: &MultiAsset,
		_who: &MultiLocation,
		_context: Option<&XcmContext>,
	) -> Result<Assets, XcmError> {
		R2::get()
	}
}
```

parachain/runtime/test-common/src/lib.rs
```diff
@@ -288,7 +296,60 @@ pub fn send_transfer_token_message_failure<Runtime, XcmConfig>(
 				destination_address,
 				fee_amount,
 			);
-			// check err is NotHoldingFees
+			assert_err!(outcome.ensure_complete(), expected_error);
+		});
+}
+
+#[allow(clippy::too_many_arguments)]
+pub fn send_transfer_token_message_with_invalid_fee_params<Runtime, XcmConfig>(
```
vgeddes, line 304: I don't think we need this test - the assumption is that the stored pricing parameters are valid after being set.

parachain/runtime/runtime-common/src/tests.rs
```diff
@@ -0,0 +1,246 @@
...
+		EthereumNetwork,
+		SuccessfulTransactor,
+		MockOkOutboundQueue,
+	>::handle_fee(fee.clone(), Some(&ctx), reason);
+	// assert fee not touched and just forward to the next handler
+	assert_eq!(result, fee)
+}
+
+#[test]
+fn handle_fee_fail_for_invalid_location() {
```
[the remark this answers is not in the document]
yrong: https://github.com/Snowfork/snowbridge/pull/1109/commits/18daf99473b759c8c278815dcf33bd4e75616a7a

parachain/runtime/runtime-common/src/tests.rs
yrong: https://github.com/Snowfork/snowbridge/pull/1109/commits/18daf99473b759c8c278815dcf33bd4e75616a7a

parachain/primitives/router/src/outbound/mod.rs
```diff
@@ -113,7 +113,7 @@ where
 		// validate the message
 		let (ticket, fee) = OutboundQueue::validate(&outbound_message).map_err(|err| {
 			log::error!(target: "xcm::ethereum_blob_exporter", "OutboundQueue validation of message failed. {err:?}");
-			SendError::Unroutable
+			SendError::Fees
```
claravanstaden, line 116: Not sure this is right. The message validation failure could be caused by a payload that is too large, an invalid channel, etc. So maybe `InvalidMessage` or something is better.

parachain/runtime/runtime-common/src/lib.rs
```diff
@@ -98,32 +112,46 @@ impl<Balance, AccountId, FeeAssetLocation, EthereumNetwork, AssetTransactor, Fee
...
+			log::error!(
+				target: LOG_TARGET,
+				"calculated refund incorrect with local_fee: {:?} and remote_fee: {:?}",
+				local_fee,
+				remote_fee,
+			);
+			return fees
+		}
+		// Refund remote component of fee to physical origin
+		let result = AssetTransactor::deposit_asset(
```
claravanstaden, line 136: Why the change from `deposit_or_burn_fee` to `deposit_asset`? Or is it the same thing? cc @alistair-singh

parachain/runtime/runtime-common/src/lib.rs
```diff
@@ -98,32 +112,46 @@ impl<Balance, AccountId, FeeAssetLocation, EthereumNetwork, AssetTransactor, Fee
...
+				result.unwrap_err()
+			);
+			return fees
+		}
+
+		// Return remaining fee to the next fee handler in the chain.
+		let mut modified_fees = fees.inner().clone();
+		modified_fees.remove(fee_index);
+		modified_fees
+			.push(MultiAsset { id: Concrete(token_location), fun: Fungible(local_fee.into()) });
```
claravanstaden, line 154: Why is the local fee passed on to the next fee handler?

parachain/runtime/runtime-common/src/tests.rs
```diff
@@ -0,0 +1,181 @@
...
+	let result = XcmExportFeeToSibling::<
+		u128,
+		u64,
+		TokenLocation,
+		EthereumNetwork,
+		SuccessfulTransactor,
+		MockOkOutboundQueue,
+	>::handle_fee(fee.clone(), Some(&ctx), reason);
+	assert_eq!(result, fee)
+}
```
claravanstaden, line 181: Nice tests!

parachain/runtime/runtime-common/src/lib.rs
alistair-singh: Only the remote fee (fee for execution on the destination) needs to be siphoned off to the parachain sovereign. The local fee (the fee charged for this xcm being processed on bridgehub) needs to go to wherever the default fees for xcm go, i.e. the next fee handler. Probably waivered because the origin is AssetHub which is a system parachain.

parachain/pallets/outbound-queue/src/test.rs
```diff
@@ -266,3 +266,47 @@ fn encode_digest_item() {
...
+		assert_eq!(fee.local, 698000000);
+		assert_eq!(fee.remote, 2680000000000);
+	});
+}
+
+#[test]
+fn test_calculate_fees() {
+	new_tester().execute_with(|| {
+		let gas_used: u64 = 250000;
+		let illegal_price_params: PricingParameters<<Test as Config>::Balance> =
```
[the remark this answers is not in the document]
yrong: Exactly! The `remote_fee` calculated in this case is zero so the sovereign account of the sibling chain is not properly refunded.

parachain/pallets/outbound-queue/src/send_message_impl.rs
```diff
@@ -57,7 +58,14 @@ where
...
 
 		let gas_used_at_most = T::GasMeter::maximum_gas_used_at_most(&message.command);
+		ensure!(
+			gas_used_at_most > T::GasMeter::MAXIMUM_BASE_GAS &&
+				gas_used_at_most < T::GasMeter::MAXIMUM_GAS_CAP,
+			SendError::MessageInvalidGasFees
+		);
 		let fee = Self::calculate_fee(gas_used_at_most, T::PricingParameters::get());
+		let zero_fee = <Self as SendMessageFeeProvider>::Balance::zero();
+		ensure!(fee.local > zero_fee && fee.remote > zero_fee, SendError::MessageInvalidGasFees);
```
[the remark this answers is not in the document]
yrong: This check is for the test case as I mentioned here in https://github.com/Snowfork/snowbridge/pull/1109#discussion_r1453721841 in which case the `remote_fee` calculated is zero which leads to the sovereign account of the sibling chain is not properly refunded. Finally we can't fund the agent on the Ethereum side.

`MAXIMUM_GAS_CAP` added here is a hard limit that applies to all commands. It's mainly for command [Upgrade](https://github.com/Snowfork/snowbridge/blob/d77201bc370ab2a4da15de701871b4a643bd5b5c/parachain/primitives/core/src/outbound.rs#L397) with a sanity dynamic gas. Otherwise the gas check in [Gateway](https://github.com/Snowfork/snowbridge/blob/d77201bc370ab2a4da15de701871b4a643bd5b5c/contracts/src/Gateway.sol#L152-L154) will fail and the bridge stuck. It also applies to the future command arbitrary transact the gas is dynamic from input of the end-user.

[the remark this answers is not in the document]
yrong: `MAXIMUM_GAS_CAP` added here is a hard limit that applies to all commands. It's mainly for command [Upgrade](https://github.com/Snowfork/snowbridge/blob/d77201bc370ab2a4da15de701871b4a643bd5b5c/parachain/primitives/core/src/outbound.rs#L397) with a sanity dynamic gas. It also applies to the future command arbitrary transact for which the gas is dynamic from the input of the end-user. Otherwise, the gas check in [Gateway](https://github.com/Snowfork/snowbridge/blob/d77201bc370ab2a4da15de701871b4a643bd5b5c/contracts/src/Gateway.sol#L152-L154) will always fail and the bridge stuck.

parachain/primitives/router/src/outbound/mod.rs
yrong: Nice catch! My first glance `Unroutable` does not make sense so changed to another one from https://github.com/Snowfork/polkadot-sdk/blob/d0e0b386187b5958b0d6e7ae9aadec9e0b88388e/polkadot/xcm/src/v3/traits.rs#L417-L439,  but as you mentioned `Fees` does not cover all the cases you mentioned.

I can't find another error more sensible in the current list, maybe we can add another PR  to `polkadot-sdk` for better error categories but not the scope of this PR so just reverted.

parachain/pallets/system/src/tests.rs
```diff
@@ -203,7 +203,7 @@ fn set_pricing_parameters_invalid() {
 		params = Parameters::get();
 		params.rewards.remote = sp_core::U256::zero();
 		assert_noop!(
-			EthereumSystem::set_pricing_parameters(origin, params),
+			EthereumSystem::set_pricing_parameters(origin.clone(), params),
```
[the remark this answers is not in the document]
yrong: https://github.com/Snowfork/snowbridge/pull/1109/commits/d567f5f7ef9cf7b0b25a50cf42ab61c1510d9420

parachain/pallets/outbound-queue/src/test.rs
```diff
@@ -266,3 +266,47 @@ fn encode_digest_item() {
...
+				fee_per_gas: 1_u32.into(),
+				rewards: Rewards { local: 1_u32.into(), remote: 1_u32.into() },
+			};
+		let fee = OutboundQueue::calculate_fee(gas_used, illegal_price_params.clone());
+		assert_eq!(fee.local, 698000000);
+		// Though none zero pricing params the remote fee calculated here is invalid
+		// which should be avoided
+		assert_eq!(fee.remote, 0);
+	});
+}
```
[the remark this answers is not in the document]
vgeddes: It seems odd that the calculated remote fee is zero. Can you investigate why this happens? Seems like it could be a bug in our code.

parachain/pallets/outbound-queue/src/send_message_impl.rs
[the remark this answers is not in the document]
vgeddes: I don't think we should introduce `MAXIMUM_GAS_CAP`, for the same reasons I outlined in https://github.com/Snowfork/snowbridge/pull/1109#discussion_r1453336525.

parachain/pallets/outbound-queue/src/test.rs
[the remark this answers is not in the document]
vgeddes: Actually I realize why the calculated remote fee is zero. Since ether has 18 decimal places and, dot has 10 decimal places, `convert_from_ether_decimals` will end up truncating to zero very small fixed-point numbers that don't fit in 10 decimals.

This is fine though, and it is expected behavior.

parachain/pallets/outbound-queue/src/send_message_impl.rs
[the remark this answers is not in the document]
vgeddes: See https://github.com/Snowfork/snowbridge/pull/1109#discussion_r1455250041. I think its fine if calculated remote fee is zero. It just means the fee is too low to be represented using DOT decimals.

parachain/runtime/runtime-common/src/tests.rs
yrong: https://github.com/Snowfork/snowbridge/pull/1109/commits/122572d74f3e49f6195754f0f81191b8ffafecd0

parachain/pallets/outbound-queue/src/send_message_impl.rs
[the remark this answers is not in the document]
yrong: Checks reverted in https://github.com/Snowfork/snowbridge/pull/1109/commits/af15053f0d01cfa7f06bd6f3119987873793297a

vgeddes (approved):
+1

PR #1112, opened 2024-01-12
Refactoring create asset and add CI tests to cover

yrong:
Resolves: SNO-817
Requires: https://github.com/Snowfork/polkadot-sdk/pull/99


parachain/primitives/router/src/inbound/mod.rs
```diff
@@ -318,3 +295,19 @@ impl<AccountId> GlobalConsensusEthereumConvertsFor<AccountId> {
 		(b"ethereum-chain", chain_id).using_encoded(blake2_256)
 	}
 }
+
+#[allow(clippy::large_enum_variant)]
+#[derive(Encode, Decode, Debug, PartialEq, Eq, Clone, TypeInfo)]
+pub enum Call {
+	/// Call create_asset on foreign assets pallet.
+	#[codec(index = 53)]
```
claravanstaden, line 303: I think the pallet index should stay configured in the runtime config. What happens if the ForeignAssets pallet indexes are different per runtime? I think they are the same and that is the convention, but I still think this is runtime related and shouldn't live in the snowbridge crates.

claravanstaden (commented):
I had a quick look, I think the polkadot-sdk companion PR should still be opened right? 😄

parachain/primitives/router/src/inbound/mod.rs
yrong: > What happens if the ForeignAssets pallet indexes are different per runtime? I think they are the same and that is the convention.

Yeah, that's the precondition, checked that all the same in multiple runtimes(e.g. polkadot/kusama). I do not see a reason why they need to break this rule in the future so prefer to make things simple at this moment.

 But I'm open if you all think that's a problem and will make some changes accordingly.

alistair-singh: Maybe we can add a test to make sure they are in line?

alistair-singh: Ah I see one is added. Ignore this.

alistair-singh (approved):

parachain/primitives/router/src/inbound/mod.rs
vgeddes: I agree with Clara here, I am a bit uncomfortable with baking runtime configuration inside our pallets & primitives crate.

Secondly, I am sensitive about changing code unnecessarily at this point. Our next audit is scheduled for Feb 8, and this change will increase the scope of the audit.

I want to reserve any scope increases to accommodate for the second round of code review that the Parity bridges team is going to perform soon.

yrong: The initial intention of this PR is for the [comment](https://github.com/paritytech/polkadot-sdk/pull/2522#discussion_r1427682795) to add the call index to some primitives which can be referenced/checked in the AH runtime tests. 

So @vgeddes do we still need this PR?  I think we already have the emulated test register token to cover this use case so maybe the refactoring here and the unit test added in AH may not be necessary at all?

claravanstaden: I think Slava wanted us to move those constants here: `polkadot-sdk/bridges/primitives/chain-asset-hub-rococo`. I think it is fine to move the pallet index, but not to our crates. It should still be in the polkadot-sdk, and it should be configured per testnet. I think it can actually go here too: `polkadot-sdk/cumulus/parachains/common/src/rococo.rs`

yrong: Thanks Clara and that makes sense!  Then I think it's also better to move the [hard-coded weight](https://github.com/Snowfork/snowbridge/blob/d77201bc370ab2a4da15de701871b4a643bd5b5c/parachain/primitives/router/src/inbound/mod.rs#L191) to some common runtime config.

But if as Vincent mentioned the primary concern is to `reserve any scope increases to accommodate for the second round of code review that the Parity` then perhaps it's not the right time to do the refactoring?

So I'll stop here and let the team decide whether to continue the change or not.

yrong: @claravanstaden Some refactoring in https://github.com/Snowfork/snowbridge/pull/1112/commits/37aa97eedcd0c3a13be72d78603a338ed1f332d0 as you suggested.

parachain/primitives/router/src/inbound/mod.rs
```diff
@@ -83,24 +83,27 @@ pub enum Destination {
...
+pub struct CreateAssetCallInfo {
+	/// Index of the create call
+	pub call_index: CallIndex,
+	/// Deposit required to create the asset
+	pub asset_deposit: u128,
+	/// Minimal balance of the asset
+	pub min_balance: u128,
+	/// Weight required for the xcm transact
+	pub transact_weight_at_most: Weight,
+}
```
yrong, line 97: So we add a `CallInfo` structure for the create asset and make it runtime config(without any hard coded params anymore), meanwhile we add runtime tests in asset-hub to check the `call_index` and `transact_weight_at_most` all as expected.

vgeddes (commented):
Hey Ron this is a very nice improvement. However we have an audit in 2-3 weeks and I want us to have some sort of code freeze for relatively non-important changes.

So lets wait till after the next audit to merge this in.

alistair-singh (approved):

PR #1114, opened 2024-01-15
Adds runtime tests

claravanstaden:
Adds runtime test for relayer extrinsics:

- [x] beacon relayer
- [x] execution relayer
- [x] check nonce and digest after sending transfer token message **(digest is empty 🤷🏻‍♀️ )**

Resolves: [SNO-811](https://linear.app/snowfork/issue/SNO-811)
Polkadot-sdk companion: https://github.com/Snowfork/polkadot-sdk/pull/98


parachain/runtime/test-common/src/lib.rs
```diff
@@ -154,6 +166,29 @@ pub fn send_transfer_token_message_success<Runtime, XcmConfig>(
...
+			assert_ok!(nonce);
+			assert_eq!(nonce.unwrap(), 1);
+
+			let digest = included_head.digest();
+
+			//let digest = frame_system::Pallet::<Runtime>::digest();
+			let digest_items = digest.logs();
+			assert!(digest_items.len() == 1 && digest_items[0].as_other().is_some());
+
+			//assert_eq!(Messages::<Test>::decode_len(), Some(4));
```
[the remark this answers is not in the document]
yrong: Seems it's because the [run_to_block](https://github.com/Snowfork/polkadot-sdk/blob/2afac316933a3b031c979af9cb8d1e36f03eccb8/cumulus/parachains/runtimes/test-utils/src/lib.rs#L287-L296) we run in this test case does not include the finalized phase.

I'll check if possible to make some fixes here.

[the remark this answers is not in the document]
yrong: https://github.com/Snowfork/snowbridge/pull/1126

@claravanstaden I've made this PR to resolve the issue and the test passed. The bad thing it requires more changes in `polkadot-sdk` repo to include the finalized phase. Though I think it is general enough but not sure worth introducing at this moment. 

Another option is to limit the scope of the change and maybe add another `run_to_block` function in our test case to only finalize the pallets as required. Something like we do in the unit test https://github.com/Snowfork/snowbridge/blob/3b17d2a89d23ea1de5889418acae7683ecd49604/parachain/pallets/outbound-queue/src/mock.rs#L122-L136

I'll leave it up to you.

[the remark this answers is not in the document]
claravanstaden: Hey @yrong! Thanks for the effort you put in here, I appreciate it. Let me try the local "run_to_end_of_next_block" in our test case first, as I guess that will be the least disruptive.

parachain/pallets/ethereum-client/Cargo.toml
```diff
@@ -73,6 +75,7 @@ std = [
 	"serde",
 	"snowbridge-core/std",
 	"snowbridge-ethereum/std",
+	"snowbridge-pallet-ethereum-client-fixtures/std",
```
yrong, line 78: Seems unnecessary.

parachain/pallets/ethereum-client/fixtures/Cargo.toml
```diff
@@ -0,0 +1,44 @@
...
+	"sp-core/std",
+	"sp-std/std",
+]
+runtime-benchmarks = [
+	"frame-benchmarking",
+	"frame-benchmarking/runtime-benchmarks",
+	"frame-support/runtime-benchmarks",
+	"frame-system/runtime-benchmarks",
+	"snowbridge-core/runtime-benchmarks",
+]
```
yrong, line 44: `runtime-benchmarks` seems not necessary here.

claravanstaden: I referenced the inbound fixtures crate you added. 😅 https://github.com/Snowfork/snowbridge/blob/main/parachain/pallets/inbound-queue/fixtures/Cargo.toml#L38 Is it necessary in that crate?

yrong: Yeah, seems also unnecessary in that crate because IIUC we're using the fixture from local setup for the benchmark?

The benefit of introducing the `runtime-benchmarks` feature is to differentiate the fixture between public network(goerli/sepolia) and local setup(hacked mainnet), As we know the fork version is different and we may want to use the real data(from public network) for the benchmark.

So it's nitpicky and feel free to ignore it.

yrong (approved):
+1

PR #1115, opened 2024-01-16
Disable system calls from xcm & Refactoring emulated tests

yrong:
- Disable the agent/channel operation from xcm(not in the scope of the next launch)
- Refactoring emulated tests with inbound fixtures
- More tests and checks

Requires: https://github.com/Snowfork/polkadot-sdk/pull/108


alistair-singh (approved):
+1

parachain/pallets/inbound-queue/src/lib.rs
```diff
@@ -360,6 +358,17 @@ pub mod pallet {
 			})?;
 			Ok(())
 		}
+
+		pub fn refund_relayer(
```
vgeddes, line 362: It seems we have now have this pattern in the emulated tests where we call these three functions in order:
* refund_relayer
* do_convert
* send_xcm

However this is problematic as internal implementation details are now leaking into the tests. It makes the tests more brittle, and less useful. The tests should treat `submit` as a black box.

I would rather refactor the `submit` extrinsic, so that it calls the following function, where verification is optional.

```rust
fn process_message(message: &Message, verify: bool) -> DispatchResult
```

And then the simulated tests can call `process_message(msg, false)`

vgeddes (changes requested):

parachain/pallets/inbound-queue/src/lib.rs
yrong: Notice @claravanstaden is making [InboundQueueFixture](https://github.com/Snowfork/snowbridge/blob/c754451c9d5672ab7e9250ea6e8e5274c121251b/parachain/pallets/inbound-queue/src/benchmarking/fixtures.rs#L8-L11) auto-generated in https://github.com/Snowfork/snowbridge/pull/1119, then I would prefer to reuse it for the emulated test here. So there is no need to add the `process_message(msg, false)`  function and just test the original `submit` function without any mock.

The problem is that `InboundQueueTest`  is for benchmark only and we want to use it also in unit/emulated tests. We did make https://github.com/Snowfork/snowbridge/pull/1013 for that purpose though closed it for some other high-priority PRs.

So maybe it's time to reopen it. Meanwhile In https://github.com/Snowfork/snowbridge/pull/1119 we also take it into consideration when generating `InboundQueueFixture`.

@vgeddes @claravanstaden WDYT?

claravanstaden: I made a comment here: https://github.com/Snowfork/snowbridge/pull/1119 I would also like to use https://github.com/Snowfork/snowbridge/tree/main/parachain/pallets/ethereum-client/tests/fixtures in the runtime tests. Maybe worthwhile to split test fixtures out to a new crate.

vgeddes: Cool, ok, this all makes sense - lets split the fixture out into a separate crate - something like `pallets/inbound-queue/fixtures`.

yrong: Addressed in https://github.com/Snowfork/polkadot-sdk/pull/106

vgeddes: Ok now that we have the fixtures, its no longer necessary to have this `refactor_relayer` function. @yrong can you revert it?

Also helps reduce the amount of code which needs to be re-audited.

yrong: Revert in https://github.com/Snowfork/snowbridge/pull/1115/commits/4f9f56e74092c566b95052caab33258d15c1d5c3

claravanstaden (approved):

vgeddes (approved):

PR #1119, opened 2024-01-17
Remove minimal spec

claravanstaden:
- [x] remove beacon-minimal-spec feature
- [x] update lodestar to use mainnet spec when running the local net
- [x] test smoke tests work
- [x] investigate error: `thread 'tests::submit_update_execution_headers_too_far_behind' has overflowed its stack`
- [x] update test fixture/benchmark data
- [x] adds script to generate test data for inbound message queue tests (will aid https://github.com/Snowfork/snowbridge/pull/1114)
- [x] generate next sync committee update
- [x] check benchmarks work
- [x] check beefy relayer starting

Polkadot-sdk companion: https://github.com/Snowfork/polkadot-sdk/pull/100


yrong:
> investigate error: thread 'tests::submit_update_execution_headers_too_far_behind' has overflowed its stack

Maybe you can try this env setup first
https://github.com/Snowfork/snowbridge/blob/2710a25504ba855ec378c6d242d8b6347794b085/.github/workflows/parachain.yml#L100

web/packages/test/scripts/deploy-ethereum.sh
```diff
@@ -87,8 +91,12 @@ deploy_local() {
     echo "Waiting for geth API to be ready"
     sleep 3
 
-    if [ "$eth_fast_mode" == "true" ]; then
-        hack_beacon_client
+    # if were are running locally speed up the seconds per slot from 12 seconds to 2 seconds. if we are not
+    # running locally, revert
+    if [ "$eth_network" == "localhost" ]; then
+        set_slot_time 12 2
```
yrong, line 97: Maybe we can further decrease the slot time to 1 second, then a sync period will be 2.27 hours.  
Anyway, it seems still a bit challenging to test the switchover process.

claravanstaden:
> > investigate error: thread 'tests::submit_update_execution_headers_too_far_behind' has overflowed its stack
> 
> Maybe you can try this env setup first
> 
> https://github.com/Snowfork/snowbridge/blob/2710a25504ba855ec378c6d242d8b6347794b085/.github/workflows/parachain.yml#L100

Thanks! This did the trick. It will be a problem when running the test in polkadot-sdk, perhaps. It is maybe worthwhile to change the test to fit in the stack size, maybe use heap-allocation instead.

web/packages/test/scripts/deploy-ethereum.sh
vgeddes: I would be careful being too aggressive, as it may mean the beacon-light-client could struggle to import headers at such a fast rate. May need relayer updates too.

parachain/pallets/inbound-queue/src/benchmarking/fixtures.rs
```diff
@@ -1,20 +1,22 @@
...
 use snowbridge_beacon_primitives::CompactExecutionHeader;
 use snowbridge_core::inbound::{Log, Message, Proof};
 use sp_std::vec;
 
 pub struct InboundQueueTest {
-	pub execution_header: CompactExecutionHeader,
-	pub message: Message,
+    pub execution_header: CompactExecutionHeader,
+    pub message: Message,
 }
```
yrong, line 11: Cool!  IIUC the `InboundQueueFixture` is also auto-generated based on some template file. I would like it's not limited to benchmark tests and also apply to unit/integration tests.

More context in https://github.com/Snowfork/snowbridge/pull/1115#discussion_r1460904896

web/packages/test/scripts/build-binary.sh
```diff
@@ -58,27 +58,26 @@ build_relayer() {
...
 
 build_lodestar() {
-    if [ ! -d "$root_dir/lodestar/packages/cli/lib" ]; then
-        pushd $root_dir/lodestar
-        if [ "$eth_fast_mode" == "true" ]; then
-            hack_beacon_client
-        fi
-        yarn install && yarn run build
-        popd
+    pushd $root_dir/lodestar
```
claravanstaden, line 73: I think it's fine to build lodestar on every run because it won't rebuild if there are no changes.

web/packages/test/scripts/deploy-ethereum.sh
```diff
@@ -52,6 +52,8 @@ start_lodestar() {
             timestamp=$(date -d'+10second' +%s)
         fi
 
+        export LODESTAR_PRESET="mainnet"
```
claravanstaden, line 55: Use the "mainnet" spec for local tests.

flake.nix
```diff
@@ -37,9 +37,10 @@
                     ps
 
                     # typescript
+                    python3
```
claravanstaden, line 40: Python is necessary to build lodestar.

flake.nix
```diff
@@ -37,9 +37,10 @@
...
                     nodePackages.pnpm
-                    nodePackages.yarn
-                    nodejs-18_x
+                    nodejs_20
+                    (yarn.override { nodejs = nodejs_20; })
```
claravanstaden, line 43: Yarn comes with node 18 so we need to override it here. Lodestar 1.12.0 doesn't build with 18.16.1

web/packages/test/scripts/deploy-ethereum.sh
claravanstaden: It looks like 1 second per slot is fine for our local setup. The bridge hub parachain sometimes drops imports but then recovers. I think that's okay, for the sake of faster tests? Otherwise we can change the `SECONDS_PER_SLOT` back to 2 seconds.

flake.nix
yrong: Btw the lodestar in our repo is the forked branch https://github.com/Snowfork/lodestar/tree/snowbridge which is based on 1.14.0 with a hack to work with node 18.16.

Anyway, the hack here is necessary for working with node 20 so happy to see it.

yrong (approved):
Awesome!

web/packages/test/scripts/build-binary.sh
yrong: Unfortunately, it did rebuild even no changes and cost almost 13s on my laptop.

yrong: Yeah, that's a problem. Maybe we can add a bool flag/env in the script to control whether to force rebuild or not. But it's very nit-picky so leave it up to you.

PR #1122, opened 2024-01-22
Schedule dencun on sepolia

yrong:
We may need a runtime upgrade on Rococo testnet before Jan.31(unix time:1706655072).


yrong:
Seems already coverer in https://github.com/Snowfork/snowbridge/pull/1119 so closed.

claravanstaden (approved):

vgeddes (approved):

PR #1125, opened 2024-01-23
Refactoring Inbound fixture

yrong:
Requires: https://github.com/Snowfork/polkadot-sdk/pull/106

Refactoring inbound fixtures to a separate crate and reuse it in emulated tests.


parachain/pallets/inbound-queue/src/fixtures.rs
```diff
@@ -3,8 +3,10 @@
 use hex_literal::hex;
 use snowbridge_beacon_primitives::CompactExecutionHeader;
 use snowbridge_core::inbound::{Log, Message, Proof};
+use sp_core::RuntimeDebug;
 use sp_std::vec;
 
+#[derive(Clone, RuntimeDebug)]
 pub struct InboundQueueTest {
```
vgeddes, line 10: Lets rename this to the describe the message content, ie RegisterAsset, SendToken, etc.

yrong: https://github.com/Snowfork/snowbridge/pull/1125/commits/278752a161d5aeff6aee531b5f4ecc39631ea636

relayer/cmd/generate_beacon_data.go
```diff
@@ -237,7 +248,7 @@ func generateBeaconTestFixture(cmd *cobra.Command, _ []string) error {
 			return fmt.Errorf("get beacon block containing header: %w", err)
 		}
 
-		beaconBlockSlot, err := strconv.ParseUint(beaconBlock.Data.Message.Body.ExecutionPayload.BlockNumber, 10, 64)
+		beaconBlockSlot, err := strconv.ParseUint(beaconBlock.Data.Message.Slot, 10, 64)
```
yrong, line 251: A bug here.

relayer/cmd/generate_beacon_data.go
```diff
@@ -529,7 +543,7 @@ func getEthereumEvent(ctx context.Context, gatewayContract *contracts.Gateway, c
 				}
 				break
 			}
-			if iter.Event.Nonce >= 1 {
+			if iter.Event.Nonce >= uint64(nonce) {
```
yrong, line 546: `nonce` as param to generate inbound fixtures for different commands.

relayer/cmd/generate_beacon_data.go
claravanstaden: Good catch! 😄

parachain/pallets/inbound-queue/fixtures/src/register_asset.rs
```diff
@@ -1,21 +1,20 @@
...
 use hex_literal::hex;
 use snowbridge_beacon_primitives::CompactExecutionHeader;
 use snowbridge_core::inbound::{Log, Message, Proof};
 use sp_std::vec;
 
-pub struct InboundQueueTest {
-	pub execution_header: CompactExecutionHeader,
-	pub message: Message,
-}
+pub type RegisterAsset = InboundQueueFixture;
```
claravanstaden, line 10: Nit: While I think "register asset" is probably more correct, everywhere else we refer to it as "register token" and so I think we should be consistent here and call it "RegisterToken".

relayer/cmd/generate_beacon_data.go
```diff
@@ -38,7 +38,9 @@ func generateBeaconDataCmd() *cobra.Command {
 	}
 
 	cmd.Flags().String("url", "http://127.0.0.1:9596", "Beacon URL")
-	cmd.Flags().Bool("wait_until_next_period", true, "Waiting until next period")
+	cmd.Flags().Bool("wait_until_next_period", false, "Waiting until next period")
```
claravanstaden, line 41: I think we should leave the default true here. The reason being is if you update all the other fixtures, but leave our regenerating the next sync committee files, the tests will likely fail because the tests will be for different runs.

The beacon test fixtures and the inbound message fixtures should be in the same run, since I need to the beacon imports + message test in the same test.

relayer/cmd/generate_beacon_data.go
```diff
@@ -38,7 +38,9 @@ func generateBeaconDataCmd() *cobra.Command {
...
+	cmd.Flags().Uint32("nonce", 1, "Nonce of the inbound message")
```
claravanstaden, line 42: Nice!

relayer/cmd/generate_beacon_data.go
```diff
@@ -345,26 +354,31 @@ func generateBeaconTestFixture(cmd *cobra.Command, _ []string) error {
 
 		log.WithFields(log.Fields{
 			"location": pathToBeaconBenchmarkData,
-			"filename": filename,
 		}).Info("writing result file")
 
-		err = writeBenchmarkDataFile(fmt.Sprintf("%s/%s", pathToBeaconBenchmarkData, filename), rendered)
+		err = writeBenchmarkDataFile(fmt.Sprintf("%s", pathToBeaconBenchmarkData), rendered)
```
claravanstaden, line 359: Why is the file name removed here?

relayer/cmd/generate_beacon_data.go
```diff
@@ -345,26 +354,31 @@ func generateBeaconTestFixture(cmd *cobra.Command, _ []string) error {
...
 		if err != nil {
 			return err
 		}
 
 		// writing inbound queue fixtures
+		testCase, _ := cmd.Flags().GetString("test_case")
```
claravanstaden, line 365:
```suggestion
		testCase, err := cmd.Flags().GetString("test_case")
```

relayer/cmd/generate_beacon_data.go
```diff
@@ -345,26 +354,31 @@ func generateBeaconTestFixture(cmd *cobra.Command, _ []string) error {
...
+		if err != nil {
+			return err
+		}
+		pathToInboundBenchmarkDataTemplate := fmt.Sprintf("parachain/templates/%s.mustache", testCase)
+		pathToInboundQueueFixtureData := fmt.Sprintf("parachain/pallets/inbound-queue/fixtures/src/%s.rs", testCase)
```
claravanstaden, line 370: Can use a constant for these paths too.

smoketest/src/helper.rs
```diff
@@ -184,7 +184,7 @@ pub async fn send_sudo_xcm_transact(
 	penpal_client: &Box<OnlineClient<PenpalConfig>>,
 	message: Box<VersionedXcm>,
 ) -> Result<ExtrinsicEvents<PenpalConfig>, Box<dyn std::error::Error>> {
-	let dest = Box::new(VersionedMultiLocation::V3(MultiLocation {
+	let dest = Box::new(VersionedLocation::V3(MultiLocation {
```
claravanstaden, line 187: Are we not using XCM V4 now, on BridgeHub? 

Thanks for updating these, by the way! I forgot about updating them.

claravanstaden (commented):
Nice work, @yrong! 🥳

parachain/pallets/inbound-queue/fixtures/src/register_asset.rs
yrong: https://github.com/Snowfork/snowbridge/pull/1125/commits/eb5fb16f585dab6646955a54da9640a9556941fa

relayer/cmd/generate_beacon_data.go
yrong: Reverted in https://github.com/Snowfork/snowbridge/pull/1125/commits/98b37582c0a63a324c203556581238e7b6e1cbf6

Btw: WDYT if we totally removed the fixtures/tests for next period, it's a long wait time here.  Since test fixtures in current period are also able to cover the switch-over process so coverage will not be decreased a lot? 

Anyway, not the scope of this PR and we can improve it later.

relayer/cmd/generate_beacon_data.go
yrong: https://github.com/Snowfork/snowbridge/pull/1125/commits/7b1948a3f7e1a585695b7e792db66d292d389587

smoketest/src/helper.rs
yrong: Yes, it's v4 now. But seems it's compatible with v3 and I tested the smoke tests still works.

I would prefer to update the smoke tests in a separate PR later.

claravanstaden (approved):
🚢

parachain/pallets/inbound-queue/fixtures/src/lib.rs
```diff
@@ -0,0 +1,14 @@
+#![cfg_attr(not(feature = "std"), no_std)]
```
claravanstaden, line 1: We need our license header here. I think perhaps in the generated fixtures too.

yrong: https://github.com/Snowfork/snowbridge/pull/1125/commits/1da51e135ae0b278568f8b8bbc4232378ec1a4e4

parachain/pallets/inbound-queue/fixtures/src/send_token.rs
```diff
@@ -0,0 +1,43 @@
...
+// Generated, do not edit!
+// See ethereum client README.md for instructions to generate
+
+use crate::InboundQueueFixture;
+use hex_literal::hex;
+use snowbridge_beacon_primitives::CompactExecutionHeader;
+use snowbridge_core::inbound::{Log, Message, Proof};
+use sp_std::vec;
+
+pub type SendToken = InboundQueueFixture;
```
vgeddes, line 12: I think its a bit unnecessary to add this alias for each fixture.

vgeddes: We should rather add a doc comment (only if necessary - the function name may be enough) saying that the returned fixture contains a send token message.

vgeddes (approved):

PR #1132, opened 2024-01-30
TokenSent Event mislabels `sender` and `token` properties

alistair-singh:
`sender` needs to be the first property. `token` needs to be the seconds. 

See: https://github.com/Snowfork/snowbridge/blob/alistair/token-sent-correct-name/contracts/src/Assets.sol#L135

### DIFF
--- a/contracts/src/interfaces/IGateway.sol
+++ b/contracts/src/interfaces/IGateway.sol
@@ -68,8 +68,8 @@ interface IGateway {
 
     /// @dev Emitted once the funds are locked and an outbound message is successfully queued.
     event TokenSent(
-        address indexed sender,
         address indexed token,
+        address indexed sender,
         ParaID indexed destinationChain,
         MultiAddress destinationAddress,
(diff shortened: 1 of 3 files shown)


contracts/src/interfaces/IGateway.sol
```diff
@@ -69,8 +69,8 @@ interface IGateway {
 
     /// @dev Emitted once the funds are locked and an outbound message is successfully queued.
     event TokenSent(
-        address indexed token,
```
vgeddes, line 72: Nice catch! Though can we rather keep `token` as the first parameter (since the event is named `TokenSent`, so `token` is most important), and instead update the code in `sendToken` that emits the event?

alistair-singh: Addressed in c8e71a3df8a1d0ea584437e45b3955d0fb3507d3

vgeddes (approved):

PR #1136, opened 2024-01-31
ORML integration

yrong:
Resolves: SNO-839

- [x] Prepare template chain as submodule
- [x] Initialize with minimal ORML traits/pallets required
- [x] Bridge adaptor for token transfer 
- [x] Emulated tests 

Primary changes in https://github.com/Snowfork/parachain-orml-template/pull/1 as reference for the ORML integration.

Emulated [test](https://github.com/Snowfork/parachain-orml-template/blob/cc5df6a01b757b05fa163d98d5048e7d753316d5/integration-tests/emulated/tests/bridges/bridge-hub-rococo/src/tests/snowbridge.rs#L546) transferring WETH from ethereum to orml-template chain and vice versa.


yrong:
Currently, the workflow requires token transfer messages all relayed from Asset Hub(AH), so from the perspective of the ORML chain there is not much difference from the integration flow with any other sibling chain(e.g. mainly to register foreign token and it's XCM Location in `asset-registry` module which the 3rd team should already be familiar with).

Previously I planned to add the template chain into our e2e setup and write smoke tests for it, but consider above seems not worthwhile for that. The emulated test and the companion PR with changes required should be pretty enough to demonstrate. 

So I prefer not to merge this PR and just leave it here as a reference for the integration flow.

vgeddes:
Sounds good Ron, I agree that we should not merge this since the `Snowfork/parachain-orml-template` repo should be sufficient.

yrong:
Leave https://github.com/Snowfork/parachain-orml-template/pull/1 and close this one.

PR #1138, opened 2024-01-31
Mandatory commitments only

alistair-singh:
Adds the `IsMandatory` flag to `SafeScanCommitmentResult`. This is just appended to logging in this PR. 

But I used this flag to skip signed commitments that are not mandatory n order to test parity https://github.com/paritytech/polkadot-sdk/pull/3108 .
```go
	if !result.IsMandatory {
		logEntry.Info("Skipped due to not mandatory commitment.")
		continue
	}
```

We do not need code changes to support the fix. Leaving the IsMandatory field in here because I think its useful for logging. In the future we could use it to save costs by limiting submissions to the light client to only mandatory commitments.

Resolves: SNO-842

### DIFF
--- a/relayer/relays/beefy/scanner.go
+++ b/relayer/relays/beefy/scanner.go
@@ -78,12 +78,12 @@ func scanBlocks(ctx context.Context, api *gsrpc.SubstrateAPI, startBlock uint64,
 }
 
 type ScanCommitmentsResult struct {
-	SignedCommitment    types.SignedCommitment
-	BlockNumber         uint64
-	BlockHash           types.Hash
-	Depth               uint64
-	MandatoryCommitment bool
-	Error               error
+	SignedCommitment types.SignedCommitment
+	BlockNumber      uint64
+	BlockHash        types.Hash
+	Depth            uint64
+	IsMandatory      bool
+	Error            error
 }
 
 func ScanCommitments(ctx context.Context, api *gsrpc.SubstrateAPI, startBlock uint64) (<-chan ScanCommitmentsResult, error) {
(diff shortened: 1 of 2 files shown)


yrong (approved):
+1

relayer/relays/beefy/scanner.go
```diff
@@ -129,6 +131,21 @@ func scanCommitments(ctx context.Context, api *gsrpc.SubstrateAPI, startBlock ui
 				return
 			}
 
+			isMandatory := false
```
vgeddes, line 134: I would rather we go all the way and add a documented `MandatoryCommitmentsOnly` config setting. Otherwise this code is going to be forgotten about it.

yrong: We may not really need it as can be already achieved by setting
https://github.com/Snowfork/snowbridge/blob/aab91ae8c8d99072deb8d8d3230306763eb46996/web/packages/test/config/beefy-relay.json#L8 with a value big enough(i.e. more than blocks in an epoch).

alistair-singh: Addressed in [a791b2a](https://github.com/Snowfork/snowbridge/pull/1138/commits/a791b2a7aef4bc3107710370a3a268a4166ac64c).

We can use the `update-period` but this is more explicit and is good for testing the mmr fix as well. Currently turning this setting on will fail untill we update from upstream polkadot-sdk. I tested this change with cherry-picking .
```
$ git cherry-pick e5bb11b008ca6c1b5a07cab94a2327538f3bf15c
```

I think this setting also works well with incentivizing beefy relayers. For instance, our relayer can run in a mode where it only sync's mandatory commitments (i.e. 1 commitment per epoch). And if there are pending messages and an incentive is provided by a user to speed up the transaction then it can sync the next commitment it sees.

PR #1141, opened 2024-02-07
Transact from eth to sub direction

yrong:


contracts/src/Gateway.sol
```diff
@@ -617,4 +621,31 @@ contract Gateway is IGateway, IInitializable {
...
+            sender = msg.sender;
+            originKind = 0x01;
+        } else if (message.originKind == OriginKind.Xcm) {
+            Channel storage channel = _ensureChannel(destinationChain.into());
+            sender = channel.agent;
+            originKind = 0x03;
+            costs.foreign = message.fee;
+        } else {
+            revert InvalidTransact();
+        }
```
vgeddes, line 644: I think there is a misunderstanding here. In both cases (OriginKind.SovereignAccount or OriginKind.Xcm), the sender should be `msg.sender`. We don't want agents to be marked as senders.

Consider the following flow:

1. Message reaches bridgeHub, and we translate it to the following:
```
UniversalOrigin(GlobalConsensus(Ethereum)),
DescendOrigin(AccountKey20 { msg.sender })
Transact { OriginKind = "SovereignAccount" OR "Xcm" }
```

2. On the destination parachain, the following converter converts the XCM origin (a MultiLocation) based on the following rules:
https://github.com/Snowfork/polkadot-sdk/blob/c23c50f2cb4697885e1dd891da6fdd33036631a8/cumulus/parachains/runtimes/bridge-hubs/bridge-hub-rococo/src/xcm_config.rs#L114

3. If `OriginKind` is `Xcm`, then the `XcmPassthrough<RuntimeOrigin>` converter is used, which just passes the multilocation through unmodified.

4. If `OriginKind` is `SovereignAccount`, then the `SovereignSignedViaLocation<LocationToAccountId, RuntimeOrigin>` converter is used, I think. The dest must make sure that `LocationToAccountId` can handle AccountKey20 junctions.

So we can see from this flow that we don't need to involve agent addresses or anything like that.

yrong: That's true. With `dest must make sure that LocationToAccountId can handle AccountKey20 junctions` it's possible to use the original `msg.sender` in both cases. 

I prefer to keep the agent mode for reasons:

- For some extrinsics the dest chain may not want any `SovereignAccount` to invoke but only allowed from a specific Origin, something like the [CreateOrigin](https://github.com/Snowfork/polkadot-sdk/blob/348a1a010481002e41594ed75e5d78b7c2dbed92/cumulus/parachains/runtimes/assets/asset-hub-rococo/src/lib.rs#L415C2-L415C38)  filter when register foreign token on AH.
- For `SovereignAccount` we don't charge them on Ethereum side for the foreign/xcm fee part by default. Instead we'll require end user to manually prefund his derived account(a keyless Address32 mapping to the ethereum address) on dest chain before they make the call.  It's always a bit challege. While for Agent mode by setting `costs.foreign = message.fee` we charge fees on Etherum on their behalf and use that to refund the Agent sovereign account on dest chain(without requiring end user to do that by themselves).

yrong: Btw:
```
UniversalOrigin(GlobalConsensus(Ethereum)),
DescendOrigin(AccountKey20 { msg.sender })
Transact { OriginKind = "SovereignAccount" OR "Xcm" }
```
won't not work without fees, my estimated Xcm(not tested yet) will be something like

```
UniversalOrigin(GlobalConsensus(Ethereum)),
DescendOrigin(AccountKey20 { msg.sender }),
WithdrawAsset((Location::parent(),fee)),
BuyExecution { fees: (Location::parent(),fee), weight_limit: Unlimited }
Transact { OriginKind = "SovereignAccount" OR "Xcm" }
```

vgeddes: > For Xcm mode there are extrinsics with high security the dest chain won't like to expose to any EOA on Ethereum. Instead only allow to invoke from a specific xcm origin(e.g. only from it's own agent through our bridge), something like the CreateOrigin filter on AH but more strict.

I don't quite agree. With Xcm mode, the extrinsic on the destination parachain can have an origin check that verifies the origin location `../AccountKey20 { }`.

This can prevent unauthorized EOA's from calling the extrinsic.

yrong: Ok, I'll use `msg.sender` in both cases.

yrong:
Could we start with the teleport pattern as before which will make things a bit easier? Or by no means that third party chain will accept that?  

I'm not sure but as we may notice to support transact it requires some custom [xcm config](https://github.com/Snowfork/polkadot-sdk/pull/114/files#diff-37d844ab36f3c061d846240105ddb24c0832ca0f7da21a89ea3e051c6d978b3b) anyway and only DOT from BH is required so maybe not a big deal?

vgeddes:
> Could we start with the teleport pattern as before which will make things a bit easier? Or by no means that third party chain will accept that?  

Nope, unfortunately teleportation is strictly not allowed between system and third party chains.

yrong:
@vgeddes To not use teleport for transact in
https://github.com/Snowfork/polkadot-sdk/pull/114/commits/51b76fe5271e67337a15da4c0b097822f8580f12

alistair-singh:
> Nope, unfortunately teleportation is strictly not allowed between system and third party chains.

What if we teleported the DOT after the `UniversalOrigin` but before `DescendOrigin`. Then the Target chain would allow teleport of DOT from the Snowbridge Location. Only the bridge inbound-queue pallet should be able to assume this location's privilege. 

```
UniversalOrigin(GlobalConsensus(Ethereum)),
ReceiveTeleportedAsset { DOT },
BuyExecution { DOT },
DescendOrigin(AccountKey20 { msg.sender })
Transact { OriginKind = "SovereignAccount" OR "Xcm" }
```

yrong:
I've removed the teleport prerequisite in recent commits, the fee flow is:
- charge fees from msg.sender to agent of the dest chain, on Ethereum
- transfer fees from sovereign of dest chain to BH, on BH
- pay fees from sovereign of BH, on dest chain

contracts/src/interfaces/IGateway.sol
```diff
@@ -105,4 +105,7 @@ interface IGateway {
         uint128 destinationFee,
         uint128 amount
     ) external payable;
+
```
vgeddes, line 108: Similar to the APIs for `registerToken` and `sendToken`, we need a `quoteTransactFee` function that calculates the amount of ether the user must provide when calling `transact`.

vgeddes (commented):
First round of review comments..

contracts/src/Gateway.sol
```diff
@@ -617,4 +621,21 @@ contract Gateway is IGateway, IInitializable {
...
     }
+
+    /// @inheritdoc IGateway
+    function transact(ParaID destinationChain, TransactMessage calldata message) external payable {
+        if (message.validate() == false) {
+            revert InvalidTransact();
+        }
+        Ticket memory ticket;
+        Costs memory costs;
+        costs.foreign = message.fee;
```
vgeddes, line 632: To confirm what we discussed on the call:

* Similarly to Moonbeam's transact implementation, the user needs to have a prefunded sovereign account on the destination chain. The fee will be withdrawn from this account, and used to buy weight.
* Therefore, on the ethereum side, the user only needs to pay a small fee to cover the relayer reward on BridgeHub. This fee should be deposited into the agent of the destination parachain.

contracts/src/Gateway.sol
```diff
@@ -617,4 +621,21 @@ contract Gateway is IGateway, IInitializable {
         assets.assetHubCreateAssetFee = config.assetHubCreateAssetFee;
         assets.assetHubReserveTransferFee = config.assetHubReserveTransferFee;
     }
+
+    /// @inheritdoc IGateway
+    function transact(ParaID destinationChain, TransactMessage calldata message) external payable {
```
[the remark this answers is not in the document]
yrong: https://github.com/Snowfork/snowbridge/pull/1141/commits/4719a4252a3409d2cd629ce7adb6299c6581c245

contracts/src/Types.sol
```diff
@@ -109,3 +109,44 @@ struct TokenInfo {
...
+        result = 0x01;
+    } else if (kind == OriginKind.Xcm) {
+        result = 0x03;
+    }
+    return result;
+}
+
+using {validate} for TransactMessage global;
+
+struct TransactMessage {
```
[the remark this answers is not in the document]
yrong: https://github.com/Snowfork/snowbridge/pull/1141/commits/4719a4252a3409d2cd629ce7adb6299c6581c245

contracts/src/interfaces/IGateway.sol
yrong: https://github.com/Snowfork/snowbridge/pull/1141/commits/c192361a3bd28513f381d683f13192773b9fe6ad

contracts/src/Gateway.sol
yrong: https://github.com/Snowfork/polkadot-sdk/pull/114/commits/a8b84355336e9968c1ebe87054b8b465ffab3606

contracts/src/interfaces/IGateway.sol
```diff
@@ -107,5 +107,11 @@ interface IGateway {
     ) external payable;
 
     /// @dev Call transact in destinationChain
-    function transact(ParaID destinationChain, TransactMessage calldata message) external payable;
+    function transact(
+        ParaID destinationChain,
+        OriginKind originKind,
+        uint128 fee,
```
vgeddes, line 113: Lets rename this to `destinationFee`, to be similar to the `destinationFee` in `sendToken`.

yrong: https://github.com/Snowfork/snowbridge/pull/1141/commits/43b5442a488392c3557abad40558cf1dae8b8396

contracts/src/Gateway.sol
```diff
@@ -617,4 +621,31 @@ contract Gateway is IGateway, IInitializable {
...
+
+    /// @inheritdoc IGateway
+    function transact(
+        ParaID destinationChain,
+        OriginKind originKind,
+        uint128 destinationFee,
+        Weight calldata weightAtMost,
+        bytes calldata call
+    ) external payable {
+        if (call.length == 0 || destinationFee == 0 || weightAtMost.refTime == 0 || weightAtMost.proofSize == 0) {
```
alistair-singh, line 633: `call.length` must be at least 2 bytes. One for moduleId, and one for callIndex.

contracts/src/Gateway.sol
```diff
@@ -617,4 +621,31 @@ contract Gateway is IGateway, IInitializable {
...
         assets.assetHubReserveTransferFee = config.assetHubReserveTransferFee;
     }
+
+    /// @inheritdoc IGateway
+    function transact(
+        ParaID destinationChain,
+        OriginKind originKind,
+        uint128 destinationFee,
+        Weight calldata weightAtMost,
+        bytes calldata call
```
alistair-singh, line 631: Do we may need to limit the size of the call data or charge a fee based on the size of call data?

yrong: I would prefer not enforce that check on chain, could be achieved from UI side, and it's totally up to third party chain.

contracts/src/Gateway.sol
yrong: https://github.com/Snowfork/snowbridge/pull/1141/commits/f8f0f57b29aae65c48ae0fe4fb577fe8cd67ac68

contracts/src/Gateway.sol
```diff
@@ -617,4 +621,31 @@ contract Gateway is IGateway, IInitializable {
...
+        if (call.length < 2 || destinationFee == 0 || weightAtMost.refTime == 0 || weightAtMost.proofSize == 0) {
+            revert InvalidTransact();
+        }
+        Ticket memory ticket;
+        Costs memory costs;
+        bytes memory payload =
+            SubstrateTypes.Transact(msg.sender, originKind.encode(), destinationFee, weightAtMost, call);
+        ticket.dest = destinationChain;
+        ticket.costs = costs;
+        ticket.payload = payload;
```
vgeddes, line 642: I would rather initialise the ticket like this to make clear that we aren't mistakenly leaving costs uninitialized
```suggestion
Ticket memory ticket = Ticket({
  dest: destinationChain,
  costs: Costs({
    native: 0,
    foreign: 0,
  }),
  payload: payload
});
```

contracts/src/Gateway.sol
```diff
@@ -617,4 +621,31 @@ contract Gateway is IGateway, IInitializable {
...
+    function transact(
+        ParaID destinationChain,
+        OriginKind originKind,
+        uint128 destinationFee,
+        Weight calldata weightAtMost,
+        bytes calldata call
+    ) external payable {
+        if (call.length < 2 || destinationFee == 0 || weightAtMost.refTime == 0 || weightAtMost.proofSize == 0) {
+            revert InvalidTransact();
+        }
```
vgeddes, line 635: I feel a bit uncomfortable having the source chain interpreting and validating these fields - I think that's the job of the destination chain. Or in other words, I think we can pass these values through safely without validation.

contracts/src/Gateway.sol
```diff
@@ -617,4 +621,31 @@ contract Gateway is IGateway, IInitializable {
         assets.assetHubCreateAssetFee = config.assetHubCreateAssetFee;
         assets.assetHubReserveTransferFee = config.assetHubReserveTransferFee;
     }
+
+    /// @inheritdoc IGateway
+    function transact(
```
vgeddes, line 626: I think in this _user-facing_ API, we should rather call this function `sendCall`. This is also consistent with the naming of `sendToken`.

```suggestion
    function sendCall(
```

yrong: https://github.com/Snowfork/snowbridge/pull/1141/commits/5084432d3b8c46ad002216bc34f0d94c8ca4dc08

contracts/src/Gateway.sol
yrong: https://github.com/Snowfork/snowbridge/commit/00abc1330b4eb6e0932ddbc7e8d1bbe1ca757eae

Do you think better to add some native costs here to avoid spamming?

contracts/test/ScaleCodec.t.sol
```diff
@@ -47,4 +47,16 @@ contract ScaleCodecTest is Test {
         vm.expectRevert(ScaleCodec.UnsupportedCompactEncoding.selector);
         ScaleCodec.checkedEncodeCompactU32(uint256(type(uint32).max) + 1);
     }
+
+    function testEncodeCompactU64() public {
+        assertEq(ScaleCodec.encodeCompactU64(0), hex"00");
```
alistair-singh, line 52: +1

alistair-singh (approved):

yrong:
Add `TransactFeeMode` in
https://github.com/Snowfork/snowbridge/pull/1141/commits/3d91da451aeae9a88475ec881cdd6a44846982d5
and https://github.com/Snowfork/polkadot-sdk/pull/114/commits/7b916e5ff30fa614c7a3f7168974c5f2bab9aab4

web/packages/test/scripts/configure-substrate.sh
```diff
@@ -84,13 +84,20 @@ config_xcm_version() {
     send_governance_transact_from_relaychain $ASSET_HUB_PARAID "$call"
 }
 
+fund_penpal_sender_sovereign() {
+    # forceSetBalance($sender_sovereign_account, 1000000000000)
+    local transact_call="0x0a0800"$sender_sovereign_account"070010a5d4e8"
+    send_governance_transact_from_relaychain $PENPAL_PARAID "$transact_call"
+}
+
```
claravanstaden, line 92: Since the [method is repeated here](https://github.com/Snowfork/snowbridge/pull/1141/files#diff-1c6c806fd0be56e4d5eaa21c28ffcc41567c65dd40c8c470bea4f50556c8efe4R7), maybe we can use the method in `configure-penpal` instead of duplicating.

web/packages/test/scripts/set-env.sh
```diff
@@ -75,12 +77,12 @@ beacon_relayer_pub_key="${BEACON_RELAYER_PUB_KEY:-0xc46e141b5083721ad5f5056ba1cd
 execution_relayer_assethub_pub_key="${EXECUTION_RELAYER_ASSETHUB_PUB_KEY:-0x34284f0c1920694dd2e798e94e378307980d0f52d556009fc451c08bd65a8b4a}"
 # Execution relay account (//ExecutionRelayPenpal 5HgmfVcc8xBUcReNJsUaJMhFWGkdYpEw4RiCX4SeZPdKXR6H in testnet)
 execution_relayer_penpal_pub_key="${EXECUTION_RELAYER_PENPAL_PUB_KEY:-0xf8aed1861e571ef861cf34fe9587a211465aa380787e0102c2e89dfb0b666d3b}"
+# Sovereign account for msg.sender in penpal (Ethereum address:0x90A987B944Cb1dCcE5564e5FDeCD7a54D3de27Fe)
+sender_sovereign_account="ee99e7e8ac49f08251154c033f827541f4fb8a5b1fc4d6d9b1ab72c103bd3023"
```
claravanstaden, line 81: How did you get these addresses? For the sovereign account, using the Penpal Para ID and then: https://github.com/Snowfork/polkadot-sdk/blob/snowbridge/bridges/snowbridge/primitives/core/src/lib.rs#L44?

And the Ethereum address?

claravanstaden (approved):
👏🏻 Awesome!

claravanstaden:
> @alistair-singh As Vincent mentioned teleportation is strictly not allowed here so I've removed that in recent commits, the current fee flow is:
> 
> * charge fees from msg.sender and deposit to agent of the dest chain, on Ethereum
> * transfer fees from sovereign of dest chain to BH, on BH
> * buy execution and pay fees from sovereign of BH, on dest chain

Does this PR still work this way?

contracts/src/Types.sol
```diff
@@ -109,3 +109,44 @@ struct TokenInfo {
...
+    OriginKind originKind;
+    uint128 fee;
+    Weight weightAtMost;
+    bytes call;
+}
+
+function validate(TransactMessage calldata message) pure returns (bool) {
+    if (
+        message.call.length == 0 || message.fee == 0 || message.weightAtMost.refTime == 0
+            || message.weightAtMost.proofSize == 0
```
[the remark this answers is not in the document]
claravanstaden: Where did this validation move to? Should it not be done [here](https://github.com/Snowfork/snowbridge/pull/1141/files#diff-74470ba8f35b42dbc8805739e92708c0696343e0b1819feb3e7a5ebb77cc623cR643) somewhere?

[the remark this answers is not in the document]
yrong: Nice catch! It seems I missed that during some refactoring. Add it back here https://github.com/Snowfork/snowbridge/pull/1141/commits/3dd1adccda2767cdfbf54a5ece51d262a437cef0

contracts/src/Gateway.sol
```diff
@@ -623,4 +626,30 @@ contract Gateway is IGateway, IInitializable {
         assets.assetHubCreateAssetFee = config.assetHubCreateAssetFee;
         assets.assetHubReserveTransferFee = config.assetHubReserveTransferFee;
     }
+
+    // Calculate cost for transact
+    function _calculateTransactCost(uint128 destinationFee) internal pure returns (Costs memory costs) {
+        return Costs({native: 0, foreign: destinationFee});
```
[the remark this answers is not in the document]
yrong: Nope, all fees paid on Ethereum in advance.

web/packages/test/scripts/set-env.sh
yrong: It's for a legacy fee mode paying fees on the destination chain which is deprecated. So just remove the script here.

https://github.com/Snowfork/snowbridge/pull/1141/commits/70413d80d4f636778f05ce3b53660bdf6105c91d

yrong:
> Does this PR still work this way?

Nope. Sorry for the confusion, I should have removed it earlier.

claravanstaden:
> > Does this PR still work this way?
> 
> Nope. Sorry for the confusion, I should have removed it earlier.

Would you mind adding a new comment with the updated functionality? That summary is helpful to understand the PR at a high level.

vgeddes:
> One caveat for Parachain to integrate is that since the fee is paid upfront on Ethereum and `UnpaidExecution` is sent to the dest chain, the risk potentially is a malicious user can spam the dest chain to call expensive transact with a tiny fee. So It's the responsibility of Parachain to inspect/check the fee passed through can cover the execution cost though not charged.
> 
> Custom Xcm config for that In [Snowfork/polkadot-sdk@c18f771](https://github.com/Snowfork/polkadot-sdk/commit/c18f77115499f2806999ecf17c80261a7c178d42) could be referenced.

Good catch. At least with paid execution, the `Transact` includes the weight required at most, and if the user's supplied execution fee does not cover the maximum transact weight, the xcm will not be executed.

contracts/src/Gateway.sol
```diff
@@ -613,4 +620,33 @@ contract Gateway is IGateway, IInitializable, IUpgradable {
         assets.assetHubCreateAssetFee = config.assetHubCreateAssetFee;
         assets.assetHubReserveTransferFee = config.assetHubReserveTransferFee;
     }
+
+    // Calculate cost for transact
+    function _calculateTransactCost(uint128) internal pure returns (Costs memory costs) {
```
alistair-singh, line 625: This is uneeded and can probably be removed?

contracts/src/Gateway.sol
```diff
@@ -613,4 +620,33 @@ contract Gateway is IGateway, IInitializable, IUpgradable {
...
+        bytes memory payload =
+            SubstrateTypes.Transact(msg.sender, originKind.encode(), destinationFee, weightAtMost, call);
+        Costs memory costs = _calculateTransactCost(destinationFee);
+        Ticket memory ticket = Ticket({dest: destinationChain, costs: costs, payload: payload});
+        _submitOutbound(ticket);
+    }
+
+    /// @inheritdoc IGateway
+    function quoteSendCallFee(uint128 destinationFee) external view returns (uint256) {
+        Costs memory costs = _calculateTransactCost(destinationFee);
```
alistair-singh, line 649: We can remove destinationfee from here because only delivery costs are paid in the new fee model.

rfc/transact_from_ethereum_to_substrate.md
```diff
@@ -0,0 +1,47 @@
+# RFC: Transact from Ethereum to Substrate
+
+
+## Summary
+
+This RFC proposes the feature to call transact from Ethereum to Substrate through our bridge, including two PRs separately with https://github.com/Snowfork/snowbridge/pull/1141 for solidity and https://github.com/Snowfork/polkadot-sdk/pull/114 for substrate. 
+
+## Explanation
+
+Basically it works as follows:
```
alistair-singh, line 10: I would add a fee flow section:

- dApp is represents `msg.sender` or its sovereign depending on context.
- Parachain represents the target parachain, its sovereign or its agent depending on context.

Sequence|Where|Who|What
-|-|-|-
1|Gateway|dApp|pays(ETH, converted to DOT here) Parachain Agent for delivery costs only (no execution cost)
2|Bridge Hub|Relayer|pays(DOT) node for execution
3|Bridge Hub|Parachain Sovereign|pays(DOT) Relayer for delivery (refund+reward)
4|Parachain|dApp|pays(DOT, Native) for execution only.

Does this balance? Yes because the parachain agent collects fees for delivery on Ethereum side and the Parachain Sovereign pays for delivery on the Bridge Hub side. 

Execution costs totally belongs to the dApp prefunding its sovereign on the Parachain.

contracts/src/Gateway.sol
yrong: https://github.com/Snowfork/snowbridge/pull/1141/commits/61d5ea5688072fe4b4d1b4a81d4e8c2ac558c5cc

contracts/src/Gateway.sol
```diff
@@ -613,4 +620,33 @@ contract Gateway is IGateway, IInitializable, IUpgradable {
...
+        }
+        bytes memory payload =
+            SubstrateTypes.Transact(msg.sender, originKind.encode(), destinationFee, weightAtMost, call);
+        Costs memory costs = _calculateTransactCost();
+        Ticket memory ticket = Ticket({dest: destinationChain, costs: costs, payload: payload});
+        _submitOutbound(ticket);
+    }
+
+    /// @inheritdoc IGateway
+    function quoteSendCallFee(uint128 destinationFee) external view returns (uint256) {
```
alistair-singh, line 648: `destinationFee` can be removed from here, as the quote should only return delivery fees.

yrong: https://github.com/Snowfork/snowbridge/pull/1141/commits/bfdff266ac2acac0a39be8f3f82897a427b1dd72

vgeddes:
Obsolete

PR #1143, opened 2024-02-20
Fix code coverage

claravanstaden:
Fixes the coverage config file not to include the parachain. States on the readme the coverage is for contracts.

### DIFF
--- a/.github/workflows/contracts.yml
+++ b/.github/workflows/contracts.yml
@@ -8,9 +8,9 @@ on:
     branches:
       - main
   pull_request:
-  #  paths:
-  #    - "contracts/**"
-  #    - "!contracts/**/README.md"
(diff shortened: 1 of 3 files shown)


vgeddes:
Ah I would remove the note that the coverage is for contracts only. It's in a very prominent position at the start of the readme, yet is not very important.

Also there is some kind of markup error - see the bullet points in the image below. Are you trying to use some special markup?

Capture-2024-02-20-182414

claravanstaden:
@vgeddes I added an asterisk to indicate a special note about the coverage. 😂 I see that's not clear, I'll remove it.

alistair-singh (approved):

vgeddes (approved):
+1

PR #113, opened 2020-09-25
E2E Tests

vgeddes:
Fleshed out the E2E tests.

Changes:
* For simplicitly and consistency, the test uses Wei units and `BigNumber.js` for arithmetic and assertions.
* Calculate actual gasCost by retrieving transaction.

PR #1144, opened 2024-02-20
Run Smoke Tests in CI

claravanstaden:
Adds:
- Http service to the beacon relayer to check if the relayer is running
- Bash script to run the smoke tests
- Bash script to check if the beacon relayer is running
- Minor changes to get the e2e stack to work in GA

Resolves: [SNO-857](https://linear.app/snowfork/issue/SNO-857)
Polkadot-sdk companion: [Smoke Tests in CI #113](https://github.com/Snowfork/polkadot-sdk/pull/113)


scripts/init.sh
```diff
@@ -36,8 +39,13 @@ echo "Installing web packages"
 echo "Download geth to replace the nix version"
 OS=$(uname -s | tr A-Z a-z)
 MACHINE_TYPE=$(uname -m | tr A-Z a-z)
+if [ "$OS" == "linux" ]; then
+  MACHINE_TYPE="amd64"
+fi
```
alistair-singh, line 44: This should work for mac 64 bit intel as well with the following modification:
```suggestion
MACHINE_TYPE=$(uname -m | tr A-Z a-z | sed 's/x86_64/amd64/')
```

relayer/cmd/run/beacon/command.go
```diff
@@ -102,3 +108,27 @@ func run(_ *cobra.Command, _ []string) error {
 
 	return nil
 }
+
+func startHttpServer(ctx context.Context) error {
+	srv := &http.Server{Addr: ":8080", Handler: nil}
```
alistair-singh, line 113: Make the port a command line flag which defaults to 8080.

web/packages/test/scripts/run-smoketests.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
+set -eu
+
+source scripts/set-env.sh
+
+pushd $root_dir/smoketest
+
+./make-bindings.sh
+echo "run register token test"
+cargo test --test register_token --release -- --nocapture
```
alistair-singh, line 10: I would build all tests without running them first to catch any build errors early.
```suggestion
echo "build tests"
cargo test --no-run
echo "run register token test"
cargo test --test register_token --release -- --nocapture
```

alistair-singh:
Why do we need the health check?

claravanstaden:
@alistair-singh The start-services.sh script is run in the background: https://github.com/Snowfork/polkadot-sdk/pull/113/files#diff-ebcd3a3b6186ad4cf5cf2337ff84cc4692492610301fefb171509d671bf9f507R62

So I needed a way to know if the E2E stack is complete. I'm open to better ideas. 😅 I guess I could just check if the relayer process is running 🤷🏻‍♀️

relayer/cmd/run/beacon/command.go
claravanstaden: I ended up removing the http service and just checked if the beacon relayer process is running.

PR #1145, opened 2024-02-20
Transact from sub to eth

yrong:
Resolves: SNO-586,SNO-866,SNO-696
Requires:  https://github.com/Snowfork/polkadot-sdk/pull/116


contracts/src/utils/Call.sol
```diff
@@ -22,4 +22,54 @@ library Call {
             }
         }
     }
+
+    /// @notice Use when you _really_ really _really_ don't trust the called
+    /// contract. This prevents the called contract from causing reversion of
+    /// the caller in as many ways as we can.
```
alistair-singh, line 28: We should understand how using zexcessivelySafeCall` can still fail? And potentially have tests for at least one case to make sure we don't poisen pill the inbound queue.

alistair-singh (commented):
Looks good! Just one comment tests.

contracts/src/utils/Call.sol
yrong: Yeah, an unsafe call could be dangerous so we need to be very careful here. Except that I would also like to introduce white list control, something like the [SafeCallFilter](https://github.com/polkadot-fellows/runtimes/blob/2ef38e75b43b2b91d160936843fa8b3e4ffbf5fc/system-parachains/bridge-hubs/bridge-hub-kusama/src/xcm_config.rs#L135) but on Ethereum side.

yrong: https://github.com/Snowfork/snowbridge/pull/1145/commits/ebc89866f4134dbd7c8e6aaf2f6ead33330a7fa5

vgeddes: My thoughts about this:

1. I think `excessivelySafeCall` (used by Nomad) is too complicated for our needs. When issuing a call, we are only interested in whether the call succeeded or not. The `returndata` we can ignore. For this reason, I prefer that we use this safe call solution: https://github.com/ethereum-optimism/optimism/blob/b36eb5515cc2a34a15383b2eee488dbac83d6caf/packages/contracts-bedrock/src/libraries/SafeCall.sol#L12

2. I am optimistic we've accounted for all the ways a contract call can be unsafe. This includes guards against re-entrancy (nonce check), gas limits, and returnbomb attacks. Therefore I don't see a reason to include a safe call filter. If we think a call can still be unsafe, we need to understand why.

yrong: > How does SafeCallFilter work? Will we have to restrict allowing transact to only explicitly specified contracts and/or functions?

It's not by us or require any sudo call from relaychain, it's a system control call from the Parachain sovereign through xcm to increase the legal transacts step by step totally in their own control.

In this way we can release the transact PR but disable it without introducing any risk.

yrong: SafeCall ignore returndata https://github.com/Snowfork/snowbridge/pull/1145/commits/0a8dc1e425d495a1bfa217cea6dde520260519ec

rfc/transact_from_substrate_to_ethereum.md
```diff
@@ -0,0 +1,55 @@
+# RFC: Transact from Substrate to Ethereum
+
+
+## Summary
+
+This RFC proposes the feature to call transact from Substrate to Ethereum through our bridge, including two PRs separately with https://github.com/Snowfork/snowbridge/pull/1145 for solidity and https://github.com/Snowfork/polkadot-sdk/pull/116 for substrate.
+
+## Explanation
+
+We use penpal for the integration, basically it works as follows:
```
alistair-singh, line 10: Fee flow:

- User represents a user who kicks off an extrinsic on the parachain.
- Parachain represents the source parachain, its sovereign or its agent depending on context.

Sequence|Where|Who|What
-|-|-|-
1|Parachain|User|pays(DOT, Native) to node to execute custom extrinsic; pays (DOT) to Treasury for delivery as part of custom extrinsic.
2|Bridge Hub|Parachain|pays(DOT) to Treasury Account for delivery(local fee), pays(DOT) to Parachain sovereign for delivery(remote fee), essentially a refund. Remote fee converted to ETH here.
3|Gateway|Relayer|pays(ETH) to validate and execute message.
4|Gateway|Parachain Agent|pays(ETH) to relayer for delivery(reward+refund) and execution.

Does this balance? No, Parachain agent pays for delivery and execution on the Ethereum side. User only pays for calling the extrinsic and delivery on the source Parachain. Because the custom extrinsic essentially allows the user to impersonate the parachain on ethereum side it becomes the Parachains job to charge appropriately for both delivery and execution. This doesn't mean there is something wrong with the implementation, just that this needs to be communication to creator of the custom extrinsic.

yrong: For step 2 only pays(DOT) for the delivery(local fee) portion.

https://github.com/Snowfork/polkadot-sdk/blob/569d638d429944fc8d4e2d7db3b653d565879f86/bridges/snowbridge/primitives/router/src/outbound/mod.rs#L119-L125

yrong: Summarize it in https://github.com/Snowfork/snowbridge/pull/1145/commits/bd88ed892f307ac79a941f0d45097ba17a8b47ee

contracts/src/Gateway.sol
```diff
@@ -274,12 +280,24 @@ contract Gateway is IGateway, IInitializable, IUpgradable {
...
+        (AgentExecuteCommand command, bytes memory commandParams) =
+            abi.decode(params.payload, (AgentExecuteCommand, bytes));
+        if (command == AgentExecuteCommand.Transact) {
+            (address target, bytes memory payload,) = abi.decode(commandParams, (address, bytes, uint64));
+            SafeCallFilterStorage.Layout storage $ = SafeCallFilterStorage.layout();
+            bytes4 selector = bytes4(payload.slice(0, 4));
+            if ($.safeCalls[target][selector] != true) {
+                revert NoPermission();
+            }
+        }
```
claravanstaden, line 292: So this is just a temporary measure to release this feature safely to allowed chains?

smoketest/tests/transact_from_penpal_to_ethereum.rs
```diff
@@ -0,0 +1,83 @@
...
+	let test_clients = initial_clients().await.expect("initialize clients");
+
+	let ethereum_client = *(test_clients.ethereum_client.clone());
+	let penpal_client = *(test_clients.penpal_client.clone());
+
+	let hello_world = HelloWorld::new(HELLO_WORLD_CONTRACT, ethereum_client.clone());
+	let contract_abi: Abi = hello_world.abi().clone();
+	let function = contract_abi.function("sayHello").unwrap();
+	let encoded_data =
+		function.encode_input(&[Token::String("Hello, Clara!".to_string())]).unwrap();
```
claravanstaden, line 31:
```suggestion
		function.encode_input(&[Token::String("Hello!".to_string())]).unwrap();
```

😅

claravanstaden (approved):

contracts/src/Gateway.sol
musnit: i worry that this safe call filter both here and in your past PRs is an anti-pattern that doesn't just limit the functionality of the bridge, but also leads us to bad architectural design decisions that will have wider impact on fees, execution/channels and our trust model overall

yrong: Actually we can't allow transact to invoke arbitrary call, at least not including the `AgentExecutor::mintToken|burnToken` which is introduced in polkadot native assets.

More context in https://github.com/Snowfork/snowbridge/pull/1155#issuecomment-2026820489

I can probably change it to a blacklist, the whitelist is just follow the pattern every parachain should declare which calls are exposed through xcm.

yrong: > The real thing we need here is a different Agent for transact than the Agent we use to lock funds in, maybe an agent derived from a pallet on parachain or user.

Yeah, I like this. The isolation with different Agent make more sense here. I'll make change accordingly.

yrong: https://github.com/Snowfork/polkadot-sdk/pull/116/commits/f4b6dba04744f6c31a589fe7ccb83cdc3be886f5#diff-952be233381eef865196222dc9f95b0d88e54cdcec924e84d03c5622d2e59ec9 and https://github.com/Snowfork/snowbridge/pull/1145/commits/606e867b7badc6d356c8f4b56e6b81ee0eb27811

vgeddes: >>The real thing we need here is a different Agent for transact than the Agent we use to lock funds in, maybe an agent derived from a pallet on parachain or user.

> Yeah, I like this. The isolation with different agents make more sense here. I'll make change accordingly.

Agree

contracts/src/Gateway.sol
```diff
@@ -381,6 +389,20 @@ contract Gateway is IGateway, IInitializable, IUpgradable {
...
+        TransactCallParams memory params = abi.decode(data, (TransactCallParams));
+        address agent = _ensureAgent(params.agentID);
+        bytes memory call =
+            abi.encodeCall(AgentExecutor.executeCall, (params.target, params.payload, params.dynamicGas));
+        (, bytes memory result) = Agent(payable(agent)).invoke(AGENT_EXECUTOR, call);
+        (bool success) = abi.decode(result, (bool));
+        if (!success) {
+            revert AgentTransactCallFailed();
+        }
+        emit TransactExecuted(params.agentID, params.target, params.payload);
```
vgeddes, line 403: `params.payload` could be very large, cheaper to hash it when emitting event

```suggestion
        emit TransactExecuted(params.agentID, params.target, keccak256(params.payload));
```

contracts/src/Params.sol
```diff
@@ -82,3 +82,15 @@ struct SetPricingParametersParams {
...
+// Payload for TransactCall
+struct TransactCallParams {
+    /// @dev The agent ID of the consensus system
+    bytes32 agentID;
+    /// @dev The target contract
+    address target;
+    /// @dev Payload of the call
+    bytes payload;
+    /// @dev Max gas cost of the call
+    uint64 dynamicGas;
```
vgeddes, line 95: The message packet already has a `maxDispatchGas`, which is the upper limit a transact can consume. So I am not sure what role this `dynamicGas` plays.

The gas estimator on BridgeHub should add a buffer to `maxDispatchGas` to cover the cost of dispatching via the agent.

contracts/src/interfaces/IGateway.sol
```diff
@@ -35,6 +35,8 @@ interface IGateway {
     // Emitted when funds are withdrawn from an agent
     event AgentFundsWithdrawn(bytes32 indexed agentID, address indexed recipient, uint256 amount);
 
+    event TransactExecuted(bytes32 indexed agentID, address indexed target, bytes payload);
```
vgeddes, line 38:
```suggestion
    event Transacted(bytes32 indexed agentID, address indexed target, bytes32 payloadHash);
```

contracts/src/utils/Call.sol
```diff
@@ -22,4 +22,76 @@ library Call {
...
+    /// to memory.
+    /// @param _target The address to call
+    /// @param _gas The amount of gas to forward to the remote contract
+    /// @param _value The value in wei to send to the remote contract
+    /// @param _maxCopy The maximum number of bytes of returndata to copy
+    /// to memory.
+    /// @param _calldata The data to send to the remote contract
+    /// @return success and returndata, as `.call()`. Returndata is capped to
+    /// `_maxCopy` bytes.
+    function excessivelySafeCall(address _target, uint256 _gas, uint256 _value, uint16 _maxCopy, bytes memory _calldata)
```
vgeddes, line 42: We're no longer using this, can we remove it?

contracts/src/Gateway.sol
yrong: https://github.com/Snowfork/snowbridge/pull/1145/commits/7880d466ac4cf20b843279b041aa490b509f0483

contracts/src/Params.sol
yrong: Yes, it's the same thing. Just rename it for less confusion.
https://github.com/Snowfork/snowbridge/pull/1145/commits/27c14e94cdba7eed261178b7f2391b055d74bee5

contracts/src/interfaces/IGateway.sol
yrong: https://github.com/Snowfork/snowbridge/pull/1145/commits/14a40cf610f2f832222bdc57df9248617bfd0c23

contracts/src/utils/Call.sol
yrong: https://github.com/Snowfork/snowbridge/pull/1145/commits/0510064a144a5d8b5a8d2e4f797471474d22f5c8

vgeddes:
Obsolete

PR #1146, opened 2024-02-21
Changes required to enable contract upgrades

alistair-singh:
The initialization code would run on every upgrade. This stops that behavior.
Resolves: SNO-870

Upgrades can now be done by:
1. Set the required parameters in the .envrc.
2. Deploy the gateway logic contract only.
```cmd
from_start_services=true sh -c "source ./scripts/deploy-contracts.sh && deploy_gateway_logic"
```
3. Run the Snowbridge control cli to create a pre-image: https://github.com/Snowfork/snowbridge/pull/1147
4. Run the opengov cli to create a proposal from pre-image: https://github.com/joepetrowski/opengov-cli

### DIFF
--- a/contracts/src/Gateway.sol
+++ b/contracts/src/Gateway.sol
@@ -86,6 +86,7 @@ contract Gateway is IGateway, IInitializable {
     error InvalidAgentExecutionPayload();
     error InvalidCodeHash();
     error InvalidConstructorParams();
+    error AlreadyInitialized();
 
     // handler functions are privileged
     modifier onlySelf() {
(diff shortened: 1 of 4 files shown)


alistair-singh:
Re-opening.

contracts/src/Gateway.sol
```diff
@@ -578,43 +578,47 @@ contract Gateway is IGateway, IInitializable {
...
-        Config memory config = abi.decode(data, (Config));
-
         CoreStorage.Layout storage core = CoreStorage.layout();
-        core.mode = config.mode;
 
-        // Initialize agent for BridgeHub
-        address bridgeHubAgent = address(new Agent(BRIDGE_HUB_AGENT_ID));
-        core.agents[BRIDGE_HUB_AGENT_ID] = bridgeHubAgent;
+        // Check if initialized and run initialization once.
+        if (core.channels[PRIMARY_GOVERNANCE_CHANNEL_ID].agent == address(0)) {
```
vgeddes, line 584: This quite an invasive way of implementing this check, introducing a lot of code churn. 

Rather invert the check:

```solidity
if (core.channels[PRIMARY_GOVERNANCE_CHANNEL_ID].agent != address(0)) {
  revert AlreadyInitialized()
}
```

As the above example shows, I think we should probably also fail the upgrade by reverting.

claravanstaden (approved):

yrong (approved):
+1

vgeddes (approved):

PR #1147, opened 2024-02-26
CLI Tool for creating governance preimages

vgeddes:
This tool creates preimages in which can be executed on the relay chain to send governance messages to BridgeHub. The tool also computes the correct weights for `Transact` and `UnpaidExecution`.

It relies on generated bindings for BridgeHub and the Relay chain, which ensures type safety.

Right now, as examples, it supports governance commands for setting the operational mode of the gateway, as well as upgrading the gateway.

```
$ target/debug/snowbridge-control 
Usage: snowbridge-control [FORMAT] <COMMAND>

Commands:
  gateway-operating-mode  Change the gateway operating mode
  upgrade                 Upgrade the Gateway contract
  help                    Print this message or the help of the given subcommand(s)

Arguments:
  [FORMAT]  Output format of preimage [default: hex] [possible values: hex, binary]

Options:
  -h, --help     Print help
  -V, --version  Print version
```

### Compilation

The tool needs to be compiled separately for each our environments:

```bash
# Rococo (default)
cargo build

# Kusama
cargo build --no-default-features --features kusama

# Polkadot
cargo build --no-default-features --features polkadot
```


control/src/main.rs
```diff
@@ -0,0 +1,162 @@
...
+}
+
+#[derive(Debug, Subcommand)]
+pub enum Command {
+    /// Change the gateway operating mode
+    GatewayOperatingMode {
+        /// Operating mode
+        #[arg(value_enum)]
+        mode: GatewayOperatingModeArg,
+    },
```
alistair-singh, line 29: Having the set operating mode command is useful but this would also need to be in the same pre-image as the `Upgrade` otherwise an upgrade would require two proposals. 
The pre-image for upgrade needs to:
1. Set to mode `RejectingOutbound`.
2. Call upgrade extrinsic.
3. Set mode to `Normal`

control/src/main.rs
```diff
@@ -0,0 +1,162 @@
...
+    #[arg(value_enum, default_value_t=Format::Hex)]
+    format: Format,
+
+    #[command(subcommand)]
+    command: Command,
+}
+
+#[derive(Debug, Subcommand)]
+pub enum Command {
+    /// Change the gateway operating mode
```
alistair-singh, line 24: We should maybe have a command for the initial deploy that creates a single preimage for the whole deploy like we would do in production, and then we consume the command from start services in order to set up the local testnet.

This would be the most useful command given the kusama deployment is soon.

alistair-singh (commented):
Related PR: https://github.com/Snowfork/snowbridge/pull/1146

control/src/main.rs
vgeddes: I designed Gateway upgrades to not need parts of the bridge to be disabled first. At least for most routine upgrades.

control/src/main.rs
vgeddes: Yeah definitely! The main helper function `wrap_calls` can already emit an XCM script that contains multiple Transacts (if we go that route instead of `Utility.batch`).

yrong: Seems `Utility.batch(_all)` is not allowed in BH runtime through xcm so this one is super useful.  The caveat is to be careful to not exceed the weight limit for xcm.
https://github.com/polkadot-fellows/runtimes/blob/2ef38e75b43b2b91d160936843fa8b3e4ffbf5fc/system-parachains/bridge-hubs/bridge-hub-kusama/src/lib.rs#L291C20-L291C29

control/runtimes/rococo/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "rococo-runtime"
```
claravanstaden, line 2: Will we ever publish these crates? https://crates.io/crates/rococo-runtime If we will, maybe we should prefix with snowbridge something.

claravanstaden (approved):
Very nice! The info you added in this PR description would be useful in a readme.

I ran it locally and it works very well - easy to use. Are you still going to generate the bindings for Kusama and Polkadot?

control/preimage/src/helpers.rs
```diff
@@ -0,0 +1,114 @@
...
+pub fn increase_weight(ref_time: &mut u64, proof_size: &mut u64) {
+    let x = multiply_by_rational_with_rounding(*ref_time as u128, 110, 100, Rounding::Up).expect("overflow");
+    let y = multiply_by_rational_with_rounding(*proof_size as u128, 110, 100, Rounding::Up).expect("overflow");
+
+    *ref_time = x.try_into().expect("overflow");
+    *proof_size = y.try_into().expect("overflow");
+}
+
+// Base weight costs for Transact and ExpectTransactStatus
+const TRANSACT_INSTR_WEIGHT: u64 = 10_000_000;
```
yrong, line 39: Could take https://github.com/polkadot-fellows/runtimes/blob/2ef38e75b43b2b91d160936843fa8b3e4ffbf5fc/system-parachains/bridge-hubs/bridge-hub-kusama/src/weights/xcm/pallet_xcm_benchmarks_generic.rs#L82
as reference.

yrong: Also some weight should be reserved for the `unpaid_execution`
https://github.com/polkadot-fellows/runtimes/blob/2ef38e75b43b2b91d160936843fa8b3e4ffbf5fc/system-parachains/bridge-hubs/bridge-hub-kusama/src/weights/xcm/pallet_xcm_benchmarks_generic.rs#L337-L343

yrong (approved):
Cool!

control/runtimes/rococo/Cargo.toml
vgeddes: Nah, I think we will only compile this tool on-demand.

control/preimage/src/helpers.rs
vgeddes: Actually in the latest upload, I just configured `UnpaidExecution` to use unlimited weight.

alistair-singh (approved):
+1

alistair-singh:
Merging as I require this change.

PR #1149, opened 2024-03-05
Testnet improvements

alistair-singh:
Originally included this in my Javascript API branch. But decided to split it off into a separate PR.

1. Update block wait times to 100 for tests. It was failing inconsistently.
2. Make relayers use unique accounts. They where previously sharing an account.
3. Merge `configure-bridgehub.sh` and `configure-assethub.sh` into a single `configure-substrate.sh`.
4. Upgrade polkadot-js-api version and standardize js dependencies. Use flags `--noWait  --nonce -1` when using `polkadot-js-api` to make testnet start faster. `2min40seconds`


yrong (approved):
Cool!

PR #1151, opened 2024-03-08
Add fallback URL for beacon state download

claravanstaden:
Resolves: [SNO-905](https://linear.app/snowfork/issue/SNO-905)


claravanstaden:
Not necessary anymore, using a beacon store instead of a fallback URL since the other consensus clients seem to have similar problems to Lodestar anyway.

PR #1155, opened 2024-03-13
Asset from polkadot

yrong:
Requires: https://github.com/Snowfork/polkadot-sdk/pull/128


contracts/src/AgentExecutor.sol
```diff
@@ -36,4 +48,19 @@ contract AgentExecutor {
     function _transferToken(address token, address recipient, uint128 amount) internal {
         IERC20(token).safeTransfer(recipient, amount);
     }
+
+    /// @dev Register native asset from polkadto as ERC20 `token`.
+    function _registerToken(bytes32 agentID, bytes32 tokenID, string memory name, string memory symbol, uint8 decimals)
+        internal
+    {
+        IERC20 token = new ERC20(name, symbol, decimals);
+        Gateway(msg.sender).registerTokenByID(tokenID, address(token), agentID);
```
vgeddes, line 57: Thought experiment:

Instead of calling a function on the Gateway, make `_registerToken` and `execute` return an ABI-encoded `calldata`. Then in `Gateway.agentExecute`, execute that `calldata` using `call`.

contracts/src/AgentExecutor.sol
```diff
@@ -7,21 +7,33 @@ import {SubstrateTypes} from "./SubstrateTypes.sol";
...
     using SafeNativeTransfer for address payable;
 
+    // Emitted when token minted
+    event TokenMinted(bytes32 indexed tokenID, address token, address recipient, uint256 amount);
+
     /// @dev Execute a message which originated from the Polkadot side of the bridge. In other terms,
     /// the `data` parameter is constructed by the BridgeHub parachain.
     ///
-    function execute(bytes memory data) external {
+    function execute(bytes32 agentID, bytes memory data) external {
```
vgeddes, line 25: See https://github.com/Snowfork/snowbridge/pull/1155/files#r1531673672 for an idea to experiment with

```suggestion
    function execute(bytes32 agentID, bytes memory data) external returns (bytes memory) {
```

contracts/src/AgentExecutor.sol
yrong: https://github.com/Snowfork/snowbridge/pull/1155/commits/780c7bd920796d22169e34406ff48ea8b4b041cf

contracts/src/AgentExecutor.sol
```diff
@@ -36,4 +48,19 @@ contract AgentExecutor {
...
+    function _registerToken(bytes32 agentID, bytes32 tokenID, string memory name, string memory symbol, uint8 decimals)
+        internal
+    {
+        IERC20 token = new ERC20(name, symbol, decimals);
+        Gateway(msg.sender).registerTokenByID(tokenID, address(token), agentID);
+    }
+
+    /// @dev Mint ERC20 token to `recipient`.
+    function _mintToken(bytes32 tokenID, address recipient, uint256 amount) internal {
+        address token = Gateway(msg.sender).getTokenAddress(tokenID);
```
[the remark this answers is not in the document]
yrong: https://github.com/Snowfork/snowbridge/pull/1155/commits/27353f0e9802d0cd04b423d1b7b13a060b80e400

yrong:
In https://github.com/Snowfork/snowbridge/pull/1155/commits/0b3344cb9e35e6df5ad9c745fd24ef1559635aab I made the change to reuse `IGateway.sendToken` for polkadot native asset. 

I'd still prefer to separate as different message types internally. There are differences in several aspects and the separating will make the code a bit cleaner IMO.

yrong:
Add 3 smoke tests to cover the basic flow


```
cargo test --test register_polkadot_token // register relay token(i.e. ROC|KSM|DOT) on Ethereum
cargo test --test transfer_polkadot_token // transfer relay token from AssetHub to Ethereum
cargo test --test send_polkadot_token // send relay token from Ethereum back to AssetHub
```

contracts/src/Assets.sol
```diff
@@ -174,4 +176,80 @@ library Assets {
...
+                info.tokenID, destinationChain, destinationAddress.asAddress20(), destinationChainFee, amount
+            );
+        } else {
+            revert Unsupported();
+        }
+
+        emit IGateway.TokenSent(info.token, sender, destinationChain, destinationAddress, amount);
+    }
+
+    function _burn(address agentExecutor, address agent, bytes32 tokenID, address sender, uint256 amount) internal {
```
vgeddes, line 245:
```suggestion
    function _burnToken(address agentExecutor, address agent, bytes32 tokenID, address sender, uint256 amount) internal {
```

contracts/src/Assets.sol
```diff
@@ -174,4 +176,80 @@ library Assets {
...
+        MultiAddress calldata destinationAddress,
+        uint128 destinationChainFee,
+        uint128 amount
+    ) external returns (Ticket memory ticket) {
+        CoreStorage.Layout storage $core = CoreStorage.layout();
+
+        address agent = $core.agents[info.agentID];
+        if (agent == address(0)) {
+            revert AgentDoesNotExist();
+        }
```
vgeddes, line 216: Rather move this to `Gateway.sendToken` and pass `agent` as a parameter.

The `Assets` library doesn't need to know how agents are stored.

contracts/src/Gateway.sol
```diff
@@ -265,12 +268,21 @@ contract Gateway is IGateway, IInitializable {
...
+        (AgentExecuteCommand command, bytes memory payload) = abi.decode(params.payload, (AgentExecuteCommand, bytes));
+
+        bytes memory call = abi.encodeCall(AgentExecutor.execute, (params.agentID, command, payload));
 
         (bool success, bytes memory returndata) = Agent(payable(agent)).invoke(AGENT_EXECUTOR, call);
         if (!success) {
             revert AgentExecutionFailed(returndata);
         }
+
+        if (command == AgentExecuteCommand.RegisterToken) {
```
vgeddes, line 280: I think the approach I suggested to return a command from from `Agent.invoke` doesn't look as neat as the Agent calling `Gateway.registerToken`.

So I think we should go back to that old way.

vgeddes (commented):
Looks good! Though in retrospect, some of my previous recommendations to you were not good ideas in the end, and so I'd like you to revert those.

contracts/src/AgentExecutor.sol
```diff
@@ -36,4 +55,29 @@ contract AgentExecutor {
...
         IERC20(token).safeTransfer(recipient, amount);
     }
+
+    /// @dev Register native asset from polkadto as ERC20 `token`.
+    function _registerToken(bytes32 agentID, bytes32 tokenID, string memory name, string memory symbol, uint8 decimals)
+        internal
+        returns (bytes memory)
+    {
+        IERC20 token = new ERC20(name, symbol, decimals);
+        Assets.registerTokenByID(tokenID, address(token), agentID);
```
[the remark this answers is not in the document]
yrong: https://github.com/Snowfork/snowbridge/pull/1155/commits/fddb9cf3b38ee6c25e4f23fc26daa70e0b9ea73e

Never mind, it's always fine to reach a better design iteratively. :relaxed:

contracts/src/Types.sol
```diff
@@ -107,5 +109,8 @@ struct Ticket {
 
 struct TokenInfo {
     bool isRegistered;
-    bytes31 __padding;
```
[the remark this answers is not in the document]
yrong: Yeah, could be a problem. I'd prefer a specific migration https://github.com/Snowfork/snowbridge/pull/1155/commits/e5d9dd847dad0977e6778af5ed9da7a1edef3eb8

contracts/src/Assets.sol
yrong: https://github.com/Snowfork/snowbridge/pull/1155/commits/edccccdf014c5d4a8934591bef649a870e430992

contracts/src/Assets.sol
yrong: https://github.com/Snowfork/snowbridge/pull/1155/commits/7c29684614addf8f404a914696c9a14b7fe64df5

contracts/src/Gateway.sol
```diff
@@ -428,9 +440,27 @@ contract Gateway is IGateway, IInitializable {
...
+                )
+            );
+        } else {
+            _submitOutbound(
+                Assets.sendToken(token, msg.sender, destinationChain, destinationAddress, destinationFee, amount)
+            );
+        }
+    }
+
+    function getTokenInfo(bytes32 tokenID) external view returns (TokenInfo memory) {
```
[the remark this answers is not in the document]
yrong: https://github.com/Snowfork/snowbridge/pull/1155/commits/6eb8344598de0b90e2b37bd2ce32f90f0deb221c

contracts/src/Assets.sol
```diff
@@ -174,4 +176,80 @@ library Assets {
...
+
+        // Polkadot-native token: burn wrapped token
+        _burn(executor, agent, info.tokenID, sender, amount);
+
+        if (destinationChainFee == 0) {
+            revert InvalidDestinationFee();
+        }
+
+        ticket.dest = destinationChain;
+        ticket.costs = _transferTokenCosts(destinationChainFee);
```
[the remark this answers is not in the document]
yrong: https://github.com/Snowfork/snowbridge/pull/1155/commits/9f7e99b807bb0f0144110d52917ab375a400893d

contracts/src/Assets.sol
```diff
@@ -174,4 +176,80 @@ library Assets {
...
+        if ($.tokenRegistryByID[tokenID].isRegistered == false) {
+            revert TokenNotRegistered();
+        }
+        return $.tokenRegistryByID[tokenID];
+    }
+
+    // @dev Transfer polkadot native tokens back
+    function sendForeignToken(
+        address executor,
+        TokenInfo memory info,
```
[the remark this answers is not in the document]
yrong: https://github.com/Snowfork/snowbridge/pull/1155/commits/724c649fd63d0f75678aa80ddeb8ad32af84f8e2

contracts/src/AgentExecutor.sol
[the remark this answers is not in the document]
yrong: Reverted in recent commit.

contracts/src/AgentExecutor.sol
yrong: Reverted.

contracts/src/interfaces/IGateway.sol
```diff
@@ -105,4 +114,7 @@ interface IGateway {
         uint128 destinationFee,
         uint128 amount
     ) external payable;
+
+    /// @dev Get tokenInfo by tokenID
+    function getTokenInfo(bytes32 tokenID) external view returns (TokenInfo memory);
```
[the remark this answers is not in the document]
yrong: The user will need to know the address of the foreign token when sending it back to substrate. Optionally could be retrieved from the `ForeignTokenRegistered` event so I removed it from `IGateway`.

contracts/src/Gateway.sol
yrong: https://github.com/Snowfork/snowbridge/pull/1155/commits/fddb9cf3b38ee6c25e4f23fc26daa70e0b9ea73e

musnit (commented):
is it necessary for this PR to be making heavy modifications to our lower-level protocol? (ie, Gateway.sol and AgentExecutor.sol)

Could it not be implemented at a higher level without modifying our core protocol, just in Assets.sol and ERC20.sol?

yrong:
> is it necessary for this PR to be making heavy modifications to our lower-level protocol? (ie, Gateway.sol and AgentExecutor.sol)

Yes. The underlying logic/mechanism for transferring Ethereum native tokens(e.g. WETH) and Polkadot native tokens(e.g. DOT) are totally different. That's why I added these new logics/types separated in a clean way from the previous.

> Could it not be implemented at a higher level without modifying our core protocol, just in Assets.sol and ERC20.sol (potentially on top of Transact)?

We can't use transact to mint ERC20 DOT, it should be a reserve-based transfer from the source chain. That also somehow explains the reason why I'd like to introduce the `SafeCallFilter` in https://github.com/Snowfork/snowbridge/pull/1145#discussion_r1504330936

vgeddes:
> is it necessary for this PR to be making heavy modifications to our lower-level protocol? (ie, Gateway.sol and AgentExecutor.sol)

This "Assets from Polkadot" feature is implemented as commands that are interpreted by the Agent contract representing the source parachain. As such I consider it a higher-level protocol. The lower-level protocol is not affected much.

Our protocol layering isn't split neatly on file boundaries, partly because the Ethereum side of the bridge has a _mostly_ single contract architecture, with functionality delegated to modular libraries where possible. However `Gateway.sol` still contains a lot of support code that is needed by both low and high-level parts of the protocol.

* Gateway.sol: God-contract containing all bridging code
  * Asset.sol: Support module for asset related functionalities
  * AgentExecutor.sol: Code which is executed within an Agent

> Could it not be implemented at a higher level without modifying our core protocol, just in Assets.sol and ERC20.sol _(potentially on top of Transact)_?

contracts/src/AgentExecutor.sol
```diff
@@ -36,4 +50,25 @@ contract AgentExecutor {
...
+    function _registerToken(bytes32 agentID, bytes32 tokenID, string memory name, string memory symbol, uint8 decimals)
+        internal
+    {
+        IERC20 token = new ERC20(name, symbol, decimals);
+        Gateway(msg.sender).registerForeignToken(tokenID, address(token), agentID);
+    }
+
+    /// @dev Mint ERC20 token to `recipient`.
+    function _mintToken(bytes32 tokenID, address recipient, uint256 amount) internal {
+        address token = Gateway(msg.sender).tokenAddressOf(tokenID);
```
alistair-singh, line 64: `token` can be resolved in Gateway and passed to AgentExecutor as a parameter. This will brake the dependency on `msg.sender`.

contracts/src/AgentExecutor.sol
```diff
@@ -36,4 +50,25 @@ contract AgentExecutor {
     function _transferToken(address token, address recipient, uint128 amount) internal {
         IERC20(token).safeTransfer(recipient, amount);
     }
+
+    /// @dev Register native asset from polkadto as ERC20 `token`.
+    function _registerToken(bytes32 agentID, bytes32 tokenID, string memory name, string memory symbol, uint8 decimals)
+        internal
+    {
+        IERC20 token = new ERC20(name, symbol, decimals);
+        Gateway(msg.sender).registerForeignToken(tokenID, address(token), agentID);
```
alistair-singh, line 59: The call stack of registering a token is:
```
Gateway.submitV1 -> AgentExecutor.execute -> AgentExecutor._registerToken -> Gateway. registerForeignToken
```
From what I can see this is done to make sure the owner of ERC20 token is the agent. This can be simplified if we add a `setOwner` method to `ERC20`. This will allow the call stack to be:
```
Gateway.submitV1 -> Gateway.registerForeignToken
```
Where `Gateway.registerForeignToken` creates a new ERC20 to which initially it is the owner, and then assigns ownership to the agent using the `setOwner` function.

This will also allow agentId to removed from the `execute` function signature. This will also probably save gas.

contracts/src/AgentExecutor.sol
```diff
@@ -36,4 +50,25 @@ contract AgentExecutor {
...
+
+    /// @dev Mint ERC20 token to `recipient`.
+    function _mintToken(bytes32 tokenID, address recipient, uint256 amount) internal {
+        address token = Gateway(msg.sender).tokenAddressOf(tokenID);
+        ERC20(token).mint(recipient, amount);
+        emit TokenMinted(tokenID, token, recipient, amount);
+    }
+
+    function burnToken(bytes32 tokenID, address sender, uint256 amount) external {
+        address token = Gateway(msg.sender).tokenAddressOf(tokenID);
```
alistair-singh, line 70: Ditto: pass `token` in from Gateway.sol.

contracts/src/AgentExecutor.sol
```diff
@@ -36,4 +50,25 @@ contract AgentExecutor {
...
+    {
+        IERC20 token = new ERC20(name, symbol, decimals);
+        Gateway(msg.sender).registerForeignToken(tokenID, address(token), agentID);
+    }
+
+    /// @dev Mint ERC20 token to `recipient`.
+    function _mintToken(bytes32 tokenID, address recipient, uint256 amount) internal {
+        address token = Gateway(msg.sender).tokenAddressOf(tokenID);
+        ERC20(token).mint(recipient, amount);
+        emit TokenMinted(tokenID, token, recipient, amount);
```
alistair-singh, line 66: Moving the events to ERC20 makes AgentExecutor simple again in terms of dependencies. It only depends on IERC20 interface.

contracts/src/ERC20.sol
```diff
@@ -0,0 +1,351 @@
...
+ * Additionally, an {Approval} event is emitted on calls to {transferFrom}.
+ * This allows applications to reconstruct the allowance for all accounts just
+ * by listening to these events. Other implementations of the EIP may not emit
+ * these events, as it isn't required by the specification.
+ *
+ * Finally, the non-standard {decreaseAllowance} and {increaseAllowance}
+ * functions have been added to mitigate the well-known issues around setting
+ * allowances. See {IERC20-approve}.
+ */
+contract ERC20 is IERC20, IERC20Permit {
```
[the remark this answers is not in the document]
vgeddes: I prefer `ERC20` at least for the base contract.

We can add a child contract `WrappedToken` that inherits from `ERC20`, with specialized functions for burning and anything else needed.

contracts/src/AgentExecutor.sol
vgeddes: Yeah, I agree that the current flow is overly complicated. We should explore alternatives like the one you suggest.

Though instead of the Gateway passing ownership of the ERC20 token to the Agent, the gateway calls something like:

```solidity
Agent(agent).invoke(AGENT_EXECUTOR, AgentExecutor.createERC20(params...))
```

contracts/src/ERC20.sol
```diff
@@ -0,0 +1,351 @@
...
+     * required by the EIP. See the note at the beginning of {ERC20}.
+     *
+     * Requirements:
+     *
+     * - `sender` and `recipient` cannot be the zero address.
+     * - `sender` must have a balance of at least `amount`.
+     * - the caller must have allowance for ``sender``'s tokens of at least
+     * `amount`.
+     */
+    function transferFrom(address sender, address recipient, uint256 amount) external virtual override returns (bool) {
```
vgeddes, line 153: I think we should move most of these external functions to a common library, to save on deployment costs. Especially since this code is going to be deployed hundreds of times. It will save on bridging fees.

contracts/src/ERC20.sol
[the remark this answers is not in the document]
vgeddes: Actually, I think we should just leave the naming alone for now while we sort out the more important protocol-level issues you raised in other comments.

contracts/src/ERC20.sol
musnit: maybe we should just be deploying simple proxies for each token rather than a full erc20 contract each time? but it's not a big deal i guess, even hundreds of times of redeploys won't cost too much

contracts/src/AgentExecutor.sol
yrong: Vincent, is that something https://github.com/Snowfork/snowbridge/pull/1155#discussion_r1541966351 we tried and then reverted?

contracts/src/AgentExecutor.sol
yrong: https://github.com/Snowfork/snowbridge/pull/1155/commits/28f513d669b7bf4e97ac60e719147694599504c6

contracts/src/AgentExecutor.sol
yrong: https://github.com/Snowfork/snowbridge/pull/1155/commits/9b7811d301f04d2417fd9fad3e4ba1bc1dfc59a5

contracts/src/AgentExecutor.sol
yrong: There is already an low level `Transfer` event in ERC20, the outside event could provide some more context like `tokenID`.

contracts/src/ERC20.sol
yrong: Yeah, maybe we can delegate these advanced features(allowance/permit) to something like https://github.com/Uniswap/permit2

Currently, the ERC20.sol is only 4KB so may not be a big issue.

vgeddes: > maybe we should just be deploying simple proxies for each token rather than a full erc20 contract each time? but it's not a big deal i guess, even hundreds of times of redeploys won't cost too much

Yeah, I was thinking about this yesterday actually. It seems [ERC-6909](https://github.com/jtriley-eth/ERC-6909) is the emerging standard for multi-token contracts (replacing ERC-1155). Its being used in Uniswap-v4 And so I was thinking Gateway can maintain a single ERC-6909 contract, and then instantiate thin ERC20 proxies over it.

I am ambivalent about it.

Positives:
* Significant gas savings
* Simplified protocol as it would no longer involve agents (the contract would be owned by the gateway rather than the agent of the origin parachain).
* The ability to upgrade the multi-token contract, to add new features and fix bugs. This is not really possible in the existing design where we instantiate separate ERC20 contracts.

Negatives:
* Gav really wanted Agents to own the wrapped tokens, as its fits with his particular vision of XCM as a cross-chain protocol. So that may be a blocker.
* ERC-6909 is fairly [opinionated](https://ethereum-magicians.org/t/erc-6909-multi-token-standard/13891/36). For example, it doesn't support [EIP-2612](https://eips.ethereum.org/EIPS/eip-2612) signed approvals, and instead has an approved operator mechanism (similar to ERC-777)

contracts/src/Gateway.sol
```diff
@@ -394,6 +416,49 @@ contract Gateway is IGateway, IInitializable {
         emit PricingParametersChanged();
     }
 
+    // @dev Register a new fungible Polkadot token for an agent
+    function _registerForeignToken(bytes32 agentID, address agent, bytes memory params) internal {
```
vgeddes, line 420: With the new approach, I was thinking that the `AgentExecuteCommand.RegisterToken` command would be replaced by a toplevel `Command.RegisterToken` command that is dispatched in `submitV1`.

Also all these new asset-specific handling functions need to be moved to `Assets.sol` to keep within the contract size limit.

contracts/src/Gateway.sol
```diff
@@ -394,6 +416,49 @@ contract Gateway is IGateway, IInitializable {
...
+        if ($.tokenRegistryByID[tokenID].isRegistered == true) {
+            revert TokenAlreadyRegistered();
+        }
+        ERC20 foreignToken = new ERC20(name, symbol, decimals);
+        address token = address(foreignToken);
+        TokenInfo memory info =
+            TokenInfo({isRegistered: true, isForeign: true, tokenID: tokenID, agentID: agentID, token: token});
+        $.tokenRegistry[token] = info;
+        $.tokenRegistryByID[tokenID] = info;
+        foreignToken.transferOwnership(agent);
```
vgeddes, line 433: Instead of transferring ownership, just pass the agent address to the constructor of the ERC20:

```solidity
public address immutable owner; 

constructor(address _owner, ...) {
  owner = _owner
}
```

This saves gas, and reduces our reliance on OpenZeppelin libraries, which I really don't want to use for various reasons, including:
* I don't like OpenZepellins widespread use of inheritance, I think it makes auditing harder.

contracts/src/Gateway.sol
```diff
@@ -394,6 +416,49 @@ contract Gateway is IGateway, IInitializable {
...
+        emit ForeignTokenRegistered(tokenID, agentID, token);
+    }
+
+    // @dev Mint foreign token from Polkadot
+    function _mintForeignToken(address agent, bytes memory params) internal {
```
vgeddes, line 438: The handler should be moved to Assets.sol to reduce contract size limit

contracts/src/Gateway.sol
yrong: https://github.com/Snowfork/snowbridge/pull/1155/commits/b00538f4ed35464f2bf270433d60e873fd540edd

contracts/src/Gateway.sol
yrong: https://github.com/Snowfork/snowbridge/pull/1155/commits/3da6be486e3c02c96e77583cc3279f7dac270481

contracts/src/AgentExecutor.sol
yrong: https://github.com/Snowfork/snowbridge/pull/1155/commits/79c1ba0f8848df15b15f3642c299e6b336cf2f81

contracts/src/Gateway.sol
yrong: https://github.com/Snowfork/snowbridge/pull/1155/commits/fdf15aa92abe3bd9d35ce8d21f2f62e8f29cd4df and https://github.com/Snowfork/polkadot-sdk/pull/128/commits/fec5a63c8d24669928986fc749e2be5e5b1365b6

contracts/src/ERC20.sol
musnit: can we make a decision here? maybe best to keep it simple and stick with existing standards?

vgeddes: Yes, I am favor the current approach - just brought the idea up in case anyone had strong opinions.

@yrong lets go with the current approach, but with most of `ERC20.sol` factored out into a library `ERC20Lib.sol`

yrong: Vincent, so which functions you'd like to move to a library?

I would assume the contract size may not be a big issue here and will not make a difference for the bridging fees?

 If we do want to slim the contract somehow, I'd prefer to remove all token permit/approve functions to make ERC20.sol minimal. 
 
 For all [EIP-2612](https://eips.ethereum.org/EIPS/eip-2612), [EIP-1271](https://eips.ethereum.org/EIPS/eip-1271) features we delegate to a separate contract like [permit2](https://github.com/Uniswap/permit2). 
 
so WDYT?

contracts/src/Assets.sol
```diff
@@ -174,4 +179,112 @@ library Assets {
...
+        MultiAddress calldata destinationAddress,
+        uint128 destinationChainFee,
+        uint128 amount
+    ) external returns (Ticket memory ticket) {
+        // Polkadot-native token: burn wrapped token
+        _burnToken(executor, agent, info.token, sender, amount);
+
+        if (destinationChainFee == 0) {
+            revert InvalidDestinationFee();
+        }
```
claravanstaden, line 199: Maybe the `revert InvalidDestinationFee();` will revert the burnToken changes, but just curious to why you don't do the `destinationChainFee` check before the burn?

contracts/src/storage/AssetsStorage.sol
```diff
@@ -15,9 +14,13 @@ library AssetsStorage {
...
         // Extra fee for registering a token, to discourage spamming (Ether)
         uint256 registerTokenFee;
+        // Token registry by token address
+        mapping(address token => TokenInfo) tokenRegistry;
+        // Token registry by tokenID
+        mapping(bytes32 tokenID => TokenInfo) tokenRegistryByID;
     }
 
-    bytes32 internal constant SLOT = keccak256("org.snowbridge.storage.assets");
+    bytes32 internal constant SLOT = keccak256("org.snowbridge.storage.assets.v2");
```
claravanstaden, line 23: Why is this called `SLOT`?

contracts/src/upgrades/rococo/GatewayWithAssetStorageV2.sol
```diff
@@ -0,0 +1,39 @@
...
+        LegacyAssetsStorage.Layout storage $ = LegacyAssetsStorage.layout();
+        AssetsStorage.Layout storage $v2 = AssetsStorage.layout();
+
+        $v2.assetHubAgent = $.assetHubAgent;
+        $v2.assetHubParaID = $.assetHubParaID;
+        $v2.assetHubCreateAssetFee = $.assetHubCreateAssetFee;
+        $v2.assetHubReserveTransferFee = $.assetHubReserveTransferFee;
+        $v2.registerTokenFee = $.registerTokenFee;
+        for (uint256 i = 0; i < tokens.length; i++) {
+            $v2.tokenRegistry[tokens[i]].isRegistered = $.tokenRegistry[tokens[i]].isRegistered;
```
claravanstaden, line 36: Why do you just set `isRegistered` and not the other token fields?

contracts/test/Gateway.t.sol
```diff
@@ -50,12 +51,15 @@ import "./mocks/GatewayUpgradeMock.sol";
...
+    event Transfer(address indexed from, address indexed to, uint256 value);
+
+    ParaID public bridgeHubParaID = ParaID.wrap(1013);
+    bytes32 public bridgeHubAgentID = 0x03170a2e7597b7b7e3d84c05391d139a62b157e78786d8c082f29dcf4c111314;
     address public bridgeHubAgent;
 
-    ParaID public assetHubParaID = ParaID.wrap(1002);
-    bytes32 public assetHubAgentID = keccak256("1002");
+    ParaID public assetHubParaID = ParaID.wrap(1000);
+    bytes32 public assetHubAgentID = 0x81c5ab2571199e3188135178f3c2c8e2d268be1313d029b30f534fa579b69b79;
```
claravanstaden, line 62: Where do you get these values? Just for reference if this need to be updated?

contracts/test/Gateway.t.sol
```diff
@@ -909,4 +919,53 @@ contract GatewayTest is Test {
...
+            name: "DOT",
+            symbol: "DOT",
+            decimals: 10
+        });
+
+        vm.expectEmit(true, true, false, false);
+        emit IGateway.ForeignTokenRegistered(bytes32(uint256(1)), assetHubAgentID, address(0));
+
+        GatewayMock(address(gateway)).registerForeignTokenPublic(abi.encode(params));
+    }
```
claravanstaden, line 936: Some negative tests would be good to add.

claravanstaden (commented):
Very nice! 👏🏻

contracts/src/Gateway.sol
```diff
@@ -265,11 +283,16 @@ contract Gateway is IGateway, IInitializable {
...
-        (bool success, bytes memory returndata) = Agent(payable(agent)).invoke(AGENT_EXECUTOR, call);
-        if (!success) {
-            revert AgentExecutionFailed(returndata);
+        if (command == AgentExecuteCommand.MintToken) {
+            (bytes32 tokenID, address recipient, uint256 amount) =
+                abi.decode(commandParams, (bytes32, address, uint256));
+            Assets.mintForeignToken(AGENT_EXECUTOR, agent, tokenID, recipient, amount);
+        } else if (command == AgentExecuteCommand.TransferToken) {
+            (address token, address recipient, uint128 amount) = abi.decode(commandParams, (address, address, uint128));
+            Assets.transferToken(AGENT_EXECUTOR, agent, token, recipient, amount);
```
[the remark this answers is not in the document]
yrong: The token address is unknown on the substrate side. 

Usually, token in Parachain is represented by the xcm `Location`, we convert it to a `tokenID` and save the mapping(tokenID, Location) on BH.

When the token is created/registered as ERC20, we save the mapping(tokenAddress,tokenID) on Ethereum storage.

Finally, when the token is sent back from Ethereum by address, we convert it to the `tokenID` and on BH map to the original `Location` which will be used to identify the token on the original parachain.

contracts/src/Gateway.sol
```diff
@@ -265,11 +283,16 @@ contract Gateway is IGateway, IInitializable {
...
         }
 
-        bytes memory call = abi.encodeCall(AgentExecutor.execute, params.payload);
+        (AgentExecuteCommand command, bytes memory commandParams) =
+            abi.decode(params.payload, (AgentExecuteCommand, bytes));
 
-        (bool success, bytes memory returndata) = Agent(payable(agent)).invoke(AGENT_EXECUTOR, call);
-        if (!success) {
-            revert AgentExecutionFailed(returndata);
+        if (command == AgentExecuteCommand.MintToken) {
```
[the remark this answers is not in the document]
yrong: `TransferToken` is the previous impl for sending the Ethereum native token(e.g. WETH) from substrate to Ethereum.

contracts/src/Assets.sol
yrong: https://github.com/Snowfork/snowbridge/pull/1155/commits/c2adf8f219d90407de1bafcad1ebaab5cf9617fb

contracts/src/Gateway.sol
```diff
@@ -428,9 +458,36 @@ contract Gateway is IGateway, IInitializable {
...
+        if (!info.isRegistered) {
+            revert TokenNotRegistered();
+        }
+        if (info.isForeign) {
+            address agent = _ensureAgent(info.agentID);
+            _submitOutbound(
+                Assets.sendForeignToken(
+                    agent,
+                    AGENT_EXECUTOR,
+                    info,
```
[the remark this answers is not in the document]
yrong: https://github.com/Snowfork/snowbridge/pull/1155/commits/b0b582dd38096f9074e4f085c79cda2934fbee40

contracts/src/Types.sol
```diff
@@ -84,11 +84,13 @@ enum Command {
     SetOperatingMode,
     TransferNativeFromAgent,
     SetTokenTransferFees,
-    SetPricingParameters
+    SetPricingParameters,
+    RegisterForeignToken
 }
 
 enum AgentExecuteCommand {
```
vgeddes, line 91: I'm not sure its worth having an `AgentExecuteCommand` enum anymore. `TransferToken` and `MintToken` could just be normal commands.

contracts/src/Gateway.sol
```diff
@@ -265,11 +283,16 @@ contract Gateway is IGateway, IInitializable {
             revert InvalidAgentExecutionPayload();
         }
 
-        bytes memory call = abi.encodeCall(AgentExecutor.execute, params.payload);
+        (AgentExecuteCommand command, bytes memory commandParams) =
```
vgeddes, line 286: Similar to my other recent comment, there isn't much sense in having an `AgentExecuteCommand` enum anymore. Instead, mintToken and TransferToken are just normal commands like `upgrade`, etc.

contracts/src/ERC20.sol
vgeddes: As per this permit2 [explainer](https://github.com/dragonfly-xyz/useful-solidity-patterns/tree/main/patterns/permit2), permit2 is really meant to be compatibility layer for older tokens that don't support EIP-2612. Its not meant to be a replacement for it.

And the user still has to manually approve the permit2 contract to spend the tokens. So there is an initial setup step that detracts from the UX experience.

vgeddes: I don't think it will take long to introduce `ERC20Lib.sol`. The hardest part will be moving the token storage into a struct like https://github.com/aragon/zeppelin-solidity/blob/23540364330f6b109980933c032238dd4d1308d1/contracts/token/ERC20Lib.sol#L8

After that's done, I would suggest moving `_update` into the library first, as its the largest function

contracts/src/Types.sol
yrong: Yeah, I think so and will refactor accordingly.

contracts/test/Gateway.t.sol
yrong: It's derived from an xcm Location with `AgentIdOf` on BH.  Yeah not necessary for a test but I'd prefer to make it consistent with the real one.

contracts/src/upgrades/rococo/GatewayWithAssetStorageV2.sol
yrong: For Ethereum native tokens(e.g. WETH) there is no value in other fields.

contracts/src/storage/AssetsStorage.sol
yrong: Just by convention as each contract has a finite amount of storage space, which is divided into 32-byte slots.

contracts/src/Types.sol
yrong: https://github.com/Snowfork/snowbridge/pull/1155/commits/b2a666ac836fd926e087580ace88acb98e58a0d4

contracts/src/Gateway.sol
yrong: https://github.com/Snowfork/snowbridge/pull/1155/commits/b2a666ac836fd926e087580ace88acb98e58a0d4

contracts/test/Gateway.t.sol
yrong: https://github.com/Snowfork/snowbridge/pull/1155/commits/6148c038b97c2a26982dd0825576438d73a21b44

contracts/src/Types.sol
```diff
@@ -76,19 +76,17 @@ enum OperatingMode {
 
 /// @dev Messages from Polkadot take the form of these commands.
 enum Command {
-    AgentExecute,
+    TransferToken,
```
vgeddes, line 79: Just realized this change is going to break compatibility, causing messages in the old format to fail processing.

So we need to some kind of solution to allow old messages and new messages to be processed.

musnit: could it be simpler to just stick to the old message format overall, even for this work?

yrong: https://github.com/Snowfork/snowbridge/pull/1155/commits/fcd7539afd1d7d36b1a5b3d2bf53b49c54fe956a and https://github.com/Snowfork/polkadot-sdk/pull/128/commits/40cc470c2d6fc2c6995e2e18b03d13b773f34664

Add `AgentExecuteCommand` back for compatibility. Meanwhile, make `TransferToken` also a top-level command so we can deprecate `AgentExecuteCommand` in later upgrading.

yrong:
### Fee Section from penpal to Ethereum

- User represents a user who kicks off an extrinsic on the parachain.
- Parachain represents the source parachain, its sovereign or its agent depending on context.

Sequence|Where|Who|What
-|-|-|-
1|Penpal|User| For `reserve_transfer_assets` pays(DOT, Native) to node to execute custom extrinsic; pays (DOT) to Treasury for both delivery cost on BH and execution cost on Ethereum(i.e. `EthereumBaseFee`).
2|Bridge Hub|Parachain|Pays(DOT) to Treasury Account for delivery(local fee), pays(DOT) to Parachain sovereign for delivery(remote fee), essentially a refund. Remote fee converted to ETH here.
3|Gateway|Relayer|pays(ETH) to validate and execute message.
4|Gateway|Parachain Agent|pays(ETH) to relayer for delivery(reward+refund) and execution.

yrong:
### Fee section from Ethereum to Penpal

- dApp is represents `msg.sender` or its sovereign depending on context.
- Parachain represents the target parachain, its sovereign or its agent depending on context.
- Ethereum Sovereign represents `Location{parent:2,interior:[GlobalConsensus(Ethereum)]}`

Sequence|Where|Who|What
-|-|-|-
1|Gateway|dApp|pays(ETH, converted to DOT here) Parachain Agent for both delivery cost on BH and  execution cost on destination(DOT,Native).
2|Bridge Hub|Relayer|pays(DOT) node for execution
3|Bridge Hub|Parachain Sovereign|pays(DOT) Relayer for delivery (refund+reward)
4|Parachain|Ethereum Sovereign|pays(DOT, Native) for execution only.

contracts/src/storage/AssetsStorage.sol
```diff
@@ -6,7 +6,6 @@ import {TokenInfo, ParaID} from "../Types.sol";
 
 library AssetsStorage {
     struct Layout {
-        mapping(address token => TokenInfo) tokenRegistry;
```
vgeddes, line 9: oops, we can't move storage fields, otherwise the upgraded gateway's code will be confused about the storage layout.

contracts/src/Types.sol
```diff
@@ -107,4 +111,5 @@ struct Ticket {
 
 struct TokenInfo {
     bool isRegistered;
```
vgeddes, line 113: To ensure storage is not corrupted after upgrades, we need to preserve the storage layou, only adding new fields at the end.

```suggestion
    bool isRegistered;
    bytes31 __padding;
```

contracts/src/storage/LegacyAssetsStorage.sol
vgeddes, line 1: i think we should preserve the existing storage layout in `AssetStorage.sol` rather than doing a full storage migration. The gas cost of migrating storage is unbounded (many tokens registered) so that could introduce errors.

We can however add new fields to the end of the `Layout` struct in `AssetStorage.sol`.

yrong: https://github.com/Snowfork/snowbridge/pull/1155/commits/64b9ef4828f2341495998d2d8a33dd05b073fabd

yrong: Removed the storage migration. The initial concern in https://github.com/Snowfork/snowbridge/pull/1155#discussion_r1541846907 is about the `TokenInfo` structure we changed. 

@vgeddes I'd assume it's safe to add the new `foreignID` without overwritten the exist storage, but not 100% sure. Please help to double check.

yrong: https://github.com/Snowfork/snowbridge/pull/1155/commits/cf596f3d840b6348df625d384e35fec8f7256b3a

vgeddes (approved):

PR #1156, opened 2024-03-13
Limit finalized header gap

claravanstaden:
Adds the following:

- an interim finalized update between sync committee updates if the headers are more than 1982 slots apart
- a MySQLite datastore to store metadata and ssz blobs on disk. Prunes after X number of states.
- command to store ssz states on disk - should be configured to run with cron twice a day or so
- extract interfaces so that we can easier write tests and get coverage up for the relayer. While the relayer is not part of our on-chain code, it is vital to our bridge and can stall the bridge if buggy


relayer/relays/beacon/header/header.go
```diff
@@ -116,6 +120,21 @@ func (h *Header) SyncCommitteePeriodUpdate(ctx context.Context, period uint64) e
...
 	}
 
+	// If the gap between the last two finalized headers is more than the sync committee period, sync an interim
+	// finalized header
+	validGapSlot := h.cache.Finalized.LastSyncedSlot + (h.slotsInEpoch * h.epochsPerSyncCommitteePeriod)
+	if validGapSlot < uint64(update.Payload.FinalizedHeader.Slot) {
+		err, finalizedUpdate := h.syncer.GetFinalizedUpdateAtAttestedSlot(validGapSlot)
+		if err != nil {
+			return fmt.Errorf("fetch finalized update at slot: %w", err)
+		}
```
yrong, line 130: Awesome!

relayer/relays/beacon/header/header.go
```diff
@@ -357,7 +383,12 @@ func (h *Header) populateClosestCheckpoint(slot uint64) (cache.Proof, error) {
 		}
 		err := h.populateFinalizedCheckpoint(checkpointSlot)
 		if err != nil {
-			return cache.Proof{}, fmt.Errorf("populate closest checkpoint: %w", err)
+			// 4513792 finalized slot
+			attestedHeaderSlot := checkpointSlot + (32 * 2.5)
```
yrong, line 387: Just curious about the magic number here.

relayer/relays/beacon/header/syncer/syncer.go
```diff
@@ -447,6 +448,121 @@ func (s *Syncer) GetHeaderUpdate(blockRoot common.Hash, checkpoint *cache.Proof)
...
+		return update, fmt.Errorf("fetch block: %w", err)
+	}
+
+	// Finalized header proof
+	stateTree, err := beaconState.GetTree()
+	if err != nil {
+		return update, fmt.Errorf("get state tree: %w", err)
+	}
+	_ = stateTree.Hash() // necessary to populate the proof tree values
+	finalizedHeaderProof, err := stateTree.Prove(FinalizedCheckpointGeneralizedIndex)
```
yrong, line 497: Cool! A silly question how to quickly calculate the `GeneralizedIndex` for a specific field in beacon state, like the `FinalizedCheckpointGeneralizedIndex` 105 here?

relayer/relays/beacon/header/header.go
claravanstaden: Just hardcoded for now to test, I'll change it. 😄

relayer/relays/beacon/header/syncer/syncer.go
claravanstaden: I think it can be calculated using the beacon state tree, but since it is well-known and won't change it is usually hardcoded, like here in the Ethereum spec: https://github.com/ethereum/consensus-specs/blob/dev/specs/altair/light-client/sync-protocol.md#constants

relayer/relays/beacon/store/datastore.go
yrong, line 1: Assuming we have configured a fallback client which can fetch the beacon state for arbitrary slots. Do we still need to store the beacon state in a local db here? Is that for another kind of fallback?

claravanstaden: I have removed the fallback client since I encountered similar issues with Prysm that we are having with Lodestar. I think fundamentally relying on the beacon node and its config is risky. I think the beacon store adds a layer of reassurance that we'll be able to get the proofs we need if our bridge goes offline for an extended period.

relayer/chain/parachain/writer.go
```diff
@@ -5,15 +5,30 @@ import (
...
+	WriteToParachainAndRateLimit(ctx context.Context, extrinsicName string, payload ...interface{}) error
+	WriteToParachainAndWatch(ctx context.Context, extrinsicName string, payload ...interface{}) error
+	GetLastFinalizedHeaderState() (state.FinalizedHeader, error)
+	GetFinalizedStateByStorageKey(key string) (scale.BeaconState, error)
+	GetLastBasicChannelBlockNumber() (uint64, error)
+	GetLastBasicChannelNonceByAddress(address common.Address) (uint64, error)
+	GetFinalizedHeaderStateByBlockRoot(blockRoot types.H256) (state.FinalizedHeader, error)
+	GetCompactExecutionHeaderStateByBlockHash(blockHash types.H256) (state.CompactExecutionHeaderState, error)
+	GetLastFinalizedStateIndex() (types.U32, error)
+	GetFinalizedBeaconRootByIndex(index uint32) (types.H256, error)
```
claravanstaden, line 29: Extract interfaces that we can mock out in tests.

relayer/chain/parachain/writer.go
```diff
@@ -331,6 +247,105 @@ func (wr *ParachainWriter) GetFinalizedHeaderStateByBlockRoot(blockRoot types.H2
...
+		return nil, err
+	}
+
+	c, err := types.NewCall(meta, extrinsicName, payload...)
+	if err != nil {
+		return nil, err
+	}
+	return &c, nil
+}
+
```
claravanstaden, line 348: Just reordered the private methods.

relayer/relays/beacon/header/header.go
```diff
@@ -319,7 +352,42 @@ func (h *Header) populateClosestCheckpoint(slot uint64) (cache.Proof, error) {
...
+		historicState, err := h.findClosestCheckPoint(slot)
+		if err != nil {
+			return 0, fmt.Errorf("get history finalized header for the checkpoint: %w", err)
+		}
+		checkpointSlot = historicState.BeaconSlot
+	} else {
+		// Setting the checkpoint slot to what is the latest finalized header on-chain, since the checkpoint should
+		// not be after the latest finalized header on-chain
+		checkpointSlot = lastFinalizedHeaderState.BeaconSlot
+	}
```
yrong, line 379: Seems if-else branch duplicated here.

relayer/relays/beacon/header/header.go
```diff
@@ -319,7 +352,42 @@ func (h *Header) populateClosestCheckpoint(slot uint64) (cache.Proof, error) {
...
+	err = h.populateFinalizedCheckpoint(checkpointSlot)
+	if err != nil {
+		return 0, fmt.Errorf("populated local cache with finalized header found on-chain: %w", err)
+	}
+
+	return 0, nil
+}
+
 func (h *Header) getHeaderUpdateBySlot(slot uint64) (scale.HeaderUpdatePayload, error) {
+	slot = slot + 1
```
yrong, line 390: Why do we add `slot+1` here?  Is that tested with the `register_token`  smoke test still worked?

relayer/relays/beacon/header/header.go
```diff
@@ -386,3 +457,40 @@ func (h *Header) isInitialSyncPeriod() bool {
 	lastFinalizedPeriod := h.syncer.ComputeSyncPeriodAtSlot(h.cache.Finalized.LastSyncedSlot)
 	return initialPeriod == lastFinalizedPeriod
 }
+
+func (h *Header) findClosestCheckPoint(slot uint64) (state.FinalizedHeader, error) {
```
yrong, line 461: IIUC it's not closest anymore instead the most distant. So the previous name `findCheckPointBackward` make more sense here?

vgeddes (approved):

relayer/relays/beacon/header/header.go
claravanstaden: I don't understand, apologies?

relayer/relays/beacon/header/header.go
claravanstaden: Good catch 😄 https://github.com/Snowfork/snowbridge/pull/1156/commits/0da2b16cf053b4cbc15baab874f93c7271cab0ae

relayer/relays/beacon/header/header.go
claravanstaden: Renamed here https://github.com/Snowfork/snowbridge/pull/1156/commits/b5334dc6aeb077546b47bb722f87c073256281b8.

relayer/relays/beacon/header/header.go
```diff
@@ -276,33 +330,12 @@ func (h *Header) populateClosestCheckpoint(slot uint64) (cache.Proof, error) {
 	case errors.Is(cache.FinalizedCheckPointNotAvailable, err) || errors.Is(cache.FinalizedCheckPointNotPopulated, err):
 		checkpointSlot := checkpoint.Slot
 		if checkpointSlot == 0 {
-			log.WithFields(log.Fields{"calculatedCheckpointSlot": checkpointSlot}).Info("checkpoint slot not available, try with slot in next sync period instead")
-			checkpointSlot = h.syncer.CalculateNextCheckpointSlot(slot)
-			lastFinalizedHeaderState, err := h.writer.GetLastFinalizedHeaderState()
+			checkpointSlot, err = h.populateCheckPointCacheWithDataFromChain(slot, checkpointSlot)
```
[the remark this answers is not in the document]
claravanstaden: That's a good point! https://github.com/Snowfork/snowbridge/pull/1156/commits/de118ca2e46827fe4b03a264fb594aab0219308a

relayer/relays/beacon/header/header.go
yrong: Sorry, I'm looking in the wrong. Just ignore it!

yrong (approved):
+1

PR #1160, opened 2024-03-19
Add bidirectional mapping storage for agent address/ID

yrong:


contracts/src/Gateway.sol
```diff
@@ -592,6 +593,7 @@ contract Gateway is IGateway, IInitializable {
         // Initialize agent for BridgeHub
         address bridgeHubAgent = address(new Agent(BRIDGE_HUB_AGENT_ID));
         core.agents[BRIDGE_HUB_AGENT_ID] = bridgeHubAgent;
+        core.agentAddresses[bridgeHubAgent] = BRIDGE_HUB_AGENT_ID;
```
alistair-singh, line 596: How are we going to initialize this storage on Rococo?

alistair-singh (approved):
Looks good.

contracts/src/Gateway.sol
yrong: I think we can make an upgrade with a custom `initialize` function.

PR #1161, opened 2024-03-20
Add fee multiplier to protocol

vgeddes:
Since our system is currently governed by static exchange rates, we need to add a safety buffer to the calculated fees.

This ensures that if exchange rates or gas prices fluctuate, users pay enough fees to cover costs.

Companion PR: https://github.com/Snowfork/polkadot-sdk/pull/127


claravanstaden (approved):

yrong (approved):
+1

contracts/test/Gateway.t.sol
```diff
@@ -875,14 +878,19 @@ contract GatewayTest is Test {
...
+                    exchangeRate: exchangeRate.mul(convert(2)),
+                    multiplier: multiplier.mul(convert(2)),
+                    deliveryCost: outboundFee
+                })
             )
         );
+        // Should expect 4x fee increase
         fee = IGateway(address(gateway)).quoteRegisterTokenFee();
-        assertEq(fee, 10000000000000000);
+        assertEq(fee, 40000000000000003);
```
alistair-singh, line 893: I checked the math and I am unsure where is the extra `3` comming from?

contracts/src/Gateway.sol
```diff
@@ -462,19 +463,24 @@ contract Gateway is IGateway, IInitializable {
...
     // Calculate the fee for accepting an outbound message
     function _calculateFee(Costs memory costs) internal view returns (uint256) {
         PricingStorage.Layout storage pricing = PricingStorage.layout();
-        return costs.native + _convertToNative(pricing.exchangeRate, pricing.deliveryCost + costs.foreign);
+
+        // The exchange rate and amount are both multiplied by `multiplier` to add a safety margin
+        // * `exchangeRate` is subject to real-world fluctuations
+        // * `amount` includes XCM execution fees in Polkadot that are also subject to fluctuations
+        UD60x18 exchangeRate = pricing.multiplier.mul(pricing.exchangeRate);
+        UD60x18 amount = pricing.multiplier.mul(convert(pricing.deliveryCost + costs.foreign));
```
alistair-singh, line 481: Since we do not have separate multipliers for `amount` and `exchangeRate` and apply the same multiplier to both, would it be better to only apply the multiplier once.

 In `_convertToNative` we essentially do `amount * exchangeRate`. 
So when we apply the multiplier `m` yields the substitution:
`(m * amount) * (m * exchangeRate)`
Or simplified:
`m^2 * amount * exchangeRate`

This makes the calculation nonlinear because of the multiplier squared. Setting the multiplier to `3` will have the effect of multiplying by `9` on the final value.

Would it not be simpler to make this linear by just scaling the result of `_convertToNative(exchangeRate, amount)` by the multiplier once. This will make this parameter easy to reason about and harder to configure incorrectly.

This also looks like how it is implemented on the other side of the bridge:
https://github.com/Snowfork/polkadot-sdk/pull/127/files#diff-80bccd165d72d122235ef3d186d56f161fcaa5ce73e2a8215987deefc070a2eaR383-R385

contracts/test/Gateway.t.sol
yrong: Seems some nuance of precision loss from `_convertToNative` but does not hurt here.

contracts/src/Gateway.sol
yrong: IIUC the multiplier will be 1.1 or something, so 1.1(for weight)*1.1(for exchangeRate)=1.21, with the extra 21% we're safe.

yrong: But do we also need to take the fluctuation of `fee_per_gas` into consideration? If so we need to apply the multiplier to https://github.com/Snowfork/polkadot-sdk/blob/c2936d77a4ee208d06d85f5c4df3c2d919b25d97/bridges/snowbridge/pallets/outbound-queue/src/lib.rs#L373-L378

contracts/test/Gateway.t.sol
vgeddes: yeah, I think its do precision loss, and as Ron says its inconsequential

contracts/src/Gateway.sol
vgeddes: > Alistair: Since we do not have separate multipliers for amount and exchangeRate and apply the same multiplier to both, would it be better to only apply the multiplier once.

To echo what Ron says, both the Substrate WeightToFee (and other AssetHub/BridgeHub costs) and the DOT/ETH exchange rate are subject to fluctuations. So we need to multiply them both by the multiplier.

> Ron: Do we need a separate multiplier for the price params on ethereum side?
> For example, take the fluctuation of fee_per_gas into consideration. Then we also need to apply the multiplier to

The `multiplier`parameter is only meant to be used for exchange rates, and substrate execution costs.

The gas price (FeePerGas) is very volatile, and so instead, as I've updated the docs to represent, `PricingParameters.FeePerGas` is now the *maximum* amount of gas  we're willing to compensate relayers for. So you can say that `PricingParameters.FeePerGas` already has a multiplier applied.

contracts/src/Gateway.sol
```diff
@@ -613,6 +621,7 @@ contract Gateway is IGateway, IInitializable {
         PricingStorage.Layout storage pricing = PricingStorage.layout();
         pricing.exchangeRate = config.exchangeRate;
         pricing.deliveryCost = config.deliveryCost;
+        pricing.multiplier = config.multiplier;
```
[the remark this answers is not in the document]
vgeddes: Good question. I copied the technique by @yrong and added this upgraded contract for Rococo: https://github.com/Snowfork/snowbridge/pull/1161/commits/111ee1e7fdb6d4e48b79aa31578e8174c3445dda

alistair-singh (commented):
+1

alistair-singh (approved):

PR #1163, opened 2024-03-25
Generate preimage for bridge initialization

vgeddes:
Run the following command to generate an initialization preimage:

```
snowbridge-control initialize <PARAMS>
```


control/preimage/src/main.rs
```diff
@@ -71,70 +151,115 @@ fn parse_hex_bytes(v: &str) -> Result<Bytes, String> {
...
+        Command::Initialize(params) => {
+            let (set_pricing_parameters, set_ethereum_fee) = commands::pricing_parameters(&context, &params.pricing_parameters).await?;
+            let call1 = send_xcm_bridge_hub(&context, vec![
+                commands::force_checkpoint(&params.force_checkpoint),
+                commands::gateway_operating_mode(&params.gateway_operating_mode),
+                commands::set_gateway_address(&params.gateway_address),
+                set_pricing_parameters,
+            ]).await?;
+            let call2 = send_xcm_asset_hub(&context, vec![set_ethereum_fee]).await?;
+            utility_batch(vec![call1, call2])
```
yrong, line 248: IIUC we also need to set the xcm version of Ethereum in AH or the `reserve_transfer_assets` will fail. 

Codes for that in https://github.com/Snowfork/polkadot-sdk/blob/5da7df7a823a88b3ff93bdc612b6e2006f30bb96/cumulus/parachains/integration-tests/emulated/tests/bridges/bridge-hub-rococo/src/tests/snowbridge.rs#L380-L383 and local setup in
https://github.com/Snowfork/snowbridge/blob/cc6f1236088dfba23bec54f8a7882d218f261a62/web/packages/test/scripts/configure-substrate.sh#L82-L85

vgeddes: Is this just for local E2E? As AssetHub is already running in Polkadot though for many years, and presumably already has XCM configured.

yrong: Not sure but it did require an entry exist in storage `supported_versions`, just checked `asset-hub` on Rococo testnet
image

alistair-singh: I had to set this in Rococo for the bridge to work. This is not a generic xcm configuration information, this sets the XCM version for our bridge location specifically. We need to add it to the preimage.

alistair-singh (changes requested):
We need to set the xcm version for the bridge.

alistair-singh (approved):
+1

control/preimage/src/helpers.rs
```diff
@@ -89,8 +168,12 @@ pub async fn query_weight(api: &OnlineClient<PolkadotConfig>, call: BridgeHubRun
     Ok((call_info.weight.ref_time, call_info.weight.proof_size))
 }
 
-pub async fn query_para_id(api: &OnlineClient<PolkadotConfig>) -> Result<u32, Box<dyn std::error::Error>> {
-    let storage_query = bridge_hub_runtime::storage().parachain_info().parachain_id();
+pub async fn query_para_id(
```
claravanstaden, line 171:
```suggestion
pub async fn query_para_id_bridge_hub(
```

control/preimage/src/main.rs
```diff
@@ -22,119 +26,242 @@ struct Cli {
...
+
+#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Debug, ValueEnum)]
+pub enum GatewayOperatingModeEnum {
+    Normal,
+    RejectingOutboundMessages,
+}
+
+#[derive(Debug, Args)]
+pub struct GatewayAddressArgs {
+    /// Path to JSON file containing checkpoint
```
claravanstaden, line 91:
```suggestion
    /// Address of the contract on Ethereum
```

control/preimage/src/main.rs
```diff
@@ -22,119 +26,242 @@ struct Cli {
...
 }
 
 struct Context {
     api: Box<OnlineClient<PolkadotConfig>>,
+    asset_hub_api: Box<OnlineClient<PolkadotConfig>>,
 }
 
+#[cfg(feature = "rococo-local")]
+static CONFIG: StaticConfig<'static> = StaticConfig {
+    api: "ws://127.0.0.1:11144",
```
claravanstaden, line 184:
```suggestion
    bridge_hub_api: "ws://127.0.0.1:11144",
```

claravanstaden (approved):
Some minor nits, looks good!

control/preimage/src/helpers.rs
```diff
@@ -101,3 +184,71 @@ pub async fn query_para_id(api: &OnlineClient<PolkadotConfig>) -> Result<u32, Bo
...
+        .fetch(&storage_query)
+        .await?
+        .expect("parachain id not set");
+
+    Ok(asset_hub_para_id.0)
+}
+
+pub fn utility_batch(calls: Vec<RelayRuntimeCall>) -> RelayRuntimeCall {
+    RelayRuntimeCall::Utility(
+        crate::relay_runtime::runtime_types::pallet_utility::pallet::Call::batch { calls },
```
yrong, line 205: Is the [batch_all](https://github.com/paritytech/polkadot-sdk/blob/994003854363b3c5fd0d7343f93aa1e54edb1ad0/substrate/frame/utility/src/lib.rs#L323) make more sense here?

control/runtimes/bridge-hub-polkadot/src/lib.rs
```diff
@@ -0,0 +1,15 @@
+#[subxt::subxt(
+    runtime_metadata_path = "bridge-hub-metadata.bin",
+    derive_for_all_types = "Clone",
+    substitute_type(
+        path = "snowbridge_beacon_primitives::updates::CheckpointUpdate",
+        with = "::subxt::utils::Static<::snowbridge_beacon_primitives::updates::CheckpointUpdate<512>>",
```
yrong, line 6: Interesting, is that because of the generic parameters `COMMITTEE_SIZE` which `subxt` can't infer?

yrong (approved):
Cool!

control/preimage/src/constants.rs
```diff
@@ -0,0 +1,27 @@
...
+#[cfg(feature = "kusama")]
+pub const POLKADOT_SYMBOL: &str = "KSM";
+
+#[cfg(feature = "polkadot")]
+pub const POLKADOT_SYMBOL: &str = "DOT";
+
+// Weights copied from https://github.com/paritytech/polkadot-sdk/blob/master/cumulus/parachains/runtimes/bridge-hubs/bridge-hub-rococo/src/weights/snowbridge_pallet_outbound_queue.rs
+// Added roughly 50%  buffer
+pub const PROCESS_MESSAGE_WEIGHT: (u64, u64) = (100_000_000, 5000);
+pub const COMMIT_SINGLE_MESSAGE_WEIGHT: (u64, u64) = (20_000_000, 2000);
```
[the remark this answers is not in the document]
vgeddes: That is now obsolete code - instead the tool now calls this API to estimate how much weight an extrinsic will consume: https://paritytech.github.io/polkadot-sdk/master/pallet_transaction_payment_rpc_runtime_api/trait.TransactionPaymentCallApi.html#method.query_call_info

control/preimage/src/commands.rs
```diff
@@ -1,44 +1,166 @@
...
+            ),
+        },
+    };
+
+    let outbound_delivery_fee = calculate_delivery_fee(&pricing_params).await?;
+
+    let total_outbound_fee = outbound_delivery_fee.local + outbound_delivery_fee.remote;
+
+    // Adjust outbound fee up by 10% as a buffer
+    let total_outbound_fee_adjusted = total_outbound_fee.saturating_add(total_outbound_fee / 10);
```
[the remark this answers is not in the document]
vgeddes: Yes, I think its enough, let me explain:

The outbound fee needs to cover both:
1. Local costs of processing of BridgeHub,
2. Cost of delivery to Ethereum, which is substantially more than (1)

Component (1) can fluctuate in ways that are outside of our control, for example, updated weights or changes to substrate that result in fee changes.

Component (2) is under our control, as long as we make sure to update the fee on AssetHub when changing the pricing parameters on BridgeHub.

This is why I think we can go with a reduced buffer, since component (1) is much much less than component (2)

control/preimage/src/helpers.rs
vgeddes: thanks, though this is now obsolete code.

control/preimage/src/helpers.rs
vgeddes: Turns out they use `force_batch` for OpenGov proposals, so I changed to that.

control/preimage/src/main.rs
vgeddes: Thanks, made that change.

PR #1164, opened 2024-03-27
Manual sync committee update

claravanstaden:
- If the sync committee update data retrieval fails (most likely because a beacon state is not available), build an update using the standard Beacon API endpoints.
- Splits the `protocol` settings and methods into its own package.
- Adds tests for the beacon data store.
- First try to get an exact beacon state matching the slot provided, and if it can't be found, provide the best next update


relayer/relays/beacon/header/header.go
```diff
@@ -63,6 +60,15 @@ func (h *Header) Sync(ctx context.Context, eg *errgroup.Group) error {
...
 	h.cache.AddCheckPointSlots([]uint64{lastFinalizedHeaderState.BeaconSlot})
 
+	// Special handling here for the initial checkpoint to sync the next sync committee which is not included in initial
+	// checkpoint.
+	if h.isInitialSyncPeriod() {
+		err = h.SyncCommitteePeriodUpdate(ctx, latestSyncedPeriod)
+		if err != nil {
+			return fmt.Errorf("sync next committee for initial sync period: %w", err)
+		}
+	}
```
claravanstaden, line 70: I moved this special case earlier. [This check](https://github.com/Snowfork/snowbridge/blob/main/relayer/relays/beacon/header/header.go#L172 ) will cause the initial checkpoint's next sync committee only to be synced when the period rolls over. I think it is better to sync it immediately when it is available.

relayer/relays/beacon/header/syncer/syncer.go
```diff
@@ -133,7 +154,16 @@ func (s *Syncer) GetSyncCommitteePeriodUpdate(from uint64) (scale.Update, error)
...
 	blockRootsProof, err := s.GetBlockRoots(uint64(finalizedHeader.Slot))
 	if err != nil {
-		return scale.Update{}, fmt.Errorf("fetch block roots: %w", err)
+		beaconStateData, err := s.store.GetBeaconStateData(uint64(finalizedHeader.Slot))
+		if err != nil {
+			return scale.Update{}, fmt.Errorf("fetch beacon state for block roots proof: %w", err)
+		}
+		beaconState, err := s.unmarshalBeaconState(uint64(finalizedHeader.Slot), beaconStateData)
+
+		blockRootsProof, err = s.GetBlockRootsFromState(beaconState)
```
claravanstaden, line 163: If the state is not available, check if it perhaps available in the beacon store.

claravanstaden (commented):
Apologies for the large change - the biggest cause of the change is moving the protocol package out.

relayer/relays/beacon/header/syncer/syncer.go
```diff
@@ -493,55 +523,127 @@ func (s *Syncer) unmarshalBeaconState(slot uint64, data []byte) (state.BeaconSta
...
+
+func (s *Syncer) GetFinalizedUpdateWithSyncCommittee(syncCommitteePeriod, lastFinalizedSlot uint64) (scale.Update, error) {
+	slot := (syncCommitteePeriod) * s.protocol.Settings.SlotsInEpoch * s.protocol.Settings.EpochsPerSyncCommitteePeriod
+
+	// If the latest finalized slot is equal or larger than the slot to be synced, the finalized header
+	// won't be imported and so it should be a newer header. 3 epochs further, because the finalized header is
+	// 2 epochs behind the attested header, and we need 1 epoch newer at least
+	if lastFinalizedSlot >= slot {
+		slot += s.protocol.Settings.SlotsInEpoch * 3
+	}
```
yrong, line 627: Seems not correct,  I'd assume the logic here is to advance `slot` by at least 3 epochs or more until `slot > lastFinalizedSlot`?

relayer/relays/beacon/header/syncer/syncer.go
```diff
@@ -493,55 +523,127 @@ func (s *Syncer) unmarshalBeaconState(slot uint64, data []byte) (state.BeaconSta
 	return beaconState, nil
 }
 
-// Sanity check the finalized and attested header are at 32 boundary blocks so we can download the beacon state
-func (s *Syncer) findAttestedAndFinalizedHeadersAtBoundary(initialSlot, lowestSlot uint64) (uint64, error) {
-	var headers []uint64
+// Sanity check the finalized and attested header are at 32 boundary blocks, so we can download the beacon state
+func (s *Syncer) FindLatestAttestedHeadersAtInterval(initialSlot, lowestSlot uint64) (uint64, error) {
```
yrong, line 527: Notice there is also another `FindOldestAttestedHeaderAtInterval`, can they be merged or only need 1 of them as semantically `FindOldest(start,upperBound)==FindLatest(upperBound,start)`?

relayer/relays/beacon/header/syncer/syncer.go
claravanstaden: I added this because I was worried that the latest finalized header would be larger than the update with the sync committee. But I think it is not a problem, since an older finalized update is allowed if the update contains a sync committee period (and NextSyncCommittee is not set): https://github.com/Snowfork/polkadot-sdk/blob/snowbridge/bridges/snowbridge/pallets/ethereum-client/src/lib.rs#L330

So I will remove this check.

yrong: > The latest finalized header would be larger than the update with the sync committee

I think it's normal in that case we just sync committee without updating finalized state
https://github.com/Snowfork/polkadot-sdk/blob/93687126c499f40d71baba83567936d56434dac0/bridges/snowbridge/pallets/ethereum-client/src/lib.rs#L465-L467

claravanstaden: Removed in https://github.com/Snowfork/snowbridge/pull/1164/commits/9cbee1db2231aed977210ccd0fbdbea17d412cf8.

yrong:
> If the sync committee update data retrieval fails (most likely because a beacon state is not available), build an update using the standard Beacon API endpoints.

That's Cool. Considering now we can always get a valid finality update from API endpoints, is the db option still necessary?

relayer/relays/beacon/header/syncer/syncer.go
```diff
@@ -493,55 +523,120 @@ func (s *Syncer) unmarshalBeaconState(slot uint64, data []byte) (state.BeaconSta
...
 
-func (s *Syncer) GetFinalizedUpdateAtAttestedSlot(attestedSlot uint64, lastSyncedFinalizedSlot uint64) (scale.Update, error) {
-	var update scale.Update
+func (s *Syncer) GetFinalizedUpdateWithSyncCommittee(syncCommitteePeriod, lastFinalizedSlot uint64) (scale.Update, error) {
+	slot := (syncCommitteePeriod) * s.protocol.Settings.SlotsInEpoch * s.protocol.Settings.EpochsPerSyncCommitteePeriod
+
+	boundary := (syncCommitteePeriod + 1) * s.protocol.Settings.SlotsInEpoch * s.protocol.Settings.EpochsPerSyncCommitteePeriod
 
-	attestedSlot, err := s.findAttestedAndFinalizedHeadersAtBoundary(attestedSlot, lastSyncedFinalizedSlot)
+	attestedSlot, err := s.FindOldestAttestedHeaderAtInterval(slot, boundary)
```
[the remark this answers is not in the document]
claravanstaden: The methods do similar but different things. `FindLatestAttestedHeadersAtInterval ` finds the latest available slot, while `FindOldestAttestedHeaderAtInterval` finds the oldest available slot. I have merged the common code into a method called `findValidUpdatePair` here: https://github.com/Snowfork/snowbridge/pull/1164/files#diff-c589b232c311e67a8690c53fe1886c463a185178fde8ed4aa9dd2def499603b3R578. I think the 2 methods are now small and different enough to warrant 2 functions.

claravanstaden:
> That's Cool. Considering now we can always get a valid finality update from API endpoints, is the db option still necessary?

The fallback finality update uses the DB option as a fallback as well. I would feel more comfortable with having the DB option as a backup. Do you have concerns with it?

yrong:
> I would feel more comfortable with having the DB option as a backup. Do you have concerns with it?

So we'll still stick to lodestar with ```--chain.archiveStateEpochFrequency=1``` which saves finality state every epoch, right?

Not much concern.  Just from the perspective of DevOps seems we need a separate cloud instance to run the command storing beacon state periodically for high availability,  then there is some maintenance cost.

relayer/relays/beacon/header/syncer/syncer.go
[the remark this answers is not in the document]
yrong: Yeah, I understand that. 

Just curious seems we use `FindLatestAttestedHeadersAtInterval` for `syncInterimFinalizedUpdate`  while `FindLatestAttestedHeadersAtInterval` for `SyncCommitteePeriodUpdate`, is that a special consideration for that?

> So can the search `s.FindOldestAttestedHeaderAtInterval(slot, boundary)` be replaced with `s.FindLatestAttestedHeadersAtInterval(boundry, slot)`

Or be more specific as mentioned above is that possible to always search for the latest in a reverse manner?

claravanstaden:
> So we'll still stick to lodestar with `--chain.archiveStateEpochFrequency=1` which already saves finality state every epoch, right?

Yes. 😄 

> Not much concern. Just from the perspective of DevOps seems we need a separate cloud instance to run the command storing beacon state periodically for high availability, then there is some maintenance cost.

I think it can be a simple cronjob on the EC2 server that will run the relayer anyway. Agreed on the cost. We can always remove it, but having it for launch gives me great assurance.

yrong:
> I think it can be a simple cronjob on the EC2 server that will run the relayer anyway

IIUC the cronjob only make sense on a seperate EC2 instance, it's for backup when the EC2 server running lodestar is down. 

Meanwhile the cronjob should point to a different lodestar node which I'd suggest to run a backup lodestar with the mock options ```--execution.engineMock --eth1=false```

claravanstaden:
> > I think it can be a simple cronjob on the EC2 server that will run the relayer anyway
> 
> IMHO the cronjob make sense only on a seperate EC2 instance, it's for backup when the EC2 server running lodestar is down.
> 
> So I'd assume it points to a different beacon node for which maybe we can run a backup lodestar with the mock options `--execution.engineMock --eth1=false`
> 
> In this way we have a more reliable fallback option but it also means more cost.

I had more the scenario in mind where just the beacon relayer is down (perhaps restarting because of a bug or some other issue) like we had that weekend where the beacon relayer was down because it could not find the beacon state. In that case, it would have been fine to have a cron just download beacon states every now and then.

I am also not sure if we will be able to justify running 2 EC instances just for this backup, especially if other people also start running relayers.

yrong:
Yeah, I'm also not keen to add another instance for this. For the case when the instance is down for a while, we can start lodestar from a history checkpoint,right?

relayer/relays/beacon/header/syncer/syncer.go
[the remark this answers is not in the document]
claravanstaden: I started typing a response and never sent it. 😅 The reason why these two are different in my opinion is:
- Finalized Update: Ideally want a newer update, so that it can cover a "larger" range of `SLOTS_PER_HISTORICAL_ROOT` (meaning, we want a finalized header as far from the previous finalized header) as we can.
- Finalized Update with Sync Committee: for this case, an earlier update is better since we'd like to update the next sync committee as soon as we can.

Let me know if you think this is kinda unnecessary, and if we should just use the latest header we can (and scrap the "earlier" update for Finalized Update with Sync Committee.

PR #1165, opened 2024-03-28
BeefyClient Improvements

vgeddes:
Adds two important & necessary improvements:

1.  The ability for a trusted operator to set an initial checkpoint after the contract has already been deployed. This is necessary so that we can parallelize the deployment of the bridge. We need to deploy the contracts next week, however BEEFY is only being activated on Polkadot at the end of April

2. The `submitInitial` step emits a `NewTicket(relayer, beefyBlockNumber)` event after completing successfully. As discussed with W3F in yesterday's call, this allows relayers to know that there is an ongoing interactive submission, and that they should not waste gas trying to submit themselves.  

Resolve: SNO-949


contracts/src/BeefyClient.sol
```diff
@@ -9,6 +9,7 @@ import {Uint16Array, createUint16Array} from "./utils/Uint16Array.sol";
 import {Math} from "./utils/Math.sol";
 import {MMRProof} from "./utils/MMRProof.sol";
 import {ScaleCodec} from "./utils/ScaleCodec.sol";
+import "forge-std/console.sol";
```
alistair-singh, line 12: This import is probably not needed.

contracts/src/BeefyClient.sol
```diff
@@ -428,6 +443,27 @@ contract BeefyClient {
         return Bitfield.subsample(ticket.prevRandao, bitfield, ticket.numRequiredSignatures, ticket.validatorSetLen);
     }
 
+    function setInitialCheckpoint(
```
alistair-singh, line 446: Maybe we should call this `forceCheckpoint` to match the equivalent in the bacon client.

contracts/src/BeefyClient.sol
```diff
@@ -428,6 +443,27 @@ contract BeefyClient {
...
+        uint64 _initialBeefyBlock,
+        ValidatorSet calldata _initialValidatorSet,
+        ValidatorSet calldata _nextValidatorSet
+    ) external {
+        if (msg.sender != bootstrapOperator) {
+            revert Unauthorized();
+        }
+        delete bootstrapOperator;
```
alistair-singh, line 454: Instead of making this function single-use by deleting `bootstrapOperator`, we could have a separate function to renounce ownership. This will be useful in rococo testnet as we will never renonce ownership making recovery from beefy issues easier.

yrong: Agree. Though a bit controversial leaving that owner control will make recovery easier, especially in Beefy's early phase applied in production.

contracts/src/BeefyClient.sol
claravanstaden: Good idea! 🥓

contracts/src/BeefyClient.sol
```diff
@@ -289,6 +302,8 @@ contract BeefyClient {
             prevRandao: 0,
             bitfieldHash: keccak256(abi.encodePacked(bitfield))
         });
+
+        emit NewTicket(msg.sender, commitment.blockNumber);
```
alistair-singh, line 306: We should add the `ticketId` to `NewTicket` and `NewMMRRoot` events to be able to link them.

contracts/src/BeefyClient.sol
```diff
@@ -462,6 +461,14 @@ contract BeefyClient {
...
         nextValidatorSet.usageCounters = createUint16Array(nextValidatorSet.length);
+
+        // Clear latest MMR Root if present
+        delete latestMMRRoot;
+    }
+
+    /// @dev Renounce all administrative privileges
+    function renounce() external {
+        delete forceCheckpointOperator;
     }
```
yrong, line 472: Not sure fully understand why we provide an external `renounce` here.

I would assume in the interim period we just make `BeefyClient` inherited from 
https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/access/Ownable.sol for access control and after production and stable for a while we remove the owner control with an upgrade.

yrong:
Another concern is do we still wanna make [BEEFY_CLIENT](https://github.com/Snowfork/snowbridge/blob/013f63723a1978c9b1802389902cb61f87f57ad0/contracts/src/Gateway.sol#L56) immutable?

contracts/.gas-snapshot
```diff
@@ -1,87 +1,91 @@
...
-BeefyClientTest:testSubmitWithHandover() (gas: 1644136)
-BeefyClientTest:testSubmitWithHandoverAnd3SignatureCount() (gas: 1970604)
-BeefyClientTest:testSubmitWithHandoverAndOldBlockFailsWithStaleCommitment() (gas: 364742)
-BeefyClientTest:testSubmitWithHandoverCountersAreCopiedCorrectly() (gas: 1671087)
-BeefyClientTest:testSubmitWithHandoverFailStaleCommitment() (gas: 1391533)
-BeefyClientTest:testSubmitWithHandoverFailWithInvalidValidatorProofWhenNotProvidingSignatureCount() (gas: 1446237)
-BeefyClientTest:testSubmitWithHandoverFailWithInvalidValidatorSet() (gas: 1393826)
-BeefyClientTest:testSubmitWithHandoverFailWithoutPrevRandao() (gas: 1276129)
-BeefyClientTest:testSubmitWithOldBlockFailsWithStaleCommitment() (gas: 364308)
+BeefyClientTest:testCommitPrevRandaoCalledInSequence() (gas: 16057)
```
claravanstaden, line 4: How do you regenerate this file?

vgeddes: using `forge snapshot`.

vgeddes:
> Another concern is do we still wanna make [BEEFY_CLIENT](https://github.com/Snowfork/snowbridge/blob/013f63723a1978c9b1802389902cb61f87f57ad0/contracts/src/Gateway.sol#L56) immutable?

This is by design, the beefy client and agent executor are so critically important that they should only be changed via a contract upgrade of Gateway.sol. It also reduces gas costs when referencing those contracts.

PR #1167, opened 2024-04-01
Make beefy client public

alistair-singh:
Resolves: SNO-906


contracts/src/Gateway.sol
```diff
@@ -53,7 +53,7 @@ contract Gateway is IGateway, IInitializable {
     address internal immutable AGENT_EXECUTOR;
```
yrong, line 53: I'd suggest making`AGENT_EXECUTOR` also public, then in https://github.com/Snowfork/snowbridge/blob/fd2fd86746ec91c8ce4e6c345cf925d6dbbd07a7/contracts/src/DeployGatewayLogic.sol#L28 when to upgrade with a new Gateway logic, there is no need to create it again.

vgeddes: Agree, lets make it public too 👍

contracts/src/Gateway.sol
```diff
@@ -97,7 +97,7 @@ contract Gateway is IGateway, IInitializable {
     }
 
     constructor(
-        address beefyClient,
+        address beefyLightClient,
```
vgeddes, line 100: I specifically dropped the "Light" qualifier. It doesn't add anything meaningful, and is too verbose.

contracts/src/Gateway.sol
alistair-singh: Addressed.

contracts/src/Gateway.sol
alistair-singh: The name shadowed the `beefyClient()` method. I removed the word light and used an underscore to avoid the name collision.

contracts/src/interfaces/IGateway.sol
```diff
@@ -78,6 +78,12 @@ interface IGateway {
     /// @dev Emitted when a command is sent to register a new wrapped token on AssetHub
     event TokenRegistrationSent(address token);
 
+    /// @dev the address of the BEEFY light client used to verify messages.
+    function beefyClient() external view returns (address);
+
+    // @dev the address of the agent executor implementaiton.
+    function agentExecutor() external view returns (address);
+
```
vgeddes, line 86: Not necessary. If a variable or constant is public it automatically has an accessor generated.

contracts/src/Gateway.sol
```diff
@@ -224,6 +224,14 @@ contract Gateway is IGateway, IInitializable {
...
      * Getters
      */
+    function beefyClient() external view returns (address) {
+        return BEEFY_CLIENT;
+    }
+
+    function agentExecutor() external view returns (address) {
+        return AGENT_EXECUTOR;
+    }
+
```
vgeddes, line 234: Not necessary. If a variable or constant is public it automatically has an accessor generated.

vgeddes (commented):
Just make the `BEEFY_CLIENT` variable public and you won't need to write an accessor for it.

contracts/src/Gateway.sol
alistair-singh: addressed in https://github.com/Snowfork/snowbridge/pull/1167/commits/022ab0e88e1fbc24fb507874b435c28ceea85252

contracts/src/interfaces/IGateway.sol
```diff
@@ -78,6 +78,12 @@ interface IGateway {
     /// @dev Emitted when a command is sent to register a new wrapped token on AssetHub
     event TokenRegistrationSent(address token);
 
+    /// @dev the address of the BEEFY light client used to verify messages.
+    function BEEFY_CLIENT() external returns (address);
+
+    // @dev the address of the agent executor implementaiton.
+    function AGENT_EXECUTOR() external returns (address);
+
```
yrong, line 86: Unnecessary?

alistair-singh: Our bindings are generated off IGateway for the relayer and the contract-types npm package. So this is helpful for that.

vgeddes: The problem is that I want `IGateway` to be more for user facing APIs, not infrastructure concerns. In fact, I want to remove more non-user APIs from IGateway.

Can't you generate bindings for `Gateway.sol`, and then use `IGateway.implemention()` to get the implementation contract?

claravanstaden (approved):

vgeddes (approved):
+1

PR #1170, opened 2024-04-08
Add upgradable `Shell` contract

vgeddes:
Implements a bare `Shell` contract that can be upgraded by a trusted operator to a fully fledged gateway.


yrong (approved):
Cool!

contracts/src/DeployScript.sol
```diff
@@ -59,9 +59,11 @@ contract DeployScript is Script {
 
         uint8 foreignTokenDecimals = uint8(vm.envUint("FOREIGN_TOKEN_DECIMALS"));
 
+        address recoveryOperator = vm.envOr("RECOVERY_OPERATOR", address(0));
+
```
yrong, line 63: Maybe nitpicking, what's the benefit of using `address(0)` as a fallback instead of explicitly `vm.envAddress`?

contracts/src/Gateway.sol
```diff
@@ -577,6 +568,29 @@ contract Gateway is IGateway, IInitializable {
...
+        if (impl.codehash != implCodeHash) {
+            revert InvalidCodeHash();
+        }
+
+        // Update the proxy with the address of the new implementation
+        ERC1967.store(impl);
+
+        // Call the initializer
+        (bool success, bytes memory returndata) =
+            impl.delegatecall(abi.encodeCall(IInitializable.initialize, initializerParams));
```
alistair-singh, line 588: If initializer is always run. Than we cannot through `AlreadyInitialized()` on line https://github.com/Snowfork/snowbridge/pull/1170/files#diff-74470ba8f35b42dbc8805739e92708c0696343e0b1819feb3e7a5ebb77cc623cR605

Wouldn't it need to skip if already initialized?

claravanstaden (approved):
Looks great! Perhaps just add some tests, if it makes sense? For example, to test `recover` and `renounce`. Perhaps to also test what @alistair-singh mentioned over here: https://github.com/Snowfork/snowbridge/pull/1170/files#r1558391427

contracts/src/DeployScript.sol
yrong: There is no fallback for `vm.envAddress` which requires the env to already explicitly be set, does that make more sense here?

vgeddes:
> Looks great! Perhaps just add some tests, if it makes sense? For example, to test `recover` and `renounce`. Perhaps to also test what @alistair-singh mentioned over here: https://github.com/Snowfork/snowbridge/pull/1170/files#r1558391427

Good point. Will add tests for those.

contracts/src/Gateway.sol
vgeddes: Well the idea is that each new version of the Gateway will have a new initializer defined, like in `contracts/src/upgrades/rococo/GatewayV2.sol`.

Can you foresee any problems with that?

contracts/src/DeployScript.sol
vgeddes: This is just a placeholder until we set up multisig to act as a recovery account.

contracts/src/Gateway.sol
alistair-singh: Ah ok. I get the pattern now!

vgeddes:
I've revised the implementation considerably. Its much simpler, easier to audit and will eliminate potential concerns from the fellowship about the security of the trusted operator. In other words, I believe this solution will be easier to fast-track through the voting process.

Changes:
* Introduced a bare `Shell` contract which only has the ability to upgraded and nothing else.

The existing `Gateway` implementation contract is left untouched.

If needed we can always add a trusted operator back to the Gateway implementation  that the shell will be upgraded to. So it retains that capability.

PR #1187, opened 2024-05-02
Fix findLatestCheckPoint

yrong:
For the issue Alistair found with the execution relayer in rococo environment bricked.

```
May 01 22:29:34 ip-172-31-41-212 start-execution-relayer.sh[1475828]: {"@timestamp":"2024-05-01T22:29:34.93758737Z","blockRoot":"0xd272b2946a1a06f641f50fd96bfe316c29511ffaf0f9b5edd0c0dc3a587c14c6","computedRoot":"0xd272b2946a1a06f641f50fd96bfe316c29511ffaf0f9b5edd0c0dc3a587c14c6","level":"info","message":"beacon header with block included found","slot":4868811}
May 01 22:29:34 ip-172-31-41-212 start-execution-relayer.sh[1475828]: {"@timestamp":"2024-05-01T22:29:34.937627911Z","calculatedCheckpointSlot":0,"level":"info","message":"checkpoint slot not available, try with slot in next sync period instead"}
May 01 22:29:34 ip-172-31-41-212 start-execution-relayer.sh[1475828]: {"@timestamp":"2024-05-01T22:29:34.982432636Z","calculatedCheckpointSlot":4874240,"lastFinalizedSlot":4905664,"level":"info","message":"fetch checkpoint on chain backward from history"}
May 01 22:29:34 ip-172-31-41-212 start-execution-relayer.sh[1475828]: {"@timestamp":"2024-05-01T22:29:34.992440215Z","error":"fetch execution header proof: get header update by slot with ancestry proof: populate closest checkpoint: get history finalized header for the checkpoint: Can't find checkpoint on chain for slot 4868811","level":"fatal","message":"Unhandled error"}
```

### DIFF
--- a/relayer/relays/beacon/header/header.go
+++ b/relayer/relays/beacon/header/header.go
@@ -436,11 +436,9 @@ func (h *Header) findLatestCheckPoint(slot uint64) (state.FinalizedHeader, error
 	}
 	startIndex := uint64(lastIndex)
 	endIndex := uint64(0)
-	if uint64(lastIndex) > h.protocol.Settings.EpochsPerSyncCommitteePeriod {
-		endIndex = endIndex - h.protocol.Settings.EpochsPerSyncCommitteePeriod
-	}
 
 	syncCommitteePeriod := h.protocol.Settings.SlotsInEpoch * h.protocol.Settings.EpochsPerSyncCommitteePeriod
+	slotPeriodIndex := slot / syncCommitteePeriod
 
 	for index := startIndex; index >= endIndex; index-- {
 		beaconRoot, err := h.writer.GetFinalizedBeaconRootByIndex(uint32(index))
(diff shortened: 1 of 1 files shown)


relayer/relays/beacon/header/header.go
```diff
@@ -436,11 +436,9 @@ func (h *Header) findLatestCheckPoint(slot uint64) (state.FinalizedHeader, error
...
 	startIndex := uint64(lastIndex)
 	endIndex := uint64(0)
-	if uint64(lastIndex) > h.protocol.Settings.EpochsPerSyncCommitteePeriod {
-		endIndex = endIndex - h.protocol.Settings.EpochsPerSyncCommitteePeriod
-	}
 
 	syncCommitteePeriod := h.protocol.Settings.SlotsInEpoch * h.protocol.Settings.EpochsPerSyncCommitteePeriod
+	slotPeriodIndex := slot / syncCommitteePeriod
 
 	for index := startIndex; index >= endIndex; index-- {
```
claravanstaden, line 443: `endIndex` will also be `0`, is that correct?

claravanstaden (commented):
Thanks for the fix, Ron! Could you please describe what the bug is here and your fix changes that?

relayer/relays/beacon/header/header.go
yrong: Yeah, that's primary change. _**By extending search range until the very beginning.**_

Basically what happened on Rococo setup is that we upgraded the beacon relayer a few weeks back, but we never upgraded the execution relayer so the bridge was down for about ~2-3 weeks.

So we need to search from the latest checkpoint backward until we find the matched checkpoint, previously we search for only 256 epochs at most which is not correct.

claravanstaden (approved):

alistair-singh (approved):
+1

PR #1189, opened 2024-05-06
Beacon state import and list feature

claravanstaden:
Adds features for:
- importing a ssz state file into the beacon store
- listing all the states in the beacon store
- fixes a bug in the beacon store module related to file paths

Needs to be cherry-picked into https://github.com/Snowfork/snowbridge/tree/release-v1.0.0 after merge.

### DIFF
--- a/relayer/relays/beacon/store/datastore.go
+++ b/relayer/relays/beacon/store/datastore.go
@@ -310,7 +310,7 @@ func (s *Store) createTable() error {
 	sqlStmt := `CREATE TABLE IF NOT EXISTS beacon_state (
 		id INTEGER PRIMARY KEY AUTOINCREMENT,
 		attested_slot INTEGER NOT NULL,
-		finalized_slot INTEGER NOT NULL,
+		finalized_slot INTEGER NOT NULL UNIQUE,
 		attested_sync_period INTEGER NOT NULL,
 		finalized_sync_period INTEGER NOT NULL,
 		attested_state_filename TEXT NOT NULL,
--- a/relayer/relays/beacon/store/datastore.go
+++ b/relayer/relays/beacon/store/datastore.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"path/filepath"
 	"time"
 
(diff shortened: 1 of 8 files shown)


relayer/relays/beacon/store/datastore.go
```diff
@@ -169,8 +174,51 @@ func (s *Store) WriteEntry(attestedSlot, finalizedSlot uint64, attestedStateData
...
 
+func (s *Store) ListBeaconStates() ([]BeaconState, error) {
+	var response []BeaconState
+
+	query := `SELECT id, attested_slot, finalized_slot, attested_sync_period, finalized_sync_period, attested_state_filename, finalized_state_filename FROM beacon_state ORDER BY attested_slot`
+
+	rows, err := s.db.Query(query)
+	if err != nil {
+		return response, fmt.Errorf("no match found")
+	}
```
yrong, line 185: Do we want to implement a pagination query here?

yrong:
May not be directly with change in this PR

https://github.com/Snowfork/snowbridge/blob/74124615060a032de59bddf2f91a78686fdba4b3/relayer/relays/beacon/store/datastore.go#L310-L320

Do we want to add an unique constraint on `finalized_slot` to make sure we're not importing duplicate entry into the db?

relayer/relays/beacon/store/datastore.go
claravanstaden: It is limited to 100 states so I thought it is not necessary yet, but will be cool for the future!

claravanstaden:
> May not be relevant with change in this PR
> 
> https://github.com/Snowfork/snowbridge/blob/74124615060a032de59bddf2f91a78686fdba4b3/relayer/relays/beacon/store/datastore.go#L310-L320
> 
> Do we want to add an unique constraint on `finalized_slot` to make sure we're not importing duplicate entry into the db?

Good idea! :) https://github.com/Snowfork/snowbridge/pull/1189/commits/65d9e6806d1469805bd9ff1d61eba26e6dd01b3f

vgeddes (approved):

PR #1190, opened 2024-05-06
Deployment scripts for BeefyClient & Gateway

vgeddes:
Fixes: SNO-994

* Adds deployment scripts and initial configuration for BeefyClient and Gateway
* Helper script to generate BEEFY checkpoint for production
* Allows Gateway to be rescued by a privileged operator

Fee parameter calculations in https://www.notion.so/snowfork/Gateway-Parameters-0cf913d089374027a86721883306ee61?pvs=4


contracts/test/Gateway.t.sol
```diff
@@ -857,6 +851,39 @@ contract GatewayTest is Test {
...
+        emit IUpgradable.Upgraded(address(newLogic));
+
+        hoax(0x4B8a782D4F03ffcB7CE1e95C5cfe5BFCb2C8e967);
+        Gateway(address(gateway)).rescue(impl, implCodeHash, initParams);
+
+        // Verify that the MockGatewayV2.initialize was called
+        assertEq(MockGatewayV2(address(gateway)).getValue(), 42);
+    }
+
+    function testRescuebyPublicFails() public {
```
alistair-singh, line 878: We could make this test use `dropRescueAbility` to relinquish the rescuer address, and then try upgrade using the rescuer address. This will test both paths.

contracts/src/storage/OperatorStorage.sol
```diff
@@ -0,0 +1,20 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2023 Snowfork <hello@snowfork.com>
+pragma solidity 0.8.23;
+
+import {UD60x18} from "prb/math/src/UD60x18.sol";
```
alistair-singh, line 5: Looks unused?

contracts/src/Gateway.sol
```diff
@@ -612,5 +615,31 @@ contract Gateway is IGateway, IInitializable, IUpgradable {
         assets.registerTokenFee = config.registerTokenFee;
         assets.assetHubCreateAssetFee = config.assetHubCreateAssetFee;
         assets.assetHubReserveTransferFee = config.assetHubReserveTransferFee;
+
+        // Initialize operator storage
+        OperatorStorage.Layout storage operatorStorage = OperatorStorage.layout();
+        operatorStorage.operator = 0x4B8a782D4F03ffcB7CE1e95C5cfe5BFCb2C8e967;
```
alistair-singh, line 621: We should parameterize this so that we can use a different address in rococo. One that we will probably never relinquish.

contracts/scripts/UpgradeShell.sol
```diff
@@ -0,0 +1,90 @@
...
+            bridgeHubAgentID: 0x03170a2e7597b7b7e3d84c05391d139a62b157e78786d8c082f29dcf4c111314,
+            foreignTokenDecimals: 10,
+            maxDestinationFee: dot(2),
+            initializerParams: Gateway.Config({
+                mode: OperatingMode.Normal,
+                deliveryCost: mDot(100), // 0.1 DOT
+                registerTokenFee: 0.002 ether,
+                assetHubParaID: ParaID.wrap(1000),
+                assetHubAgentID: 0x81c5ab2571199e3188135178f3c2c8e2d268be1313d029b30f534fa579b69b79,
+                assetHubCreateAssetFee: mDot(100), // 0.1 DOT
```
alistair-singh, line 58: Hey I calculated this to `20.19` DOT.

1. https://github.com/polkadot-fellows/runtimes/blob/main/system-parachains/asset-hubs/asset-hub-polkadot/src/lib.rs#L362
2. https://github.com/polkadot-fellows/runtimes/blob/main/system-parachains/asset-hubs/asset-hub-polkadot/src/lib.rs#L288
3. https://github.com/polkadot-fellows/runtimes/blob/d63ed882383e5c972757182e1125818666c9d660/system-parachains/constants/src/polkadot.rs#L68
4. https://github.com/polkadot-fellows/runtimes/blob/d63ed882383e5c972757182e1125818666c9d660/relay/polkadot/constants/src/lib.rs#L36

Python implementation:
```py
UNITS=10_000_000_000
DOLLARS=UNITS
CENTS=DOLLARS/100
MILLICENTS=CENTS/1_000

def deposit(items, bytes):
    return items * 20 * DOLLARS + bytes * 100 * MILLICENTS

deposit(1, 190) / 10**10 # 10 decimal places
```

I remember you saying that they are dropping the fee but I cannot find a related issue or PR.

contracts/scripts/UpgradeShell.sol
```diff
@@ -0,0 +1,90 @@
...
+                assetHubReserveTransferFee: mDot(100), // 0.1 DOT
```
alistair-singh, line 59: Not sure how to get this value. If you share how I can double check this for you as well.

web/packages/test-helpers/src/generateBeefyCheckpointProd.ts
```diff
@@ -0,0 +1,58 @@
...
+}
+
+async function generateBeefyCheckpoint() {
+    let beefyBlock = parseInt(process.env.BEEFY_BLOCK!, 10)
+    if (!Number.isInteger(beefyBlock) || beefyBlock == 0) {
+        throw new Error("InvalidBeefyBlock")
+    }
+
+    let _api = await ApiPromise.create({
+        provider: new WsProvider("wss://polkadot-rpc.dwellir.com"),
```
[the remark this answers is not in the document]
vgeddes: Good catch, yes it can be. Will try make the change.

contracts/src/Gateway.sol
vgeddes: good point, will re-evaluate this.

contracts/scripts/UpgradeShell.sol
vgeddes: According to my understanding, this is really just bad naming in the Gateway contract which we should fix.

It turns out that `registerTokenFee` actually covers the token registration fee, and is credited to the AssetHub agent/sovereign after exchange rate conversions.

The `assetHubCreateAssetFee` item just covers the XCM execution fee on AssetHub

For example, in BridgeHub, the registration fee is added to `assetHubCreateAssetFee` on line https://github.com/paritytech/polkadot-sdk/blob/d237adfb13ceab8724fabee74469788e19e0e6af/bridges/snowbridge/primitives/router/src/inbound/mod.rs#L167,

vgeddes: > https://github.com/polkadot-fellows/runtimes/blob/d63ed882383e5c972757182e1125818666c9d660/relay/polkadot/constants/src/lib.rs#L36

That's relay chain config, AssetHub has different storage deposit config.

vgeddes: Can see on Polkadot-JS that the asset registration deposit 0.2 DOT:
Screenshot 2024-05-16 at 02 29 01

alistair-singh: Ok but then `registerTokenFee` is too low. `0.002 ETH` (6 USD).

`CreateForeignAssetDeposit` is set to `deposit(1, 190)` which is `20.19` DOT.

https://github.com/polkadot-fellows/runtimes/blob/d63ed882383e5c972757182e1125818666c9d660/system-parachains/asset-hubs/asset-hub-polkadot/primitives/src/lib.rs#L55

Asset hub just defers to relaychain config.
https://github.com/polkadot-fellows/runtimes/blob/d63ed882383e5c972757182e1125818666c9d660/system-parachains/constants/src/polkadot.rs#L68

And this test ensures that its the set the same as asset hub.
https://github.com/polkadot-fellows/runtimes/blob/d63ed882383e5c972757182e1125818666c9d660/system-parachains/asset-hubs/asset-hub-polkadot/src/lib.rs#L1748

alistair-singh: Ah ok. There is an extra divide by 100 that I missed here:
https://github.com/polkadot-fellows/runtimes/blob/d63ed882383e5c972757182e1125818666c9d660/system-parachains/constants/src/polkadot.rs#L68

contracts/scripts/DeployBeefyClient.sol
```diff
@@ -0,0 +1,61 @@
...
+
+    function readConfig() internal pure returns (Config memory config) {
+        // TODO: When we are ready to commit to checkpoint, run the following to compute the checkpoint
+        //   (cd web/packages/test-helpers; BEEFY_BLOCK=... npx npx ts-node src/generateBeefyCheckpointProd.ts)
+        // Substitute `startBlock`, `current`, `next` below
+        config = Config({
+            startBlock: 0,
+            current: BeefyClient.ValidatorSet({id: 0, length: 0, root: 0}),
+            next: BeefyClient.ValidatorSet({id: 0, length: 0, root: 0}),
+            randaoCommitDelay: 128,
```
yrong, line 42: Is `128` finalized for the production? Seems Rococo also use this value, met one issue `the beefy syncing is slow and the latency keeps increasing` https://snowfork.slack.com/archives/G01BGMGLAC9/p1715598123882029
maybe related.

contracts/scripts/UpgradeShell.sol
yrong: Yeah, so it's about 0.2 DOT(0.000464 ETH), then the `registerTokenFee` as 0.002 should be pretty enough.

contracts/src/Gateway.sol
```diff
@@ -612,5 +615,31 @@ contract Gateway is IGateway, IInitializable, IUpgradable {
...
+    }
+
+    /// @dev Temporary rescue ability for the initial bootstrapping phase of the bridge
+    function rescue(address impl, bytes32 implCodeHash, bytes calldata initializerParams) external {
+        OperatorStorage.Layout storage operatorStorage = OperatorStorage.layout();
+        if (msg.sender != operatorStorage.operator) {
+            revert Unauthorized();
+        }
+        Upgrade.upgrade(impl, implCodeHash, initializerParams);
+    }
```
yrong, line 631: Just curious what's the workflow for emergency shutdown assuming bridge under attack? Are we gonna to upgrade to another Gateway contract untradable？

contracts/scripts/DeployBeefyClient.sol
alistair-singh: Doesnt this have to be 128  (at least 4 epoch) to limit biasability of randao?

yrong: Then the question is are we gonna to sync every commitment on Polkadot to Ethereum? If no what's the interval is. I'm not sure with the default config the relayer can catch up with the beefy commitment on Polakdot.

web/packages/test-helpers/src/generateBeefyCheckpointProd.ts
```diff
@@ -0,0 +1,58 @@
...
+    })
+
+    let api = await _api.at(await _api.rpc.chain.getBlockHash(beefyBlock))
+
+    let authorities = await api.query.beefyMmrLeaf.beefyAuthorities<AuthoritySet>()
+    let nextAuthorities = await api.query.beefyMmrLeaf.beefyNextAuthorities<AuthoritySet>()
```
yrong, line 34: Assuming we reuse it for local setup, 
https://github.com/Snowfork/snowbridge/blob/238a8fcf304dab8d824c0c62adb07a2786df5012/web/packages/test-helpers/src/generateBeefyCheckpoint.ts#L72-L78 FYI.

contracts/scripts/DeployBeefyClient.sol
vgeddes: > Is 128 finalized for the production? 

Yeah, as Alistair says, 128 is the final value for production, according to previous analysis we performed.

> Then the question is are we gonna to sync every commitment on Polkadot to Ethereum

We should try to sync only BEEFY commitment per Polkadot validator session.

Hoping you can figure out a solution to make the beefy relayers compatible with this.

contracts/src/Gateway.sol
vgeddes: This rescue ability is temporary and will be removed after a few weeks. After that we're at the mercy of the code.

contracts/src/Gateway.sol
vgeddes: fixed in latest upload

contracts/test/Gateway.t.sol
vgeddes: added a test in the latest upload

PR #1191, opened 2024-05-06
Separate endpoint for beacon state

claravanstaden:
Configure a separate endpoint for the beacon state download, because Lodestar does not backfill ssz states if the node goes offline for a period longer than the weak subjectivity period. This is problematic because we should be able to get historic states. This enables us to switch to a different consensus node for ssz states, if necessary. If all is well, both the `endpoint` and `stateEndpoint` config values can point to Lodestar.

### DIFF
--- a/relayer/cmd/generate_beacon_data.go
+++ b/relayer/cmd/generate_beacon_data.go
@@ -40,7 +40,7 @@ func generateBeaconFixtureCmd() *cobra.Command {
 		RunE:  generateBeaconTestFixture,
 	}
 
-	cmd.Flags().String("config", "", "Path to the beacon relay config")
+	cmd.Flags().String("config", "/tmp/snowbridge/beacon-relay.json", "Path to the beacon relay config")
 	cmd.Flags().Bool("wait_until_next_period", true, "Waiting until next period")
 	cmd.Flags().Uint32("nonce", 1, "Nonce of the inbound message")
 	return cmd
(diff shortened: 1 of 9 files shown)


relayer/relays/beacon/header/syncer/api/api.go
```diff
@@ -398,7 +400,7 @@ func (b *BeaconClient) GetLatestFinalizedUpdate() (LatestFinalisedUpdateResponse
 
 func (b *BeaconClient) GetBeaconState(stateIdOrSlot string) ([]byte, error) {
 	var data []byte
-	req, err := http.NewRequest("GET", fmt.Sprintf("%s/eth/v2/debug/beacon/states/%s", b.endpoint, stateIdOrSlot), nil)
+	req, err := http.NewRequest("GET", fmt.Sprintf("%s/eth/v2/debug/beacon/states/%s", b.stateEndpoint, stateIdOrSlot), nil)
```
claravanstaden, line 403: This is the main change.

relayer/cmd/generate_beacon_data.go
```diff
@@ -656,45 +666,50 @@ func generateInboundFixture(cmd *cobra.Command, _ []string) error {
 	err := func() error {
 		ctx := context.Background()
 
-		endpoint, err := cmd.Flags().GetString("url")
+		beaconConfig, err := cmd.Flags().GetString("beacon-config")
 		if err != nil {
 			return err
 		}
 
-		viper.SetConfigFile("/tmp/snowbridge/beacon-relay.json")
```
claravanstaden, line 664: Just some cleanup to remove these hardcoded config file paths.

vgeddes (approved):

yrong (approved):
+1

PR #1193, opened 2024-05-08
Instant rebalancing with fee asset always in WETH

yrong:
This is a POC for an instant rebalancing with fee asset always in WETH based upon [discussion](https://snowfork.slack.com/archives/G01BGMGLAC9/p1714995738905699?thread_ts=1714699858.812779&cid=G01BGMGLAC9) in channel.

companion change on substrate side in https://github.com/Snowfork/polkadot-sdk/commit/64b1d4453e356866395a6284b3e6c05cc499557e

Destination parachain should be configured to 

- register WETH as sufficient 
- a custom `IsReserve` to allow AssetHub as trusted reserve for WETH
- a custom `Trader` allow to use WETH to pay fees for the xcm execution

### DIFF
--- a/contracts/src/Assets.sol
+++ b/contracts/src/Assets.sol
@@ -75,6 +75,8 @@ library Assets {
         AssetsStorage.Layout storage $ = AssetsStorage.layout();
         if ($.assetHubParaID == destinationChain) {
             costs.foreign = $.assetHubReserveTransferFee;
+            // We don't charge any extra fees beyond delivery costs
+            costs.native = 0;
         } else {
             // Reduce the ability for users to perform arbitrage by exploiting a
             // favourable exchange rate. For example supplying Ether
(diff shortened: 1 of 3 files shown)


yrong:
As I mentioned in the discussion, on most substrate chains the xcm execution cost is so tiny (less than 0.1$) comparing with the gas cost on Ethereum. I'm a bit conservative to add any extra steps on Ethereum(including things like e.g. an on chain swap, or syncing the exchange rate periodically from dex, or an extra flow for an instant rebalancing, ....).

Still prefer an on-chain fee configuration allows parachain to:

- use any fee asset as they want(e.g. the native token), 
- charge any extra Ether for avoid spamming. Ideally it should cover the execution cost on substrate chains but not required(e.g. Acala/Moonbeam may set it as zero at first stage to win the customer).
- leave the rebalancing option to parachain(e.g. they may choose to do it in a periodical/batch manner every 2 weeks or more).

PR #1194, opened 2024-05-08
Fix interim finalized update

claravanstaden:
The interim finalized update code was buggy and overcomplicated - it tried to find the latest slot for the finalized header update and the oldest slot for the finalized header with a sync committee, which is unnecessary.

This PR simplifies the interim update (when the last 2 finalized headers are more than 8192 headers apart) by just providing a range of slots that will be sufficient and finding the first slot in the range that matches.

### DIFF
--- a/relayer/relays/beacon/header/header.go
+++ b/relayer/relays/beacon/header/header.go
@@ -126,12 +126,18 @@ func (h *Header) SyncCommitteePeriodUpdate(ctx context.Context, period uint64) e
 
 	// If the gap between the last two finalized headers is more than the sync committee period, sync an interim
 	// finalized header
-	maxLatency := h.cache.Finalized.LastSyncedSlot + (h.protocol.Settings.SlotsInEpoch * h.protocol.Settings.EpochsPerSyncCommitteePeriod)
-	if maxLatency < uint64(update.Payload.FinalizedHeader.Slot) {
-		err = h.syncInterimFinalizedUpdate(ctx, h.cache.Finalized.LastSyncedSlot)
-		if err != nil {
-			return fmt.Errorf("sync interim finalized header update: %w", err)
+	if uint64(update.Payload.FinalizedHeader.Slot) > h.cache.Finalized.LastSyncedSlot {
+		diff := uint64(update.Payload.FinalizedHeader.Slot) - h.cache.Finalized.LastSyncedSlot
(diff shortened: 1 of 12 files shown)


vgeddes (approved):

relayer/relays/beacon/header/header.go
```diff
@@ -237,14 +243,18 @@ func (h *Header) SyncHeaders(ctx context.Context) error {
...
 
-func (h *Header) syncInterimFinalizedUpdate(ctx context.Context, lastSyncedSlot uint64) error {
-	checkpointSlot := h.protocol.CalculateNextCheckpointSlot(lastSyncedSlot)
+func (h *Header) syncInterimFinalizedUpdate(ctx context.Context, lastSyncedSlot, newCheckpointSlot uint64) error {
+	// Calculate the range that the interim finalized header update may be in
+	minSlot := newCheckpointSlot - h.protocol.SlotsPerHistoricalRoot
+	maxSlot := lastSyncedSlot + h.protocol.SlotsPerHistoricalRoot
 
-	finalizedUpdate, err := h.syncer.GetLatestPossibleFinalizedUpdate(checkpointSlot, lastSyncedSlot)
+	finalizedUpdate, err := h.syncer.GetFinalizedUpdateAtAttestedSlot(minSlot, maxSlot, false)
```
yrong, line 251: What if `newCheckpointSlot` is far more than the `lastSyncedSlot`?

yrong:
https://github.com/Snowfork/snowbridge/blob/062b7f43ab60d4a2a4f217dbe9c3e0323c462c95/relayer/relays/beacon/header/syncer/api/api.go#L407

May not be directly relevant to the change of this PR. I would suggest to first download ssz state from `stateEndpoint` and if error then fallback to the previous `endpoint`, ant the last fallback to the local db.

relayer/relays/beacon/header/header.go
claravanstaden: Are you asking what if the `lastSyncedSlot ` is perhaps more than 1 sync committee period behind `newCheckpointSlot `? I also thought about this and I think it can't happen, because the sync committee updates are guaranteed to be sequential, right? Meaning, there will at least be one sync committee update per period.

yrong: Yeah, it's true. Just ignore it.

PR #1197, opened 2024-05-13
Generate beefy checkpoint from runtime storage

yrong:
Resolves: https://linear.app/snowfork/issue/SNO-982

### DIFF
--- a/web/packages/test-helpers/src/generateBeefyCheckpoint.ts
+++ b/web/packages/test-helpers/src/generateBeefyCheckpoint.ts
@@ -9,7 +9,7 @@ import { u32, u64 } from "@polkadot/types-codec"
 import { H256 } from "@polkadot/types/interfaces"
 import { Struct } from "@polkadot/types"
 
-interface NextAuthoritySet extends Struct {
+interface AuthoritySet extends Struct {
     id: u64
     len: u32
     keysetCommitment: H256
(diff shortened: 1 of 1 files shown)


web/packages/test-helpers/src/generateBeefyCheckpoint.ts
```diff
@@ -50,23 +47,42 @@ async function generateBeefyCheckpoint() {
...
-        let publicKey = publicKeyConvert(authorities[i], false).slice(1)
-        let publicKeyHashed = createKeccakHash("keccak256").update(Buffer.from(publicKey)).digest()
-        addrs.push(publicKeyHashed.slice(12))
+        console.log("index is:" + i + ",authority is:" + authorities[i])
+        try {
+            let publicKey = publicKeyConvert(authorities[i], false).slice(1)
+            let publicKeyHashed = createKeccakHash("keccak256")
+                .update(Buffer.from(publicKey))
+                .digest()
+            addrs.push(publicKeyHashed.slice(12))
```
alistair-singh, line 62: Wont this change the length of the list of validator addresses, which in turn will affect the merkle tree and eventual root? If the same logic is applied in the relayer then I think its fine.

yrong: As the title mentioned, we don't manually construct the merkle root from the addresses as before, instead get `currentAuthorities` from runtime storage. The printing here is only for debug purpose.

yrong:
Closed in favor of https://github.com/Snowfork/snowbridge/pull/1190

PR #1196, opened 2024-05-10
More metrics for monitoring & Fix for the local setup

yrong:
Fixes: https://linear.app/snowfork/issue/SNO-965

Raise alarm when:
- (latestPolkadotBlockOnEthereum == previousPolkadotBlockOnEthereum && beefyLatency > Threshold) which means beefy relayer stuck or something error in beefy client
- (latestEthereumBlockOnPolkadot == previousEthereumBlockOnPolkadot && beaconLatency > Threshold) which means beacon relayer stuck or something error in beacon client
- toPolkadot: outbound > inbound && inbound == previous_inbound which means execution relayer stuck or something error in `InboundQueue.submit` on BH
- toEthereum: outbound > inbound && inbound == previous_inbound which means parachain relayer stuck or something error in `Gateway.submitV1` on Ethereum
- In both direction when outbound < inbound which seems impossible or we're attacked

PR #1199, opened 2024-05-15
Format ts codes

yrong:
Merge https://github.com/Snowfork/snowbridge/pull/1196 (already reviewed) with format changes.


alistair-singh (approved):
Thanks for sorting this out!

PR #1202, opened 2024-05-20
Updates for long range sync

claravanstaden:
- Interim finalized headers will also be in the current store period, not in the next (accompanying change made for https://github.com/Snowfork/polkadot-sdk/pull/145)
- Adds methods to get a beacon checkpoint payload for:
    - a certain slot
    - from the provided file
- Better validation for finalized updates (check if the attested header was signed by the supermajority, etc)
- Execution header now downloads and imports only 8 headers at a time, instead of downloading all the headers in a sync committee period and then only submitting them on-chain. This was error-prone and would try to complete uploading all the execution headers even though it might be failing on-chain


relayer/relays/beacon/header/header.go
```diff
@@ -307,42 +323,57 @@ func (h *Header) SyncExecutionHeaders(ctx context.Context) error {
...
 }
 
-func (h *Header) syncInterimFinalizedUpdate(ctx context.Context, lastSyncedSlot, newCheckpointSlot uint64) error {
+func (h *Header) syncInterimFinalizedUpdate(ctx context.Context, lastSyncedSlot, newCheckpointSlot uint64) (scale.Update, error) {
+	currentPeriod := h.protocol.ComputeSyncPeriodAtSlot(lastSyncedSlot)
+
 	// Calculate the range that the interim finalized header update may be in
 	minSlot := newCheckpointSlot - h.protocol.SlotsPerHistoricalRoot
-	maxSlot := lastSyncedSlot + h.protocol.SlotsPerHistoricalRoot
+	maxSlot := ((currentPeriod + 1) * h.protocol.SlotsPerHistoricalRoot) - h.protocol.Settings.SlotsInEpoch // just before the new sync committee boundary
```
yrong, line 359: Nice! Would it be better to add an extra check something like:
```
ensure!(lastSyncedSlot<minSlot<maxSlot<lastSyncedSlot + h.protocol.SlotsPerHistoricalRoot)
```

yrong (approved):
Awesome!

vgeddes (approved):

alistair-singh (approved):

PR #1204, opened 2024-05-21
Filter bad beefy keys

yrong:
Resolves: SNO-982

Requires: https://github.com/Snowfork/go-substrate-rpc-client/pull/63

More context in https://snowfork.slack.com/archives/G01BGMGLAC9/p1715552817012679 and https://matrix.to/#/!gxqZwOyvhLstCgPJHO:matrix.parity.io/$5BUJTxwRg8WQjjOhqUfvCHUhB3BPomiH9wVowUAJEZc?via=matrix.parity.io&via=parity.io&via=matrix.org


yrong:
Tested the beefy relayer with our own relaychain node, generating mmr proof is ok now. 

Met an issue the validator root generated from our code is not the same as polkadot.  So the validatation failed with [InvalidValidatorProof](https://github.com/Snowfork/snowbridge/blob/ac87b56f605524c290c04b4764b55dcca7e0baf5/contracts/src/BeefyClient.sol#L274) on beefy LC.

So it seems we still need some filter logic for the bad keys/addresses, prefer to do it in upper-stream code, so add https://github.com/Snowfork/polkadot-sdk/pull/146 accordingly.

vgeddes (approved):

yrong:
Met another issue when sync from a history block

```
./relayer/build/snowbridge-relay scan-beefy --beefy-block 20876565 --validator-set-id 556 --polkadot-url ws://127.0.0.1:9944
{"@timestamp":"2024-05-22T16:24:08.54544+08:00","level":"info","logger":"stdlib","message":"2024/05/22 16:24:08 Connecting to ws://127.0.0.1:9944..."}
{"@timestamp":"2024-05-22T16:24:13.050988+08:00","endpoint":"ws://127.0.0.1:9944","level":"info","message":"Connected to chain","metaVersion":14}
{"@timestamp":"2024-05-22T16:24:13.051034+08:00","beefy-block":20876565,"fast-forward-depth":100,"level":"info","message":"Connected to relaychain.","polkadot-url":"ws://127.0.0.1:9944","validator-set-id":556}
{"@timestamp":"2024-05-22T16:24:18.580698+08:00","error":"scan safe commitments: fetch block: type *types.ExtrinsicSignatureV4 does not support Decodeable interface and could not be decoded field by field, error: expected more bytes, but could not decode any more","level":"fatal","message":"Unhandled error"}
```

relayer/go.mod
```diff
@@ -111,3 +113,5 @@ require (
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 	rsc.io/tmplfunc v0.0.3 // indirect
 )
+
+// replace github.com/snowfork/go-substrate-rpc-client/v4 => /Users/yangrong/Projects/go-substrate-rpc-client
```
alistair-singh, line 117: remove commented code.

relayer/relays/beefy/parameters.go
```diff
@@ -97,19 +99,26 @@ func cleanSignature(input types.BeefySignature) (uint8, [32]byte, [32]byte) {
...
-	for i, rawAddress := range r.Validators {
+	var leaves [][]byte
+	var err error
+	var invalidAddress []string
+	for _, rawAddress := range r.Validators {
 		address, err := rawAddress.IntoEthereumAddress()
 		if err != nil {
-			return nil, fmt.Errorf("convert to ethereum address: %w", err)
+			invalidAddress = append(invalidAddress, util.BytesToHexString(rawAddress[:]))
+			leaves = append(leaves, make([]byte, 0))
```
alistair-singh, line 109: Why do we append empty leaf on error as opposed to leaving it out?

alistair-singh (approved):

relayer/relays/beefy/parameters.go
yrong: The `BeefyEcdsaToEthereum` impl on Polkadot will map the invalid AuthorityId to default(an empty address). More context in https://snowfork.slack.com/archives/G01BGMGLAC9/p1716289299447179?thread_ts=1715552817.012679&cid=G01BGMGLAC9

web/packages/test/config/beefy-relay.json
```diff
@@ -3,9 +3,8 @@
     "polkadot": {
       "endpoint": "ws://127.0.0.1:9944"
     },
-    "beefy-activation-block": 0,
-    "fast-forward-depth": 20,
-    "update-period": 0
+    "fast-forward-depth": 3000,
+    "update-period": 3000
```
yrong, line 7: These are the config for production to only syncing mandatory/handover commitments.

The downside is for local setup(Rococo) it will also skip the non-handover commitment.  IMO it should be fine because the session length for Rococo is 10 blocks which is not long.

web/packages/test-helpers/src/generateBeefyCheckpoint.ts
```diff
@@ -47,23 +47,42 @@ async function generateBeefyCheckpoint() {
...
+                .update(Buffer.from(publicKey))
+                .digest()
+            addrs.push(publicKeyHashed.slice(12))
+        } catch (err) {
+            console.log(err)
+        }
     }
 
-    const tree = createMerkleTree(addrs)
+    console.log("valid authority length is:" + addrs.length)
```
vgeddes, line 68: This code doesn't look like its used for anything other than to print the length of the validator set. So maybe we can remove it?

web/packages/test/config/beefy-relay.json
vgeddes: Cool, though we store prod config in this repo: https://github.com/Snowfork/snowbridge-infra-suite/

I've added a PR to move the prod config there:
https://github.com/Snowfork/snowbridge-infra-suite/pull/6

So you can revert these changes to the local testnet config.

web/packages/test-helpers/src/generateBeefyCheckpoint.ts
yrong: It's for debug purpose to print all invalid addresses, would prefer to keep it.

web/packages/test/config/beefy-relay.json
yrong: https://github.com/Snowfork/snowbridge/pull/1204/commits/1df8be1578dc3a189199d55d5db953870d636450

vgeddes (approved):

PR #1206, opened 2024-05-24
Sync finalized headers on demand

claravanstaden:
Removing syncing finalized headers as they finalize. Only sync sync committee updates and interim finalized headers if the gap between headers are larger than 8192.

When receiving a message in the Ethereum relayer, check if there is a relevant on-chain finalized header to use for the ancestry proof, otherwise sync the finalized header in a batch call together with the message and its proof.

Resolves: [SNO-1015](https://linear.app/snowfork/issue/SNO-1015)


relayer/chain/parachain/writer.go
```diff
@@ -69,7 +69,7 @@ func (wr *ParachainWriter) Start(ctx context.Context, eg *errgroup.Group) error
 	return nil
 }
 
-func (wr *ParachainWriter) BatchCall(ctx context.Context, extrinsic string, calls []interface{}) error {
```
claravanstaden, line 72: Change function to allow calling different extrinsics.

relayer/relays/beacon/header/header.go
```diff
@@ -71,7 +71,7 @@ func (h *Header) Sync(ctx context.Context, eg *errgroup.Group) error {
 
 	log.Info("starting to sync finalized headers")
 
-	ticker := time.NewTicker(time.Second * 10)
+	ticker := time.NewTicker(time.Minute * 1)
```
claravanstaden, line 74: The beacon relayer can check for new updates every minute, since we only sync sync committee updates. Can even make this even 10 minutes or so.

relayer/relays/beacon/header/header.go
```diff
@@ -187,7 +186,7 @@ func (h *Header) SyncFinalizedHeader(ctx context.Context) error {
 		}
 	}
 
-	return h.updateFinalizedHeaderOnchain(ctx, update)
```
claravanstaden, line 190: No longer sync the latest header on-chain.

yrong:
We may also need some revamp for [delivery_cost](https://github.com/Snowfork/polkadot-sdk/blob/d95751bcd6956602b10eea848f067864da37846e/bridges/snowbridge/pallets/inbound-queue/src/lib.rs#L269) to take cost of finality update into consideration.

claravanstaden:
> We may also need some revamp for [delivery_cost](https://github.com/Snowfork/polkadot-sdk/blob/d95751bcd6956602b10eea848f067864da37846e/bridges/snowbridge/pallets/inbound-queue/src/lib.rs#L269) to take cost of finality update into consideration.

That's a fair point. We haven't implemented relayer rewards for the consensus relayers like the beacon and beefy relayer, right? Perhaps we should consider implementing that soon and handle it with that work? The reason why I think so, instead of handling it in this PR is:

- The goal with this PR is to make the beacon relayer less expensive to run.
- We are trying to limit this change to an off-chain change.

@vgeddes would like your input here too.

alistair-singh:
> > We may also need some revamp for [delivery_cost](https://github.com/Snowfork/polkadot-sdk/blob/d95751bcd6956602b10eea848f067864da37846e/bridges/snowbridge/pallets/inbound-queue/src/lib.rs#L269) to take cost of finality update into consideration.
> 
> That's a fair point. We haven't implemented relayer rewards for the consensus relayers like the beacon and beefy relayer, right? Perhaps we should consider implementing that soon and handle it with that work? The reason why I think so, instead of handling it in this PR is:
> 
> * The goal with this PR is to make the beacon relayer less expensive to run.
> * We are trying to limit this change to an off-chain change.
> 
> @vgeddes would like your input here too.

The execution relayer is sending a consensus update to "speed up" the transaction so that the message is included is delivered now instead of waiting 27 hours or so for the next consensus update. If the the execution relayer is paying for speed up then we should not re-imburse them as part of the delivery cost because technically they could wait ~27 hours and get free consensus delivery.

The issue is that waiting 27 hours is too long. We maybe should have a way to make the beacon relayer do intervals in between sync committee updates such that every 4 hours a consensus update occurs (matching BEEFY). That way the execution relayer can wait 4 hours or pay more out there own pocket to speed up the message.

claravanstaden:
@alistair-singh great point! Added over here: https://github.com/Snowfork/snowbridge/pull/1206/commits/11aadb8cdf04215ea85724e6889fabb51aca53c4

claravanstaden:
@yrong @alistair-singh I added a flag to disable sending a beacon header with the message, in case a channel wants to wait for the beacon relayer finalized header: https://github.com/Snowfork/snowbridge/pull/1206/commits/60d029cbf9bf05b956fcfe8b1e01265f1d80d949

claravanstaden:
@alistair-singh improved error messages here: https://github.com/Snowfork/snowbridge/pull/1206/commits/0c22608c3d77f608fb9c7eda530fbe644194e827 The fallback methods hid the initial error.

Validation for config files here: https://github.com/Snowfork/snowbridge/pull/1206/commits/4a76149383e09190b234cc22312ebf9325fda46d It looks like Viper doesn't have this built in: https://github.com/spf13/viper/issues/702#issuecomment-1865885801

relayer/chain/parachain/writer.go
```diff
@@ -80,13 +80,13 @@ func (wr *ParachainWriter) BatchCall(ctx context.Context, extrinsic string, call
...
 		for k := range slicedCalls {
-			call, err := wr.prepCall(extrinsic, slicedCalls[k])
+			call, err := wr.prepCall(extrinsic[k], slicedCalls[k])
 			if err != nil {
 				return err
 			}
 			encodedCalls[k] = *call
 		}
-		err := wr.WriteToParachainAndRateLimit(ctx, "Utility.batch_all", encodedCalls)
+		err := wr.WriteToParachainAndWatch(ctx, "Utility.batch_all", encodedCalls)
```
yrong, line 89: Will this call get blocked?

claravanstaden: Yes, it will wait until the `batchall` request has been completed and it is a block.

relayer/relays/beacon/header/header.go
```diff
@@ -73,7 +75,7 @@ func (h *Header) Sync(ctx context.Context, eg *errgroup.Group) error {
 
 	log.Info("starting to sync finalized headers")
 
-	ticker := time.NewTicker(time.Second * 10)
+	ticker := time.NewTicker(time.Minute * 2)
```
yrong, line 78: Will this interval too long for local setup?

claravanstaden: I don't think so, no, because a finalized header will be [synced every 316 blocks](https://github.com/Snowfork/snowbridge/pull/1206/files#diff-92cc554d5b7218e7576839ca25a3b71ae0bbfa8200552beaf703a95a1a0da2a3R554) (approximately 6 times in a sync committee period).

relayer/relays/execution/main.go
```diff
@@ -198,6 +195,41 @@ func (r *Relay) Start(ctx context.Context, eg *errgroup.Group) error {
...
+	log.WithFields(logrus.Fields{
+		"finalized_slot": proof.FinalizedPayload.Payload.FinalizedHeader.Slot,
+		"finalized_root": proof.FinalizedPayload.FinalizedHeaderBlockRoot,
+		"message_slot":   proof.HeaderPayload.Header.Slot,
+	}).Debug("Batching finalized header update with message")
+
+	extrinsics := []string{"EthereumBeaconClient.submit", "EthereumInboundQueue.submit"}
+	payloads := []interface{}{proof.FinalizedPayload.Payload, inboundMsg}
+	// Batch the finalized header update with the inbound message
+	err := r.writer.BatchCall(ctx, extrinsics, payloads)
```
yrong, line 225: Seems we use execution relayer to submit the beacon finality update. 

I would suggest not do like this, in recent change of beefy relayer we add a one-shot command to sync finality update.  Maybe we can follow the same pattern.

More context in https://github.com/Snowfork/snowbridge/pull/1217#issue-2326816048

yrong (approved):
+1

PR #1216, opened 2024-05-30
Split BEEFY relayer into two separate relayers

vgeddes:
The BEEFY relayer is overly complicated, slow, and brittle as it tries to do too much with a core synchronization algorithm that is inappropriate:
1. Downloading all block data, including extrinsics, which is hugely wasteful
2. Syncing mandatory commitments always
3. Optionally sync non-mandatory commitments, but only if these commitments are near the finalized head, and try to skip every N non-mandatory commitments.

Instead, we should split the relayer into services:
1. A mandatory commitment relayer, which simply syncs mandatory commitments
2. A one-shot command which syncs the commitment for the latest non-mandatory finalized BEEFY head, *only if* the beefy light client is in the same validator session as finalized beefy head on Polkadot.

This PR currently implements (1). We should also implement (2) to allow anyone to sync instantly so they can have their messages relayed. 

Suggested design for (2): Use GSRPC `beefy_getFinalizedHead` to get the latest finalized BEEFY head, then `chain_getBlock`, which I suspect should contain a beefy justification.

Fixes: SNO-1017


relayer/relays/beefy/ethereum-writer.go
```diff
@@ -79,32 +96,49 @@ func (wr *EthereumWriter) Start(ctx context.Context, eg *errgroup.Group, request
 	return nil
 }
 
-func (wr *EthereumWriter) submit(ctx context.Context, task Request) error {
+type FilterMode struct {
+	MandatoryCommitmentsOnly bool
+	All                      bool
+	DiscardDepth             uint64
+}
```
yrong, line 103: Unused?

vgeddes: Yeah it's unused 👍

relayer/relays/beefy/polkadot-listener.go
```diff
@@ -41,7 +40,7 @@ func (li *PolkadotListener) Start(
 	}
 	li.beefyAuthoritiesKey = storageKey
 
-	requests := make(chan Request)
+	requests := make(chan Request, 1)
```
yrong, line 43: Is the channel size 1 here nessessary?

relayer/relays/beefy/polkadot-listener.go
```diff
@@ -162,8 +123,22 @@ func (li *PolkadotListener) queryBeefyAuthorities(blockHash types.Hash) ([]subst
...
+	if !ok {
+		return authoritySet, fmt.Errorf("beefy authoritySet not found")
+	}
+
+	return authoritySet, nil
+}
+
+func (li *PolkadotListener) queryBeefyNextAuthoritySet(blockHash types.Hash) (BeefyAuthoritySet, error) {
+	var nextAuthoritySet BeefyAuthoritySet
 	storageKey, err := types.CreateStorageKey(li.conn.Metadata(), "MmrLeaf", "BeefyNextAuthorities", nil, nil)
```
yrong, line 142: Seems `BeefyMmrLeaf` is the pallet prefix for polkadot network.

relayer/relays/beefy/scanner.go
```diff
@@ -15,61 +15,113 @@ import (
...
+	}
+
+	finalizedHeader, err := fetchFinalizedBeefyHeader()
+	if err != nil {
+		emitError(err)
+		return
+	}
+	current := startBlock
+	for {
+		log.Info("foo")
```
yrong, line 79: Remove this line.

relayer/relays/beefy/scanner.go
```diff
@@ -15,61 +15,113 @@ import (
...
 			return
 		}
 
+		if sessionIndex > currentSessionIndex {
+			log.Info("BOO")
+			currentSessionIndex = sessionIndex
+		} else {
+			current++
+			continue
+		}
```
yrong, line 119: I don't think we need to get Session index to check whether it's handover commitment or not.

The request itself is enough to tell as `nextValidatorSetID(in proof) == validatorSetID(in commitment) + 1`

relayer/relays/beefy/scanner.go
```diff
@@ -154,99 +200,45 @@ func scanCommitments(ctx context.Context, api *gsrpc.SubstrateAPI, startBlock ui
...
 	}
 }
 
+type ScanProvableCommitmentsResult struct {
+	SignedCommitment types.SignedCommitment
+	MMRProof         merkle.SimplifiedMMRProof
+	BlockHash        types.Hash
+	Depth            uint64
+	Error            error
+}
```
yrong, line 240: Unused?

relayer/relays/beefy/scanner.go
```diff
@@ -15,61 +15,113 @@ import (
...
+
+		var sessionIndex uint32
+		_, err = api.RPC.State.GetStorage(sessionCurrentIndexKey, &sessionIndex, blockHash)
+		if err != nil {
+			emitError(fmt.Errorf("fetch session index: %w", err))
 			return
 		}
 
+		if sessionIndex > currentSessionIndex {
+			log.Info("BOO")
```
yrong, line 114: Remove this debug.

relayer/relays/beefy/polkadot-listener.go
vgeddes: Setting the channel buffer size to 0 makes it a blocking channel, meaning that when `polkadot-listener` writes to the channel it will block until `ethereum-writer` has completed submitting the last commitment.

Setting the channel buffer size to 1 allows `polkadot-listener` to post a new commitment, and then immediately start looking for the next one without waiting for `ethereum-writer`.

If it takes `polkadot-listener` a long time to scan for the next commitment, this optimization could reduce the overall sync time.

relayer/relays/beefy/scanner.go
vgeddes: Isn't that true for all commitments though?

relayer/relays/beefy/polkadot-listener.go
vgeddes: This function is actually unused now.

relayer/relays/beefy/scanner.go
yrong: Sorry, it's `validatorSetID(in LC) < validatorSetID(in commitment)` then I'd assume it can be treated as a handover commitment, even it's not the session boundry on relay chain.

relayer/relays/beefy/polkadot-listener.go
```diff
@@ -161,17 +122,3 @@ func (li *PolkadotListener) queryBeefyAuthorities(blockHash types.Hash) ([]subst
...
-	ok, err := li.conn.API().RPC.State.GetStorage(storageKey, &nextAuthoritySet, blockHash)
-	if err != nil {
-		return nextAuthoritySet, err
-	}
-	if !ok {
-		return nextAuthoritySet, fmt.Errorf("beefy nextAuthoritySet not found")
-	}
-
-	return nextAuthoritySet, nil
-}
```
yrong, line 177: I need this function for https://github.com/Snowfork/snowbridge/pull/1215, but no need to revert here I can address it in that PR.

yrong (approved):
+1

PR #1217, opened 2024-05-31
Sync beefy commitment on demand

yrong:
A one-shot relayer at this moment and how we would integrate the one-shot relayer in future as follows from @vgeddes:

1. User on Polkadot signals that they want to speed up their message by depositing a fee in a special account on AssetHub/BridgeHub
2. An offchain service witnesses that deposit, and invokes the one-shot relayer

### DIFF
--- a/relayer/cmd/sync_beefy_commitment.go
+++ b/relayer/cmd/sync_beefy_commitment.go
@@ -20,7 +20,6 @@ func syncBeefyCommitmentCmd() *cobra.Command {
 	}
 
 	cmd.Flags().String("config", "/tmp/snowbridge/beefy-relay.json", "Path to configuration file")
-	cmd.MarkFlagRequired("config")
 	cmd.Flags().String("private-key", "", "Ethereum private key")
 	cmd.Flags().String("privateKeyFile", "", "The file from which to read the private key")
 	cmd.Flags().Uint64P("relay-block", "b", 0, "Relay block number which contains a Parachain message")
--- a/relayer/cmd/sync_beefy_commitment.go
+++ b/relayer/cmd/sync_beefy_commitment.go
@@ -1,0 +1,65 @@
+package cmd
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/sirupsen/logrus"
+	"github.com/snowfork/snowbridge/relayer/chain/ethereum"
+	"github.com/snowfork/snowbridge/relayer/relays/beefy"
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+)
+
+func syncBeefyCommitmentCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "sync-latest-beefy-commitment",
+		Short: "Sync beefy commitment on demand",
+		Args:  cobra.ExactArgs(0),
+		RunE:  SyncBeefyCommitmentFn,
+	}
+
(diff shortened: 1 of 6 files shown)


relayer/cmd/sync_beefy_commitment.go
```diff
@@ -0,0 +1,64 @@
...
+		Use:   "sync-latest-beefy-commitment",
+		Short: "Sync beefy commitment on demand",
+		Args:  cobra.ExactArgs(0),
+		RunE:  SyncBeefyCommitmentFn,
+	}
+
+	cmd.Flags().String("config", "/tmp/snowbridge/beefy-relay.json", "Path to configuration file")
+	cmd.Flags().String("private-key", "", "Ethereum private key")
+	cmd.Flags().String("privateKeyFile", "", "The file from which to read the private key")
+	cmd.Flags().Uint64P("relay-block", "b", 0, "Relay block number which contains a Parachain message")
```
vgeddes, line 25: I prefer this naming, since in the context of beefy, its always the relay chain 

```suggestion
	cmd.Flags().Uint64P("block-number", "b", 0, "Relay block number which contains a Parachain message")
```

relayer/cmd/sync_beefy_commitment.go
```diff
@@ -0,0 +1,64 @@
...
+	cmd := &cobra.Command{
+		Use:   "sync-latest-beefy-commitment",
+		Short: "Sync beefy commitment on demand",
+		Args:  cobra.ExactArgs(0),
+		RunE:  SyncBeefyCommitmentFn,
+	}
+
+	cmd.Flags().String("config", "/tmp/snowbridge/beefy-relay.json", "Path to configuration file")
+	cmd.Flags().String("private-key", "", "Ethereum private key")
+	cmd.Flags().String("privateKeyFile", "", "The file from which to read the private key")
```
vgeddes, line 24:
```suggestion
	cmd.Flags().String("private-key-file", "", "The file from which to read the private key")
```

relayer/cmd/sync_beefy_commitment.go
```diff
@@ -0,0 +1,64 @@
...
+	keypair, err := ethereum.ResolvePrivateKey(privateKey, privateKeyFile)
+	if err != nil {
+		return err
+	}
+
+	relay, err := beefy.NewRelay(&config, keypair)
+	if err != nil {
+		return err
+	}
+	relayBlockNumber, _ := cmd.Flags().GetUint64("relay-block")
```
vgeddes, line 61:
```suggestion
	relayBlockNumber, _ := cmd.Flags().GetUint64("block-number")
```

vgeddes:
Still need to adjust for the change in `parameters.go`: https://github.com/Snowfork/snowbridge/pull/1216

Only submissions for mandatory commitments need to include an MMR leaf and proof

relayer/cmd/sync_beefy_commitment.go
yrong: https://github.com/Snowfork/snowbridge/pull/1217/commits/98e3e74a58119717134797cfecec8a8895e43124

yrong:
> Only submissions for mandatory commitments need to include an MMR leaf and proof. The one-shot relay doesn't need to include those.

I'd prefer to include the MMR leaf and proof for the oneshot case, I vaguely recalled it won't cost much to add some more data to calldata without validation. And with that it's possible to extend the oneshot case for even the mandatory commitments.

vgeddes (approved):

PR #1219, opened 2024-06-04
Fix for e2e setup

yrong:


claravanstaden (commented):
Works for me! There was a problem with geth but `--state.scheme=hash` fixes it, thanks!

I did notice that the beacon relayer restarts with `write to parachain: parachain write status was dropped, invalid, usurped or finality timed out`, but when I look at Polkadot JS, the beacon headers are imported. I wonder if it is a problem with the Go substrate client. Did you see the same thing?

yrong:
> Did you see the same thing?

Yeah, same issue from my side. Still investigating and guess something related to the fast slot hack?

vgeddes (approved):
Looks great, 🚀

claravanstaden (approved):
Nice, thanks for looking into this!

PR #1222, opened 2024-06-06
Improve monitoring

yrong:
- Change the way to calculate latency to use finality update between source chain and light client
- Add insufficient balance alarm as a seperate topic, only monitor the substrate accounts(the ethereum accounts will be monitored by Tenderly)
- More refactoring


vgeddes (approved):

alistair-singh (approved):
+1

PR #1223, opened 2024-06-06
reduce polling interval

vgeddes:
The beefy relayer is spamming Infura 2x a second, which I suspect is causing intermittent rate limiting errors.


alistair-singh (approved):
Nice catch!

PR #1224, opened 2024-06-10
Generate beefy test fixture by command

yrong:
### Context 

Currently the beefy test fixture is some mock data and it's challenging to generate the data. Since beefy is already enabled on polkadot,  seems make more sense to just use the data from production.

Another use case is to test the compatibility with `polkadot-sdk`, after updating from upperstream repo we make an E2E test, if syncing beefy still works then it should be fine for the upgrade.

Resolves: https://linear.app/snowfork/issue/SNO-1057


vgeddes:
Sounds like a good idea :+1:

PR #1233, opened 2024-06-17
Decrease Beacon alarm latency

claravanstaden:
The beacon client is expected to be no more than 4 epochs behind. Update the latency setting to reflect the current config.

### DIFF
--- a/web/packages/api/src/status.ts
+++ b/web/packages/api/src/status.ts
@@ -67,8 +67,8 @@ export type Sovereign = { name: string; account: string; balance: bigint; type: 
 export const BlockLatencyThreshold = {
     // Syncing beefy finality update every 4 hours(2400 blocks) so we set 3000 blocks at most.
     ToEthereum: 3000,
-    // Syncing beacon finality update every 4 hours(1200 blocks) so we set 1500 blocks at most.
-    ToPolkadot: 1500,
+    // Syncing beacon finality update every 6.4 minutes(32 blocks) so we set 128 blocks (4 epochs) at most.
+    ToPolkadot: 128,
 }
 
 export const InsufficientBalanceThreshold = {


claravanstaden:
@yrong is this the correct way to edit the alarm? How is it deployed so the alarm is updated?

yrong:
> is this the correct way to edit the alarm
Yeah, that's the right configuration to change.

>  How is it deployed so the alarm is updated?
Currently we don't have a deploy script for that, there is a cron script here  https://github.com/Snowfork/snowbridge/blob/main/web/packages/operations/src/cron.ts
which I ran on one of our ec2 instance.

yrong (approved):
+1

web/packages/api/src/status.ts
```diff
@@ -67,8 +67,8 @@ export type Sovereign = { name: string; account: string; balance: bigint; type:
 export const BlockLatencyThreshold = {
     // Syncing beefy finality update every 4 hours(2400 blocks) so we set 3000 blocks at most.
     ToEthereum: 3000,
-    // Syncing beacon finality update every 4 hours(1200 blocks) so we set 1500 blocks at most.
-    ToPolkadot: 1500,
+    // Syncing beacon finality update every 6.4 minutes(32 blocks) so we set 128 blocks (4 epochs) at most.
+    ToPolkadot: 128,
```
vgeddes, line 71: Shouldn't we add a small buffer to give the relayer some time to sync?

claravanstaden: Technically, we should never be more than 3 epochs behind. So I have added 1 epoch buffer in the alarm to allow for syncing.  I have been keeping an eye on the status dashboard and hardly see the latency go above 20 minutes. The alarm latency here is set to 25.6 minutes. 

What do you think is an acceptable latency?

claravanstaden:
> > is this the correct way to edit the alarm
> 
> Yeah, that's the right place to change.
> 
> > How is it deployed so the alarm is updated?
> 
> Currently we don't have a deploy script for that, there is a cron script here https://github.com/Snowfork/snowbridge/blob/main/web/packages/operations/src/cron.ts which I ran on one of our ec2 instance as a dameon.

Cool! @yrong will you please update this on the EC2 instance?

PR #1234, opened 2024-06-18
Set minimum Weth balance to be transferred

claravanstaden:
Sets the minimum Weth balance that needs to be transferred, to match the on-chain `min_balance` value.

Resolves: [SNO-1056](https://linear.app/snowfork/issue/SNO-1056)

### DIFF
--- a/web/packages/api/src/environment.ts
+++ b/web/packages/api/src/environment.ts
@@ -85,7 +85,7 @@ export const SNOWBRIDGE_ENV: { [id: string]: SnowbridgeEnvironment } = {
                     {
                         id: "WETH",
                         address: "0x87d1f7fdfEe7f651FaBc8bFCB6E086C278b77A7d",
-                        minimumTransferAmount: 1n,
+                        minimumTransferAmount: 15_000_000_000_000n,
                     },
                 ],
             },
@@ -214,7 +214,7 @@ export const SNOWBRIDGE_ENV: { [id: string]: SnowbridgeEnvironment } = {
                     {
                         id: "WETH",
                         address: "0xfff9976782d46cc05630d1f6ebab18b2324d6b14",
-                        minimumTransferAmount: 1n,
+                        minimumTransferAmount: 15_000_000_000_000n,
                     },
                     {
                         id: "vETH",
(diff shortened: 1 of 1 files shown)


claravanstaden:
@alistair-singh please check if I added it in the right place. I added to the Ethereum settings, with destination AH.

web/packages/api/src/environment.ts
```diff
@@ -317,7 +317,7 @@ export const SNOWBRIDGE_ENV: { [id: string]: SnowbridgeEnvironment } = {
                     {
                         id: "WETH",
                         address: "0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2",
-                        minimumTransferAmount: 1n,
+                        minimumTransferAmount: 15_000_000_000_000n,
```
alistair-singh, line 320: Needs to be added on line as well.
https://github.com/Snowfork/snowbridge/pull/1234/files#diff-0c721ef12fe6fb53cd42b43781e67cd2000aa1927affa80e849954d3cdca186cR371

web/packages/api/src/environment.ts
```diff
@@ -183,7 +183,7 @@ export const SNOWBRIDGE_ENV: { [id: string]: SnowbridgeEnvironment } = {
                     {
                         id: "WETH",
                         address: "0xfff9976782d46cc05630d1f6ebab18b2324d6b14",
-                        minimumTransferAmount: 1n,
+                        minimumTransferAmount: 15_000_000_000_000n,
```
alistair-singh, line 186: This line also: 

https://github.com/Snowfork/snowbridge/pull/1234/files#diff-0c721ef12fe6fb53cd42b43781e67cd2000aa1927affa80e849954d3cdca186cL217

web/packages/api/src/environment.ts
alistair-singh: It doesnt have to `AH->Ethereum` direction as well. But currently you have added it to `AH->Ethereum` and not `Ethereum->AH`. So it reversed in the PR. So location in the config is configured with the tokens it can receive on its chain i.e `...->AH`. So you would go to AH location, and modify the minimum there if you want to set the minimum `Ethereum->AH`.

I would do it in both directions because I want to unify this config at some point. The issue is that some tokens such as MYTH are only allowed to be sent to Mythos and not back to Ethereum. But that is an edge case. Ideally I want to get this from onchain first if it is set.

claravanstaden: Thanks for explaining! Let me know if this works: https://github.com/Snowfork/snowbridge/pull/1234/commits/dbb15c5416fa64806e220bbbbf78d7d5d271546e

alistair-singh (approved):

PR #1240, opened 2024-06-20
Mev protect

yrong:
Just mirror https://github.com/Snowfork/snowbridge/pull/1239 with some config change for the local setup.


claravanstaden (approved):
Awesome, thanks! 🙏🏻

PR #1242, opened 2024-06-21
Check sync committee supermajority

claravanstaden:
Verify that the finalized update was signed by the sync committee super majority. If not, don't send the update to the light client and just log a warning. Wait for the next update (should be within the next checking interval because Lodestar updates the finalized update as better one becomes available.

Resolves: [SNO-1025](https://linear.app/snowfork/issue/SNO-1025)


alistair-singh (approved):
+1

PR #1243, opened 2024-06-21
Upgrade solc version for release 1.0.0

claravanstaden:
Building the contracts on branch `release-v1.0.0` fails with:

```
Claras-MacBook-Pro-2:contracts claravanstaden$ forge build
[⠊] Compiling...
[⠆] Compiling 93 files with 0.8.25
[⠰] Solc 0.8.25 finished in 163.67ms
Error: 
Compiler run failed:
Error (5333): Source file requires different compiler version (current compiler is 0.8.25+commit.b61c2a91.Darwin.appleclang) - note that nightly builds are considered to be strictly less than the released version
 --> src/Agent.sol:3:1:
  |
3 | pragma solidity 0.8.23;
  | ^^^^^^^^^^^^^^^^^^^^^^^
```

Resolves: [SNO-1063](https://linear.app/snowfork/issue/SNO-1063)


claravanstaden:
I have for now just downgraded the solc version from `0.8.25` to `0.8.23` so that we can built the relayer on branch `release-v1.0.0`.

I am not sure how they should be upgraded to `0.8.25` and if that is the better fix (and the impact on deployment), maybe you can weigh in @yrong.

claravanstaden:
Edit: Looks like it should just be upgraded: https://github.com/Snowfork/snowbridge/pull/1207 It was updated in foundry.toml but not in the contracts.

PR #1244, opened 2024-06-21
Handle beacon state download error

claravanstaden:
The `REGEN_ERROR_NO_SEED_STATE` Lodestar log error may be returning a `500` instead of a `404`, which means https://github.com/Snowfork/snowbridge/pull/1241 will not work, so I need to catch all beacon state download errors.


yrong (approved):
+1

yrong:
Seems we used the same lodestar node to get beacon state from.  It reminds me of the previous discussion to avoid the single point of failure. Is that better to add another beacon state node(a mock lodestar or prysm) as fallback?

claravanstaden:
@yrong yes, we probably need to add another Lodestar node as a backup and load balance over them.

PR #1249, opened 2024-06-28
Monitoring threshold adjustment

yrong:
Resolves: https://linear.app/snowfork/issue/SNO-1067


claravanstaden (approved):

web/packages/api/src/status.ts
```diff
@@ -72,8 +72,8 @@ export const BlockLatencyThreshold = {
 }
 
 export const InsufficientBalanceThreshold = {
-    // Minimum as 10 DOT
-    Substrate: 100_000_000_000,
+    // Minimum as 100 DOT
+    Substrate: 1_000_000_000_000,
```
alistair-singh, line 76: Let's make it 300 DOT.

alistair-singh (changes requested):

web/packages/api/src/status.ts
```diff
@@ -72,8 +72,8 @@ export const BlockLatencyThreshold = {
...
     // Minimum as 0.1 Ether
     Ethereum: 100_000_000_000_000_000,
```
alistair-singh, line 78: I changed all the tenderly alerts to alert if lower than 0.3 ETH. https://linear.app/snowfork/issue/SNO-1067#comment-d4b0c3c5

Maybe we want to do the same here.

web/packages/api/src/status.ts
yrong: https://github.com/Snowfork/snowbridge/pull/1249/commits/3c60b685a6024edb5e377a51bdcaf87fb0118d18

web/packages/api/src/status.ts
yrong: https://github.com/Snowfork/snowbridge/pull/1249/commits/5c0137ed53f9ccc67d73730c0de626a1d3f5d6d2

alistair-singh (approved):

PR #1269, opened 2024-08-15
Dockerize relayer

yrong:
Resolves: https://linear.app/snowfork/issue/SNO-978


yrong:
Though I'd still prefer to release relayer in binary format, more details in https://linear.app/snowfork/issue/SNO-978#comment-03b533ff

Btw: the `docker-compose` is just used for a local setup to test the image built,  for production usage 3rd team can choose whatever tools to orchestrate multiple containers(e.g. k8s or podman)

relayer/vector.yaml
```diff
@@ -0,0 +1,19 @@
+---
```
vgeddes, line 1: I don't think we need this logging stuff in the container

relayer/.env.example
```diff
@@ -0,0 +1,6 @@
+PARACHAIN_RELAY_ASSETHUB_ETH_KEY=0x3646505e08a0f3a61417d91db18f2911b2d17c01563044f3e2f106bf36679a6a
+EXECUTION_RELAY_ASSETHUB_SUB_KEY=//ExecutionRelayAssetHub
```
vgeddes, line 2: rather leave these unset in case someone mistakenly puts real secrets here

yrong: https://github.com/Snowfork/snowbridge/pull/1269/commits/756c0bc23e80896c64e0e8da9d85c9acbf998b76

relayer/vector.yaml
yrong: https://github.com/Snowfork/snowbridge/pull/1269/commits/1175db4498249bb05a86ae711bcde1f6be0f50ea

yrong:
In this PR we also release docker image from CI, met an issue when manually invoke the CI to test, it happens when pushing the image to `ghcr.io` with error details ```ERROR: unauthorized: access token has insufficient scopes```

More details in https://github.com/Snowfork/snowbridge/actions/runs/10472492391/job/29002087164.
 
I'd assume should be fine invoked from other team member with sufficient scopes.

vgeddes (approved):

PR #1273, opened 2024-08-23
Fix e2e environment and smoketests

alistair-singh:
All these fixes were made from clean build of nix environment and a clean checkout.

For some reason the `polkadot-js-api` does not obey the `--nonce -1` flag anymore so I needed to remove it so that the testnet starts up correctly.

Tests Fixes:
- [x] register_token
- [x] send_token
- [x] transfer_token
- [ ] set_price_params
- [ ] set_token_transfer_fees
- [ ] send_token_to_penpal


flake.nix
```diff
@@ -37,7 +37,7 @@
                     ps
 
                     # typescript
-                    python3
+                    python311
```
alistair-singh, line 40: This is required to build `lodestar`. Python version 3.12 deprecates some required packages.

yrong:
> For some reason the polkadot-js-api does not obey the --nonce -1 flag anymore so I needed to remove it so that the testnet starts up correctly.

@alistair-singh  I can't reproduce the issue on my laptop with your branch, `--nonce -1` does work and the testnet starts up  as expected.

Btw: for local test I would suggest to use a short epoch length with https://github.com/Snowfork/polkadot-sdk/pull/170 included and I made another PR in https://github.com/Snowfork/snowbridge/pull/1274 based on yours.

alistair-singh:
> > For some reason the polkadot-js-api does not obey the --nonce -1 flag anymore so I needed to remove it so that the testnet starts up correctly.
> 
> @alistair-singh I can't reproduce the issue on my laptop with your branch, `--nonce -1` does work and the testnet starts up as expected.


Ok I will revert my changes to the nonce.

yrong (approved):
Awesome!

vgeddes (approved):

PR #1275, opened 2024-08-25
Improve solidity code for PNA 

vgeddes:
Various improvements and cleanups

Changes:
* More consistent and descriptive naming style
* Simplified token registry
* AgentExecutor.sol: Remove deprecated `execute` method
* Foreign tokens are now owned by the gateway rather than the assethub agent. This simplifies code without sacrificing any security & robustness
* Moved more asset-specific code to Assets.sol

FIXES: SNO-1133

### DIFF
--- a/contracts/src/Assets.sol
+++ b/contracts/src/Assets.sol
@@ -247,8 +247,8 @@ library Assets {
             // The funds will be minted into the receiver's account on AssetHub
             if (destinationAddress.isAddress32()) {
                 // The receiver has a 32-byte account ID
-                ticket.payload = SubstrateTypes.SendTokenToAssetHubAddress32(
-                    token, destinationAddress.asAddress32(), $.assetHubReserveTransferFee, amount
+                ticket.payload = SubstrateTypes.SendForeignTokenToAssetHubAddress32(
+                    foreignID, destinationAddress.asAddress32(), $.assetHubReserveTransferFee, amount
                 );
             } else {
                 // AssetHub does not support 20-byte account IDs
@@ -261,12 +261,22 @@ library Assets {
             if (destinationAddress.isAddress32()) {
                 // The receiver has a 32-byte account ID
                 ticket.payload = SubstrateTypes.SendForeignTokenToAddress32(
-                    foreignID, destinationChain, destinationAddress.asAddress32(), destinationChainFee, amount
+                    foreignID,
+                    destinationChain,
+                    destinationAddress.asAddress32(),
+                    $.assetHubReserveTransferFee,
+                    destinationChainFee,
+                    amount
                 );
             } else if (destinationAddress.isAddress20()) {
                 // The receiver has a 20-byte account ID
                 ticket.payload = SubstrateTypes.SendForeignTokenToAddress20(
-                    foreignID, destinationChain, destinationAddress.asAddress20(), destinationChainFee, amount
+                    foreignID,
+                    destinationChain,
+                    destinationAddress.asAddress20(),
+                    $.assetHubReserveTransferFee,
+                    destinationChainFee,
+                    amount
                 );
             } else {
                 revert Unsupported();
--- a/contracts/src/Assets.sol
+++ b/contracts/src/Assets.sol
@@ -16,7 +16,7 @@ import {Address} from "./utils/Address.sol";
 import {AgentExecutor} from "./AgentExecutor.sol";
 import {Agent} from "./Agent.sol";
 import {Call} from "./utils/Call.sol";
-import {ERC20} from "./ERC20.sol";
+import {Token} from "./Token.sol";
 
 /// @title Library for implementing Ethereum->Polkadot ERC20 transfers.
 library Assets {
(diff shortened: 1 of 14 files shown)


acatangiu:
> Foreign tokens are now owned by the gateway rather than the assethub agent.

What about already registered foreign assets? Are they still using the agent as admin account? Or do you plan to switch the registered admin on AH for them (need to check if that is possible)?

yrong:
Cool, much simple!  

Btw: It seems we want to support multi-hop for Ethereum->AH->Para direction. So is the fee token on Para still be relay token? If so will destination chain accept that? Can we always assume it's reserve-backed transfers beween AH and Para(more discussion in https://snowfork.slack.com/archives/G01BGMGLAC9/p1724456977726639thread_ts=1724432403.659139&cid=G01BGMGLAC9)?

My concern here is seems not easy to find a common pattern to support multi-hop which applies to all parachains.

contracts/src/Token.sol
```diff
@@ -28,33 +28,29 @@ import {ERC20Lib} from "./ERC20Lib.sol";
...
+    using TokenLib for TokenLib.Token;
 
-    error Unauthorized();
-
-    ERC20Lib.TokenStorage token;
-
-    address public immutable OWNER;
-
-    uint8 public immutable decimals;
+    TokenLib.Token token;
```
yrong, line 34: Would suggest to move this dynamic slots to the very end.

contracts/src/Assets.sol
```diff
@@ -159,6 +187,42 @@ library Assets {
...
+        uint128 destinationChainFee,
+        uint128 amount
+    ) internal returns (Ticket memory ticket) {
+        if (destinationChainFee == 0) {
+            revert InvalidDestinationFee();
+        }
+
+        Token(token).burn(sender, amount);
+
+        ticket.dest = destinationChain;
```
yrong, line 206: Is the dest here supposed to be `assetHubParaID`?

contracts/src/Assets.sol
```diff
@@ -159,6 +187,42 @@ library Assets {
...
+                foreignID, destinationChain, destinationAddress.asAddress32(), destinationChainFee, amount
+            );
+        } else if (destinationAddress.isAddress20()) {
+            // The receiver has a 20-byte account ID
+            ticket.payload = SubstrateTypes.SendForeignTokenToAddress20(
+                foreignID, destinationChain, destinationAddress.asAddress20(), destinationChainFee, amount
+            );
+        } else {
+            revert Unsupported();
+        }
```
yrong, line 221: Need to double check the execution cost on AH(i.e. `assetHubReserveTransferFee`) is included.

vgeddes:
> > Foreign tokens are now owned by the gateway rather than the assethub agent.
> 
> What about already registered foreign assets? Are they still using the agent as admin account? Or do you plan to switch the registered admin on AH for them (need to check if that is possible)?

This is an Ethereum-specific addition that does not involve AH. Currently there are only native tokens (aka Ethereum-native ERC20s) registered in the Gateway contract, and on AH too.

This change gives our gateway contract the ability to be aware of foreign assets on AH for example the Hydra or Kilt token in the ForeignAssets pallets.

So it does affect the code paths that involve ethereum-native assets such as WETH or WBTC.

acatangiu:
> > > Foreign tokens are now owned by the gateway rather than the assethub agent.
> > 
> > 
> > What about already registered foreign assets? Are they still using the agent as admin account? Or do you plan to switch the registered admin on AH for them (need to check if that is possible)?
> 
> This is an Ethereum-specific addition that does not involve AH. Currently there are only native tokens (aka Ethereum-native ERC20s) registered in the Gateway contract, and on AH too.
> 
> This change gives our gateway contract the ability to be aware of foreign assets on AH for example the Hydra or Kilt token in the ForeignAssets pallets.
> 
> So it does affect the code paths that involve ethereum-native assets such as WETH or WBTC.

I see. Sounds good! Thanks for the explanation 👍🏻

contracts/src/Assets.sol
vgeddes: Its meant to be the final destination for multi-hop transfers. 

I see now that we need to add a `SendForeignTokenToAssetHubAddress32` message type for PNA 👍

contracts/src/Token.sol
vgeddes: Moved fields around a bit

contracts/src/Assets.sol
vgeddes: Yeah good catch. I've reimplemented `_sendForeignToken` to mirror `_sendNativeToken`. So that PNA assets can be sent to assethub or a further hop from assethub.

contracts/src/Assets.sol
vgeddes: For now I'm assuming `assetHubReserveTransferFee` (which is configured to have considerable buffer) will cover costs for xcm execution on AH for both ENA and PNA flows.

We should probably rename the field to something like `assetHubXcmFee`

vgeddes:
> Cool, much simple!
> 
> Btw: It seems we want to support multi-hop for Ethereum->AH->Para direction. So is the fee token on Para still be relay token? If so will destination chain accept that? Can we always assume it's reserve-backed transfers beween AH and Para(more discussion in https://snowfork.slack.com/archives/G01BGMGLAC9/p1724456977726639thread_ts=1724432403.659139&cid=G01BGMGLAC9)?
> 
> My concern here is seems not easy to find a common pattern to support multi-hop which applies to all parachains.

One idea I had for the Ethereum->AH->Parachain flow regarding fees on the final destination parachain.

We already take DOT from AH's SA on BH to pay for XCM execution on AH. So we could go on step further and on AH we swap some of that DOT for the native token of the destination parachain.

yrong:
> go on step further and on AH we swap some of that DOT for the native token of the destination parachain.

Sounds good. Can we make the fee asset of destination Para configurable when register token [here](https://github.com/Snowfork/polkadot-sdk/blob/f1155b124813eb535e19e0ad5d2b31f0135b6a8b/bridges/snowbridge/pallets/system/src/lib.rs#L742)?  

So for chains does allow AH-DOT as fee asset, swap is not required.

contracts/src/Assets.sol
```diff
@@ -159,6 +194,88 @@ library Assets {
...
+        ticket.costs = _sendForeignTokenCosts(destinationChain, destinationChainFee, maxDestinationChainFee);
+
+        // Construct a message payload
+        if (destinationChain == $.assetHubParaID) {
+            // The funds will be minted into the receiver's account on AssetHub
+            if (destinationAddress.isAddress32()) {
+                // The receiver has a 32-byte account ID
+                ticket.payload = SubstrateTypes.SendTokenToAssetHubAddress32(
+                    token, destinationAddress.asAddress32(), $.assetHubReserveTransferFee, amount
+                );
```
yrong, line 252: I would always expect `foreignID` for PNA, not the token address.

https://github.com/Snowfork/polkadot-sdk/blob/f1155b124813eb535e19e0ad5d2b31f0135b6a8b/bridges/snowbridge/primitives/router/src/inbound/mod.rs#L63

yrong: Fixed in https://github.com/Snowfork/snowbridge/pull/1275/commits/57fa08c1ad101f46a12fc469b75d30bd76658290

yrong (approved):

PR #1279, opened 2024-08-30
Simulated upgrade test for production Gateway contract

vgeddes:
Uses Foundry's ability to run tests using storage forked from production networks

Requirements:
* Add `MAINNET_RPC_URL` to environment (I've updated Github Actions to include this secret)
* Most recent foundry/forge version

Tasks
- [*] Upgrade production gateway use forking
- [ ] Add sanity tests to ensure upgraded gateway is not compromised (storage corruption, etc).

### DIFF
--- a/contracts/test/Upgrade202408.t.sol
+++ b/contracts/test/Upgrade202408.t.sol
@@ -1,0 +1,99 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity 0.8.25;
+
+import {Test} from "forge-std/Test.sol";
+import {Strings} from "openzeppelin/utils/Strings.sol";
+import {console} from "forge-std/console.sol";
+
+import {IGateway} from "../src/interfaces/IGateway.sol";
+import {IInitializable} from "../src/interfaces/IInitializable.sol";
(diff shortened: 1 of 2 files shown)


contracts/test/Upgrade202408.t.sol
```diff
@@ -0,0 +1,99 @@
...
+            0x6eD05bAa904df3DE117EcFa638d4CB84e1B8A00C,
+            address(executor),
+            ParaID.wrap(1002),
+            0x03170a2e7597b7b7e3d84c05391d139a62b157e78786d8c082f29dcf4c111314,
+            10,
+            dot(2)
+        );
+
+        /// Mock call to Verification.verifyCommitment to bypass BEEFY verification
+        vm.mockCall(VERIFICATION_ADDR, abi.encodeWithSelector(Verification.verifyCommitment.selector), abi.encode(true));
```
yrong, line 79: Cool! Another option is to use `vm.startPrank(GatewayProxy)` and test upgrade function directly. https://github.com/Snowfork/snowbridge/commit/cf596f3d840b6348df625d384e35fec8f7256b3a

PR #1280, opened 2024-08-30
Log config file validations so that they are picked up by monitoring

alistair-singh:


relayer/cmd/run/beacon/command.go
```diff
@@ -62,7 +60,8 @@ func run(_ *cobra.Command, _ []string) error {
 
 	err = config.Validate()
 	if err != nil {
-		return fmt.Errorf("config file validation failed: %w", err)
```
alistair-singh, line 65: We only return errors from inner functions. This is the top level most function so we can safely log fatal here.

yrong (approved):
+1

claravanstaden (approved):
Nice catch!

claravanstaden:
@alistair-singh what config file item was missing?

PR #1286, opened 2024-09-13
Initialize for westend

yrong:

PR #1287, opened 2024-09-13
Daily Smoketests

claravanstaden:
- Moves the smoke test GA to the snowbridge repo instead of the polkadot-sdk repo (because our team will no longer work off a forked repo).
- Daily run of the smoketests on the production release polkadot-sdk branch

Resolves: [SNO-1080](https://linear.app/snowfork/issue/SNO-1080)

### DIFF
--- a/.github/workflows/smoke-tests.yml
+++ b/.github/workflows/smoke-tests.yml
@@ -4,11 +4,10 @@ on:
   push:
     branches:
       - main
-      - smoketests-on-snowbridge
   workflow_dispatch:
   schedule:
     # Runs at 8:00 AM every day
-    - cron: '0 9 * * *'
+    - cron: '0 8 * * *'
 
 jobs:
   smoketests:
@@ -23,7 +22,7 @@ jobs:
       - uses: actions/checkout@v2
         with:
           repository: snowfork/snowbridge
-          ref: smoketests-on-snowbridge # TODO update to "main" once approved
+          ref: main
       - run: rm -rf polkadot-sdk
       - uses: actions/checkout@v2
         with:


claravanstaden:
@yrong This should work for running smoketests daily. To see if the tests pass, I used our [polkadot-sdk fork](https://github.com/Snowfork/snowbridge/pull/1287/files#diff-ebcd3a3b6186ad4cf5cf2337ff84cc4692492610301fefb171509d671bf9f507R33). I think we should be using paritytech/polkadot-sdk branches https://github.com/paritytech/polkadot-sdk/tree/release-crates-io-v1.7.0 (currently in production) and https://github.com/paritytech/polkadot-sdk/tree/release-crates-io-v1.14.0 (being upgraded to in https://polkadot.polkassembly.io/referenda/1143). Neither of those branches have the Westend config? Perhaps we can add it?

yrong:
For westend I'll just run some smoke tests(maybe only a subset) against the live network.  That's why I initialized light client in https://github.com/Snowfork/snowbridge/pull/1286. 

I think should be fine switch to another network other than local testnet.  Just need some change to make sure [envs](https://github.com/Snowfork/snowbridge/blob/main/smoketest/src/constants.rs) point to the config on westend. 

I'll do that in https://github.com/Snowfork/snowbridge/pull/1286.

.github/workflows/smoketest.yml
```diff
@@ -0,0 +1,81 @@
...
+          repository: snowfork/snowbridge
+          ref: smoketests-on-snowbridge # TODO update to "main" once approved
+      - run: rm -rf polkadot-sdk
+      - uses: actions/checkout@v2
+        with:
+         # repository: paritytech/polkadot-sdk
+         # ref: release-crates-io-v1.7.0 # TODO change back to this branch once Westend code is merged upstream
+          repository: Snowfork/polkadot-sdk
+          ref: snowbridge
+          path: ./polkadot-sdk
```
yrong, line 34: Could we just run against master branch of paritytech/polkadot-sdk?

claravanstaden: We could, but it would be best to run against the branch that is in production. The reason for this is to catch imminent production changes. For example, the checked hash extension change was merged into master, which we knew, but it was also merged into the production branch, which caught us off-guard.

claravanstaden:
> envs

Sure, makes sense. I think we should have this PR as well, for the reason explained in https://github.com/Snowfork/snowbridge/pull/1287#discussion_r1764663726.

.github/workflows/smoketest.yml
yrong: > run against the branch that is in production.

Or run against the next/upcoming release for production? I would assume there is no need to test the current release which should definitely work?

claravanstaden: The current release changes all the time as changes are merged in (fixes/backported features). So it is definitely needed to run against the current production release branch.

yrong (approved):
+1

alistair-singh (approved):

vgeddes (approved):

PR #1288, opened 2024-09-13
Add Para threads to MMR Root

alistair-singh:
Fixes related to: https://github.com/paritytech/polkadot-sdk/pull/4751

We may need to time this to make sure we only include this when the PR release is live.

### DIFF
--- a/relayer/chain/relaychain/connection.go
+++ b/relayer/chain/relaychain/connection.go
@@ -198,7 +198,7 @@ func (co *Connection) FetchMMRLeafCount(relayBlockhash types.Hash) (uint64, erro
 //	Key: hash_twox_128("Paras") + hash_twox_128("Heads") + hash_twox_64(ParaId) + Encode(ParaId)
 const ParaIDOffset = 16 + 16 + 8
 
-// Fetch heads for all Paras. Included parachains and parathreads.
+// Fetch heads for all Paras. Included are parachains and parathreads.
 func (co *Connection) FetchParasHeads(blockHash types.Hash) ([]ParaHead, error) {
 	keyPrefix := types.CreateStorageKeyPrefix("Paras", "Heads")
 	keys, err := co.fetchKeys(keyPrefix, blockHash)
--- a/relayer/chain/relaychain/connection.go
+++ b/relayer/chain/relaychain/connection.go
@@ -131,50 +131,6 @@ type ParaHead struct {
 	Data   types.Bytes
 }
 
-func (co *Connection) fetchKeys(keyPrefix []byte, blockHash types.Hash) ([]types.StorageKey, error) {
-	const pageSize = 200
-	var startKey *types.StorageKey
-
-	if pageSize < 1 {
-		return nil, fmt.Errorf("page size cannot be zero")
-	}
-
-	var results []types.StorageKey
-	log.WithFields(log.Fields{
-		"keyPrefix": keyPrefix,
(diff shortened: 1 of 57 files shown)


relayer/relays/parachain/beefy-listener.go
```diff
@@ -258,8 +262,10 @@ func (li *BeefyListener) generateProof(ctx context.Context, input *ProofInput, h
 	// Generate a merkle proof for the parachain head with input ParaId
 	// and verify with merkle root hash of all parachain heads
 	// Polkadot uses the following code to generate merkle root from parachain headers:
-	// https://github.com/paritytech/polkadot/blob/2eb7672905d99971fc11ad7ff4d57e68967401d2/runtime/rococo/src/lib.rs#L706-L709
-	merkleProofData, err := CreateParachainMerkleProof(input.ParaHeads, input.ParaID)
+	// https://github.com/paritytech/polkadot-sdk/blob/d66dee3c3da836bcf41a12ca4e1191faee0b6a5b/polkadot/runtime/westend/src/lib.rs#L453-L460
+	// Truncate the ParaHeads to the 1024
+	// https://github.com/paritytech/polkadot-sdk/blob/d66dee3c3da836bcf41a12ca4e1191faee0b6a5b/polkadot/runtime/parachains/src/paras/mod.rs#L1305-L1311
+	merkleProofData, err := CreateParachainMerkleProof(input.ParaHeads[:MaxParaHeads], input.ParaID)
```
yrong, line 268: So are we always generate proof against a static length array(i.e. 1024)? Or could we use `min(len(paraHeads),1024)`?

alistair-singh: Good catch! Added in [1d4c23a](https://github.com/Snowfork/snowbridge/pull/1288/commits/1d4c23a45c97563a828951ac35a06026e6b11f22).

alistair-singh:
Tested successfully in rococo and it has fixed the issue.

yrong (approved):
+1

relayer/chain/relaychain/connection.go
```diff
@@ -292,7 +218,8 @@ func (co *Connection) fetchParaHeads(blockHash types.Hash) (map[uint32]ParaHead,
 		return nil, err
 	}
 
-	heads := make(map[uint32]ParaHead)
+	const numParas = 16
```
claravanstaden, line 221: Where does this number come from?

yrong: Just initial compacity of the array which can be extended as required.

relayer/chain/relaychain/connection.go
```diff
@@ -221,58 +191,15 @@ func (co *Connection) FetchMMRLeafCount(relayBlockhash types.Hash) (uint64, erro
 	return mmrLeafCount, nil
 }
 
-func (co *Connection) fetchKeys(keyPrefix []byte, blockHash types.Hash) ([]types.StorageKey, error) {
```
yrong, line 224: Seems just move position of this function, there is no internal change?

If so would suggest to leave it untouched to make diff a bit easier.

alistair-singh: addressed in https://github.com/Snowfork/snowbridge/pull/1288/commits/f6b090c7bcedde5f79d34eeacc4de80d21dc0614

vgeddes (approved):

PR #1290, opened 2024-09-16
Use paritytech polkadot-sdk with optional personal fork override

claravanstaden:
Uses your personal fork of the polkadot-sdk, instead of the Snowfork fork. PRs will be made directory to https://github.com/paritytech/polkadot-sdk instead of to the Snowfork fork and then to the original repo.

### DIFF
--- a/scripts/init.sh
+++ b/scripts/init.sh
@@ -5,7 +5,14 @@ set -eux
 echo "Checkout polkadot-sdk Snowfork fork"
 pushd ..
   if [ ! -d "polkadot-sdk" ]; then
-    git clone https://github.com/Snowfork/polkadot-sdk.git
+    gitUserName=$(git config user.name)
+    git clone "https://github.com/$gitUserName/polkadot-sdk.git"
+
+    if [ $? -ne 0 ]; then
+        echo "Error: The repository 'polkadot-sdk' does not exist for user '$gitUserName'. Please fork the 'polkadot-sdk' repository first before cloning."
+        exit 1
+    fi
+
     cd snowbridge && ln -sf ../polkadot-sdk polkadot-sdk
   fi
   pushd  polkadot-sdk
--- a/scripts/init.sh
+++ b/scripts/init.sh
@@ -4,15 +4,18 @@ set -eux
 
 echo "Checkout polkadot-sdk"
 pushd ..
-  if [ ! -d "polkadot-sdk" ]; then
+  if [[ -d ../polkadot-sdk ]] && (cd ../polkadot-sdk && git rev-parse --is-inside-work-tree) 2>&1 >/dev/nu
+  ls; then
+     echo "polkadot-sdk already exists"
+  else
     repoURL="${POLKADOT_SDK_REPO:-https://github.com/paritytech/polkadot-sdk.git}"
 
     git clone "$repoURL" polkadot-sdk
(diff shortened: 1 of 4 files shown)


acatangiu (approved):

scripts/init.sh
```diff
@@ -5,12 +5,10 @@ set -eux
 echo "Checkout polkadot-sdk Snowfork fork"
 pushd ..
   if [ ! -d "polkadot-sdk" ]; then
-    git clone https://github.com/Snowfork/polkadot-sdk.git
+    gitUserName=$(git config user.name)
```
vgeddes, line 8: I think it's going to simpler to just pass the whole repo url, instead of re-interpreting git variables. The git user.name var doesn't necessarily map to a GitHub account.

vgeddes: Also by default the paritytech polkadot-sdk repo should be cloned unless optionally overridden by a persona fork.

claravanstaden: Good point, changed over here: https://github.com/Snowfork/snowbridge/pull/1290/commits/13e8468e4cb51a494e54b2c4b4f55b0891f5ac26

scripts/init.sh
```diff
@@ -2,14 +2,16 @@
 
 set -eux
 
-echo "Checkout polkadot-sdk Snowfork fork"
+echo "Checkout polkadot-sdk"
 pushd ..
   if [ ! -d "polkadot-sdk" ]; then
```
vgeddes, line 7: Is there any value in having this `polkadot-sdk` symlink? I struggle to see any, and so the code here to manage it adds unnecessary complexity.

Typically modern editors such as VS Code and Zed allow one to create a virtual workspace containing many subprojects and it is easy to navigate between them.

vgeddes: Also we can improve the check to see if the clone already exists.

```
if [[ -d ../polkadot-sdk ]] && (cd ../polkadot-sdk && git rev-parse --is-inside-work-tree) 2>&1 >/dev/nu
ll; then 
   echo "repo already exists"
fi
```

claravanstaden: > Is there any value in having this `polkadot-sdk` symlink? I struggle to see any, and so the code here to manage it adds unnecessary complexity.
> 
> Typically modern editors such as VS Code and Zed allow one to create a virtual workspace containing many subprojects and it is easy to navigate between them.

It's mainly because we need to know where to build the binary for the E2E stack: https://github.com/Snowfork/snowbridge/blob/main/web/packages/test/scripts/build-binary.sh#L7 I guess we can be less opinionated and add an env for where the project is (likely ../polkadot-sdk, maybe that can be the default value). Let me know what you think about that.

yrong:
Do we want to add a [rust-toolchain.toml](https://github.com/Snowfork/polkadot-sdk/blob/snowbridge/rust-toolchain.toml) same as the forked repo?

vgeddes (approved):

claravanstaden:
> Do we want to add a [rust-toolchain.toml](https://github.com/Snowfork/polkadot-sdk/blob/snowbridge/rust-toolchain.toml) same as the forked repo?

Nah I don't think so, since we won't symlink the polkadot-sdk into the snowbridge repo anymore (it won't have any effect because the Rust code will be outside the snowbridge directory).

PR #1291, opened 2024-09-18
Smoke test on westend

yrong:
Resolves: https://linear.app/snowfork/issue/SNO-1089, https://linear.app/snowfork/issue/SNO-1097


web/packages/operations/ecosystem.config.js
```diff
@@ -3,7 +3,20 @@ module.exports = {
...
+      name: "transferToPolkadot",
+      node_args: "--require=dotenv/config",
+      script: "./dist/src/transfer_to_polkadot.js",
+      args: "cron"
+    },
+    {
+      name: "transferToEthereum",
+      node_args: "--require=dotenv/config",
+      script: "./dist/src/transfer_to_ethereum.js",
+      args: "cron"
```
yrong, line 19: Periodically run the transfer tests.

smoketest/tests/send_token.rs
```diff
@@ -89,7 +89,7 @@ async fn send_token() {
 
 	assert_eq!(receipt.status.unwrap().as_u64(), 1u64);
 
-	let wait_for_blocks = 100;
+	let wait_for_blocks = 1000;
```
claravanstaden, line 92: This increase will cause test failures to be super slow. Is it necessary?

smoketest/make-bindings.sh
```diff
@@ -15,13 +15,24 @@ command -v subxt || cargo install subxt-cli \
...
     --tag v0.37.0
 
-if ! lsof -Pi :11144 -sTCP:LISTEN -t >/dev/null; then
-    echo "substrate nodes not running, please start with the e2e setup and rerun this script"
-    exit 1
+eth_network="${ETH_NETWORK:-localhost}"
+polkadot_network="${POLKADOT_NETWORK:-localhost}"
+
+if [ "$polkadot_network" == "westend" ]; then
+  # Todo: There is no penpal nodes on westend yet
```
claravanstaden, line 22: Doesn't look like Penpal was deployed on Rococo either. Perhaps we can use another 3rd party parachain? Unfortunately it doesn't look like there are any other parachains besides the system chains. :/

web/packages/api/src/environment.ts
```diff
@@ -532,4 +532,138 @@ export const SNOWBRIDGE_ENV: { [id: string]: SnowbridgeEnvironment } = {
...
+                    {
+                        id: "MUSE",
+                        address: "0xb34a6924a02100ba6ef12af1c798285e8f7a16ee",
+                        minimumTransferAmount: 1n,
+                    },
+                ],
+            },
+            {
+                id: "muse",
+                name: "Muse",
```
claravanstaden, line 595: It doesn't look like Muse is deployed on Westend?

web/packages/operations/src/transfer_to_ethereum.ts
```diff
@@ -0,0 +1,89 @@
...
+    let env = "local_e2e"
+    if (process.env.NODE_ENV !== undefined) {
+        env = process.env.NODE_ENV
+    }
+    const snwobridgeEnv = environment.SNOWBRIDGE_ENV[env]
+    if (snwobridgeEnv === undefined) {
+        throw Error(`Unknown environment '${env}'`)
+    }
+
+    const { config } = snwobridgeEnv
```
claravanstaden, line 22:
```suggestion
    const { config } = snowbridgeEnv
```

web/packages/operations/src/transfer_to_polkadot.ts
```diff
@@ -0,0 +1,103 @@
...
+    let env = "local_e2e"
+    if (process.env.NODE_ENV !== undefined) {
+        env = process.env.NODE_ENV
+    }
+    const snwobridgeEnv = environment.SNOWBRIDGE_ENV[env]
+    if (snwobridgeEnv === undefined) {
+        throw Error(`Unknown environment '${env}'`)
+    }
+
+    const { config } = snwobridgeEnv
```
claravanstaden, line 23:
```suggestion
    const { config } = snowbridgeEnv
```

claravanstaden (commented):
Awesome! ✨

smoketest/tests/send_token.rs
yrong: It's nessessary for testnet/production as the finality latency is about 1 hour there, so we need to wait a bit longer.

Maybe better to load from some env and decrease it for local setup.

smoketest/make-bindings.sh
yrong: Yeah, it's not available on testnet. I think we can focus on AH first and test non-system parachain transfers later.

claravanstaden: Sounds good!

smoketest/tests/send_token.rs
claravanstaden: > Maybe better to load from some env and decrease it for local setup.

Good idea!

web/packages/api/src/environment.ts
yrong: Yeah, but would prefer to leave it here as placeholder as Muse will migrate to Westend sooner or later?

relayer/chain/relaychain/connection.go
```diff
@@ -130,37 +131,6 @@ type ParaHead struct {
...
-	// filter out parathreads
-	var parachainHeads []ParaHead
-	for _, v := range parachainIDs {
-		if head, ok := paraHeads[v]; ok {
-			parachainHeads = append(parachainHeads, head)
-		}
-	}
-	return parachainHeads, nil
-}
-
```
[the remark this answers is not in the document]
yrong: Yeah, though would prefer to address that in https://github.com/Snowfork/snowbridge/pull/1288

@alistair-singh I would assume the fix in #1288 will also be compatible with polkadot production?

web/packages/operations/src/transfer_to_ethereum.ts
yrong: https://github.com/Snowfork/snowbridge/pull/1291/commits/e5dec6b39d8383791c363e133b5ae705bccd6561

web/packages/api/src/environment.ts
alistair-singh: I think we should remove it as it will appear in the user interface. We can add it when the integration works.

web/packages/operations/src/transfer_token.ts
```diff
@@ -114,60 +116,71 @@ const monitor = async () => {
...
 }
 
-monitor()
-    .then(() => process.exit(0))
-    .catch((error) => {
-        console.error("Error:", error)
-        process.exit(1)
-    })
+if (process.argv.length != 3) {
+    console.error("Expected one argument with Enum from `start|cron|init`")
```
alistair-singh, line 172: `init` is not an option.

web/packages/operations/src/transfer_token.ts
```diff
@@ -8,22 +9,23 @@ import {
 } from "@snowbridge/api"
 import { WETH9__factory } from "@snowbridge/contract-types"
 import { Wallet } from "ethers"
+import cron from "node-cron"
 
 const monitor = async () => {
```
alistair-singh, line 14: Do we need to modify this file if we have added the two new files:
https://github.com/Snowfork/snowbridge/pull/1291/files#diff-1029e29210cb968c5341a4a12d25f34ae82529fc48371f1cfedce3f221f30439R1
and
https://github.com/Snowfork/snowbridge/pull/1291/files#diff-96a9d881f4ba07609e35a3cc2fffd6be3edcab8079fae643ef9cc66f299e6474R1

It seems like there is overlap. And if so I rather revert this file.

web/packages/operations/ecosystem.config.js
alistair-singh: How do we know if they fail?

web/packages/api/src/environment.ts
```diff
@@ -532,4 +532,138 @@ export const SNOWBRIDGE_ENV: { [id: string]: SnowbridgeEnvironment } = {
...
+                    maxConsumers: 16,
+                },
+                erc20tokensReceivable: [
+                    {
+                        id: "WETH",
+                        address: "0xfff9976782d46cc05630d1f6ebab18b2324d6b14",
+                        minimumTransferAmount: 15_000_000_000_000n,
+                    },
+                    {
+                        id: "vETH",
```
alistair-singh, line 582: Lets remove vETH until its registered. This case was for bifrost in rococo. They are not in Westend so I do not think we need this.

web/packages/api/src/environment.ts
```diff
@@ -532,4 +532,138 @@ export const SNOWBRIDGE_ENV: { [id: string]: SnowbridgeEnvironment } = {
...
+                        address: "0xc3d088842dcf02c13699f936bb83dfbbc6f721ab",
+                        minimumTransferAmount: 1n,
+                    },
+                    {
+                        id: "MUSE",
```
alistair-singh, line 587: Same for MUSE.

web/packages/api/src/environment.ts
```diff
@@ -532,4 +532,138 @@ export const SNOWBRIDGE_ENV: { [id: string]: SnowbridgeEnvironment } = {
...
+                        address: "0xfff9976782d46cc05630d1f6ebab18b2324d6b14",
+                        minimumTransferAmount: 1_000_000_000_000n,
+                    },
+                    {
+                        id: "vETH",
+                        address: "0xc3d088842dcf02c13699f936bb83dfbbc6f721ab",
+                        minimumTransferAmount: 1n,
+                    },
+                    {
+                        id: "MUSE",
```
alistair-singh, line 556: Not in westend.

web/packages/api/src/environment.ts
```diff
@@ -532,4 +532,138 @@ export const SNOWBRIDGE_ENV: { [id: string]: SnowbridgeEnvironment } = {
...
+                type: "ethereum",
+                destinationIds: ["assethub", "muse"],
+                erc20tokensReceivable: [
+                    {
+                        id: "WETH",
+                        address: "0xfff9976782d46cc05630d1f6ebab18b2324d6b14",
+                        minimumTransferAmount: 1_000_000_000_000n,
+                    },
+                    {
+                        id: "vETH",
```
alistair-singh, line 551: Not in westend.

web/packages/operations/src/transfer_token.ts
yrong: https://github.com/Snowfork/snowbridge/pull/1291/commits/583bc9094f2c9bd640420f632d05dd1f2f84b70b

web/packages/api/src/environment.ts
yrong: https://github.com/Snowfork/snowbridge/commit/59b82da59180ef118b6540a1a932d1de4d47a6e7

web/packages/operations/ecosystem.config.js
yrong: Yeah, I remove the wait logic there. Instead we depend on alarming on relayer logs & bridge/channel status as we do for the production.

Alarms forwarded to https://snowfork.slack.com/archives/C07BG558XE1

smoketest/tests/send_token.rs
yrong: https://github.com/Snowfork/snowbridge/pull/1291/commits/2d97157e224d69db5a35de1fad87b211c0ab3560

claravanstaden (approved):

PR #1294, opened 2024-09-26
Fix find on-chain checkpoint

claravanstaden:
There were 2 bugs in the find on-chain checkpoint method:
- The check constrains the checkpoint to be in the same sync committee period, which is not a requirement (a checkpoint can span over 2 sync committee periods, for example, header 8288 can prove slots 8192 - 8288 in period 1 and slots 96 - 1892 in period 0).
- The search does not traverse the ring buffer correctly (it only searches from the latest index to 0. It should wrap around and search to index + 1)

Added new config in Ansible scripts: https://github.com/Snowfork/snowbridge-infra-suite/pull/24/files#diff-0bdba71d49616abfe1a14b43ead54c6df64cdbfca68c8e4e5fe47bd4a1e3ae2bR29

### DIFF
--- a/relayer/relays/beacon/config/config.go
+++ b/relayer/relays/beacon/config/config.go
@@ -41,7 +41,9 @@ type SinkConfig struct {
 type ParachainConfig struct {
 	Endpoint             string `mapstructure:"endpoint"`
 	MaxWatchedExtrinsics int64  `mapstructure:"maxWatchedExtrinsics"`
-	HeaderRedundancy     uint64 `mapstructure:"headerRedundancy"`
+	// The max number of header in the FinalizedBeaconStateBuffer on-chain.
+	// https://github.com/paritytech/polkadot-sdk/blob/master/bridges/snowbridge/pallets/ethereum-client/src/types.rs#L23
+	HeaderRedundancy uint64 `mapstructure:"headerRedundancy"`
 }
 
 func (c Config) Validate() error {
--- a/relayer/relays/beacon/config/config.go
+++ b/relayer/relays/beacon/config/config.go
@@ -3,7 +3,6 @@ package config
 import (
 	"errors"
 	"fmt"
-	"github.com/snowfork/snowbridge/relayer/config"
 )
 
 type Config struct {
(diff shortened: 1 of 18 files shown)


relayer/relays/beacon/config/config.go
```diff
@@ -35,8 +34,14 @@ type BeaconConfig struct {
...
-	Parachain          config.ParachainConfig `mapstructure:"parachain"`
-	UpdateSlotInterval uint64                 `mapstructure:"updateSlotInterval"`
+	Parachain          ParachainConfig `mapstructure:"parachain"`
+	UpdateSlotInterval uint64          `mapstructure:"updateSlotInterval"`
+}
+
+type ParachainConfig struct {
+	Endpoint             string `mapstructure:"endpoint"`
+	MaxWatchedExtrinsics int64  `mapstructure:"maxWatchedExtrinsics"`
+	HeaderRedundancy     uint64 `mapstructure:"headerRedundancy"`
```
claravanstaden, line 44: Setting to capture the max headers in the ringbuffer: // https://github.com/paritytech/polkadot-sdk/blob/master/bridges/snowbridge/pallets/ethereum-client/src/lib.rs#L75

relayer/relays/beacon/header/header.go
```diff
@@ -523,32 +523,32 @@ func (h *Header) findLatestCheckPoint(slot uint64) (state.FinalizedHeader, error
...
 	}
 	startIndex := uint64(lastIndex)
-	endIndex := uint64(0)
+	endIndex := startIndex + 1
 
 	syncCommitteePeriod := h.protocol.Settings.SlotsInEpoch * h.protocol.Settings.EpochsPerSyncCommitteePeriod
-	slotPeriodIndex := slot / syncCommitteePeriod
-
-	for index := startIndex; index >= endIndex; index-- {
+	totalStates := syncCommitteePeriod * h.protocol.HeaderRedundancy // Total size of the circular buffer,
```
yrong, line 529: Is total size of the circular buffer be 256*20=5120?

```
totalStates := h.protocol.Settings.EpochsPerSyncCommitteePeriod * h.protocol.HeaderRedundancy
```

web/packages/test/config/beacon-relay.json
```diff
@@ -18,7 +18,8 @@
   "sink": {
     "parachain": {
       "endpoint": "ws://127.0.0.1:11144",
-      "maxWatchedExtrinsics": 8
+      "maxWatchedExtrinsics": 8,
+      "headerRedundancy": 20
```
yrong, line 22: Can we just move the new config `headerRedundancy` to `config.SpecSettings`, then not much change required.

claravanstaden: That's true, but it is not really related to the Ethereum spec settings, which is why I didn't add it there.

relayer/relays/beacon/config/config.go
vgeddes: Can we add a comment in the code describing the config item?

vgeddes (approved):
Looks good just plz add a comment describing the new config item.

relayer/relays/beacon/config/config.go
claravanstaden: Sure, https://github.com/Snowfork/snowbridge/pull/1294/commits/1956d878b677cd2a3b74b1d7313c27b3c15e6e87.

PR #1299, opened 2024-10-07
feat: adding KILT Protocol to snowbridge environment

Dudleyneedham:
Added KILT Protocol to the Snowbridge environment configuration for integration with the Snowbridge app

### DIFF
--- a/web/packages/api/src/environment.ts
+++ b/web/packages/api/src/environment.ts
@@ -250,7 +250,27 @@ export const SNOWBRIDGE_ENV: { [id: string]: SnowbridgeEnvironment } = {
                         minimumTransferAmount: 10_000_000_000_000_000n,
                     },
                 ],
-            },
+            },{
+                id: "RILT",
+                name: "RILT",
+                type: "substrate",
+                destinationIds: ["assethub"],
+                paraInfo: {
+                  paraId: 4504,
+                  destinationFeeDOT: 0n,
+                  skipExistentialDepositCheck: false,
+                  addressType: "32byte",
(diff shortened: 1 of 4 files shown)


web/packages/api/src/environment.ts
```diff
@@ -341,9 +377,35 @@ export const SNOWBRIDGE_ENV: { [id: string]: SnowbridgeEnvironment } = {
...
+                  destinationFeeDOT: 0n,
+                  skipExistentialDepositCheck: false,
+                  addressType: "32byte",
+                  decimals: 15,
+                  maxConsumers: 16,
+                },
+                erc20tokensReceivable: [
+                  {
+                    id: "PILT",
+                    address: "0x99E743964C036bc28931Fb564817db428Aa7f752",
```
alistair-singh, line 404: What would happen if someone tries to directly send this token to PILT from ethereum not using polar path?

alistair-singh (changes requested):
Also you could add the KILT parachain endpoint to config.

For Example:
https://github.com/Snowfork/snowbridge/pull/1299/files#diff-0c721ef12fe6fb53cd42b43781e67cd2000aa1927affa80e849954d3cdca186cR766

But if you want to manage it outside the context than that is also fine.

web/packages/api/src/environment.ts
alistair-singh: Also maybe this should live outside the context/environment config? As it is only used outside the API package.

alistair-singh (approved):

PR #1300, opened 2024-10-07
Contracts V2

vgeddes:
Tasks:
- [x] V2 types
- [x] V2 inbound messaging
- [x] V2 outbound messaging
- [x] Refactor to support both V1 & V2
- [x] Re-implement native token registration for V2 & Kusama compatibility
- [x] Auto-wrap ether for Ethereum->Polkadot
- [x] Auto-unwrap ether for Polkadot->Ethereum
- [ ] Generate encoded XCM fragments for token registration
- [x] Split `IGateway` into `IGatewayV1` and `IGatewayV2`
- [x] Remove duplicated error definitions scattered throughout code
- [x] Unit tests
- [ ] Code documentation


contracts/src/utils/SparseBitmap.sol
```diff
@@ -0,0 +1,16 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity 0.8.25;
+
+struct SparseBitmap {
```
yrong, line 4: What's the benifit comparing with https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/structs/BitMaps.sol?

contracts/src/types/V2.sol
```diff
@@ -0,0 +1,93 @@
...
+    // Message nonce
+    uint64 nonce;
+    // Commands
+    Command[] commands;
+}
+
+struct Command {
+    uint8 kind;
+    bytes payload;
+}
```
yrong, line 29: Seems we remove `maxDispatchGas` from the Command,  so we depend on some api service dryRun https://docs.metamask.io/wallet/reference/eth_estimategas/  for the gas cost?

contracts/src/Gateway.sol
yrong, line 1: Could we leave V1 code untouched and add GatewayV2 separately? We can even add a new branch for that. I would assume GatewayV1 will still be running for a long time side by side with GatewayV2?

contracts/src/interfaces/IGateway.sol
```diff
@@ -11,11 +13,26 @@ interface IGateway {
...
+    );
 
     // Emitted when an outbound message has been accepted for delivery to a Polkadot parachain
-    event OutboundMessageAccepted(ChannelID indexed channelID, uint64 nonce, bytes32 indexed messageID, bytes payload);
+    event OutboundMessageAccepted(
+        ChannelID indexed channelID,
+        uint64 nonce,
+        bytes32 indexed messageID,
+        bytes payload
+    );
```
yrong, line 35: Seems fmt changed?  Would suggest to remove the custom fmt section from `foundry.toml` which makes diff review a bit easier.

contracts/src/types/V2.sol
```diff
@@ -0,0 +1,95 @@
...
+
+// Inbound message from a Polkadot parachain (via BridgeHub)
+struct InboundMessage {
+    // origin
+    bytes32 origin;
+    // Message nonce
+    uint64 nonce;
+    // Commands
+    Command[] commands;
+}
```
yrong, line 24: Do we need `messge_id` or `tx_hash`  generated on source chain for tracing purpose?

vgeddes: I think `nonce` is sufficient for tracing purposes, when one takes this flow into account.

1. BH receives XCM
2. BH processes XCM and emits `MessageProcessed(xcm_id, nonce)`
3. Gateway receives message and logs the nonce.

Then the UX layer and other indexers can use the nonce logged in 2 and 3 to link the message to the XCM topic or hash.

contracts/src/interfaces/IGateway.sol
vgeddes: yeah I will try to fix this.

contracts/src/types/V2.sol
vgeddes: latest commit updates the Command to include a `gas` param.

contracts/src/utils/SparseBitmap.sol
vgeddes: Ah nice find. Will try to incorporate it if its better. Though I don't want to start relying too much on OpenZeppelin.

contracts/src/Gateway.sol
yrong: Or make GatewayV2 inherit from GatewayV1, something like

```
contract GatewayV2 is Gateway {
function submitV2() external
...
}
```

vgeddes: Unfortunately the multiple inheritance approach just didn't work for various reasons.

We also can't leave V1 untouched as there is a lot of duplicate code between V1 and V2 that needed to be factored out.

In the latest commit I've accommodated V1 and V2 using libraries.

contracts/src/v2/Calls.sol
```diff
@@ -0,0 +1,145 @@
...
+        address assetHubAgent = Functions.ensureAgent(Constants.ASSET_HUB_AGENT_ID);
+        WETH9(payable(WETH_ADDRESS)).deposit{value: msg.value}();
+        IERC20(WETH_ADDRESS).safeTransfer(assetHubAgent, msg.value);
+
+        $.outboundNonce = $.outboundNonce + 1;
+
+        bytes memory payload =
+            SubstrateTypes.encodePayloadV2(ticket.origin, ticket.assets, ticket.xcm);
+
+        emit IGateway.OutboundMessageAccepted($.outboundNonce, msg.value, payload);
```
yrong, line 100: Cool, A lot more simple now!

With `ticket.xcm` does that mean for V2 there is no need to construct xcm on chain any more?

It's the responsibility of relayer to validate the xcm and check profitable or not.

On BH we just decode the xcm with
```VersionedXcm::<()>::decode_with_depth_limit(
						MAX_XCM_DECODE_DEPTH,
						&mut data,
					)
``` and forward it to AH transparently.

File: contracts/src/v2/Calls.sol [hunk 6]
```diff
@@ -0,0 +1,145 @@
...
+
+    // Submit an outbound message to Polkadot, after taking fees
+    function _submitOutbound(Ticket memory ticket) internal {
+        CoreStorage.Layout storage $ = CoreStorage.layout();
+
+        // Ensure outbound messaging is allowed
+        _ensureOutboundMessagingEnabled();
+
+        // Wrap provided ether and transfer to AssetHub agent
+        address assetHubAgent = Functions.ensureAgent(Constants.ASSET_HUB_AGENT_ID);
```
yrong, line 91: Since we know agent of assethub do exists in production. Is this check nessessary?

contracts/src/v2/Calls.sol
```diff
@@ -0,0 +1,145 @@
...
+        WETH9(payable(WETH_ADDRESS)).deposit{value: msg.value}();
+        IERC20(WETH_ADDRESS).safeTransfer(assetHubAgent, msg.value);
```
yrong, line 93: Why not use raw ether here?

contracts/src/SubstrateTypes.sol
```diff
@@ -196,4 +205,64 @@ library SubstrateTypes {
...
+    //   assets: Vec<Asset>
+    //   xcm: Vec<u8>
+    // }
+    //
+    function encodePayloadV2(address origin, bytes[] memory assets, bytes memory xcm)
+        internal
+        pure
+        returns (bytes memory)
+    {
+        return bytes.concat(abi.encodePacked(origin), VecAsset(assets), VecU8(xcm));
```
yrong, line 223: Is `assets` required here?  I would assume On BH we just need the encoded xcm and forward to AH transparently?

contracts/src/v2/Calls.sol
```diff
@@ -0,0 +1,145 @@
...
+
+        $.outboundNonce = $.outboundNonce + 1;
+
+        bytes memory payload =
+            SubstrateTypes.encodePayloadV2(ticket.origin, ticket.assets, ticket.xcm);
```
yrong, line 98: Cool, A lot more simple now!

With `ticket.xcm` does that mean for V2 there is no need to construct xcm on chain any more?

UI will build the xcm and it's the responsibility of relayer to validate the xcm and check the relay is profitable or not.

On BH we just decode the xcm with
```
VersionedXcm::<()>::decode_with_depth_limit(
    MAX_XCM_DECODE_DEPTH,
    &mut data,
)
``` 
and forward it to AH transparently.

contracts/src/v2/Calls.sol
yrong, line 100: Do we need to add an extra `reward_address` into the event?

contracts/src/v2/Calls.sol
vgeddes: Yeah, exactly.

Though for registration of native tokens, we need to control the XCM and so we need to build it on chain. See more details in https://github.com/Snowfork/snowbridge/pull/1300/commits/149c3e425bbcaf023a9564977921bd355cf129fd.

contracts/src/SubstrateTypes.sol
vgeddes: Yes the on-chain code needs to know what assets are being transferred so that it can lock/burn them. I added more documentation on this in https://github.com/Snowfork/snowbridge/pull/1300/commits/149c3e425bbcaf023a9564977921bd355cf129fd.

contracts/src/v2/Calls.sol
vgeddes: We need to wrap the ether as it is going to be sent across the bridge to BH in the form of WETH.

contracts/src/v2/Calls.sol
vgeddes: Yeah the check is not necessary, I'll add a TODO.

contracts/src/v2/Calls.sol
vgeddes: That mechanism is only required for the Polkadot->Ethereum flow which involves delivery receipts.

For the Ethereum->Polkadot direction we can use the address of the relayer that delivered the message to BH.

contracts/src/v2/Calls.sol
yrong: > Though for registration of native tokens, control the XCM and so we need to build it on chain

Still not sure fully understand why `ticket.assets` is required here, why not just build the xcm off chain?

@vgeddes I've removed the `MessageConverter` completely in https://github.com/yrong/polkadot-sdk/pull/3 which I'd assume is not in use any more, correct me if I'm wrong.

vgeddes: > Still not sure fully understand why ticket.assets is required here, why not just build the xcm off chain?

When the message gets to BH, we convert the `payload.assets` to instructions like ReserveAssetDeposited,etc, and then append the instructions in `payload.xcm`.

Or in other words, BH builds final XCM using:
```
finalXCM = concat(
assets_to_xcm(payload.asssets),
UniversalOrigin(Ethereum)
DescendOrigin(payload.origin)
payload.xcm
)
```

We can't trust the user to build the entire xcm offchain as they could mint any assets they want. Rather, they can only supply a fragment of the xcm.

yrong: > We can't trust the user to build the entire xcm offchain as they could mint any assets they want.

Instead of convert the assets to xcm, can we just do some basic check on BH to verify the xcm is constructed based upon the assets correctly(e.g.  `ensure!(xcm contains ReserveAssetDeposited(asset)|WithdrawAsset(asset)`)?

So the point here is validate rather than construct.

yrong: Btw: for transfer use case I'm not sure `DescendOrigin(userOrigin)` make sense. For PNA we need to burn the asset locked in Ethereum sovereign on AH. That's why we need to change origin to `UniversalOrigin(GlobalConsensus(network))` before `WithdrawAsset` [here](https://github.com/paritytech/polkadot-sdk/blob/21b3a46b11c15ff3915557201657322f42527a2b/bridges/snowbridge/primitives/router/src/inbound/mod.rs#L440-L441).

contracts/src/Gateway.sol
```diff
@@ -493,199 +391,182 @@ contract Gateway is IGateway, IInitializable, IUpgradable {
...
+    uint256 public constant DISPATCH_OVERHEAD_GAS_V2 = 32_000;
 
-        // Ensure outbound messaging is allowed
-        _ensureOutboundMessagingEnabled(channel);
+    /// @dev Submit a message from Polkadot for verification and dispatch
+    /// @param message A message produced by the OutboundQueue pallet on BridgeHub
+    /// @param leafProof A message proof used to verify that the message is in the merkle tree committed by the OutboundQueue pallet
+    /// @param headerProof A proof that the commitment is included in parachain header that was finalized by BEEFY.
+    /// @param rewardAddress Account on BH to credit delivery rewards
+    function v2_submit(
```
yrong, line 419: should it also be in IGateway.sol? Meanwhile we also need another function `isRelayed(nonce)` to check if the nonce has been relayed.

vgeddes: Good catch!

contracts/src/Gateway.sol
```diff
@@ -493,199 +391,182 @@ contract Gateway is IGateway, IInitializable, IUpgradable {
...
-        if ($.mode != OperatingMode.Normal || ch.mode != OperatingMode.Normal) {
-            revert Disabled();
+    function v2_dispatch(InboundMessageV2 calldata message) internal returns (bool) {
+        for (uint256 i = 0; i < message.commands.length; i++) {
+            if (gasleft() * 63 / 64 < message.commands[i].gas + DISPATCH_OVERHEAD_GAS_V2)
+            {
+                assembly {
+                    invalid()
+                }
+            }
```
yrong, line 455: Is this check required for V2?

vgeddes: Yes it is. It there to ensure that relayers pass enough gas to allow each command in the msg to dispatch.

vgeddes: So we need to ensure that the outbound-queue pallet continues to compute the maximum dispatch gas for each command in the msg.

contracts/src/v2/Types.sol
```diff
@@ -0,0 +1,101 @@
...
+    bytes32 origin;
+    // Message nonce
+    uint64 nonce;
+    // Commands
+    Command[] commands;
+}
+
+struct Command {
+    uint8 kind;
+    uint256 gas;
```
yrong, line 19: Gas is `uint64`.

contracts/src/v2/Calls.sol
yrong: > it is too late because we have already changed the state on the Ethereum side

Just curious what will happen in this case. Is this something acceptable assuming we will implement error handing mechanism any way? The errored xcm will get trapped on BH and we allow end user to repair it? 

Something like we allow user to add fee for message pending from Substrate to Ethereum direction, assuming there is no relayer who will pick it up if it's not profitable.

contracts/src/v2/Calls.sol
```diff
@@ -0,0 +1,211 @@
...
+    error ChannelDoesNotExist();
+    error InvalidChannelUpdate();
+    error AgentExecutionFailed(bytes returndata);
+    error InvalidAgentExecutionPayload();
+    error InvalidConstructorParams();
+    error AlreadyInitialized();
+    error TokenNotRegistered();
+    error InvalidAsset();
+
+    address public constant WETH_ADDRESS = 0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2;
```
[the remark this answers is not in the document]
vgeddes: yeah, why would it need to be mutable?

contracts/src/Constants.sol
```diff
@@ -0,0 +1,21 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2023 Snowfork <hello@snowfork.com>
+pragma solidity 0.8.25;
+
+import {ParaID, ChannelID} from "./v1/Types.sol";
+
+library Constants {
+    ParaID constant ASSET_HUB_PARA_ID = ParaID.wrap(1000);
+    bytes32 constant ASSET_HUB_AGENT_ID =
+        0x81c5ab2571199e3188135178f3c2c8e2d268be1313d029b30f534fa579b69b79;
```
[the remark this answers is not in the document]
vgeddes: That's the case right now, and I think I prefer the convention vs configurability approach.

contracts/src/v2/Calls.sol
alistair-singh: > Just curious what will happen in this case. Is this something acceptable assuming we will implement error handing mechanism any way?

I think if the message is queued and if the message is profitable, than once submitted to BH we should not fail for any reason (such as validating the xcm for example) and we should try to always push the assets to AH where they can get trapped if the user passes invalid xcm. And we should trust that we constrain the origin to stop users from doing things that they cannot within the xcm, instead of validating.

contracts/src/Gateway.sol
```diff
@@ -493,199 +391,182 @@ contract Gateway is IGateway, IInitializable, IUpgradable {
...
 
-    /// @dev Ensure that the specified parachain has a channel allocated
-    function _ensureChannel(ChannelID channelID) internal view returns (Channel storage ch) {
-        ch = CoreStorage.layout().channels[channelID];
-        // A channel always has an agent specified.
-        if (ch.agent == address(0)) {
-            revert ChannelDoesNotExist();
-        }
+    // See docs for `IGateway.sendMessage`
+    function sendMessage(bytes calldata xcm, bytes[] calldata assets) external payable {
```
alistair-singh, line 504: I think we should add a MultiAddress parameter that can be the asset claimer. If provided then we can use this account with the `SetClaimer` XCM instruction. If it is zero/empty address then we SetClaimer to the origin(msg.sender).

contracts/src/Gateway.sol
```diff
@@ -493,199 +391,182 @@ contract Gateway is IGateway, IInitializable, IUpgradable {
...
+        if (!_verifyCommitment(commitment, headerProof)) {
+            revert IGateway.InvalidProof();
         }
 
-        // Generate a unique ID for this message
-        bytes32 messageID = keccak256(abi.encodePacked(channelID, channel.outboundNonce));
+        bool success = v2_dispatch(message);
 
-        emit IGateway.OutboundMessageAccepted(channelID, channel.outboundNonce, messageID, ticket.payload);
+        emit IGateway.InboundMessageDispatched(message.nonce, success, rewardAddress);
```
yrong, line 445: Would it be better to store the delivery status on chain? e.g:

```
mapping(uint64 nonce => DeliveryStatus) public nonces;
struct DeliveryStatus {
    uint64 blockNumber;
    bool result;
}
```

With the `blockNumber` it's easy for relayer to filter this event efficiently when submit the delivery proof.

vgeddes: That's a good point Ron. Though there are some simplifying assumptions which I think allow us to save 20000 gas by not storing that status on-chain.

The plan I had in mind was that the relayer who calls `v2_submit` is also responsible for sending delivery proof back to BH. Since this relayer witnessed the result for `v2_submit` it can generate a proof for `InboundMessageDispatched`, and there isn't a need for additional metadata on-chain.

For other cases when the delivery proof could not be sent back immediately, we could create a command in `snowbridge-relay` that queries for `InboundMessageDispatched(rewardAddress=X)` against some public ethereum node or event indexer, and sends the delivery proofs for the matching events. 

I will make a `rewardAddress` an indexed event parameter, which allows this query to be much faster to execute.

contracts/src/v2/Types.sol
```diff
@@ -0,0 +1,97 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2023 Snowfork <hello@snowfork.com>
+pragma solidity 0.8.25;
+
+import {OperatingMode} from "./../types/Common.sol";
+
+// Inbound message from a Polkadot parachain (via BridgeHub)
+struct InboundMessage {
```
yrong, line 8: would suggest to rename as `InboundMessageV2` or error when generate go bindings.

```
 > mage build
2024/10/28 15:53:26 INFO: Skipped ssz generated object: relays/beacon/state/beacon_encoding.go
2024/10/28 15:53:26 INFO: Skipped ssz generated object: relays/beacon/state/beacon_deneb_encoding.go
# github.com/snowfork/snowbridge/relayer/contracts
contracts/gateway.go:47:6: InboundMessage redeclared in this block
	contracts/gateway.go:40:6: other declaration of InboundMessage
Error: running "go build -o build/snowbridge-relay main.go" failed with exit code 1
```

yrong: https://github.com/Snowfork/snowbridge/pull/1321/commits/af374a65f57ea9ea7d4705f63b78a353fd6a350c FYI

contracts/src/interfaces/IGateway.sol
```diff
@@ -110,4 +171,56 @@ interface IGateway {
...
+    // Register Ethereum-native token on AHK, using `xcmFeeAHP` and `xcmFeeAHK`
+    // of `msg.value` to pay for execution on AHP and AHK respectively.
+    function v2_registerTokenOnKusama(
+        address token,
+        uint128 xcmFeeAHP,
+        uint128 xcmFeeAHK
+    ) external payable;
+
+    // Check if an inbound message was previously accepted and dispatched
+    function v2_isDispatched(uint64 nonce) external returns (bool);
```
yrong, line 225: Should add `view` keyword here.

https://github.com/Snowfork/snowbridge/pull/1321/commits/2e8d519e9c35b6f7cc41eb4a4d159af491a8c0e0

contracts/src/Gateway.sol
yrong: https://github.com/Snowfork/snowbridge/pull/1321/commits/7187bf3ef7eaffa793b269b88736ecae8b8ac7dc

contracts/src/interfaces/IGateway.sol
```diff
@@ -110,4 +171,56 @@ interface IGateway {
...
+        bytes calldata claimer
+    ) external payable;
+
+    // Register Ethereum-native token on AHP, using `xcmFeeAHP` of `msg.value`
+    // to pay for execution on AHP.
+    function v2_registerToken(address token, uint128 xcmFeeAHP) external payable;
+
+    // Register Ethereum-native token on AHK, using `xcmFeeAHP` and `xcmFeeAHK`
+    // of `msg.value` to pay for execution on AHP and AHK respectively.
+    function v2_registerTokenOnKusama(
```
alistair-singh, line 218: Just a thought on registration while thinking about our Security model. Kusama token registration can be built as a separate contract on top of the gateway.

The implementation of a contract called KusamaRegistration, that can build the registration Transact call onchain, and then submit it to the gateway. 

On BH we would decend origin to that Location so the origin would be `{ parents:2 interior: X2(Ethreum{chain_id:1}, KusamaRegistration)}`. 

This origin would be forwarded to AHK using AliasOrigin, and can be used secure the register extrinsic on the foreignAssets pallet on AHK. This is currently how we secure token registration now on AHP, using the gateway contracts origin (Snowbridge Sovereign). 

Upgrade would be just deploying a new version of the KusamaRegistration contract and then updating the origin which is allowed to call register on AHK. And it would be separate from the Gateway.

contracts/src/upgrades/rococo/RococoGatewayV2.sol
```diff
@@ -16,16 +16,7 @@ contract RococoGatewayV2 is Gateway {
         bytes32 bridgeHubAgentID,
```
claravanstaden, line 16: nit: File should probably be for Westend now that Rococo is deprecated.

contracts/src/v1/Calls.sol
```diff
@@ -34,40 +48,215 @@ library Assets {
...
+    function tokenAddressOf(bytes32 tokenID) external view returns (address) {
+        AssetsStorage.Layout storage $ = AssetsStorage.layout();
+        return $.tokenAddressOf[tokenID];
+    }
+
+    /*
+    * Internal functions
+    */
+
+    // Convert foreign currency to native currency (ROC/KSM/DOT -> ETH)
```
claravanstaden, line 188:
```suggestion
    // Convert foreign currency to native currency (WND/PAS/KSM/DOT -> ETH)
```

contracts/src/v2/Calls.sol
```diff
@@ -0,0 +1,222 @@
...
+        internal
+    {
+        // Build XCM for token registration on AHP and possibly AHK
+        // The XCM includes the necessary `PaysFee` instructions that:
+        // 1. Buy `xcmFeeAHP` worth of execution on AHP
+        // 2. Buy `xcmFeeAHK` worth of execution on AHK
+        bytes memory xcm;
+        if (xcmFeeAHK > 0) {
+            // Build XCM that is executed on AHP and forwards another message to AHK
+            xcm = bytes.concat(hex"deadbeef", abi.encodePacked(token), hex"deadbeef");
```
claravanstaden, line 134: `hex"deadbeef"` is this a placeholder for something?

contracts/src/v2/Types.sol
```diff
@@ -0,0 +1,97 @@
...
+
+library CommandKind {
+    uint8 constant Upgrade = 0;
+    uint8 constant SetOperatingMode = 1;
+    uint8 constant UnlockNativeToken = 2;
+    uint8 constant RegisterForeignToken = 3;
+    uint8 constant MintForeignToken = 4;
+    uint8 constant CreateAgent = 5;
+    uint8 constant CallContract = 6;
+}
```
claravanstaden, line 31: Is this only used for testing?

claravanstaden (commented):
Nice!

I was wondering if it would be feasible to add Transact support in this release, given that all the building blocks are there?

contracts/src/Gateway.sol
```diff
@@ -493,199 +391,192 @@ contract Gateway is IGateway, IInitializable, IUpgradable {
...
+    /// @dev Submit a message from Polkadot for verification and dispatch
+    /// @param message A message produced by the OutboundQueue pallet on BridgeHub
+    /// @param leafProof A message proof used to verify that the message is in the merkle tree committed by the OutboundQueue pallet
+    /// @param headerProof A proof that the commitment is included in parachain header that was finalized by BEEFY.
+    /// @param rewardAddress Account on BH to credit delivery rewards
+    function v2_submit(
+        InboundMessageV2 calldata message,
+        bytes32[] calldata leafProof,
+        Verification.Proof calldata headerProof,
+        bytes32 rewardAddress
```
claravanstaden, line 423: Won't we just just derive the account from the `origin` on BH?

contracts/src/v2/Calls.sol
claravanstaden: Will the contract build this XCM packet?

vgeddes: yeah, its gonna be baked into the contract.

contracts/src/v2/Types.sol
vgeddes: No, these contants are used in the dispatcher in `Gateway.v2_submit`

contracts/src/v2/Calls.sol
```diff
@@ -0,0 +1,222 @@
...
+    uint8 public constant MAX_ASSETS = 8;
+
+    // Send an XCM with arbitrary assets to Polkadot Asset Hub
+    //
+    // Params:
+    //   * `xcm` (bytes): SCALE-encoded XCM message
+    //   * `assets` (bytes[]): Array of asset specs.
+    //
+    // Supported asset specs:
+    // * ERC20: abi.encode(0, tokenAddress, value)
```
claravanstaden, line 63: What is the 0 for?

vgeddes: Its a tag used to descriminate between different assets kinds. 

Its very much like a Rust enum. Since Solidity does not have those, we have to emulate them.

contracts/src/Gateway.sol
```diff
@@ -170,316 +146,238 @@ contract Gateway is IGateway, IInitializable, IUpgradable {
...
+    // Execute code within an agent
+    function v1_handleAgentExecute(bytes calldata data) external onlySelf {
+        HandlersV1.agentExecute(AGENT_EXECUTOR, data);
+    }
 
-        ch.mode = params.mode;
-        emit ChannelUpdated(params.channelID);
+    /// @dev Create an agent for a consensus system on Polkadot
+    function v1_handleCreateAgent(bytes calldata data) external onlySelf {
+        HandlersV1.createAgent(data);
```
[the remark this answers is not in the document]
yrong: try / catch can only catch errors from external function calls.

contracts/src/Gateway.sol
yrong: It's different. The origin is derived from account on source chain, while the rewardAddress is account of the relayer for accepting reward.

contracts/src/v1/Handlers.sol
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    /// @dev Perform an upgrade of the gateway
+    function upgrade(bytes calldata data) external {
+        UpgradeParams memory params = abi.decode(data, (UpgradeParams));
+        Upgrade.upgrade(params.impl, params.implCodeHash, params.initParams);
+    }
+
+    // @dev Set the operating mode of the gateway
+    function setOperatingMode(bytes calldata data) external {
```
alistair-singh, line 78: Both v1 and v2 paths modify the same storage. So essentially v1 and v2 handlers are just two ways to set the bridge operating mode, and they are not ways to disable v1 and v2 separately.

contracts/src/storage/AssetsStorage.sol
```diff
@@ -18,6 +19,14 @@ library AssetsStorage {
...
         // Foreign token registry by token ID
         mapping(bytes32 foreignID => address) tokenAddressOf;
+        uint8 foreignTokenDecimals;
+        // The maximum fee that can be sent to a destination parachain to pay for execution (DOT).
+        // Has two functions:
+        // * Reduces the ability of users to perform arbitrage using a favourable exchange rate
+        // * Prevents users from mistakenly providing too much fees, which would drain AssetHub's
+        //   sovereign account here on Ethereum.
+        uint128 maxDestinationFee;
+        address weth;
```
alistair-singh, line 29: Move WETH to the first item. I will be adding this as part of the AutoWrapping/Unwrapping feature. So this change will help migration got smoothly.
```suggestion
        address weth;
        uint8 foreignTokenDecimals;
        // The maximum fee that can be sent to a destination parachain to pay for execution (DOT).
        // Has two functions:
        // * Reduces the ability of users to perform arbitrage using a favourable exchange rate
        // * Prevents users from mistakenly providing too much fees, which would drain AssetHub's
        //   sovereign account here on Ethereum.
        uint128 maxDestinationFee;
```

https://github.com/Snowfork/snowbridge/pull/1343/files#diff-3c18b55dea02438f70b3d8fe9ddfb126767f57a5eb4b6840d884c8883f73f521R21-R22

contracts/scripts/DeployLocal.sol
```diff
@@ -78,30 +75,34 @@ contract DeployLocal is Script {
...
-            assetHubAgentID: assetHubAgentID,
             assetHubCreateAssetFee: uint128(vm.envUint("CREATE_ASSET_FEE")),
             assetHubReserveTransferFee: uint128(vm.envUint("RESERVE_TRANSFER_FEE")),
             exchangeRate: ud60x18(vm.envUint("EXCHANGE_RATE")),
             multiplier: ud60x18(vm.envUint("FEE_MULTIPLIER")),
-            rescueOperator: address(0)
+            rescueOperator: address(0),
+            foreignTokenDecimals: foreignTokenDecimals,
+            maxDestinationFee: maxDestinationFee,
+            weth: weth
```
alistair-singh, line 92: Make weth first item.
```suggestion
            rescueOperator: address(0),
            weth: weth,
            foreignTokenDecimals: foreignTokenDecimals,
            maxDestinationFee: maxDestinationFee
```
Commit:
https://github.com/Snowfork/snowbridge/pull/1343/commits/e1a85782ec1e58d8fa1f8bbdfc6cd4c6b247d470

contracts/src/Initializer.sol
```diff
@@ -0,0 +1,128 @@
...
+        uint128 assetHubReserveTransferFee;
+        /// @dev extra fee to discourage spamming
+        uint256 registerTokenFee;
+        /// @dev Fee multiplier
+        UD60x18 multiplier;
+        /// @dev Optional rescueOperator
+        address rescueOperator;
+        uint8 foreignTokenDecimals;
+        uint128 maxDestinationFee;
+        address weth;
```
alistair-singh, line 59: Move weth to the first member.
```suggestion
        /// @dev The WETH address used for wrapping ETH
        address weth;
        uint8 foreignTokenDecimals;
        uint128 maxDestinationFee;
```
Commit:
https://github.com/Snowfork/snowbridge/pull/1343/commits/e1a85782ec1e58d8fa1f8bbdfc6cd4c6b247d470

contracts/src/Functions.sol
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    function withdrawWrappedEther(
+        address executor,
+        address agent,
+        address payable recipient,
+        uint128 amount
+    ) internal {
+        withdrawNativeToken(executor, agent, weth(), address(this), amount);
+        WETH9(payable(weth())).withdraw(amount);
```
yrong, line 108: This won't work as Gateway is not allowed to receive Ether.
https://github.com/Snowfork/snowbridge/blob/4e17456e3d774b3176966abdb80e1323d027d9d8/contracts/src/GatewayProxy.sol#L40-L44

A potential fix in https://github.com/Snowfork/snowbridge/pull/1321/commits/9081585234d8d8f9e6d0a6dede0a9918f2c0b030

contracts/src/v2/Calls.sol
```diff
@@ -0,0 +1,164 @@
...
+
+        CoreStorage.Layout storage $ = CoreStorage.layout();
+        $.outboundNonce = $.outboundNonce + 1;
+
+        Payload memory payload = Payload({
+            origin: origin,
+            assets: preparedAssets,
+            xcm: xcm,
+            claimer: claimer,
+            value: uint128(msg.value) - executionFee - relayerFee,
```
claravanstaden, line 119: What would this value be used for?

contracts/src/Functions.sol
vgeddes: Good catch! yeah plz merge in

contracts/src/Constants.sol
```diff
@@ -0,0 +1,21 @@
...
+
+    ParaID constant BRIDGE_HUB_PARA_ID = ParaID.wrap(1002);
```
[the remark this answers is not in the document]
vgeddes: yeah, indeed.

contracts/src/Gateway.sol
```diff
@@ -170,316 +146,238 @@ contract Gateway is IGateway, IInitializable, IUpgradable {
...
+        } else if (message.command == CommandV1.CreateAgent) {
+            try Gateway(this).v1_handleCreateAgent{gas: maxDispatchGas}(message.params) {}
             catch {
                 success = false;
             }
-        } else if (message.command == Command.CreateChannel) {
-            try Gateway(this).createChannel{gas: maxDispatchGas}(message.params) {}
-            catch {
+        } else if (message.command == CommandV1.CreateChannel) {
+            success = false;
```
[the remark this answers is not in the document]
vgeddes: yeah, channels no more.

contracts/src/Gateway.sol
```diff
@@ -170,316 +146,238 @@ contract Gateway is IGateway, IInitializable, IUpgradable {
...
+            amount
+        );
     }
 
-    /// @dev Update the configuration for a channel
-    function updateChannel(bytes calldata data) external onlySelf {
-        UpdateChannelParams memory params = abi.decode(data, (UpdateChannelParams));
+    // @dev Get token address by tokenID
+    function tokenAddressOf(bytes32 tokenID) external view returns (address) {
+        return CallsV1.tokenAddressOf(tokenID);
```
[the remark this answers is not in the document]
vgeddes: in certain cases v1 & v2 underlying implementation are the same, so we just use v1 code.

contracts/src/v2/Calls.sol
```diff
@@ -0,0 +1,164 @@
...
+        uint128 executionFee,
+        uint128 relayerFee
+    ) internal {
+        // Ensure outbound messaging is allowed
+        _ensureOutboundMessagingEnabled();
+
+        require(msg.value <= type(uint128).max, IGatewayV2.ExceededMaximumValue());
+        require(msg.value >= executionFee + relayerFee, IGatewayV2.InsufficientValue());
+        address assetHubAgent = Functions.ensureAgent(Constants.ASSET_HUB_AGENT_ID);
+        payable(assetHubAgent).safeNativeTransfer(msg.value);
```
alistair-singh, line 103: This one line does a lot, it essentially takes the fee, reward and any locked eth and passes it to the asset hub agent. Maybe we should collect fees and reward in a separate account like the gateway proxy contract address and just leave locked funds in the asset hub agent.

contracts/src/v1/Calls.sol
```diff
@@ -34,40 +48,203 @@ library Assets {
...
+
+        // Ensure the user has enough funds for this message to be accepted
+        if (msg.value < fee) {
+            revert FeePaymentToLow();
+        }
+
+        channel.outboundNonce = channel.outboundNonce + 1;
+
+        // Deposit total fee into agent's contract
+        payable(channel.agent).safeNativeTransfer(fee);
```
alistair-singh, line 228: Same here about the fees, we should collect the fees into the gateway contracts asset hub account.

contracts/src/v1/Calls.sol
```diff
@@ -34,40 +48,203 @@ library Assets {
...
+
+        // Reimburse excess fee payment
+        if (msg.value > fee) {
```
alistair-singh, line 231: Check dust before transfer.

contracts/src/Gateway.sol
```diff
@@ -493,199 +399,187 @@ contract Gateway is IGateway, IInitializable, IUpgradable {
...
     }
 
-    /// @dev Ensure that the specified parachain has a channel allocated
-    function _ensureChannel(ChannelID channelID) internal view returns (Channel storage ch) {
-        ch = CoreStorage.layout().channels[channelID];
-        // A channel always has an agent specified.
-        if (ch.agent == address(0)) {
-            revert ChannelDoesNotExist();
-        }
+    function v2_isDispatched(uint64 nonce) external view returns (bool) {
```
alistair-singh, line 510: This method is general enough that we do not have to prefix it with `v2_`

yrong:
@vgeddes I just merged several changes required for the smoke test https://github.com/Snowfork/snowbridge/pull/1300/commits/d77abeef882ff4217d6d52dd9790986a5ef5f2de please check.

contracts/src/Gateway.sol
```diff
@@ -170,316 +162,230 @@ contract Gateway is IGateway, IInitializable, IUpgradable {
...
 
         // Calculate a gas refund, capped to protect against huge spikes in `tx.gasprice`
         // that could drain funds unnecessarily. During these spikes, relayers should back off.
-        uint256 gasUsed = _transactionBaseGas() + (startGas - gasleft());
+        uint256 gasUsed = v1_transactionBaseGas() + (startGas - gasleft());
         uint256 refund = gasUsed * Math.min(tx.gasprice, message.maxFeePerGas);
 
         // Add the reward to the refund amount. If the sum is more than the funds available
         // in the channel agent, then reduce the total amount
         uint256 amount = Math.min(refund + message.reward, address(channel.agent).balance);
```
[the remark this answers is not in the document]
vgeddes: yeah, this makes sense, now that we are supporting native eth. Will fix 👍

contracts/src/v2/Calls.sol
vgeddes: Well for E->P flow in V2, fees and rewards are both ultimately redeemable as WETH (or native ETH) on AH. So they are actually locked funds.

PR #1301, opened 2024-10-08
Fix find checkpoint

yrong:


claravanstaden (approved):
Nice, thanks!

relayer/relays/beacon/header/header.go
```diff
@@ -555,7 +555,7 @@ func (h *Header) findLatestCheckPoint(slot uint64) (state.FinalizedHeader, error
 	endIndex := startIndex + 1
```
yrong, line 555: Btw: The CI tests failed for running too long,  what about `startIndex` is the last item of the ring buffer, will it cause the endless loop?

claravanstaden: You're right! Fix here :) https://github.com/Snowfork/snowbridge/pull/1304

yrong:
Closed in favor of https://github.com/Snowfork/snowbridge/pull/1304

PR #1302, opened 2024-10-08
Fix undefined event params

claravanstaden:
Westend Subscan sometimes returns events without the "params" json response. This PR skips those entries so that it does not break the transfer UI.


alistair-singh (approved):

web/packages/api/src/subscan.ts
```diff
@@ -152,6 +152,11 @@ export const fetchEvents = async <T>(
         }
 
         for (const { event_index, params } of paramsResponse.json.data) {
+            if (params === undefined) {
+                console.log("Event does not have any params", event_index)
```
alistair-singh, line 156:
```suggestion
                console.warn("Event does not have any params", event_index)
```

alistair-singh (approved):

PR #1304, opened 2024-10-08
Fix find checkpoint

claravanstaden:


relayer/relays/beacon/header/header.go
```diff
@@ -552,12 +552,21 @@ func (h *Header) findLatestCheckPoint(slot uint64) (state.FinalizedHeader, error
...
 	}
 	startIndex := uint64(lastIndex)
+
 	endIndex := startIndex + 1
+	iterations := uint64(0)
 
 	syncCommitteePeriod := h.protocol.Settings.SlotsInEpoch * h.protocol.Settings.EpochsPerSyncCommitteePeriod
 	totalStates := h.protocol.Settings.EpochsPerSyncCommitteePeriod * h.protocol.HeaderRedundancy // Total size of the circular buffer,
 	// https://github.com/paritytech/polkadot-sdk/blob/master/bridges/snowbridge/pallets/ethereum-client/src/lib.rs#L75
 	for index := startIndex; index != endIndex; index = (index - 1 + totalStates) % totalStates {
```
yrong, line 562: I'd suggest to simplify the loop pattern to end at the startIndex.

```
for {
    ....
    index = (index - 1 + totalStates) % totalStates
    if index == startIndex {
	    log.Debug("loop finished, max iterations reached")
	    break
    }
}
...
```

claravanstaden: That's a good point. 😅 Let me know what you think of https://github.com/Snowfork/snowbridge/pull/1304/commits/ba114cf1b9e28f0074f35d9be860eac8b6fe6769.

relayer/relays/beacon/header/header.go
```diff
@@ -552,12 +552,19 @@ func (h *Header) findLatestCheckPoint(slot uint64) (state.FinalizedHeader, error
...
 	totalStates := h.protocol.Settings.EpochsPerSyncCommitteePeriod * h.protocol.HeaderRedundancy // Total size of the circular buffer,
 	// https://github.com/paritytech/polkadot-sdk/blob/master/bridges/snowbridge/pallets/ethereum-client/src/lib.rs#L75
-	for index := startIndex; index != endIndex; index = (index - 1 + totalStates) % totalStates {
+	for {
+		index = (index - 1 + totalStates) % totalStates
+		if index == startIndex {
+			log.WithError(err).Debug("unable to find a relevant on-chain header, max iterations reached")
+			break
+		}
+
```
yrong, line 567: Better to move this to the end or we may miss the first item at startIndex.

claravanstaden: Good catch: https://github.com/Snowfork/snowbridge/pull/1304/commits/757581a6f48249b488aeed8cbb61cec9e8695f66 Added a test for the condition too.

yrong (approved):
+1

PR #1311, opened 2024-10-14
Audit-4: Remove rescue operator

yrong:
Fixes: SNO-1196


alistair-singh (commented):
Should we remove `"./storage/OperatorStorage.sol"` as well?

vgeddes (changes requested):
See https://linear.app/snowfork/issue/SNO-1196#comment-2b6ed3d6. Removing the rescue operator would defeat the purpose of this contract.

What we need to do rather is better document the purpose of this contract.

yrong:
Closed in favor of https://github.com/Snowfork/snowbridge/pull/1318

PR #1313, opened 2024-10-16
Relayers OFAC

claravanstaden:
Resolves: [SNO-1156](https://linear.app/snowfork/issue/SNO-1156)

TODO remaining for a follow-up PR: SNO-1214


relayer/chain/parachain/message.go
```diff
@@ -125,3 +126,114 @@ func removeLeadingZeroHashForSlice(s []string) []string {
...
+func decodeDestination(variant types.U8, destinationBytes []byte) (string, error) {
+	switch variant {
+	case 0:
+		// Account32
+		account32 := &types.H256{}
+		err := types.DecodeFromBytes(destinationBytes, account32)
+		if err != nil {
+			return "", fmt.Errorf("failed to decode destination: %v", err)
+		}
+		return account32.Hex(), nil
```
alistair-singh, line 219: So in the UI we do not post the hex version of the account. We post the SS58 version. Not sure what the correct thing to do there is?

relayer/relays/execution/main.go
```diff
@@ -412,6 +416,26 @@ func (r *Relay) doSubmit(ctx context.Context, ev *contracts.GatewayOutboundMessa
...
+	destination, err := parachain.GetDestination(ev.Payload)
+	if err != nil {
+		return err
+	}
+
+	log.WithFields(log.Fields{
+		"destination": destination,
+		"payload":     common.Bytes2Hex(ev.Payload),
+	}).Info("extracted destination from message")
+	banned, err := r.ofac.IsBanned(ev.Raw.Address.Hex(), destination)
```
alistair-singh, line 428: Is `ev.Raw.Address.Hex()` equivalent to `msg.sender` or `tx.origin`. Because I think we need the latter. The original signer is OFAC banned.

relayer/relays/parachain/scanner.go
```diff
@@ -186,6 +192,15 @@ func (s *Scanner) findTasksImpl(
 			if err != nil {
 				return nil, fmt.Errorf("decode message error: %w", err)
 			}
+			isBanned, err := s.IsBanned(m)
+			if err != nil {
+				//return nil, fmt.Errorf("banned check: %w", err)
+				log.WithError(err).Fatal("error checking banned address found")
```
alistair-singh, line 198: Return error after logging.

relayer/relays/parachain/scanner.go
```diff
@@ -464,3 +479,159 @@ func (s *Scanner) findLatestNonce(ctx context.Context) (uint64, error) {
...
 	return ethInboundNonce, err
 }
+
+func (s *Scanner) IsBanned(m OutboundQueueMessage) (bool, error) {
+	destination, err := GetDestination(m)
+	if err != nil {
+		return true, err
+	}
+
+	return s.ofac.IsBanned(destination, "")
```
alistair-singh, line 489: Maybe add a comment that we will in a follow up ban the source address.

yrong:
When OFAC banned detected, since it's ordered channel the message which can't skipped, I would assume the bridge is actually bricked?

So do we plan to add an governance call to reset the nonce/message which is banned?  Or we wait for V2 with non order messaging be ready first?

relayer/chain/parachain/message.go
claravanstaden: Just saw here Chainalysis does not support non-EVM chains. 😬  https://go.chainalysis.com/chainalysis-oracle-docs.html `Compatible networks`

relayer/relays/parachain/scanner.go
claravanstaden: >Does that require AliasOrigin with V2 first?

To extract it from the on-chain message, yes. In the interim we can fetch the event from Subscan and pull out the source from there.

relayer/relays/execution/main.go
claravanstaden: Addressed in https://github.com/Snowfork/snowbridge/pull/1313/commits/1e9558eccbc14a06e658fb86ac3c0733c6a79f61.

relayer/chain/parachain/message.go
claravanstaden: Addressed in https://github.com/Snowfork/snowbridge/pull/1313/commits/1e9558eccbc14a06e658fb86ac3c0733c6a79f61.

relayer/relays/parachain/scanner.go
claravanstaden: https://github.com/Snowfork/snowbridge/pull/1313/commits/9363111b50c2cf4e9d3e631eef70de6307d869eb

vgeddes (approved):

vgeddes:
> When OFAC banned detected, since it's ordered channel the message can't be skipped, I would assume the bridge is actually bricked?
> 
> So do we plan to add an governance call to reset the nonce/message which is banned? Or we wait for V2 with non-order messaging be ready first?

We will work around this with an off-chain process/policy until V2 ships.

PR #1315, opened 2024-10-16
Audit 10: Remove Duplicate Fee Function

alistair-singh:
`_sendTokenCosts` and `_sendForeignTokenCosts` do the exact same thing. We can just re-use `_sendTokenCosts`.

Resolves: SNO-1202

### DIFF
--- a/contracts/src/Assets.sol
+++ b/contracts/src/Assets.sol
@@ -194,36 +194,6 @@ library Assets {
         emit IGateway.TokenSent(token, sender, destinationChain, destinationAddress, amount);
     }
 
-    function _sendForeignTokenCosts(
-        ParaID destinationChain,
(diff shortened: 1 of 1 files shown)


claravanstaden:
@alistair-singh it looks like this a duplicate of issue 11. Just confirming, reusing the same method fixes "10. Querying costs of sending a native asset is not exposed" as well, correct?

claravanstaden (approved):

alistair-singh:
> @alistair-singh it looks like this a duplicate of issue 11. Just confirming, reusing the same method fixes "10. Querying costs of sending a native asset is not exposed" as well, correct?

Yes it does, because there is no point exposing the duplicate function. Rather remove the duplicate.

PR #1317, opened 2024-10-17
Audit Issue 9: Remove unused code

claravanstaden:
Resolves: [SNO-1201](https://linear.app/snowfork/issue/SNO-1201)

### DIFF
--- a/contracts/src/Gateway.sol
+++ b/contracts/src/Gateway.sol
@@ -97,24 +97,13 @@ contract Gateway is IGateway, IInitializable, IUpgradable {
     error ChannelAlreadyCreated();
     error ChannelDoesNotExist();
     error InvalidChannelUpdate();
-    error AgentExecutionFailed(bytes returndata);
     error InvalidAgentExecutionPayload();
     error InvalidConstructorParams();
-    error AlreadyInitialized();
     error TokenNotRegistered();
 
     // Message handlers can only be dispatched by the gateway itself
     modifier onlySelf() {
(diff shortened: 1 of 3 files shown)


contracts/src/Gateway.sol
```diff
@@ -650,7 +631,6 @@ contract Gateway is IGateway, IInitializable, IUpgradable {
         // Initialize agent for BridgeHub
         address bridgeHubAgent = address(new Agent(BRIDGE_HUB_AGENT_ID));
         core.agents[BRIDGE_HUB_AGENT_ID] = bridgeHubAgent;
-        core.agentAddresses[bridgeHubAgent] = BRIDGE_HUB_AGENT_ID;
```
claravanstaden, line 653: I think we should rather leave these in, in case we need this data later. I also understand we can't remove the [storage mapping](https://github.com/Snowfork/snowbridge/blob/main/contracts/src/storage/CoreStorage.sol#L16) without a migration, so it might not be worthwhile to remove this in the first place.

alistair-singh: Also dont we still need agents for Transact support?

claravanstaden: Probably yeah. I'll revert these.

claravanstaden: Reverted in https://github.com/Snowfork/snowbridge/pull/1317/commits/d15b6dfe69ad7ec4764e64946ddc5bb28442edd5.

alistair-singh (approved):

PR #1318, opened 2024-10-21
Audit-4: Comments on Shell

yrong:
Resolves: https://linear.app/snowfork/issue/SNO-1196

### DIFF
--- a/contracts/src/Shell.sol
+++ b/contracts/src/Shell.sol
@@ -8,6 +8,11 @@ import {IUpgradable} from "./interfaces/IUpgradable.sol";
 import {IShell} from "./interfaces/IShell.sol";
 import {ERC1967} from "./utils/ERC1967.sol";
 
+/**
+ * @title Shell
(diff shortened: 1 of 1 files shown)


contracts/src/Shell.sol
```diff
@@ -8,6 +8,11 @@ import {IUpgradable} from "./interfaces/IUpgradable.sol";
 import {IShell} from "./interfaces/IShell.sol";
 import {ERC1967} from "./utils/ERC1967.sol";
 
+/**
+ * @title Shell
+ * Only used for initial trusted bootstrapping, the operator is removed after that
+ * so in fact no longer has any use for production.
```
vgeddes, line 14:
```suggestion
 * Only used in the initial deployment of the GatewayProxy, which was deployed along
 * with this Shell as its logic contract. The Shell was then upgraded using a trusted
 * operator to the full Gateway contract. Currently this code is no longer in use but is
 * kept around for archival purposes.
```

PR #1319, opened 2024-10-22
Fix batch size unfit

yrong:
More context in https://snowfork.slack.com/archives/C07BG558XE1/p1729535072433709


alistair-singh (approved):
+1

PR #1320, opened 2024-10-22
Deposit WETH and Send

alistair-singh:


vgeddes:
Thanks Alistair. This prototype is a good option for V1 that doesn't require any complicated Gateway contract upgrades :+1:

For V2 though, probably makes sense to add a more specialized `Gateway.sendEther(xcm)` API that only sends ETH and an `xcm` to handle that Ether on AH. What do you think?

alistair-singh:
> Thanks Alistair. This prototype is a good option for V1 that doesn't require any complicated Gateway contract upgrades 👍
> 
> For V2 though, probably makes sense to add a more specialized `Gateway.sendEther(xcm)` API that only sends ETH and an `xcm` to handle that Ether on AH. What do you think?

Yeah I do not think its possible to auto unwrap WETH to ETH with V1 either. I think we have to specifically build it into V2 like you said.

contracts/src/utils/AutoWETH9.sol
```diff
@@ -0,0 +1,85 @@
...
+        IGateway(GATEWAY_PROXY).sendToken{value: fee}(
+            WETH_INSTANCE, destinationChain, destinationAddress, destinationFee, amount
+        );
+    }
+
+    // Unwrap WETH to ETH and transfer to destination
+    function transferFrom(address src, address payable dst, uint256 wad) public returns (bool) {
+        if (WETH9(WETH_INSTANCE).transferFrom(src, address(this), wad)) {
+            WETH9(WETH_INSTANCE).withdraw(wad);
+            dst.transfer(wad);
```
tutu1997, line 42: Would it be better to use

```
(bool sent, ) = dst.call{value: wad}("");
require(sent, "Failed to send funds");
```

yrong, line 42: Would it be better change to

```
(bool sent, ) = dst.call{value: wad}("");
require(sent, "Failed to send funds");
```

contracts/src/utils/AutoWETH9.sol
```diff
@@ -0,0 +1,85 @@
...
+        emit Wrapped(msg.sender, amount);
+
+        uint256 fee = msg.value - amount;
+        IGateway(GATEWAY_PROXY).sendToken{value: fee}(
+            WETH_INSTANCE, destinationChain, destinationAddress, destinationFee, amount
+        );
+    }
+
+    // Unwrap WETH to ETH and transfer to destination
+    function transferFrom(address src, address payable dst, uint256 wad) public returns (bool) {
```
yrong, line 39: Is this function internal or required to be public?  would also suggest to remove the `src` as input and replace it with `msg.sender`.

vgeddes (commented):
I think the issue with introducing a wrapper token is that it fragments ETH liquidity within Polkadot, as AutoWETH and WETH have different XCM locations. It also adds confusion for developers and users about which one to use.

Auto-wrapping for Ethereum->Polkadot: We can easily implement without breaking V1 API. See the example interface further below.

Auto-unwrapping for Polkadot->Ethereum: Hard to implement this without breaking V1 compat or having to make changes on AH and BH and wait for releases.

```solidity
    /// @dev Quote a fee in Ether for sending Ether
    /// 1. Delivery costs to BridgeHub
    /// 2. XCM execution costs on destinationChain
    function quoteSendEtherFee(ParaID destinationChain, uint128 destinationFee) external view returns (uint256);

    /// @dev Send Ether to parachain `destinationChain` and deposit into account `destinationAddress`
    ///
    /// Params:
    /// * `amount`: The amount of ether to send
    /// 
    /// The Ether passed in the transaction as `msg.value` must cover both `amount` as well the bridging
    /// fee calculated by `quoteSendEtherFee`.
    function sendEther(
        ParaID destinationChain,
        MultiAddress calldata destinationAddress,
        uint128 destinationFee,
        uint128 amount
    ) external payable;
```

We should definitely support full auto-(un)wrap for v2 though.

alistair-singh:
Closing in favor of breaking backward compatibility and auto wrapping/unwrapping WETH in v1 by default.

PR #1323, opened 2024-10-25
Fixes building/testing on Nix

Lederstrumpf:
some missing components required for running on nix environments


vgeddes (approved):
Looks good!

PR #1331, opened 2024-11-11
Remove rococo config deprecated

yrong:


web/packages/api/src/environment.ts
```diff
@@ -698,7 +554,7 @@ export const SNOWBRIDGE_ENV: { [id: string]: SnowbridgeEnvironment } = {
             BEACON_HTTP_API: "https://lodestar-sepolia.chainsafe.io",
             ETHEREUM_API: (key) => `https://eth-sepolia.g.alchemy.com/v2/${key}`,
             RELAY_CHAIN_URL: "https://westend-rpc.polkadot.io",
-            ASSET_HUB_URL: "wss://westend-asset-hub-rpc.polkadot.io",
+            ASSET_HUB_URL: "https://westend-asset-hub-rpc.polkadot.io",
```
alistair-singh, line 557: This will break transfers, @polkadot/api has this weird issue where by it doesn't subscribe for finalization on transactions when using https.  Using https means the code hangs forever waiting for an event that never comes. 

If you want to use https override it when creating the context, see below for example.
https://github.com/Snowfork/snowbridge-app/blob/82a99e0aebe4146dd8b139780fbc788ab5405def/lib/snowbridge.ts#L222

alistair-singh (changes requested):
+1 for this clean up!

web/packages/api/src/environment.ts
yrong: Revert in https://github.com/Snowfork/snowbridge/pull/1331/commits/df25226590cd2336cdbdac0c95f955b646dcd08c

I change to https because met an issue today the monitor process fail to connect to AH through the WS endpoint, and it works with the https endpoint.

 Anyway, we can still make WS default.  It's also possible to overide the value with an env variable.
https://github.com/Snowfork/snowbridge/blob/b4f04733bf269e2d2dc906616c1d31814fb00816/web/packages/operations/src/monitor.ts#L28

alistair-singh (approved):

PR #1333, opened 2024-11-19
Fix relayer build GA task

claravanstaden:
There is a known issue when building abigen tools with Go 1.23.x. Upgrading to abigen 1.14.8 with includes a fix: https://gitlab.alpinelinux.org/alpine/aports/-/issues/16401

https://github.com/ethereum/go-ethereum/issues/30100


yrong (approved):
Cool!

PR #1334, opened 2024-11-19
Auto label PRs and Relayer component notes only

claravanstaden:
Adds a labeller action to label repo changes based on the changed component.

Generate relayer release changelog notes only based on relayer components.


yrong (approved):
Cool!

PR #1336, opened 2024-11-20
Improve alarms

yrong:
- Move alarm configuration out of api package
- Make alarm evaluation variables configurable
- Monitor with `SCAN_INTERVAL` more sensitive

Resolves: https://linear.app/snowfork/issue/SNO-1233

### DIFF
--- a/web/packages/operations/.env.example
+++ b/web/packages/operations/.env.example
@@ -1,5 +1,5 @@
 # Network config
-NODE_ENV=rococo_sepolia
+NODE_ENV=polkadot_mainnet
 
 # AWS Config
 AWS_ACCESS_KEY_ID=
@@ -15,8 +15,8 @@ REACT_APP_INFURA_KEY=
 # GRAPHQL config
 GRAPHQL_API_URL=https://data.snowbridge.network/graphql
 
-# Scan interval(in minutes)
-SCAN_INTERVAL=60 
+# Scan interval(30 minutes by default)
+SCAN_INTERVAL=30
 
 # Keys
 ETHEREUM_KEY=
@@ -28,3 +28,15 @@ CRON_EXPRESSION=0 0 * * *
 # Dashboard Url
 LATENCY_DASHBOARD_URL=
 BALANCE_DASHBOARD_URL=
+
+# Latency Threshold
+BlockLatencyToEthereum=
+BlockLatencyToPolkadot=
+
+# Alarm Evaluation Configs
+ToEthereumStaleEvaluationPeriods=
+ToEthereumStalePeriod=
+ToEthereumStaleDatapointsToAlarm=
+ToPolkadotStaleEvaluationPeriods=
+ToPolkadotStalePeriod=
+ToPolkadotStaleDatapointsToAlarm=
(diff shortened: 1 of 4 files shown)


yrong:
Reinitialize alarms with threshold more sensitive, monitor process restarted and we can wait to see in CloudWatch if make sense.

web/packages/operations/.env.example
```diff
@@ -15,8 +15,8 @@ REACT_APP_INFURA_KEY=
 # GRAPHQL config
 GRAPHQL_API_URL=https://data.snowbridge.network/graphql
 
-# Scan interval(in minutes)
-SCAN_INTERVAL=60 
+# Scan interval(30 minutes by default)
+SCAN_INTERVAL=30
```
alistair-singh, line 19: Do you think it would be valuable to do 15 minutes?

web/packages/operations/src/alarm.ts
```diff
@@ -117,21 +165,17 @@ export const sendMetrics = async (metrics: status.AllMetrics) => {
...
+                channel.toEthereum.outbound > channel.toEthereum.inbound &&
+                    channel.toEthereum.inbound == channel.toEthereum.previousInbound
             ),
         })
         metricData.push({
             MetricName: AlarmReason.ToEthereumChannelAttacked.toString(),
-            Value: Number(
-                channel.toEthereum.outbound < channel.toEthereum.inbound     
-            ),
+            Value: Number(channel.toEthereum.outbound < channel.toEthereum.inbound),
```
alistair-singh, line 174: Instead of logging booleans, maybe we should log the difference. This would then be a metric of undelivered messages.

web/packages/operations/src/alarm.ts
```diff
@@ -35,11 +89,8 @@ export const sendMetrics = async (metrics: status.AllMetrics) => {
     metricData.push({
         MetricName: AlarmReason.BeefyStale.toString(),
         Value: Number(
-            metrics.bridgeStatus.toEthereum.blockLatency >
-            (process.env["CheckIntervalToEthereum"]
-            ? parseInt(process.env["CheckIntervalToEthereum"])
-            : status.BlockLatencyThreshold.ToEthereum) &&
-                metrics.bridgeStatus.toEthereum.latestPolkadotBlockOnEthereum <=
+            metrics.bridgeStatus.toEthereum.blockLatency > BlockLatencyThreshold.ToEthereum &&
```
alistair-singh, line 92: Maybe should also just log the blockLatency as a plain number instead of a boolean. Easier threshold management.

alistair-singh (approved):
+1, approving. Although I think we should move away from using boolean style metrics and we should just log numbers where possible. Maybe we can address in a separate PR.

web/packages/operations/.env.example
yrong: I'd prefer to start from 30 minutes first.

web/packages/operations/src/alarm.ts
yrong: https://github.com/Snowfork/snowbridge/pull/1336/commits/785b59ba70fde21abf0301b937dacc61e4a02e1a

web/packages/operations/src/alarm.ts
yrong: https://github.com/Snowfork/snowbridge/pull/1336/commits/785b59ba70fde21abf0301b937dacc61e4a02e1a metrics for the difference.

PR #1347, opened 2024-12-06
Check channel agent Id

alistair-singh:


yrong (approved):
+1

contracts/src/Assets.sol
```diff
@@ -280,7 +280,11 @@ library Assets {
     }
 
     // @dev Mint foreign token from Polkadot
-    function mintForeignToken(bytes32 foreignTokenID, address recipient, uint256 amount) external {
+    function mintForeignToken(address agent, bytes32 foreignTokenID, address recipient, uint256 amount) external {
```
vgeddes, line 283: As discussed before, I prefer if we check the channel id instead of the agent id. Ultimately what we are doing is adding a restriction that the mint message can only be passed via the Channel owned by AssetHub.

Example:
```solidity
function mintForeignToken(ChannelID channelID, ...) {
   if (channelID == $.assetHubParaID.into()) { ... }
}
```

alistair-singh: Addressed in https://github.com/Snowfork/snowbridge/pull/1347/commits/bbf6532502054242e168d7433d4e68441a4a31b7

vgeddes (approved):

PR #1350, opened 2024-12-08
Ringfence PNA by agent id

alistair-singh:
Tokens are now owned by Agents instead of the Gateway. Tokens are now completely ringfenced by Agents.


alistair-singh:
Close in favor of https://github.com/Snowfork/snowbridge/pull/1347

PR #1354, opened 2024-12-12
Support Native ETH in v1

alistair-singh:
Changes
1. Use the `0x0000000000000000000000000000000000000000` token address as Native ETH.
2. When sendToken is called with the `0x00..` an amount of `msg.value` is locked in asset hub agent.
3. To avoid clashing with fees/rewards and locked funds, fees/rewards are collected and dispensed from the Gateway Proxy account. 
4. The existing funds in the agents will be migrated on upgrade. 
5. Scripts which mentions FundAgent are now called FundGateway

Related polkadot-sdk: https://github.com/paritytech/polkadot-sdk/pull/6855

TODO

- [x] Fix tests
- [x] Add tests for the Eth asserting msg.value.
- [x] Migrate agent funds to gateway proxy on upgrade


contracts/src/Gateway.sol
```diff
@@ -530,18 +531,17 @@ contract Gateway is IGateway, IInitializable, IUpgradable {
...
-        // Deposit total fee into agent's contract
-        payable(channel.agent).safeNativeTransfer(fee);
-
+        // The fee is already collected into the gateway contract
         // Reimburse excess fee payment
-        if (msg.value > fee) {
-            payable(msg.sender).safeNativeTransfer(msg.value - fee);
+        if (msg.value > totalEther) {
+            payable(msg.sender).safeNativeTransfer(msg.value - totalEther);
         }
```
yrong, line 545: Is this refund process still required since Gateway now is the vault of Ether? I'd suggest to remove it. 

Or for security do we need to add some reentrantlock on outbound calls?

contracts/src/Types.sol
```diff
@@ -107,6 +107,7 @@ struct Ticket {
     ParaID dest;
     Costs costs;
     bytes payload;
+    uint128 etherAmount;
```
vgeddes, line 110: I'd just call this `value` to match V2 nomenclature.

```suggestion
    uint128 value;
```

contracts/src/GatewayProxy.sol
```diff
@@ -37,9 +37,6 @@ contract GatewayProxy is IInitializable {
         }
     }
 
-    // Prevent users from unwittingly sending ether to the gateway, as these funds
-    // would otherwise be lost forever.
-    receive() external payable {
-        revert NativeCurrencyNotAccepted();
-    }
```
vgeddes, line 44: We can't revert this. `GatewayProxy` is immutable forever. Instead you'll need to find a workaround, such as creating an explicit function such as `depositEther` to receive ether.

contracts/src/Types.sol
alistair-singh: addressed in 0e40439370417481613eeccb5e70386649a10c73

contracts/src/GatewayProxy.sol
alistair-singh: I am not sure I understand this? Surely this is exactly what we want, to use the immutable GatewayProxy address as the storage for the Ether. Should it not be the Gateway implementation that we want to reject `receive` on?

alistair-singh: Ah I get it, it is already created and cannot be changed in mainnet. Sorry ignore above comment.

alistair-singh:
@vgeddes we have a Command called `TransferNativeFromAgent` which is a management-style command, issuable from bridgehub only, that can empty an agent account of ETH. We would need to add a similar command `TransferNativeFromGateway` to do the same thing for the Gateway. We should also maybe disable `TransferNativeFromAgent` so that there is no way to remove ETH from an agent other than an Unlock command.

contracts/src/GatewayProxy.sol
alistair-singh: addressed in [a9768e1](https://github.com/Snowfork/snowbridge/pull/1354/commits/a9768e1562ebaa3eec223e878c7c869a4bf0bd91)

contracts/src/Gateway.sol
yrong: @alistair-singh I added a PR https://github.com/Snowfork/snowbridge/pull/1356 with tests to demonstrate the reentrancy behavior through the refund process.

It will cost a bit more gas and the test shows that even without the guard the bridge can't be exploited. So I'm still not sure if it's necessary.

Please let me know what you think.

alistair-singh: Nice catch! How is this different from the `nonentrancy` keyword in solidity 0.28 that we use in Contracts v2?

yrong: Yeah, the [Transient storage](https://soliditylang.org/blog/2024/01/26/transient-storage/) version is better.

alistair-singh: I address this in a different PR here:
https://github.com/Snowfork/snowbridge/pull/1358

I wonder if we should use OZ or stay with our version which has way less boilerplate but we can discuss on that PR.

yrong (approved):
+1

contracts/src/AgentExecutor.sol
```diff
@@ -16,11 +17,15 @@ contract AgentExecutor {
 
     /// @dev Transfer ether to `recipient`. Unlike `_transferToken` This logic is not nested within `execute`,
     /// as the gateway needs to control an agent's ether balance directly.
-    ///
     function transferNative(address payable recipient, uint256 amount) external {
         recipient.safeNativeTransfer(amount);
     }
 
+    /// @dev Transfer ether to Gateway. Used once off for migration purposes. Can be removed after version 1.
+    function transferNativeToGateway(address payable gateway, uint256 amount) external {
```
vgeddes, line 25: In this context, `msg.sender` will always be the gateway address, as verified by `Agent.invoke`. So having a `gateway` parameter is redundant.

Also, prefer this naming in this context:
```suggestion
    function transferEtherToGateway(address payable gateway, uint256 amount) external {
```

contracts/src/Assets.sol
```diff
@@ -60,7 +62,7 @@ library Assets {
     ) external view returns (Costs memory costs) {
         AssetsStorage.Layout storage $ = AssetsStorage.layout();
         TokenInfo storage info = $.tokenRegistry[token];
-        if (!info.isRegistered) {
+        if (!info.isRegistered && token != address(0)) {
```
vgeddes, line 65: Why not just register ether as a token, so we can avoid the complexity of this check?

contracts/src/Assets.sol
```diff
@@ -144,7 +148,18 @@ library Assets {
         AssetsStorage.Layout storage $ = AssetsStorage.layout();
 
         // Lock the funds into AssetHub's agent contract
-        _transferToAgent($.assetHubAgent, token, sender, amount);
+        if (token != address(0)) {
```
vgeddes, line 151: I think the way you've introduced native ether support adds complexity by adding a lot of decision branches at the lowest level.

We already have branches at a higher level to choose between foreign, native tokens. We can add a branch for eth support there.

Concretely, I would suggest adding a new `_sendEther` function instead of adding branches in `_sendNativeToken`.

contracts/src/AgentExecutor.sol
alistair-singh: Addressed in aca38466c8d25da51ecf7629014f20b9f47d8871

contracts/src/Assets.sol
alistair-singh: Addressed in a3e5aa4bc4dcf33624c5b6f90e32a9be5279cc8f

contracts/src/Assets.sol
alistair-singh: Addressed in [d8a9a58](https://github.com/Snowfork/snowbridge/pull/1354/commits/d8a9a58d6572f619ef00c8a57f44c2022ffa9b56)

contracts/src/Assets.sol
```diff
@@ -145,19 +149,77 @@ library Assets {
...
-            _transferToAgent($.assetHubAgent, token, sender, amount);
-            ticket.value = 0;
+        // Lock Native Ether into AssetHub's agent contract
+        if (msg.value < amount) {
+            revert InvalidAmount();
+        }
+        payable($.assetHubAgent).safeNativeTransfer(amount);
+        ticket.value = amount;
+
+        ticket.dest = $.assetHubParaID;
```
vgeddes, line 159: It seems that both `_sendEther` and `_sendNativeToken` share this ticket/payload creation code. Can we not factor that out?

alistair-singh: addressed in 0921ff9777a86e0be660bd46d36b6cef965dc171

web/packages/api/src/environment.ts
```diff
@@ -65,7 +65,7 @@ export const SNOWBRIDGE_ENV: { [id: string]: SnowbridgeEnvironment } = {
                 erc20tokensReceivable: [
                     {
                         id: "WETH",
-                        address: "0x87d1f7fdfEe7f651FaBc8bFCB6E086C278b77A7d",
+                        address: "0x774667629726ec1FaBEbCEc0D9139bD1C8f72a23",
```
claravanstaden, line 68: Do you know why the WETH address changed?

claravanstaden (approved):
Awesome feature! 🥳

contracts/src/upgrades/Gateway202410.sol
```diff
@@ -25,10 +25,27 @@ contract Gateway202410 is Gateway {
...
     // Override parent initializer to prevent re-initialization of storage.
-    function initialize(bytes memory) external view override {
+    function initialize(bytes memory) external override {
         // Ensure that arbitrary users cannot initialize storage in this logic contract.
         if (ERC1967.load() == address(0)) {
             revert Unauthorized();
         }
+
+        // We expect version 0, deploying version 1.
+        CoreStorage.Layout storage $ = CoreStorage.layout();
```
vgeddes, line 35: What's this for?

For the next upgrade after this, we'll deploy a new contract like `Gateway202506.sol` with different initializer code.

contracts/src/Assets.sol
```diff
@@ -211,6 +275,7 @@ library Assets {
 
         ticket.dest = $.assetHubParaID;
         ticket.costs = _sendTokenCosts(destinationChain, destinationChainFee, maxDestinationChainFee);
+        ticket.value = 0;
```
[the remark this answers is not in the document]
alistair-singh: The Assets module needs to transfer native Ether as an asset to the agent. The _submitOutbound method needs to transfer native Ether as the fee. The _submitOutbound method does not know what was previously transferred by the Assets module to the agent, so this value in the ticket represents this so it can be used for fee accounting later down in the message submission process. All outbound messages share the same ticket struct so its best to initialize them explicitly to 0.

contracts/src/Assets.sol
```diff
@@ -194,6 +197,67 @@ library Assets {
...
+        address sender,
+        ParaID destinationChain,
+        MultiAddress calldata destinationAddress,
+        uint128 destinationChainFee,
+        uint128 maxDestinationChainFee,
+        uint128 amount
+    ) internal returns (Ticket memory ticket) {
+        AssetsStorage.Layout storage $ = AssetsStorage.layout();
+
+        // Lock Native Ether into AssetHub's agent contract
```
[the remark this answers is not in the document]
alistair-singh: addressed in ddad89faf0db1a75a54aa4246e4e577918d74ddb

contracts/src/Gateway.sol
```diff
@@ -215,10 +215,8 @@ contract Gateway is IGateway, IInitializable, IUpgradable {
                 success = false;
             }
         } else if (message.command == Command.TransferNativeFromAgent) {
-            try Gateway(this).transferNativeFromAgent{gas: maxDispatchGas}(message.params) {}
-            catch {
-                success = false;
-            }
+            // Disable this bridge command because now native Ether can be locked in an agent.
+            success = true;
```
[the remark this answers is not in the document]
alistair-singh: Rather be explicit as documentation in code that we now completely ignore this message.

contracts/src/upgrades/Gateway202410.sol
alistair-singh: It stops us from deploying this instance of Gateway202410 twice and hence running the same migration twice. I know this is controlled by governance, but an explicit way to do this in code is better.

vgeddes: Sure, we can consider adding a version check, but I don't think this PR is the right place for it.

I will remove the version check for the time being.

contracts/src/Assets.sol
```diff
@@ -241,21 +229,66 @@ library Assets {
...
+                    $.assetHubReserveTransferFee,
+                    destinationChainFee,
+                    amount
+                );
+            } else {
+                revert Unsupported();
+            }
+        }
+
+        emit IGateway.TokenSent(token, sender, destinationChain, destinationAddress, amount);
```
alistair-singh, line 291: This block can be replaced by calling `_emitSendNativeTokenTicket`.

contracts/src/Assets.sol
alistair-singh: It looks like you reverted the `_sendEther` and `_sendNativeToken` split and merged it again. Just out of interest why?

vgeddes: changed my mind, your original way was cleaner :)

web/packages/api/src/environment.ts
vgeddes: this seems to have been reverted.

vgeddes:
@alistair-singh ready for another round of review - I improved the validation of the passed ether in `Gateway.submitOutbound`

PR #1357, opened 2024-12-24
Fix send to penpal e2e test for XCM v5

alistair-singh:
If running the e2e tests agains Parity master branch than the tests fail because Penpal has been updated to xcm v5 and registering of assets now require sudo privileges or xcm origins.

We should only merge when we are ready to move off release `polkadot-stable2409`.


claravanstaden (approved):

alistair-singh:
This has been patches as part of 2412.

PR #1359, opened 2024-12-30
Ignore AgentTransferFromNative bridge command

alistair-singh:
Because we now lock Ether in the agent, we should never allow transferring native Ether unless it is part of an unlock bridge command.

### DIFF
--- a/contracts/src/Gateway.sol
+++ b/contracts/src/Gateway.sol
@@ -38,7 +38,6 @@ import {
     CreateChannelParams,
     UpdateChannelParams,
     SetOperatingModeParams,
-    TransferNativeFromAgentParams,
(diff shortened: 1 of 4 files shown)


yrong:
It may not be related to this PR but reminds me in V2, for transact, we allow the creation of a user/custom agent and dispatch transact with some Ether attached from that agent.

https://github.com/yrong/polkadot-sdk/blob/a3c5d872019313fdee8f188ff5fba854b4c20e8a/cumulus/parachains/integration-tests/emulated/tests/bridges/bridge-hub-westend/src/tests/snowbridge_v2_outbound.rs#L284

So we may still need to allow the user to claim Ether from the agent.

yrong (approved):
+1

PR #1361, opened 2024-12-31
Add register ether command to control tool

alistair-singh:
Resolves: SNO-1222, SNO-1251

### DIFF
--- a/control/preimage/src/commands.rs
+++ b/control/preimage/src/commands.rs
@@ -337,6 +337,7 @@ pub fn force_set_metadata(params: &UpdateAssetArgs) -> AssetHubRuntimeCall {
 }
 
 pub fn register_ether(params: &RegisterEtherArgs) -> (AssetHubRuntimeCall, AssetHubRuntimeCall) {
+    use subxt::utils::AccountId32;
     let chain_id = crate::bridge_hub_runtime::CHAIN_ID;
     #[cfg(feature = "paseo")]
     use asset_hub_paseo_types::*;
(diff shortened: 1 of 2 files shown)


control/preimage/src/commands.rs
```diff
@@ -31,9 +31,18 @@ use crate::bridge_hub_runtime::RuntimeCall as BridgeHubRuntimeCall;
...
+        junction::NetworkId,
+        junctions::Junctions::{X1, X2},
+        location::Location,
     };
+    pub fn get_ether_id(chain_id: u64) -> Location {
+        return Location {
+            parents: 2,
+            interior: X1([GlobalConsensus(NetworkId::Ethereum { chain_id })]),
+        };
+    }
```
yrong, line 45: Please check https://github.com/paritytech/polkadot-sdk/pull/6855#discussion_r1900322845

alistair-singh: This was merged by parity before discussing with Vincent. But I think we would like it to be `{ parents: 2, interior: X1([GlobalConsensus(Ethereum { chain_id: 1 })]) }` because in XCM this is the convention. I.e. The relaychain token and the destination of the relaychain share the same location. So in our case the bridge location and the bridge native token also shares the same location.

yrong (approved):
+1

control/preimage/src/main.rs
```diff
@@ -210,6 +220,22 @@ pub struct TreasuryProposal2024Args {
...
+#[derive(Debug, Args)]
+pub struct RegisterEtherArgs {
+    /// The minimum balance of the Ether asset that users are allowed to hold
+    #[arg(long, value_name = "WEI", default_value_t = 1u128)]
+    ether_min_balance: u128,
+    /// The Ether asset display name
+    #[arg(long, value_name = "ASSET_DISPLAY_NAME", default_value_t = String::from("Ether"))]
+    ether_name: String,
+    /// The Ether asset symbol
+    #[arg(long, value_name = "ASSET_SYMBOL", default_value_t = String::from("ETH.e"))]
```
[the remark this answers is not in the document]
alistair-singh: Resolved in https://github.com/Snowfork/snowbridge/pull/1361/commits/9132e649758d4faefe376b4f862490fcfd1703f3

vgeddes (approved):

PR #1362, opened 2025-01-10
History data from subsquid indexer

yrong:
Requires: https://github.com/Snowfork/snowbridge-subsquid/pull/2 with PNA&Multiple-hops transfers.

In this PR also update the js libraries to be consistent with the UI.


vgeddes (approved):
This is off-chain so I'm happy to have this merged in without an in-depth review.

PR #1363, opened 2025-01-16
Adds Aave token metadata to mainnet environment

claravanstaden:


vgeddes (approved):
Looks good.

Though we need to move away from a having the UI depend on a hardcoded list of tokens. Otherwise we will repeat this work every few weeks.

Surely we can update the UI to query the the token metadata on demand for unknown tokens?

@yrong do think we populate the indexer database with token metadata 

Either way @claravanstaden can add you a ticket to track this work?

alistair-singh (approved):

yrong:
> do think we can populate the indexer database with token metadata?

Yeah, we can index the tokens registered through our bridge.

claravanstaden:
@vgeddes I added [SNO-1270](https://linear.app/snowfork/issue/SNO-1270).

PR #1367, opened 2025-01-22
Api for moonbeam integration

yrong:
Resolves: https://linear.app/snowfork/issue/SNO-1265
Just a thin wrapper on top of Moonbeam's [xcm precompile](https://github.com/Moonsong-Labs/moonkit/blob/93ebaff0f5d69b507b8a44aced932d790a677723/precompiles/pallet-xcm/XcmInterface.sol#L142-L143)


alistair-singh (commented):
I am happy with these changes. Great work. However maybe we can support this transfer with a simple Ethers,js abi interface instead of a separate package.

https://docs.ethers.org/v6/getting-started/#starting-contracts
```
abi = [
  "function decimals() view returns (string)",
  "function symbol() view returns (string)",
  "function balanceOf(address addr) view returns (uint)"
]

// Create a contract
contract = new Contract("dai.tokens.ethers.eth", abi, provider)
```

This means that moonbeam can just be a module in the existing API package.

yrong:
> Support this transfer with a simple Ethers.js abi interface

Does that imply you prefer to add another high-level precompile on top of the [original one](https://github.com/Moonsong-Labs/moonkit/blob/93ebaff0f5d69b507b8a44aced932d790a677723/precompiles/pallet-xcm/XcmInterface.sol#L142-L143) into Moombeam's code base?

alistair-singh:
> > Support this transfer with a simple Ethers.js abi interface
> 
> Does that imply you prefer to add another high-level precompile on top of the [original one](https://github.com/Moonsong-Labs/moonkit/blob/93ebaff0f5d69b507b8a44aced932d790a677723/precompiles/pallet-xcm/XcmInterface.sol#L142-L143) into Moombeam's code base?

Can we simply use untyped bindings like so?

```ts
let abi = [
  "transferAssetsUsingTypeAndThenAddress((uint8,bytes[]),(address,uint256)[],uint8,uint8,uint8,bytes)",
]
const xcmInterfacePrecompile = "0x000000000000000000000000000000000000081A"
let contract = new Contract(xcmInterfacePrecompile, abi, provider)

const tx = await contract[
        "transferAssetsUsingTypeAndThenAddress((uint8,bytes[]),(address,uint256)[],uint8,uint8,uint8,bytes)"
    ](dest, assets, assetTransferType, remoteFeesIdIndex, feesTransferType, customXcmOnDest)

```

yrong:
> Can we simply use untyped bindings like so?

I tried, and it seems won't work.

```
Error: TypeError: xcmInterface.transferAssetsUsingTypeAndThenAddress((uint8,bytes[]),(address,uint256)[],uint8,uint8,uint8,bytes) is not a function
    at executeTransfer (/home/ubuntu/projects/snowbridge/web/packages/integrations/moonbeam/src/index.ts:223:6)
    at processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async run (/home/ubuntu/projects/snowbridge/web/packages/integrations/moonbeam/src/testTransfer.ts:31:5)
 ELIFECYCLE  Command failed with exit code 1.
```

Btw: The API binding is Autogenerated with [typechain](https://github.com/Snowfork/snowbridge/blob/3ac2cd9004f5b8bd74cc49b94901b8c27f66ffb0/web/packages/integrations/moonbeam/package.json#L17).

alistair-singh:
Closing in favor of https://github.com/Snowfork/snowbridge/pull/1398

PR #1370, opened 2025-01-26
Query the status of a single transfer & add comments for the indexer api

yrong:
Resolves: https://linear.app/snowfork/issue/SNO-1289


alistair-singh (commented):
Looks good. But we are missing one method and example, how to query the status of a single bridge transfer via message id.

yrong:
> we are missing one method and example, how to query the status of a single bridge transfer via message id.

Addressed in https://github.com/Snowfork/snowbridge/pull/1370/commits/d341b69b2dec0d6428b75c4d461629090fb6d533

alistair-singh (approved):

PR #1321, opened 2024-10-22
Parachain Relayer V2

yrong:
Resolves: https://linear.app/snowfork/issue/SNO-1230
Requires: https://github.com/paritytech/polkadot-sdk/pull/6706

PR #1371, opened 2025-02-02
V2

vgeddes:
Implementation of the Snowbridge V2 [specification](https://github.com/paritytech/polkadot-sdk/blob/master/bridges/snowbridge/docs/v2.md).

Companion PR for Polkadot changes: https://github.com/paritytech/polkadot-sdk/pull/7402

# Overview

Compatibility with the V1 protocol has been retained, and the two protocols co-exist in the same `Gateway` contract. However the V1 implementation has been heavily refactored.

## Design

The bridge mostly has a single-contract architecture, and makes heavy use of libraries for code organisation. 

All APIs are available in the main `Gateway` contract.

### V1 
* `src/v1/IGateway.sol`: Interface for V1 API
* `src/v1/Types.sol`: Data Structures
* `src/v1/Calls.sol`: Direct calls on the Gateway itself.
* `src/v1/Handlers.sol`: Handlers for cross-chain bridge messages

### V2
* `src/v2/IGateway.sol`: Interface for V2 API
* `src/v2/Types.sol`: Data Structures
* `src/v2/Calls.sol`: Direct calls on the Gateway itself.
* `src/v2/Handlers.sol`: Handlers for cross-chain bridge messages

### Common Functions

Both V1 and V2 have some common functionality:

* `src/interfaces/IGatewayBase.sol`: Interface for shared APIs
* `src/Functions.sol`: Common implementation functions

### Deprecated Commands

These cross-chain bridge commands are no longer available:

* `CreateChannel`
* `UpdateChannel`
* `CreateAgent`
* `TransferNativeFromAgent`

For all intensive purposes there exists a single Channel and Agent, corresponding to the AssetHub parachain in Polkadot

### Unordered Message Delivery

A design flaw in V1 was that messages had to be delivered in the order they were committed on the source chain. This had numerous bad properties:

* Relayers were forced to deliver all messages, otherwise the bridge would halt.
* Higher-priority governance messages could get stuck behind lower-priority user messages.

In V2, messages can be delivered in any order. As a consequence, the concept of messaging _channels_ was made redundant, and no longer exists in V2.

### Fees & Incentives

A design flaw in V1 was that all fee & incentive calculations were performed on-chain, using static exchange rates. This resulted in on-chain complexity and high fees.

In V2, this is all moved offchain, and its the responsibility of the UX layer to direct the user to pass enough fees to cover execution and incentives.

All relaying incentives in both directions are claimable on BridgeHub.

#### Ethereum->Polkadot

The bridge fee is native ether. This can be used to pay for execution on the remote side in AssetHub, since it has a local DEX that convert wrapped ether to DOT (the native currency of AssetHub).

1. User calls `v2_sendMessage`, passing ether to cover execution on the remote destination, as well as incentives for the relayer.
2. This ether is locked up on the Ethereum side. 
3. When the message is received on BridgeHub, the ether included in the message (as `execution_fee` and `relayer_fee`) is used for execution and incentives respectively.
4. The relayer can claim the incentives on BridgeHub, and the reward is paid out on AssetHub.

#### Polkadot->Ethereum

The bridge fee on the Polkadot side is wrapped ether.

1. Relayer calls `v2_submit`, with the message and proof. They also need to supply an BridgeHub address which can later claim the relaying incentive.
2. The relayer must forward a delivery receipt back to BridgeHub. This is proof of emission of the `InboundMessageDispatched` event.
3. The relaying incentive (wrapped ether) is rendered claimable on BridgeHub, and is paid out on AssetHub

### Command Batches

In the Polkadot->Ethereum direction, a message can include multiple commands.

For example, a user may wish to transfer assets and call a contract using a single bridge message.

 ### Better support for XCM processing idioms

In the Ethereum->Polkadot polkadot direction, the bridging API is much more low-level and flexible.

The user can call `v2_sendMessage`, supplying an arbitrary XCM that perform any arbitrary action within the Polkadot network, including forwarding XCMs to other parachains. They can even supply other assets to be used for execution fees on these third-party parachains.

We do not expect end users to call `v2_sendMessage` directly.


yrong:
Cool! 
@claravanstaden @vgeddes May I merge https://github.com/Snowfork/snowbridge/pull/1321, I've tested the smoke tests against https://github.com/paritytech/polkadot-sdk/pull/7402 including legacy V1 and some V2 tests, all work as expected

```
// ENA
cargo test --test register_token
cargo test --test send_token
cargo test --test v2 transfer_ena::transfer_ena

// PNA
cargo test --test register_polkadot_token
cargo test --test v2 transfer_pna::transfer_pna
cargo test --test send_polkadot_token

```

claravanstaden:
@yrong yes please! :D

relayer/cmd/run/parachain/command.go
```diff
@@ -14,17 +14,19 @@ import (
...
 var (
-	configFile     string
-	privateKey     string
-	privateKeyFile string
-	privateKeyID   string
+	configFile          string
+	privateKey          string
+	privateKeyFile      string
+	privateKeyID        string
+	parachainPrivateKey string
```
claravanstaden, line 29: @yrong do we still need the substrate private key for the parachain relayer? Maybe for the delivery proofs? I don't see it in your [Parachain Relayer V2 PR](https://github.com/Snowfork/snowbridge/pull/1321) anymore, but perhaps it was in the v2 branch already.

Reason why I am asking is I want to add the key to the start-relayer scripts but just want to make sure it is still necessary.

yrong: Yeah, we need that for submitting delivery proofs to BH. 

I wrap a script specific for the V2 message relayer in https://github.com/Snowfork/snowbridge/blob/v2/web/packages/test/scripts/start-relayer-v2.sh and test by running a legacy V1 relayer side by side with the V2 message relayer.

In that way, I can test the V1 smoke tests still work as expected for the V2 setup.

Anyway, feel free to add that to start-relayer scripts.

contracts/src/v2/Types.sol
```diff
@@ -0,0 +1,170 @@
...
+    Asset[] assets;
+    Xcm xcm;
+    bytes claimer;
+    // ether value
+    uint128 value;
+    // additional ether value for execution fees
+    uint128 executionFee;
+    // additional ether value for relayer fees
+    uint128 relayerFee;
+}
```
yrong, line 44: I'd suggest adding an `id` field as V1 for tracing/indexing purposes.

The on-chain change is detailed in
https://github.com/paritytech/polkadot-sdk/pull/7402#discussion_r1976823135

alistair-singh: I do not think we need to do this. I think generating a hash based on the nonce is a good enough solution. However if the UX designs to set the topic id manually in the XCM we should obey it. I have added more details here:
https://github.com/paritytech/polkadot-sdk/pull/7402/files/27189594b6ef9affbda2d2c6f870c67de6fdce3a#r1980442243

web/packages/api/src/index.ts
```diff
@@ -4,8 +4,8 @@ import { AbstractProvider, JsonRpcProvider, WebSocketProvider } from "ethers"
 import {
     BeefyClient,
     BeefyClient__factory,
-    IGateway,
-    IGateway__factory,
+    IGatewayV1 as IGateway,
+    IGatewayV1__factory as IGateway__factory,
```
yrong, line 8: Breaking changes.

yrong (approved):
+1

PR #1381, opened 2025-02-12
Gateway should no longer build token registration XCM

vgeddes:
Instead, I've modified the protocol so that:

1. Users can still pass _raw_ xcm blobs
2. The gateway itself can pass a "template id" like `CreateAssetXCM` of raw xcm.

Then the inbound-queue on BH can either passthrough the raw xcm or build the create asset XCM.


contracts/src/v2/Types.sol
```diff
@@ -44,6 +44,34 @@ struct Payload {
...
+    uint8 kind;
+    bytes data;
+}
+
+library XcmKind {
+    /// SCALE-encoded raw bytes for `VersionedXcm`
+    uint8 constant Raw = 0;
+    /// Create a new asset in the ForeignAssets pallet of AssetHub
+    uint8 constant CreateAsset = 1;
+}
```
vgeddes, line 57: @alistair-singh @claravanstaden @yrong to give more context, this change opens up the possibility of users not strictly needing to pass raw XCM through the bridge to Polkadot.

Instead they can pass in a simpler ABI-encoded data, like:

```
Transfer {
  destination: Location
  destination_fee: (Location, uint128),
  beneficiary: Location
  transact: Option<Bytes>
}
```

Then the inbound-queue can build an XCM that does the following:
* Swap native ether included in the message for `destination_fee`
* Adds an `InitiateTransfer` instruction that forwards assets to destination
  * Deposits into beneficiary account
  * Optionally executes an XCM::Transact

Would this be useful?

yrong: IMHO, it will introduce additional complexity to the on-chain logic. I still prefer the raw XCM version, which offloads the complexity to off-chain.

yrong:
@vgeddes @claravanstaden Since we've already merged https://github.com/vgeddes/polkadot-sdk/pull/13, we may also need to merge this PR to V2 to pass the smoke tests.

Let me know if we need any further tweaks!

PR #1384, opened 2025-02-16
Improve ERC20 implementation

vgeddes:
Various improvements:
* Use OZ EDCSA library instead of own buggy homebrew crypto
* Remove obsolete `increaseAllowance`, `decreaseAllowance` methods (even OZ erc20 implementation no longer implements them)
* Improve error codes
* Add unit and fuzz tests. Turns out we had no test coverage.


contracts/src/TokenLib.sol
```diff
@@ -6,178 +6,48 @@ pragma solidity 0.8.28;
...
-     * Requirements:
-     *
-     * - `to` cannot be the zero address.
-     */
     function mint(Token storage token, address account, uint256 amount) external {
-        if (account == address(0)) {
-            revert IERC20.InvalidAccount();
-        }
-
+        require(account != address(0), IERC20.InvalidReceiver(address(0)));
```
yrong, line 20: Why not use `revert` as it’s used in other cases?

contracts/src/TokenLib.sol
```diff
@@ -6,178 +6,48 @@ pragma solidity 0.8.28;
 
 import {IERC20} from "./interfaces/IERC20.sol";
 import {IERC20Permit} from "./interfaces/IERC20Permit.sol";
+import {ECDSA} from "openzeppelin/utils/cryptography/ECDSA.sol";
 
 library TokenLib {
```
alistair-singh, line 11: Do we still gain anything from having a separate token library? This will get inlined into the Token.sol? So is it worth the extra indirection and mental overhead?

contracts/test/Token.t.sol
```diff
@@ -0,0 +1,348 @@
...
+        uint256 transferAmount
+    ) public {
+        // Avoid zero address and ensure distinct addresses
+        vm.assume(owner != address(0) && spender != address(0) && receiver != address(0));
+        vm.assume(owner != spender && spender != receiver && owner != receiver);
+
+        // Ensure amounts are valid
+        vm.assume(mintAmount > 0 && allowanceAmount > 0 && transferAmount > 0);
+        vm.assume(mintAmount >= transferAmount);
+        vm.assume(allowanceAmount >= transferAmount);
```
alistair-singh, line 98: We need a simple test that rejects if the allowance is less than the amount transferred. The fuzz tests assume that the amount is always greater than allowance so we are not covered there.

contracts/src/TokenLib.sol
vgeddes: It gets inlined into `TokenLib.sol`, which is shared by instances of `Token.sol`.

contracts/src/TokenLib.sol
vgeddes: `require` does revert with the custom error `IERC20.InvalidReceiver` if the condition is not met.

contracts/src/TokenLib.sol
```diff
@@ -186,90 +56,85 @@ library TokenLib {
...
+        return keccak256(
+            abi.encode(
+                keccak256("EIP712Domain(string name,string version,uint256 chainId,address verifyingContract)"),
+                keccak256(bytes(name)),
+                keccak256(bytes("1")),
+                block.chainid,
+                address(this)
+            )
+        );
+    }
```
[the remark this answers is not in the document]
vgeddes: I benchmarked it, and creating a domainSeparator without caching only costs 4500 gas. I don't think its worth adding complexity for such little saving.

contracts/src/Token.sol
```diff
@@ -5,39 +5,23 @@
 pragma solidity 0.8.28;
 
 import {IERC20} from "./interfaces/IERC20.sol";
+import {IERC20Metadata} from "./interfaces/IERC20Metadata.sol";
```
claravanstaden, line 8: Nice!

contracts/src/Token.sol
```diff
@@ -48,147 +32,61 @@ contract Token is IERC20, IERC20Permit {
...
-    /**
-     * @dev Creates `amount` tokens and assigns them to `account`, increasing
-     * the total supply. Can only be called by the owner.
-     *
-     * Emits a {Transfer} event with `from` set to the zero address.
-     *
-     * Requirements:
-     *
-     * - `account` cannot be the zero address.
-     */
```
claravanstaden, line 79: Any reason why the docs were removed?

contracts/src/Token.sol
```diff
@@ -48,147 +32,61 @@ contract Token is IERC20, IERC20Permit {
...
-    {
-        token.permit(DOMAIN_SEPARATOR, issuer, spender, value, deadline, v, r, s);
+    function allowance(address _owner, address spender) external view returns (uint256) {
+        return token.allowance[_owner][spender];
     }
 
-    function balanceOf(address account) external view returns (uint256) {
-        return token.balance[account];
-    }
+    // IERC20Permit
```
claravanstaden, line 77: What does this comment mean in this place? Should it not be above `permit`?

contracts/src/TokenLib.sol
claravanstaden: Can `amount` be 0 here?

claravanstaden (approved):

contracts/src/Token.sol
vgeddes: To reduce redundancy. `Token.sol` is just the implementation of the `IERC20.sol` interface, which itself is documented.

contracts/test/Token.t.sol
vgeddes: Added in https://github.com/Snowfork/snowbridge/pull/1384/commits/ded3f51d0d854b7b45eb98d984beccc7a7f3c384

contracts/src/TokenLib.sol
vgeddes: Yes, it seems most ERC20 implementations allow 0-value transfers, to allow for better composability with DeFI apps.

More convo here: https://x.com/PaulRBerg/status/1644318310165626880

vgeddes: Added a test in https://github.com/Snowfork/snowbridge/pull/1384/commits/254cb361fe198777a5ea674a49dadbf955329148 to ensure zero-value transfers are allowed.

alistair-singh (approved):
+1

PR #1393, opened 2025-02-26
Improve fork tests for upgrade

vgeddes:
I've improved the fork tests to reduce the amount of code that was being mocked out.


alistair-singh (approved):
+1

PR #1396, opened 2025-02-28
Improve smoke tests

yrong:
Smoke test listed in https://www.notion.so/snowfork/Snowbridge-Tests-1a7296aaabef80cfb6e1ca1641051d0c

- [x] Fix existing tests
- [ ] Add the missing ones

### DIFF
--- a/smoketest/tests/send_token_to_penpal_v2.rs
+++ b/smoketest/tests/send_token_to_penpal_v2.rs
@@ -109,10 +109,9 @@ async fn send_token_to_penpal_v2() {
 			SetTopic(H256::random().into()),
 		]),
 	}]));
-	let encoded_xcm = message.encode();
 
-	let xcm = Bytes::from(encoded_xcm);
-	let claimer = Bytes::from(vec![]);
+	let xcm = Bytes::from(message.encode());
+	let claimer = Bytes::from(beneficiary().encode());
 	let assets = vec![weth_asset];
 
 	let receipt = gateway


smoketest/tests/v2/send_ena_to_penpal.rs
```diff
@@ -109,10 +109,9 @@ async fn send_token_to_penpal_v2() {
 			SetTopic(H256::random().into()),
 		]),
 	}]));
-	let encoded_xcm = message.encode();
 
-	let xcm = Bytes::from(encoded_xcm);
-	let claimer = Bytes::from(vec![]);
+	let xcm = Bytes::from(message.encode());
+	let claimer = Bytes::from(beneficiary().encode());
```
yrong, line 114: The claimer should be set with a 32-byte address, or the test will fail since the default H160 claimer does not work on AH.

claravanstaden: Cool, I think is fine for now. We discussed in the call on Friday to set the Snowbridge sovereign as the claimer, if the claimer is not provided.

smoketest/tests/v2/send_pna.rs
```diff
@@ -0,0 +1,110 @@
...
+	let expected_owner: AccountId32Substrate = (*SUBSTRATE_RECEIVER).into();
+
+	let mut event_found = false;
+	while let Some(Ok(block)) = blocks.next().await {
+		println!("Polling assethub block {} for minted event.", block.number());
+
+		let events = block.events().await.unwrap();
+		for event in events.find::<Minted>() {
+			let minted = event.unwrap();
+			assert_eq!(minted.who, expected_owner);
```
claravanstaden, line 102: Would be good to check the minted amount as well.

claravanstaden (approved):
Nice! 🥳

PR #1404, opened 2025-03-06
Log topic on ethereum side

vgeddes:
We missed this.


alistair-singh (approved):

yrong:
Some tests failed
```
forge test --match-test testSubmitRegisterPNA -vvvv
```
which requires test fixture from local setup.

@vgeddes You can leave it to me, I'll update the relayer in this PR.

yrong:
Tested this PR alongside https://github.com/vgeddes/polkadot-sdk/pull/35 using a local setup. The outbound smoke test works as expected.

I think we can merge these two PRs now.

PR #1411, opened 2025-03-14
Update indexer endpoint

yrong:
- Update the indexer endpoint to the newly deployed one
- Remove the deprecated subscan config


web/packages/api/src/environment.ts
```diff
@@ -10,11 +10,6 @@ export type Config = {
     SECONDARY_GOVERNANCE_CHANNEL_ID: string
     RELAYERS: Relayer[]
     PARACHAINS: { [paraId: string]: string }
-    SUBSCAN_API?: {
```
vgeddes, line 13: why do all these config keys need to be in uppercase? These arent environment variables.

alistair-singh (approved):

alistair-singh:
We should update the package.json so that we can create a release with the new urls.

PR #1415, opened 2025-03-17
Do not check for ED if dry run was successful

alistair-singh:
We play it safe by always checking for ED. But if the dry run succeeds it means the ED was covered by left over fees or by some other sufficient asset.

Resolves: SNO-1421


yrong (approved):
+1

PR #1419, opened 2025-03-18
Support PNA

yrong:
- [x] Transfer relay token to Ethereum
- [ ] Send relay token back
- [ ] Transfer PAL(penpal native) to Ethereum
- [ ]  Send PAL back

### DIFF
--- a/web/packages/api/package.json
+++ b/web/packages/api/package.json
@@ -8,7 +8,7 @@
     "url": "https://github.com/Snowfork/snowbridge.git",
     "directory": "web/packages/api"
   },
-  "main": "src/index.ts",
+  "main": "dist/index.js",
   "types": "dist/index.d.ts",
   "scripts": {
     "build": "tsc",
--- a/web/packages/api/package.json
+++ b/web/packages/api/package.json
@@ -8,7 +8,7 @@
     "url": "https://github.com/Snowfork/snowbridge.git",
     "directory": "web/packages/api"
   },
-  "main": "dist/index.js",
+  "main": "src/index.ts",
   "types": "dist/index.d.ts",
   "scripts": {
     "build": "tsc",
--- a/web/packages/api/src/assets_v2.ts
+++ b/web/packages/api/src/assets_v2.ts
@@ -77,6 +77,7 @@ export type Asset = {
     decimals: number
     isSufficient: boolean
     xc20?: string
+    location?: any
 }
 
 export type RegistryOptions = {
@@ -1156,6 +1157,21 @@ function addOverrides(envName: string, result: RegistryOptions) {
                 ],
             }
             break
+        }
+        case "westend_sepolia": {
+            result.assetOverrides = {
+                "1000": [
+                    {
+                        token: "0xF50fb50d65C8C1f6c72E4D8397c984933AfC8F7e".toLowerCase(),
+                        name: "WND",
+                        minimumBalance: 1n,
+                        symbol: "WND",
+                        decimals: 18,
+                        isSufficient: true,
+                        location: DOT_LOCATION,
+                    },
+                ],
+            }
         }
     }
 }
--- a/web/packages/api/src/environment.ts
+++ b/web/packages/api/src/environment.ts
@@ -588,7 +588,7 @@ export const SNOWBRIDGE_ENV: { [id: string]: SnowbridgeEnvironment } = {
             },
             RELAY_CHAIN_URL: "https://westend-rpc.polkadot.io",
             PARACHAINS: {
-                "1000": "wss://westend-asset-hub-rpc.polkadot.io",
+                "1000": "https://asset-hub-westend.dotters.network",
                 "1002": "https://westend-bridge-hub-rpc.polkadot.io",
             },
             GATEWAY_CONTRACT: "0x9ed8b47bc3417e3bd0507adc06e56e2fa360a4e9",
--- a/web/packages/api/src/toEthereum_v2.ts
+++ b/web/packages/api/src/toEthereum_v2.ts
@@ -168,7 +168,14 @@ export async function createTransfer(
     let messageId: string | undefined
     let tx: SubmittableExtrinsic<"promise", ISubmittableResult>
     if (sourceParaId === assetHubParaId) {
-        tx = createERC20AssetHubTx(parachain, ethChainId, tokenAddress, beneficiaryAccount, amount)
+        tx = createERC20AssetHubTx(
+            parachain,
+            ethChainId,
+            tokenAddress,
+            beneficiaryAccount,
+            amount,
+            ahAssetMetadata
+        )
     } else {
         messageId = await buildMessageId(
             parachain,
@@ -364,11 +371,19 @@ export async function getDeliveryFee(
     let assetHubExecutionFeeDOT = 0n
     let returnToSenderExecutionFeeDOT = 0n
     let returnToSenderDeliveryFeeDOT = 0n
-    const bridgeHubDeliveryFeeDOT = await calculateDeliveryFee(
-        assetHub,
-        registry.bridgeHubParaId,
-        xcm
-    )
+    // Todo: confirm why AssetHub Westend does not support payment apis
+    // 0.1 DOT as workaround
+    let bridgeHubDeliveryFeeDOT =
+        registry.parachains[registry.assetHubParaId].estimatedDeliveryFeeDOT || 1_000_000_000n
+    try {
+        bridgeHubDeliveryFeeDOT = await calculateDeliveryFee(
+            assetHub,
+            registry.bridgeHubParaId,
+            xcm
+        )
+    } catch (e) {
+        console.warn(`Asset Hub does not support payment apis. Error.`, e)
+    }
     if (parachain !== registry.assetHubParaId) {
         const returnToSenderXcm = buildParachainERC20ReceivedXcmOnDestination(
             assetHub.registry,
@@ -489,13 +504,16 @@ export async function validateTransfer(
     const [nativeBalance, dotBalance, tokenBalance] = await Promise.all([
         getNativeBalance(sourceParachain, sourceAccountHex),
         getDotBalance(sourceParachain, source.info.specName, sourceAccountHex),
-        getTokenBalance(
-            sourceParachain,
-            source.info.specName,
-            sourceAccountHex,
-            registry.ethChainId,
-            tokenAddress
-        ),
+        transfer.computed.ahAssetMetadata.location?.parents == DOT_LOCATION.parents &&
+        transfer.computed.ahAssetMetadata.location?.interior == DOT_LOCATION.interior
+            ? getNativeBalance(sourceParachain, sourceAccountHex)
+            : getTokenBalance(
+                  sourceParachain,
+                  source.info.specName,
+                  sourceAccountHex,
+                  registry.ethChainId,
+                  tokenAddress
+              ),
     ])
 
     if (amount > tokenBalance) {
@@ -999,9 +1017,14 @@ function createERC20AssetHubTx(
     ethChainId: number,
     tokenAddress: string,
     beneficiaryAccount: string,
-    amount: bigint
+    amount: bigint,
+    asset: Asset
 ): SubmittableExtrinsic<"promise", ISubmittableResult> {
-    const assetLocation = erc20Location(ethChainId, tokenAddress)
+    let assetLocation = erc20Location(ethChainId, tokenAddress)
+    // Asset with location not null for PNA
+    if (asset.location) {
+        assetLocation = asset.location
+    }
     const assets = {
         v4: [
             {
--- a/web/packages/api/src/xcmBuilder.ts
+++ b/web/packages/api/src/xcmBuilder.ts
@@ -2,7 +2,7 @@ import { Registry } from "@polkadot/types/types"
 import { beneficiaryMultiAddress } from "./utils"
 import { ETHER_TOKEN_ADDRESS } from "./assets_v2"
 
-export const DOT_LOCATION = { parents: 1, interior: { here: null } }
+export const DOT_LOCATION = { parents: 1, interior: "Here" }
 
 const ethereumNetwork = (ethChainId: number) => ({
     GlobalConsensus: { Ethereum: { chain_id: ethChainId } },
--- a/web/packages/operations/.gitignore
+++ b/web/packages/operations/.gitignore
@@ -2,3 +2,5 @@ node_modules
 dist
 .env
 config
+.env.*
+!.env.example
--- a/web/packages/operations/.westend_sepolia.registry.json
+++ b/web/packages/operations/.westend_sepolia.registry.json
@@ -1,0 +1,107 @@
+{
+  "environment": "westend_sepolia",
+  "ethChainId": 11155111,
+  "gatewayAddress": "0x9ed8b47bc3417e3bd0507adc06e56e2fa360a4e9",
+  "assetHubParaId": 1000,
+  "bridgeHubParaId": 1002,
+  "relaychain": {
+    "tokenSymbols": "WND",
+    "tokenDecimals": 12,
+    "ss58Format": 42,
+    "isEthereum": false,
+    "accountType": "AccountId32",
+    "name": "Westend",
+    "specName": "westend",
+    "specVersion": 1018001
+  },
+  "bridgeHub": {
+    "tokenSymbols": "WND",
+    "tokenDecimals": 12,
+    "ss58Format": 42,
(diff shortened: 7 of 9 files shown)


yrong:
@alistair-singh Please check https://github.com/vgeddes/polkadot-sdk/pull/64#issue-2935626093 when you're back.

web/packages/api/src/toEthereum_v2.ts
```diff
@@ -347,38 +374,84 @@ export async function getDeliveryFee(
...
-    )
-    if (parachain !== registry.assetHubParaId) {
-        const returnToSenderXcm = buildParachainERC20ReceivedXcmOnDestination(
+    let xcm: any
+
+    if (sourceAssetMetadata.location) {
+        xcm = buildResultXcmAssetHubPNATransferFromParachain(
             assetHub.registry,
             registry.ethChainId,
+            sourceAssetMetadata,
```
alistair-singh, line 386: Since this is just an estimate of the fee, you do not need to pass in the asset token and rather just pass in a dummy token address or multilocaiton.

web/packages/api/src/toEthereum_v2.ts
```diff
@@ -347,38 +374,84 @@ export async function getDeliveryFee(
...
+            "0x0000000000000000000000000000000000000000",
+            "0x0000000000000000000000000000000000000000000000000000000000000000",
+            340282366920938463463374607431768211455n,
+            340282366920938463463374607431768211455n,
+            340282366920938463463374607431768211455n,
+            parachain,
+            340282366920938463463374607431768211455n
+        )
+    }
+    console.log("xcm on asset hub for pna transfer from 3rd parachain:", xcm.toHuman())
```
alistair-singh, line 411: this will show in the debug console in the browser. So maybe remove logs like this.

web/packages/api/src/toEthereum_v2.ts
```diff
@@ -347,38 +374,84 @@ export async function getDeliveryFee(
...
+
+    let assetHubExecutionFeeDOT = 0n
+    let returnToSenderExecutionFeeDOT = 0n
+    let returnToSenderDeliveryFeeDOT = 0n
+    let bridgeHubDeliveryFeeDOT =
+        registry.parachains[registry.assetHubParaId].estimatedDeliveryFeeDOT || 1_000_000_000n
+    const ahParachain = registry.parachains[registry.assetHubParaId]
+    if (ahParachain.features.hasXcmPaymentApi) {
+        // Todo: Seems incorrect--the XCM here is the one executed on AH,
```
alistair-singh, line 420: You are correct, the original version was incorrect. So basically now we are over-estimating the size of the blob by measuring  a larger blob and in effect charging the user a slightly higher delivery fee.

web/packages/api/src/toEthereum_v2.ts
```diff
@@ -347,38 +374,84 @@ export async function getDeliveryFee(
...
+                source.registry,
+                registry.ethChainId,
+                "0x0000000000000000000000000000000000000000",
+                340282366920938463463374607431768211455n,
+                340282366920938463463374607431768211455n,
+                "0x0000000000000000000000000000000000000000000000000000000000000000",
+                "0x0000000000000000000000000000000000000000000000000000000000000000"
+            )
+        }
+        console.log("xcm received on 3rd parachain for pna transfer:", returnToSenderXcm.toHuman())
```
alistair-singh, line 453: Ditto: Remove unnecessary logging.

web/packages/api/src/toEthereum_v2.ts
```diff
@@ -481,22 +554,36 @@ export async function validateTransfer(
...
 
-    const [nativeBalance, dotBalance, tokenBalance] = await Promise.all([
+    const [nativeBalance, dotBalance] = await Promise.all([
         getNativeBalance(sourceParachain, sourceAccountHex),
         getDotBalance(sourceParachain, source.info.specName, sourceAccountHex),
-        getTokenBalance(
+    ])
+    let tokenBalance: any
+    if (
+        transfer.computed.ahAssetMetadata.location?.parents == DOT_LOCATION.parents &&
```
alistair-singh, line 573: Maybe we should use the token address for DOT for this. I am not sure if this will match every version of xcm. We need to find a way to normalize multi-locations. I think in a separate PR we need to have a runtime API on bridgehub which takes in a versioned multi-location and returns the token_id. And when we index the assets on each chain, we normalize it to the token_id and use the token id as a stable id for multi-locations.

web/packages/api/src/toEthereum_v2.ts
```diff
@@ -1198,3 +1309,55 @@ async function buildMessageId(
...
+        parachain.registry,
+        ethChainId,
+        sourceAccount,
+        beneficiaryAccount,
+        asset,
+        messageId,
+        sourceParaId,
+        returnToSenderFeeInDOT
+    )
+    console.log("custom xcm for pna transfer from 3rd parachain:" + customXcm.toPrimitive())
```
alistair-singh, line 1353: Ditto: remove uneeded logging.

web/packages/api/src/assets_v2.ts
```diff
@@ -77,6 +78,13 @@ export type Asset = {
     decimals: number
     isSufficient: boolean
     xc20?: string
+    // Location on source Parachain
+    location?: any
+    // Location reanchored on AH
+    locationOnAH?: any
+    // Location reanchored on Ethereum
+    locationOnEthereum?: any
```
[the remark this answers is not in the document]
yrong: It's a bit challenge to implement the reanchor logic in javascript, I prefer to make things easy for now.

web/packages/api/src/assets_v2.ts
```diff
@@ -77,6 +78,13 @@ export type Asset = {
...
+    assetId?: number
```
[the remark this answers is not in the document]
yrong: https://github.com/Snowfork/snowbridge/pull/1419/commits/2f2256fc7894fc47f3428d26997002d09c4bd569

web/packages/api/src/assets_v2.ts
```diff
@@ -542,7 +551,19 @@ export async function getLocationBalance(
             ).toPrimitive() as any
             return BigInt(accountData?.free ?? 0n)
         }
-        case "penpal-parachain":
+        case "penpal-parachain": {
```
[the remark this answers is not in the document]
yrong: TBH I did't.  

For Hydration and Bifrost, it seems the `assetRegistry` pallet contains the PNA so I'd assume the legacy codes (quering AssetId by MultiLocation) should just work.

For Moonbeam, it seems the `assetRegistry` only contains the foreignAssets, so we may need to pass the assetId directly, as Penpal does.

Since we can't really support PNA from these Parachains as we discussed, I'm not too worried about this at the moment. We can revist and test it once these chains have integrate with AH.

web/packages/api/src/assets_v2.ts
```diff
@@ -610,9 +631,16 @@ export function getTokenBalance(
...
     ethChainId: number,
-    tokenAddress: string
+    tokenAddress: string,
+    asset?: Asset
 ) {
-    return getLocationBalance(provider, specName, erc20Location(ethChainId, tokenAddress), account)
+    return getLocationBalance(
+        provider,
+        specName,
+        asset?.location || erc20Location(ethChainId, tokenAddress),
```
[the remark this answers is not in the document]
yrong: Not much difference between the nullish coalescing operator and the logical OR operator in this case?

web/packages/api/src/environment.ts
```diff
@@ -588,7 +588,7 @@ export const SNOWBRIDGE_ENV: { [id: string]: SnowbridgeEnvironment } = {
             },
             RELAY_CHAIN_URL: "https://westend-rpc.polkadot.io",
             PARACHAINS: {
-                "1000": "wss://westend-asset-hub-rpc.polkadot.io",
+                "1000": "https://asset-hub-westend.dotters.network",
```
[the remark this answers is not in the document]
yrong: https://github.com/Snowfork/snowbridge/pull/1419/commits/e88c94121b790c9eb5cd77aea1a7900ac92531ab

web/packages/api/src/xcmBuilder.ts
```diff
@@ -529,3 +529,458 @@ export function buildResultXcmAssetHubERC20TransferFromParachain(
...
+            ),
+        ],
+    })
+}
+
+function buildAssetHubXcmForPNAFromParachain(
+    ethChainId: number,
+    sourceAccount: string,
+    beneficiary: string,
+    asset: Asset,
```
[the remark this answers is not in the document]
yrong: https://github.com/Snowfork/snowbridge/pull/1419/commits/9e8a169725a78b6d9a4006c0708cb1dfc447fbb5

web/packages/api/src/xcmBuilder.ts
```diff
@@ -529,3 +529,458 @@ export function buildResultXcmAssetHubERC20TransferFromParachain(
...
+            },
+            { setTopic: topic },
+        ],
+    })
+}
+
+export function buildAssetHubPNAReceivedXcm(
+    registry: Registry,
+    ethChainId: number,
+    asset: Asset,
```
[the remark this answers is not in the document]
yrong: Not sure I fully understand the point here. The function is in use, and the asset is used to build the `WithdrawAsset` instruction to withdraw reserved PNA from the bridge sovereign on AH.

web/packages/api/src/toEthereum_v2.ts
yrong: I prefer to use the actual value if possible.

web/packages/api/src/toEthereum_v2.ts
yrong: https://github.com/Snowfork/snowbridge/pull/1419/commits/d46b66e5ec02da79143dfb32d4a3080b94baf0f9

web/packages/api/src/toEthereum_v2.ts
yrong: https://github.com/Snowfork/snowbridge/pull/1419/commits/8b70d62602591da474babccf8fe2ab9f64a4f026

web/packages/api/src/toEthereum_v2.ts
yrong: > have a runtime API on bridgehub which takes in a versioned multi-location and returns the token_id

This branch is only for transferring `DOT` from AH, which is a special case that retrieves the token balance from the `Balance` pallet. I also tested that the `{ parents: 1, interior: "Here" }` representation works for both XCMV4 and XCMV5. So, I prefer to keep things simple here.

web/packages/api/src/xcmBuilder.ts
```diff
@@ -289,21 +289,12 @@ export function buildParachainERC20ReceivedXcmOnAssetHub(
             },
             {
                 depositReserveAsset: {
+                    // Should use `AllCounted` here. Reference:
+                    // https://github.com/paritytech/polkadot-sdk/blob/f5de39196e8c30de4bc47a2d46b1a0fe1e9aaee0/bridges/snowbridge/primitives/inbound-queue/src/v1.rs#L357-L359
```
yrong, line 293: Seems like a bug in the previous version. Made a potential fix here.

web/packages/api/src/xcmBuilder.ts
```diff
@@ -289,21 +289,12 @@ export function buildParachainERC20ReceivedXcmOnAssetHub(
...
-                            {
-                                id: erc20Location(ethChainId, tokenAddress),
-                                fun: {
-                                    Fungible: transferAmount,
-                                },
-                            },
-                        ],
+                        wild: {
+                            AllCounted: 2,
+                        },
```
yrong, line 297: Seems like a bug in the previous version. Made a potential fix here.

web/packages/api/src/toEthereum_v2.ts
```diff
@@ -347,38 +385,97 @@ export async function getDeliveryFee(
...
+            340282366920938463463374607431768211455n,
             340282366920938463463374607431768211455n,
             340282366920938463463374607431768211455n,
+            parachain,
+            340282366920938463463374607431768211455n
+        )
+        forwardedXcm = buildExportXcmForERC20(
+            assetHub.registry,
+            registry.ethChainId,
+            tokenAddress,
```
alistair-singh, line 433: Do we need this? Can we not just pass in empty token for estimation?
```suggestion
            "0x0000000000000000000000000000000000000000",
```

web/packages/api/src/toEthereum_v2.ts
```diff
@@ -1198,3 +1341,99 @@ async function buildMessageId(
...
+    const assets = {
+        v4: [
+            {
+                id: asset.location,
+                fun: { Fungible: amount },
+            },
+        ],
+    }
+    const feeAsset = {
+        v4: asset.location,
```
alistair-singh, line 1410: Here we implicitly swap "KSM" (or any other bridged asset) for DOT during execution. The UI doesn't really inform the user of the swap and the cost implications associated with it. So let's pay with DOT for this release. It is ok to hardcode DOT as the fee here.

web/packages/api/src/toEthereum_v2.ts
```diff
@@ -1198,3 +1341,99 @@ async function buildMessageId(
     ])
     return blake2AsHex(entropy)
 }
+
+function createPNASourceParachainTx(
```
alistair-singh, line 1345: We need to make sure this case is disabled. We do not support it for initial release.

web/packages/api/src/toEthereum_v2.ts
yrong: There is no swap in this case. It's for transfer KSM from AH to Ethereum. The fee asset is same as the PNA in V1.

web/packages/api/src/toEthereum_v2.ts
yrong: I'd prefer to leave it here. The first released PNAs only covers the native asset on AH plus KSM from AH. So they won't touch this branch.

alistair-singh: I am not saying we should remove the code path. I am saying we should make sure users never hit this code path until we officially support it. Are we sure that the only assets that we list as transferable in mainnet do not trigger this code path?

web/packages/api/src/toEthereum_v2.ts
alistair-singh: Ah ok my mistake, the issue is with my understanding is that when asset hub is the source, the delivery fee is taken directly from the users DOT balance and not from inside the XCM holding/fee register.

web/packages/api/src/assets_v2.ts
```diff
@@ -1154,6 +1203,178 @@ function addOverrides(envName: string, result: RegistryOptions) {
...
+                        decimals: 10,
+                        isSufficient: false,
+                        location: {
+                            parents: 0,
+                            interior: { X2: [{ palletInstance: 50 }, { generalIndex: 31337 }] },
+                        },
+                        assetId: "31337",
+                    },
+                ],
+            }
```
alistair-singh, line 1287: Why does asset discovery not pick up these assets?

yrong: The reason mentioned in https://linear.app/snowfork/issue/SNO-1207#comment-de6709cc

web/packages/api/src/toEthereum_v2.ts
yrong: > Are we sure that the only assets that we list as transferable in mainnet do not trigger this code path?

Yeah, it's only for multi-hop transfers from a source parachain to Ethereum.

web/packages/api/src/toEthereum_v2.ts
yrong: Same as https://github.com/Snowfork/snowbridge/pull/1419#discussion_r2026997065?

web/packages/api/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@snowbridge/api",
-  "version": "0.1.46",
+  "version": "0.1.47",
```
alistair-singh, line 3: This clashes with my PR https://github.com/Snowfork/snowbridge/pull/1450. May need to sync here.

web/packages/api/src/toEthereum_v2.ts
alistair-singh: But this introduces a new parameter to getting the fee estimate. Not sure if we really need it.

alistair-singh (approved):
1. Please make tickets for the asset discovery.
2. The runtime endpoint we need to resolve versioned multi-locations to token ids.

PR #1422, opened 2025-03-20
Merge from V1

yrong:


contracts/src/v2/IGateway.sol
```diff
@@ -29,9 +29,7 @@ interface IGatewayV2 {
     event AgentCreated(bytes32 agentID, address agent);
 
     /// Emitted when inbound message has been dispatched
-    event InboundMessageDispatched(
-        uint64 indexed nonce, bytes32 topic, bool success, bytes32 rewardAddress
-    );
+    event InboundMessageDispatched(uint64 indexed nonce, bool success, bytes32 rewardAddress);
```
yrong, line 32: Adding the topic ID here is good for the indexer, but we missed the change on Substrate side. Otherwise, the `submit_delivery_proof` extrinsic will fail.

Since the upstream PR is almost finalized, just revert it here.

yrong: Reverted

smoketest/src/helper.rs
```diff
@@ -235,6 +238,41 @@ pub async fn fund_account(
 	Ok(())
 }
 
+pub async fn construct_create_agent_call(
```
vgeddes: We removed these three governance calls from latest Polkadot-SDK code. We should probably remove the smoketests as well, right?

contracts/src/upgrades/westend/Gateway20250311.sol
vgeddes, line 1: not necessary for the V2 branch, where it will just be dead code. This only makes sense in the V1 stable branch which we will create.

contracts/scripts/upgrades/westend/DeployGateway20250311.sol
vgeddes, line 1: not necessary for the V2 branch, where it will just be dead code. This only makes sense in the V1 stable branch which we will create.

contracts/scripts/DeployLocalGatewayLogic.sol
vgeddes, line 1: not necessary for the V2 branch, where it will just be dead code. This only makes sense in the V1 stable branch which we will create.

smoketest/src/helper.rs
yrong: Already removed - seems like the comment on code is outdated.

contracts/scripts/DeployLocalGatewayLogic.sol
yrong: https://github.com/Snowfork/snowbridge/pull/1422/commits/e4aecdbb59f5e75ce827247554cc50d99121cb24

vgeddes (approved):

PR #1434, opened 2025-04-02
Tune alarm configurations

yrong:


web/packages/api/src/environment.ts
```diff
@@ -491,7 +491,7 @@ export const SNOWBRIDGE_ENV: { [id: string]: SnowbridgeEnvironment } = {
             },
             RELAY_CHAIN_URL: "https://polkadot-rpc.dwellir.com",
             PARACHAINS: {
-                "1000": "wss://asset-hub-polkadot-rpc.dwellir.com",
+                "1000": "wss://sys.ibp.network/asset-hub-polkadot",
```
alistair-singh, line 494: Why not use dwellir here?

alistair-singh (commented):
Looks good. Minor comments.

web/packages/api/src/environment.ts
yrong: When I tried to reproduce https://snowfork.slack.com/archives/C072WCDQ67M/p1743598481917979?thread_ts=1743597070.201479&cid=C072WCDQ67M, I tested the RPC endpoint with and without the ".n." endpoints `wss://asset-hub-polkadot-rpc.n.dwellir.com`, but neither worked. This caused the monitoring service to be blocked and triggered a `Missing data` false alarm.

alistair-singh: We should maybe be able to override this in the monitoring service via variable instead of this URL which gets used by the UI as well.

claravanstaden: I think its better to rely on Dwellir, with whom we have a relationship. Let's hear what they say about this instead of changing it

yrong: I've reverted that change in https://github.com/Snowfork/snowbridge/pull/1434/commits/a8546e72102fc427c709ac8c6d379612c801451a

PR #1437, opened 2025-04-04
Hack to force relay commitment [DO NOT MERGE]

alistair-singh:


relayer/relays/beefy/scanner.go
```diff
@@ -19,7 +20,7 @@ type ScanBlocksResult struct {
 
 func ScanBlocks(ctx context.Context, meta *types.Metadata, api *gsrpc.SubstrateAPI, startBlock uint64) (chan ScanBlocksResult, error) {
 	results := make(chan ScanBlocksResult)
-	go scanBlocks(ctx, meta, api, startBlock, results)
+	go scanBlocks(ctx, meta, api, 25420890, results)
```
alistair-singh, line 23: Hard code block number to 5 blocks before the block that has the bad commitment so that we can skip directly to the block we want.

relayer/relays/beefy/scanner.go
```diff
@@ -108,7 +109,7 @@ func scanBlocks(ctx context.Context, meta *types.Metadata, api *gsrpc.SubstrateA
 		}
 
 		if sessionIndex > currentSessionIndex {
-			currentSessionIndex = sessionIndex
+			//currentSessionIndex = sessionIndex
```
alistair-singh, line 112: Do not increment the session, which essentially makes it try every block in the next session, pushing all found commitments through instead of the first commitment per session.

relayer/relays/beefy/scanner.go
```diff
@@ -171,8 +172,10 @@ func scanCommitments(ctx context.Context, meta *types.Metadata, api *gsrpc.Subst
 
 			commitment, proof, err := fetchCommitmentAndProof(meta, api, result.BlockHash)
 			if err != nil {
-				emitError(fmt.Errorf("fetch commitment and proof: %w", err))
-				return
+				logrus.WithError(err).Error("scanned commitment")
+				//emitError(fmt.Errorf("fetch commitment and proof: %w", err))
+				//return
+				continue
```
alistair-singh, line 178: When there is an error skip to the next commitment.

yrong:
Cool!  Though it seems more like a workaround. I noticed that in https://github.com/Snowfork/snowbridge/pull/1342 we removed the `on-demand` command, which allows one-shot syncing from an arbitrary Beefy block-not only the boundry session block.

I'd suggest to add it back-it could be useful for scenarios like this.

yrong:
>  allows one-shot syncing from an arbitrary Beefy block-not just the boundry session block.

@alistair-singh I added it back in https://github.com/Snowfork/snowbridge/pull/1449

PR #1440, opened 2025-04-08
Audit Issue 8: Adds contract address validation when registering a native token

claravanstaden:
Resolves: [SNO-1461](https://linear.app/snowfork/issue/SNO-1461)


contracts/src/Functions.sol
```diff
@@ -94,6 +94,11 @@ library Functions {
     }
 
     function registerNativeToken(address token) internal {
+        // Validate that the token address is a contract
```
vgeddes, line 97: Should remove the check in the legacy v1 implementation since that version also calls into this function: https://github.com/Snowfork/snowbridge/blob/3388275efafc2753abe0c614288e8b4b3c777057/contracts/src/v1/Calls.sol#L72

claravanstaden: Addressed in https://github.com/Snowfork/snowbridge/pull/1440/commits/3c97fc077f42f20229f4d71370b86d6716b2b591. There is a change in behaviour for v1 here, where previously token could be re-registered. We are changing this behaviour by [reverting](https://github.com/Snowfork/snowbridge/blob/main/contracts/src/Functions.sol#L105) if the token has already been registered.

vgeddes: That's only for foreign tokens. Notice the `&& info.isForeign()` check.

vgeddes: I'm a bit confused because the code in `main` has this implementation, which does allow re-registration for native tokens:

```solidity
    function registerNativeToken(address token) internal {
        // NOTE: Explicitly allow a native token to be re-registered. This offers resiliency
        // in case a previous registration attempt of the same token failed on the remote side.
        // It means that registration can be retried.

        if (info.isRegistered && info.isForeign()) {
            // Prevent re-registration of foreign tokens as native tokens
            revert IGatewayBase.TokenAlreadyRegistered();
        }

        info.isRegistered = true;
    }
```

claravanstaden: @vgeddes you're right, reverted the comment in https://github.com/Snowfork/snowbridge/pull/1440/commits/86532d35cbccd34b03d6e980822044cd29173a6a.

vgeddes (approved):

PR #1442, opened 2025-04-09
Audit Issue 22: Optimize sparse bitmap

claravanstaden:
Resolves: [SNO-1475](https://linear.app/snowfork/issue/SNO-1475)

### DIFF
--- a/contracts/src/utils/SparseBitmap.sol
+++ b/contracts/src/utils/SparseBitmap.sol
@@ -2,19 +2,19 @@
 pragma solidity 0.8.28;
 
 struct SparseBitmap {
-    mapping(uint256 bucket => uint256) data;
+    mapping(uint64 => uint128) data;
 }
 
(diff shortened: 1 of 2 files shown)


yrong (approved):
Cool!

vgeddes (commented):
I think the audit feedback isn't applicable to the Solidity implementation of `SparseBitmap`.

This implementation is intended to be general-purpose because it supports all key types, not just `uint64`, without sacrificing performance.

In any case, as you may know, in Solidity, the atomic word length is always 32-bytes and so `uint64` occupies the same space as `uint256`.

Also not sure why you made the bucket value `uint128`? Since `uint256` is the default numeric type in Solidity, and because it can store 256 bits, it can store more nonces than `uint128`.

claravanstaden:
@vgeddes the audit issue does mention the Solidity code as well, but I agree with you, it is perhaps not necessary.

PR #1443, opened 2025-04-09
Audit Issue 23: Handle 0 length case for makeIndex

claravanstaden:
Resolves: [SNO-1476](https://linear.app/snowfork/issue/SNO-1476)


yrong:
Not sure this is nessessary. I'd assume the validate data set is updated by Beefy client through proof verification, so it shouldn't be zero in any case.

vgeddes (approved):

claravanstaden:
@yrong perhaps, but the defensive check can't hurt so I am going to add it.

PR #1445, opened 2025-04-09
Audit Issue 25: Remove outdated comment

claravanstaden:
Resolves: [SNO-1478](https://linear.app/snowfork/issue/SNO-1478)

### DIFF
--- a/contracts/src/Functions.sol
+++ b/contracts/src/Functions.sol
@@ -94,9 +94,6 @@ library Functions {
     }
 
     function registerNativeToken(address token) internal {
-        // NOTE: Explicitly allow a native token to be re-registered. This offers resiliency
-        // in case a previous registration attempt of the same token failed on the remote side.
-        // It means that registration can be retried.
         AssetsStorage.Layout storage $ = AssetsStorage.layout();
         TokenInfo storage info = $.tokenRegistry[token];
 


yrong:
The comment reminds me that in V1, we allowed PNA to be re-registered with modified metadata, which does't seem possible in V2?

claravanstaden:
> The comment reminds me that in V1, we allowed PNA to be re-registered with modified metadata, which does't seem possible in V2?

Yes, related to https://github.com/Snowfork/snowbridge/pull/1440#discussion_r2036835954.

contracts/src/Functions.sol
```diff
@@ -94,9 +94,6 @@ library Functions {
     }
 
     function registerNativeToken(address token) internal {
-        // NOTE: Explicitly allow a native token to be re-registered. This offers resiliency
-        // in case a previous registration attempt of the same token failed on the remote side.
-        // It means that registration can be retried.
         AssetsStorage.Layout storage $ = AssetsStorage.layout();
         TokenInfo storage info = $.tokenRegistry[token];
```
vgeddes, line 98: but the code in main does seem to allow re-registration?

claravanstaden: You're right, closing this issue.

PR #1449, opened 2025-04-10
Fix on-demand relaying of Beefy commitments

yrong:
Resolves: https://linear.app/snowfork/issue/SNO-1449 & https://linear.app/snowfork/issue/SNO-1499


alistair-singh (commented):
This looks like it only handles v2?

yrong:
> This looks like it only handles v2?

I'll cherry pick https://github.com/Snowfork/snowbridge/pull/1449/commits/e107c77398bc351998732a748916eacdc9d22090 to V1.

vgeddes (commented):
I think the approach here needs to be improved:

1. The on-demand syncer needs to support both V1 & V2
2. For V2, instead of comparing nonces, consider querying `OutboundQueue.PendingOrders`, looking for the highest nonce, and making sure the Beefy commitment covering that nonce is synced.

contracts/src/Gateway.sol
```diff
@@ -422,6 +422,9 @@ contract Gateway is IGatewayBase, IGatewayV1, IGatewayV2, IInitializable, IUpgra
         }
 
         $.inboundNonce.set(message.nonce);
+        if (message.nonce > $.maxInboundNonce) {
+            $.maxInboundNonce = message.nonce;
```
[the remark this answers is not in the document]
claravanstaden: Just because of this check? Or is it the storage that is expensive? Just wondering if it also affects https://github.com/Snowfork/snowbridge/pull/1443.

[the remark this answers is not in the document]
yrong: Yeah, I did not relealize will be increased that much. I'll try to find a solution that doesn’t require any on-chain changes.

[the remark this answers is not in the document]
yrong: https://github.com/Snowfork/snowbridge/pull/1449/commits/9b978098b5a6a34c74128d7567282b1c3691f73b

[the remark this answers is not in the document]
vgeddes: Looks better though the algorithm needs to maintain support for V1. Or in other words, you need to check nonces for v1 and nonces for v2, and then decide whether to sync.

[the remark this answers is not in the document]
yrong: Support both V1 and V2 with the same Beefy relayer.

https://github.com/Snowfork/snowbridge/pull/1449/commits/f4d0dc56de41df983215895ee9a3ba77e84b66f7

web/packages/test/scripts/start-relayer.sh
```diff
@@ -79,6 +79,19 @@ start_relayer() {
             sleep 20
         done
     ) &
+    # Launch beefy on demand relay
```
vgeddes, line 82: In the local env, the session length is like only 10 blocks, which means the on-demand relay isn't necessary and could conflict with the main beefy relayer.

What behavior did you observe in the local env?

Maybe we could add a check in the on-demand relay, where if it sees that the session length is less than 10 minutes it does nothing?

relayer/relays/beefy/on-demand-sync.go
```diff
@@ -72,101 +83,93 @@ func (relay *OnDemandRelay) Start(ctx context.Context) error {
...
-	}
-
-}
-
-func sleep(ctx context.Context, d time.Duration) {
-	select {
-	case <-ctx.Done():
-		return
-	case <-time.After(d):
+		time.Sleep(time.Second * 10)
```
vgeddes, line 172: Why change the implementation of `sleep`? This new implementation doesn't handle the case when `ctx` is cancelled, and it means that the relayer process won't exit properly on error.

relayer/relays/beefy/on-demand-sync.go
```diff
@@ -72,101 +83,93 @@ func (relay *OnDemandRelay) Start(ctx context.Context) error {
...
+		log.Info("Sync completed")
+
+		relay.waitUntilMessagesSynced(ctx, paraNonce)
 	}
+	return nil
 }
 
 func (relay *OnDemandRelay) waitUntilMessagesSynced(ctx context.Context, paraNonce uint64) {
-	sleep(ctx, time.Minute*10)
+	waitingTime := 0
```
vgeddes, line 161: This new approach where the waiting times out creates a problem: In the next iteration of the loop within `relay.Start`, the same commitment will be synced, costing money. And this will continue forever until the message relayer finally syncs the user message.

So for V1, we should rather have `waitUntilMessagesSynced` wait forever (i.e revert these changes). Since we run the user message relayers we need to keep them up an running always.

web/packages/test/scripts/start-relayer.sh
yrong: > the session length is like only 10 blocks

Actually, it's 20 blocks (2 mins), so I can test it within a reasonable time window. Since it's a local setup, I assume it's not a big deal even if it might cause conflicts, and I'd prefer to add it.

relayer/cmd/run/beefy/command.go
```diff
@@ -94,24 +94,20 @@ func run(_ *cobra.Command, _ []string) error {
...
 
-		err = relay.Start(ctx)
-		if err != nil {
-			logrus.WithError(err).Fatal("Unhandled error")
-			cancel()
+		eg.Go(func() error {
+			err = relay.Start(ctx)
 			return err
-		}
+		})
```
yrong, line 110: Run the on-demand relay in a separate goroutine.

relayer/relays/beefy/on-demand-sync.go
yrong: > This new implementation doesn't handle the case when ctx is cancelled, and it means that the relayer process won't exit properly on error.

This won't happen, as we will run the on-demand relay in a separate goroutine. As mentioned in https://github.com/Snowfork/snowbridge/pull/1449#discussion_r2079728457.

That way, we can remove the internal checks everywhere and make error handling cleaner.

relayer/relays/beefy/on-demand-sync.go
```diff
@@ -72,101 +83,93 @@ func (relay *OnDemandRelay) Start(ctx context.Context) error {
...
+		// Check if we are rate-limited
+		if !relay.tokenBucket.TryConsume(1) {
+			return fmt.Errorf("Rate-limit exceeded")
+		}
 
-			beefyBlockHash, err := relay.relaychainConn.API().RPC.Beefy.GetFinalizedHead()
-			if err != nil {
-				if errors.Is(err, context.Canceled) {
-					return nil
-				}
```
yrong, line 115: Since we've moved the relay into an external goroutine,  checking for context cancellation here is not nessessary.

relayer/relays/beefy/on-demand-sync.go
yrong: https://github.com/Snowfork/snowbridge/pull/1449/commits/e09f957af4bd3077785c348d0a82aae1530897f1

web/packages/test/scripts/start-relayer.sh
vgeddes: It is problematic in a local env if it causes conflicts and undefined behavior. We want local env to be repeatable and consistent.

relayer/cmd/run/beefy/command.go
```diff
@@ -94,24 +94,20 @@ func run(_ *cobra.Command, _ []string) error {
 			return err
 		}
 
-		err = relay.Start(ctx, eg)
```
vgeddes, line 97: This changing boilerplate code which is shared by all our Go relayers. Its also unnecessary, since `relay.Start` spawns its own go routines internally.

relayer/cmd/run/beefy/command.go
vgeddes: The convention with all our relayers is that `Relay.Start` should spawn its own go routines internally.

relayer/relays/beefy/on-demand-sync.go
vgeddes: fyi, The purpose of this check is to exit the process cleanly when I press 'Ctrl-C'. I've setup the relayers to handle these these signals using context cancellation.

So you got to make sure to test that flow.

relayer/relays/beefy/on-demand-sync.go
```diff
@@ -72,101 +83,91 @@ func (relay *OnDemandRelay) Start(ctx context.Context) error {
...
-		if err != nil {
-			if errors.Is(err, context.Canceled) {
-				return nil
+		select {
+		case <-ctx.Done():
+			return nil
+		case <-time.After(time.Second * 30):
+			log.Info("Starting check")
+			err = relay.syncV1(ctx)
+			if err != nil {
```
vgeddes, line 107: add the check for cancelled context here maybe?

relayer/relays/beefy/on-demand-sync.go
yrong: Yeah, I've tested the exit flow - it responds to `Ctrl-C` correctly.

relayer/relays/beefy/on-demand-sync.go
yrong: Not nessessary, as mentioned https://github.com/Snowfork/snowbridge/pull/1449#discussion_r2103607607

relayer/cmd/run/beefy/command.go
yrong: https://github.com/Snowfork/snowbridge/pull/1449/commits/b072474d551f6ccdce907c7e5cdc1cb69ff4b453

vgeddes:
Looks better now! Can you let it run on westend and test that all works well?

PR #1463, opened 2025-04-29
Send Token to Kusama

claravanstaden:
- Adds API layer `forKusama`
- Test scripts for transferring Weth and Dot from Polkadot to Kusama and back.


web/packages/api/src/assets_v2.ts
```diff
@@ -5,7 +5,7 @@ import { SnowbridgeEnvironment } from "./environment"
 import { Context } from "./index"
 import {
     buildParachainERC20ReceivedXcmOnDestination,
-    DOT_LOCATION,
+    DOT_LOCATION, dotLocationOnKusamaAssetHubLocation,
```
yrong, line 8: The formatting doesn’t seem correct here. I’d suggest running `pnpm format` to format the code before committing.

web/packages/operations/src/transfer_for_kusama.ts
```diff
@@ -0,0 +1,148 @@
...
+    const polkadot_keyring = new Keyring({ type: "sr25519" })
+
+    const SUBSTRATE_ACCOUNT = process.env["SUBSTRATE_KEY"]
+        ? polkadot_keyring.addFromUri(process.env["SUBSTRATE_KEY"])
+        : polkadot_keyring.addFromUri("//Ferdie")
+
+    const registry = await fetchRegistry(env, context)
+
+    let sourceAccountHex = "0x460411e07f93dc4bc2b3a6cb67dad89ca26e8a54054d13916f74c982595c2e0e";
+    let beneficiaryAccountHex = "0x460411e07f93dc4bc2b3a6cb67dad89ca26e8a54054d13916f74c982595c2e0e";
```
yrong, line 79: I'd suggest not hardcoding the account here. Instead, consider loading it from an env variable so others can test with their own accounts.

web/packages/api/src/forKusama.ts
```diff
@@ -0,0 +1,667 @@
...
+        DOT_LOCATION,
+        dotLocationOnKusamaAssetHubLocation(),
+        erc20Location(registry.ethChainId, "0x0000000000000000000000000000000000000000"), // actual token location doesn't matter here, just weighing the message
+        xcmBridgeFee,
+        xcmBridgeFee,
+        registry.assetHubParaId,
+        340282366920938463463374607431768211455n,
+        "0x0000000000000000000000000000000000000000000000000000000000000000",
+        "0x0000000000000000000000000000000000000000000000000000000000000000");
+
```
yrong, line 119: Is this `forwardedXcm` specific to the direction from AHP to AHK (i.e., local-reserve-based)? 

I assume it's not the same as the forwardXCM used for the reverse direction, which is destination-reserve-based.

claravanstaden: Yip, only DOT and ERC-20s are supported atm, mentioned here: https://github.com/Snowfork/snowbridge/pull/1463#issuecomment-2893421450

web/packages/api/src/assets_v2.ts
claravanstaden: Addressed in https://github.com/Snowfork/snowbridge/pull/1463/commits/98d2efd9cdcb350d14ce2f2c6e342d8d25620bff.

web/packages/operations/src/transfer_for_kusama.ts
claravanstaden: Added in https://github.com/Snowfork/snowbridge/pull/1463/commits/a0bca9365d718e6cfeae8bc7e2b110ed0ce5008b.

yrong:
> I had some trouble with KSM, it looks like DOT as the fee asset isn't supported yet. I have parked it for now, will maybe pick it up again later.

I'd assume KSM would be used as the fee asset in this case. There's a DOT-KSM pool on AHP for that. In my previous tests transferring KSM from AHP to Ethereum, I recall swapping KSM out from that pool.

yrong:
@claravanstaden Just a minor fix in https://github.com/Snowfork/snowbridge/pull/1479

web/packages/api/src/assets_v2.ts
```diff
@@ -105,11 +106,18 @@ export type RegistryOptions = {
     ethchains: (string | AbstractProvider)[]
     relaychain: string | ApiPromise
     bridgeHub: string | ApiPromise
+    kusama?: KusamaOptions
```
alistair-singh, line 109: Better name than KusamaOptions could be PKBridgeOptions or whatever the name of their bridge is because technically this code path should work with anybridge collocated with us on Asset Hub.

web/packages/api/src/assets_v2.ts
```diff
@@ -547,6 +617,20 @@ export async function fromContext(context: Context): Promise<RegistryOptions> {
         ethchains: context.ethChains().map((ethChainId) => context.ethChain(ethChainId)),
         parachains,
     }
+
+    if (context.config.kusama) {
+        const [kusamaAssetHub] = await Promise.all([context.kusamaAssetHub()])
```
alistair-singh, line 622:
```suggestion
        const kusamaAssetHub = await context.kusamaAssetHub()
```

web/packages/api/src/assets_v2.ts
```diff
@@ -611,6 +695,12 @@ export async function getLocationBalance(
             }
             return BigInt(accountData?.balance ?? 0n)
         }
+        case "statemine": {
```
alistair-singh, line 698: You should be able to merge this case with the one above which may support PNAs in the future as well.

web/packages/api/src/xcmBuilder.ts
```diff
@@ -2,13 +2,21 @@ import { Registry } from "@polkadot/types/types"
...
-export const HERE_LOCATION = { parents: 0, interior: "Here"  }
-export const DOT_LOCATION = { parents: 1, interior: "Here"  }
+export const HERE_LOCATION = { parents: 0, interior: "Here" }
+export const DOT_LOCATION = { parents: 1, interior: "Here" }
 
 const ethereumNetwork = (ethChainId: number) => ({
     GlobalConsensus: { Ethereum: { chain_id: ethChainId } },
 })
 
+export const polkadotNetwork = () => ({
```
alistair-singh, line 12: Just use constants instead of lambda.

web/packages/api/src/xcmBuilder.ts
```diff
@@ -2,13 +2,21 @@ import { Registry } from "@polkadot/types/types"
...
+    GlobalConsensus: { Polkadot: { network: null } },
+})
+
+export const kusamaNetwork = () => ({
```
alistair-singh, line 16: Just use constants instead of lambda.

web/packages/api/src/xcmBuilder.ts
```diff
@@ -23,6 +31,34 @@ export function parachainLocation(paraId: number) {
...
+}
+
+export function polkadotAssetHubLocation(parachainId: number) {
+    return {
+        parents: 2,
+        interior: { x2: [{ GlobalConsensus: { Polkadot: null } }, { parachain: parachainId }] },
+    }
+}
+
+export function dotLocationOnKusamaAssetHubLocation() {
```
alistair-singh, line 55: Just use constants instead of lambda.

web/packages/operations/src/transfer_for_kusama.ts
```diff
@@ -0,0 +1,151 @@
...
+        : polkadot_keyring.addFromUri("//Ferdie")
+
+    const registry = await fetchRegistry(env, context)
+
+    const SOURCE_ACCOUNT_PUBLIC = SOURCE_ACCOUNT.address
+    const DEST_ACCOUNT_PUBLIC = DEST_ACCOUNT.address
+
+    let sourceAssetHub
+    let destAssetHub
+    if (direction == Direction.ToPolkadot) {
```
alistair-singh, line 93: Nice, like this enum, boths sides are symmetrical this makes sense.
+1

alistair-singh (changes requested):
Happy with this. Good to merge after some changes.

web/packages/api/src/forKusama.ts
yrong: For transferring DOT and ERC-20s back from AHK to AHP, as mentioned above the forwardedXcm should be different ( destination-reserve-based) and I did't see a branch for that?

web/packages/api/src/forKusama.ts
```diff
@@ -0,0 +1,800 @@
...
+        let amount = xcmBridgeFee + bridgeHubDeliveryFee
+        totalFeeInDot = await quoteFeeSwap(
+            sourceAssetHub,
+            { parents: 1, interior: "Here" },
+            { parents: 2, interior: { x1: [{ GlobalConsensus: { Polkadot: null } }] } },
+            amount
+        )
+    } else {
+        totalFeeInDot = totalFee
+    }
```
yrong, line 158: For the P-K bridge, it seems to be a bit more complicated than our P-E bridge. It looks like they take the message fee [here](https://github.com/paritytech/polkadot-sdk/blob/e44b89fb7ca3385f314803c733ad97b26cd14e9f/bridges/modules/xcm-bridge-hub-router/src/lib.rs#L324-L325) into consideration. So it's `base_fee + message_fee + delivery_fee` in total.

web/packages/api/src/forKusama.ts
```diff
@@ -0,0 +1,800 @@
...
+    )
+
+    let bridgeHubDeliveryFee = await calculateDeliveryFee(
+        sourceAssetHub,
+        registry.bridgeHubParaId,
+        forwardedXcm
+    )
+
+    // In either DOT or KSM
+    let totalFee = xcmBridgeFee + bridgeHubDeliveryFee
```
yrong, line 143: For the P-K bridge, it seems to be a bit more complicated than our P-E bridge. It looks like they take the message-fee [here](https://github.com/paritytech/polkadot-sdk/blob/e44b89fb7ca3385f314803c733ad97b26cd14e9f/bridges/modules/xcm-bridge-hub-router/src/lib.rs#L324-L325) into consideration. So it's `base_fee + message_fee + delivery_fee` in total.

web/packages/api/src/forKusama.ts
```diff
@@ -0,0 +1,800 @@
...
+    // Convert KSM to DOT
+    if (direction == Direction.ToPolkadot) {
+        console.info("Converting KSM to DOT")
+        let amount = xcmBridgeFee + bridgeHubDeliveryFee
+        totalFeeInDot = await quoteFeeSwap(
+            sourceAssetHub,
+            { parents: 1, interior: "Here" },
+            { parents: 2, interior: { x1: [{ GlobalConsensus: { Polkadot: null } }] } },
+            amount
+        )
```
yrong, line 155: Well, for the Kusama->Polkadot direction, as discussed earliar, the fee asset is KSM - a runtime parameter declared [here](https://github.com/polkadot-fellows/runtimes/blob/9859c5f95c9031526c4dbe728e3f95b262327d63/system-parachains/asset-hubs/asset-hub-kusama/src/xcm_config.rs#L485), I'm not sure swapping it to DOT here make sense.

yrong (changes requested):
https://github.com/Snowfork/snowbridge/pull/1463#issuecomment-2898659181
https://github.com/Snowfork/snowbridge/pull/1463#discussion_r2101461555
https://github.com/Snowfork/snowbridge/pull/1463#discussion_r2101466364

web/packages/api/src/forKusama.ts
claravanstaden: **Polkadot to Kusama:** 
- https://github.com/Snowfork/snowbridge/pull/1463/files#diff-ed53112a03c400d74149228af80e7c512c73c7ef37a3d84562db55e9728ddc35R571-R578
- https://github.com/Snowfork/snowbridge/pull/1463/files#diff-ed53112a03c400d74149228af80e7c512c73c7ef37a3d84562db55e9728ddc35R418-R425
- https://github.com/Snowfork/snowbridge/blob/05ca21b78569544c2f966d083a2be5eb22a06158/web/packages/api/src/xcmBuilder.ts#L1133

**Kusama to Polkadot:**
- https://github.com/Snowfork/snowbridge/pull/1463/files#diff-ed53112a03c400d74149228af80e7c512c73c7ef37a3d84562db55e9728ddc35R626-R633
- https://github.com/Snowfork/snowbridge/pull/1463/files#diff-ed53112a03c400d74149228af80e7c512c73c7ef37a3d84562db55e9728ddc35R408-R415
- https://github.com/Snowfork/snowbridge/blob/05ca21b78569544c2f966d083a2be5eb22a06158/web/packages/api/src/xcmBuilder.ts#L1212

web/packages/api/src/forKusama.ts
claravanstaden: I think we need to decide what the fee asset for Kusama to Polkadot should be. I kept it DOT because it is more consistent with the rest of the bridge, but I don't have strong feelings about this. WDYT @alistair-singh @vgeddes?

web/packages/api/src/assets_v2.ts
claravanstaden: I kind of agree, I had the same thought but then decided against it because:

- The P<>K bridge is the only bridge that can actually be tested against.
- Then all the other code should probably also refer to PKBridge right? And it refers to Polkadot-Kusama anyway.

But I'm open to changing it if you think its way better. I just didn't think it made sense to made it generic here and concrete everywhere else.

web/packages/api/src/forKusama.ts
yrong: I'd assume the bridge fee in `T::FeeAsset` [here](https://github.com/paritytech/polkadot-sdk/blob/ff5c67c8a3d60ccb3f0af114a07db8f4bec485a0/bridges/modules/xcm-bridge-hub-router/src/lib.rs#L330) is always the native token of the source chain.  It’s taken from the end user and sent to the Treasury on the source chain, meaning the bridge transfer can’t proceed without it.

web/packages/api/src/assets_v2.ts
claravanstaden: Addressed in https://github.com/Snowfork/snowbridge/pull/1463/commits/f309c26ea81cff0207f542a9ba8cfecc82e561de.

web/packages/api/src/xcmBuilder.ts
claravanstaden: Addressed in https://github.com/Snowfork/snowbridge/pull/1463/commits/f309c26ea81cff0207f542a9ba8cfecc82e561de.

web/packages/api/src/xcmBuilder.ts
claravanstaden: Addressed in https://github.com/Snowfork/snowbridge/pull/1463/commits/f309c26ea81cff0207f542a9ba8cfecc82e561de.

web/packages/api/src/xcmBuilder.ts
claravanstaden: Addressed in https://github.com/Snowfork/snowbridge/pull/1463/commits/f309c26ea81cff0207f542a9ba8cfecc82e561de.

acatangiu:
> Hey @yrong! Thanks for the review.
> 
> > Just curious - do we support transferring KSM from AHK to AHP and back? If so, could we also add a test script for that?
> 
> I had some trouble with KSM, it looks like DOT as the fee asset isn't supported yet. I have parked it for now, will maybe pick it up again later.

What is this based on? on both sides anything can be used as fees through autoswap as long as there is a liquid pool for it.

claravanstaden:
@acatangiu I wrote an integration test for sending KSM with DOT as fee, it works (I made an unrelated mistake last week, which caused the test to fail).

However, I did check what @yrong mentioned:

> I'd assume the bridge fee in T::FeeAsset [here](https://github.com/paritytech/polkadot-sdk/blob/ff5c67c8a3d60ccb3f0af114a07db8f4bec485a0/bridges/modules/xcm-bridge-hub-router/src/lib.rs#L330) is always the native token of the source chain. It’s taken from the end user and sent to the Treasury on the source chain, meaning the bridge transfer can’t proceed without it.

I sent WETH from Kusama to Polkadot, and my KSM balance was lowered, even though I used DOT as fee. So I think KSM makes more sense to use as fee, right? Otherwise the user needs some DOT and some KSM, which is bad UX.

yrong:
> on both sides anything can be used as fees through autoswap as long as there is a liquid pool for it

Yeah, we know that we can use the swap call or the `ExchangeAsset` instruction to obtain any asset, provided a liquid pool exists. What we're discussing here is how to estimate the fee along the route - more specifically, the bridge fee for the AHK-AHP direction.

claravanstaden:
@yrong sure, https://assethub-kusama.subscan.io/extrinsic/9407589-2. There is a KSM amount minted yeah.

acatangiu:
> I sent WETH from Kusama to Polkadot, and my KSM balance was lowered, even though I used DOT as fee. So I think KSM makes more sense to use as fee, right? Otherwise the user needs some DOT and some KSM, which is bad UX.

yes, try to use native whenever possible, I expect requiring KSM on Kusama side to be better UX than requiring DOT there - either way, if the source is KAH or PAH, the UI could check pools and do some swap before to cover all fees using only WETH

claravanstaden:
To summarize the changes that I'll make:
- Change the fee asset from K->P to KSM
- Use the `Balances::Minted` event from the dry run (same for P->K direction) as the execution fee (so total fee would be Balances::Minted + delivery fee) @alistair-singh let me know if you agree with this? This will replace getting the `XcmBridgeHubRouterBaseFee` from storage (and calculating the message fee per byte, which I have not implemented)

claravanstaden:
@yrong I changed the fee from K->P to KSM, and added KSM as a transferable asset (with scripts): http://github.com/Snowfork/snowbridge/pull/1463/commits/cd0462175656a8494ca42314758cd13f4b332a31

Awaiting feedback on how to determine the fee: https://snowfork.slack.com/archives/G01BGMGLAC9/p1747993120325679 Once decided, I'll implement it.

yrong:
> Awaiting feedback on how to determine the fee

Either way works for me as long as the fee estimation is accurate. I don’t have a strong preference, so I’ll leave the decision to you or Alistair.

I may do some cleanup and refactoring to support Snowbridge V2, but that will be handled independently.

web/packages/api/src/forKusama.ts
```diff
@@ -137,23 +154,28 @@ export async function getDeliveryFee(
...
             registry.assetHubParaId,
             100000000000n,
             "0x0000000000000000000000000000000000000000000000000000000000000000",
             "0x0000000000000000000000000000000000000000000000000000000000000000"
         )
     }
 
-    // Message on Dest AH
-    //calculateDestinationFee
+    let bytes = forwardedXcm.toHex().length / 2;
```
yrong, line 166: Should we use `forwardedXcm.toU8a().length` instead?

claravanstaden: Yeah, that is better. https://github.com/Snowfork/snowbridge/pull/1463/commits/df04168327da242314e628ed4dec88e9bbd0bdb6

web/packages/api/src/xcmBuilder.ts
```diff
@@ -332,6 +368,27 @@ export function buildParachainERC20ReceivedXcmOnAssetHub(
     })
 }
 
+function buildAssetHubXcmFromParachainKusama(beneficiary: string, topic: string) {
+    return [
```
yrong, line 372: Would it be better to move these functions to a separate file specific to kusama, e.g., kusamaXcmBuilder.ts?

claravanstaden: Definitely, addressed in https://github.com/Snowfork/snowbridge/pull/1463/commits/2fed1d60dffa676c1263bb2d6de4df985e8f1576.

web/packages/api/src/assets_v2.ts
```diff
@@ -750,6 +856,7 @@ export function padFeeByPercentage(fee: bigint, padPercent: bigint) {
     if (padPercent < 0 || padPercent > 100) {
         throw Error(`padPercent ${padPercent} not in range of 0 to 100.`)
     }
+    console.log("PAD: ", fee * ((100n + padPercent) / 100n))
```
alistair-singh, line 859: Remove this logging?

alistair-singh (approved):
Just a dangling debug log command that needs to be removed.

web/packages/api/src/assets_v2.ts
claravanstaden: Removed in https://github.com/Snowfork/snowbridge/pull/1463/commits/90a7ddca2b9b88ec1a5c6b0001c5e774b478b3a1.

PR #1469, opened 2025-05-09
Audit Issue 6: Allow independent execution of each command

claravanstaden:
Allows execution of all inbound commands on Ethereum, even if the preceding command fails. The `InboundMessageDispatched ` event success field will be `true` if all commands succeeded, or `false` if all or some of the commands failed. For each failed command, a `CommandFailed` event will be emitted, capturing the message nonce and the command index.

Resolves: [SNO-1459](https://linear.app/snowfork/issue/SNO-1459)

### DIFF
--- a/contracts/src/Gateway.sol
+++ b/contracts/src/Gateway.sol
@@ -428,9 +428,11 @@ contract Gateway is IGatewayBase, IGatewayV1, IGatewayV2, IInitializable, IUpgra
             revert IGatewayBase.InvalidProof();
         }
 
-        // Dispatch the message payload
+        // Dispatch the message payload - now returns whether all commands succeeded
         bool success = v2_dispatch(message);
 
+        // Emit the event with a success value of:
+        // 1 = all commands succeeded, 0 = some commands failed
         emit IGatewayV2.InboundMessageDispatched(
             message.nonce, message.topic, success, rewardAddress
         );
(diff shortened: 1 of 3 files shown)


contracts/src/Gateway.sol
```diff
@@ -509,56 +511,67 @@ contract Gateway is IGatewayBase, IGatewayV1, IGatewayV2, IInitializable, IUpgra
...
     function v2_dispatch(InboundMessageV2 calldata message) internal returns (bool) {
+        bool allCommandsSucceeded = true;
+
         for (uint256 i = 0; i < message.commands.length; i++) {
             // check that there is enough gas available to forward to the command handler
             if (gasleft() * 63 / 64 < message.commands[i].gas + DISPATCH_OVERHEAD_GAS_V2) {
                 revert IGatewayV2.InsufficientGasLimit();
             }
+
+            bool commandSucceeded = true;
```
vgeddes, line 525: This change has introduced some complexity with managing these success flags. I suspect the code be easier to read and maintain if we factor out the the body of the `for` loop into a separate internal`_dispatchCommand(Command calldata command)` function that returns a boolean.

This is just an experiment. It may be worthwhile to try out.

Then the the contents of `_dispatchCommand` can just `return false`. One could say this a more functional style.

vgeddes (commented):
Looks great,

contracts/src/Gateway.sol
claravanstaden: I tried it - it was a good suggestion - looks simpler. :) Let me know what you think.

yrong (approved):
Nice!

contracts/src/Gateway.sol
```diff
@@ -428,9 +428,11 @@ contract Gateway is IGatewayBase, IGatewayV1, IGatewayV2, IInitializable, IUpgra
             revert IGatewayBase.InvalidProof();
         }
 
-        // Dispatch the message payload
+        // Dispatch the message payload. The boolean returned indicates whether all commands succeeded.
         bool success = v2_dispatch(message);
 
+        // Emit the event with a success value "true" if all commands successfully executed, otherwise "false"
+        // if all or some of the commands failed.
```
vgeddes, line 435: I think this documentation should be moved to the the comment for `InboundMessageDespatched` in `IGateway.sol`.

claravanstaden: Addressed in https://github.com/Snowfork/snowbridge/pull/1469/commits/37a2a4352403831e0dd149f66c7f85aac9c188ae.

contracts/src/Gateway.sol
```diff
@@ -509,58 +511,72 @@ contract Gateway is IGatewayBase, IGatewayV1, IGatewayV2, IInitializable, IUpgra
...
-        for (uint256 i = 0; i < message.commands.length; i++) {
-            // check that there is enough gas available to forward to the command handler
-            if (gasleft() * 63 / 64 < message.commands[i].gas + DISPATCH_OVERHEAD_GAS_V2) {
-                revert IGatewayV2.InsufficientGasLimit();
+    // Internal helper to dispatch a single command
+    function _dispatchCommand(
+        CommandV2 calldata command,
+        bytes32 origin,
+        uint64 nonce,
+        uint256 index
```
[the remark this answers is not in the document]
claravanstaden: Addressed in https://github.com/Snowfork/snowbridge/pull/1469/commits/37a2a4352403831e0dd149f66c7f85aac9c188ae.

contracts/src/Gateway.sol
```diff
@@ -509,58 +511,72 @@ contract Gateway is IGatewayBase, IGatewayV1, IGatewayV2, IInitializable, IUpgra
...
 
+    // Dispatch all the commands within the batch of commands in the message payload. Each command is processed
+    // independently, and failures emits event CommandFailed without stopping execution of subsequent commands.
+    function v2_dispatch(InboundMessageV2 calldata message) internal returns (bool) {
+        bool allCommandsSucceeded = true;
+
+        for (uint256 i = 0; i < message.commands.length; i++) {
+            bool commandSucceeded =
+                _dispatchCommand(message.commands[i], message.origin, message.nonce, i);
+            if (!commandSucceeded) {
```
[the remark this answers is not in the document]
claravanstaden: Addressed in https://github.com/Snowfork/snowbridge/pull/1469/commits/37a2a4352403831e0dd149f66c7f85aac9c188ae.

vgeddes (approved):

PR #1471, opened 2025-05-12
Skip proof validation when performing estimation

yrong:
More context and details in https://linear.app/snowfork/issue/SNO-1397#comment-5693a6c2

### DIFF
--- a/contracts/src/Gateway.sol
+++ b/contracts/src/Gateway.sol
@@ -416,16 +416,18 @@ contract Gateway is IGatewayBase, IGatewayV1, IGatewayV2, IInitializable, IUpgra
             revert IGatewayBase.InvalidNonce();
         }
 
-        bytes32 leafHash = keccak256(abi.encode(message));
-
         $.inboundNonce.set(message.nonce);
 
-        // Produce the commitment (message root) by applying the leaf proof to the message leaf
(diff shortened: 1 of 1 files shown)


contracts/src/Gateway.sol
```diff
@@ -416,16 +416,18 @@ contract Gateway is IGatewayBase, IGatewayV1, IGatewayV2, IInitializable, IUpgra
             revert IGatewayBase.InvalidNonce();
         }
 
-        bytes32 leafHash = keccak256(abi.encode(message));
-
         $.inboundNonce.set(message.nonce);
 
-        // Produce the commitment (message root) by applying the leaf proof to the message leaf
-        bytes32 commitment = MerkleProof.processProof(leafProof, leafHash);
+        if (msg.sender != address(this)) {
```
alistair-singh, line 421: This should be safe because of the `nonreentrant` modifier. There is no way to re-enter this code path address of the gateway proxy. But we may need a test maybe to covert that we cannot do this. 

For instance if we remove the `nonreentrant` modifier and maybe try use sending Ether to a contract address to try and verify a message without proof. Like try and exploit it without `nonreentrant` and then we can add back `nonreentrant` and we know we are safe.

yrong:
Closed, as this will be handled in https://github.com/Snowfork/snowbridge/pull/1478

PR #1477, opened 2025-05-16
refactor: use the built-in max/min to simplify the code

studystill:
In Go 1.21, the standard library includes built-in [max/min](https://pkg.go.dev/builtin@go1.21.0#max) function, which can greatly simplify the code.

### DIFF
--- a/relayer/chain/ethereum/receipt.go
+++ b/relayer/chain/ethereum/receipt.go
@@ -21,10 +21,7 @@ func GetAllReceipts(ctx context.Context, conn *Connection, block *etypes.Block) 
 
 	for i := 0; i < numTransactions; i += receiptFetchBatchSize {
 		eg, ctx := errgroup.WithContext(ctx)
-		upper := i + receiptFetchBatchSize
-		if upper >= numTransactions {
-			upper = numTransactions
-		}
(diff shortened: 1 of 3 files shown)


relayer/chain/ethereum/receipt.go
```diff
@@ -21,10 +21,7 @@ func GetAllReceipts(ctx context.Context, conn *Connection, block *etypes.Block)
 
 	for i := 0; i < numTransactions; i += receiptFetchBatchSize {
 		eg, ctx := errgroup.WithContext(ctx)
-		upper := i + receiptFetchBatchSize
-		if upper >= numTransactions {
-			upper = numTransactions
-		}
+		upper := min(i+receiptFetchBatchSize, numTransactions)
```
claravanstaden, line 24: Shouldn't this be 
```suggestion
		upper := max(i+receiptFetchBatchSize, numTransactions)
```

studystill:
Sorry for the late reply.

Actually, there is no mistake here. For example, if i + receiptFetchBatchSize is 30 and numTransactions is 20, the final upper needs to be 20, which is the minimum of the two values.

yrong (approved):
+1

PR #1482, opened 2025-05-30
Fix moonbeam asset registry

claravanstaden:
Moonbeam changed their assets to native EVM assets, which requires some changes to our asset registry.

https://x.com/MoonbeamNetwork/status/1928192602140258530

### DIFF
--- a/web/packages/api/src/assets_v2.ts
+++ b/web/packages/api/src/assets_v2.ts
@@ -1378,8 +1378,8 @@ function defaultPathFilter(envName: string): (_: Path) => boolean {
                 if (
                     path.asset === MYTHOS_TOKEN_ID &&
                     (
-                        (path.destination !== 3369 && path.type === "ethereum") ||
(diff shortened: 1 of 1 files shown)


alistair-singh (approved):

web/packages/api/src/assets_v2.ts
```diff
@@ -1398,8 +1378,8 @@ function defaultPathFilter(envName: string): (_: Path) => boolean {
                 if (
                     path.asset === MYTHOS_TOKEN_ID &&
                     (
-                        (path.destination !== 3369 && path.type === "ethereum") || 
-                        (path.source !== 3369 && path.type === "substrate")
+                        // TODO: Disable Mythos to Eth until mythos is ready to enable
+                        (path.destination !== 3369)
```
alistair-singh, line 1382: Unrelated. But we need to disable this in order to push.

PR #1485, opened 2025-06-03
Add acala

alistair-singh:
Resolves: SNO-1502

### DIFF
--- a/web/packages/api/src/parachains/assethub.ts
+++ b/web/packages/api/src/parachains/assethub.ts
@@ -26,7 +26,11 @@ export class AssetHubParachain extends ParachainBase {
         return this.getNativeBalance(account)
     }
 
-    async getAssets(ethChainId: number, pnas: PNAMap): Promise<AssetMap> {
+    getAssets(ethChainId: number, pnas: PNAMap): Promise<AssetMap> {
+        return this.getAssetsFiltered(ethChainId, pnas, bridgeablePNAsOnAH)
+    }
+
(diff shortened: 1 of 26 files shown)


web/packages/api/src/parachains/assethub.ts
```diff
@@ -0,0 +1,244 @@
+import { AssetMap, PNAMap } from "src/assets_v2"
```
yrong, line 1: Why not use the relative path `../assets_v2`?

yrong:
@alistair-singh  I made [PR#1486](https://github.com/Snowfork/snowbridge/pull/1486) to fix the formatting. 

I suggest running `pnpm format` before committing, as it makes later reviews and conflict resolution easier.

web/packages/api/src/parachains/assethub.ts
alistair-singh: will fix

alistair-singh:
> I suggest running `pnpm format` before committing, as it makes later reviews and conflict resolution easier.

Where does the format settings come from, I ran fmt on this branch but i still see diffs with your branch?

yrong:
> Where does the format settings come from

It's from https://github.com/Snowfork/snowbridge/blob/main/web/packages/api/.prettierrc

yrong (approved):
Cool!

yrong:
@alistair-singh, I made some additional refactoring in [PR#1490](https://github.com/Snowfork/snowbridge/pull/1490). Please take a look at your convenience — no pressure at all. I’m totally fine if you’d prefer to focus on the Acala work for now and review it later after this PR is merged.

PR #1492, opened 2025-06-09
Snowbridge V2: Outbound transfer API

yrong:
Resolves: https://linear.app/snowfork/issue/SNO-1521

### Changelog

- Revamp `getDeliveryFee` to include 
  - The execution fee on source chain estimated by dry-running the `xcmPaymentApi`
  - The execution fee on Ethereum based on the gas price and a static gas estimation

- Revamp `createTransfer` to include fees calculated in the previous step when building XCM in V5, and use a low-level `PolkadotXCM::execute` extrinsic


web/packages/api/src/toEthereum_v3.ts
```diff
@@ -0,0 +1,856 @@
...
+            getAssetHubConversationPalletSwap(
+                assetHub,
+                paraLoc,
+                DOT_LOCATION,
+                returnToSenderExecutionFeeDOT
+            ),
+        ])
+        totalFeeInNative = totalFeeInNativeRes
+        assetHubExecutionFeeNative = assetHubExecutionFeeNativeRes
+        returnToSenderExecutionFeeNative = returnToSenderExecutionFeeNativeRes
```
yrong, line 332: @alistair-singh 
Just curious - with the Relay chain migrating key user functionalities (Staking,Governance,Balances) to Asset Hub, AH will effectively become the de facto reserve location for DOT. I assume all parachains are now expected to accept DOT in that form. 

By default, can we assume parachains will accept Ether/DOT (with AH as the reserve) for fee payments? If so, maybe we can remove the corresponding branch in the code to simplify things.

This would also help promote native Ether usage within the ecosystem. Additionally, we could further improve the API to let end users choose their preferred fee asset, assuming the necessary liquidity pools are available on AH — though that would be more of an API-level enhancement rather than a global chain-level property.

yrong:
> Cool. Can you describe the algorithm you use to estimate gas on ethereum (and point me to the relevant code)?

Sure.   It's straightforward [here](https://github.com/Snowfork/snowbridge/blob/57bd962193f50c67a6a8aa47bf595420518fba1b/web/packages/api/src/toEthereum_v3.ts#L335-L340), calculated as `gasPrice * (tokenDeliverGas + baseDeliveryGas)`.

To keep things simple at first, both `tokenDeliverGas` and `baseDeliveryGas` are static configuration values. 

The first (tokenDeliverGas) configures the gas cost of a local transfer for a specific token. The second (baseDeliveryGas) represents the gas cost of `v2_submit`, excluding command execution - it's a static value that mainly covers proof verification.

web/packages/api/src/assets_v2.ts
```diff
@@ -15,6 +15,8 @@ export type ERC20Metadata = {
     symbol: string
     decimals: number
     foreignId?: string
+    // The gas cost of a local transfer
+    deliveryGas?: bigint
```
claravanstaden, line 19: How will we know the value this should be set to?

web/packages/api/src/parachains/assethub.ts
```diff
@@ -126,23 +131,36 @@ export class AssetHubParachain extends ParachainBase {
...
                             this.provider.query.foreignAssets.asset(assetType),
                             this.provider.query.foreignAssets.metadata(assetType),
                         ])
                     ).map((encoded) => encoded.toPrimitive() as any)
                     if (!assetInfo) {
-                        console.warn(
-                            `Asset '${JSON.stringify(
-                                locationOnAH
-                            )}' is not a registered foregin asset on ${this.specName}.`
+                        assetType = this.provider.registry.createType(
```
claravanstaden, line 145:
```suggestion
                        // Query assets using XCM V5, if XCM V4 did not return anything
                        assetType = this.provider.registry.createType(
```

web/packages/api/src/toEthereum_v3.ts
```diff
@@ -0,0 +1,856 @@
...
+    connections: { assetHub: ApiPromise; source: ApiPromise; ethereum: AbstractProvider },
+    parachain: number,
+    registry: AssetRegistry,
+    tokenAddress: string,
+    padPercentage?: bigint,
+    defaultFee?: bigint
+): Promise<DeliveryFee> {
+    const { assetHub, source, ethereum } = connections
+    // Fees stored in 0x5fbc5c7ba58845ad1f1a9a7c5bc12fad
+    const feePadPercentage = padPercentage ?? 70n
```
claravanstaden, line 136: 70% fee padding seems kinda of high, is there a reason for that?

web/packages/api/src/toEthereum_v3.ts
```diff
@@ -0,0 +1,856 @@
...
+    const feeStorageKey = xxhashAsHex(":BridgeHubEthereumBaseFeeV2:", 128, true)
+    const feeStorageItem = await assetHub.rpc.state.getStorage(feeStorageKey)
+    let leFee = new BN((feeStorageItem as Codec).toHex().replace("0x", ""), "hex", "le")
+
+    let snowbridgeDeliveryFeeDOT = 0n
+    if (leFee.eqn(0)) {
+        console.warn("Asset Hub onchain BridgeHubEthereumBaseFee not set. Using default fee.")
```
claravanstaden, line 143:
```suggestion
        console.warn("Asset Hub onchain BridgeHubEthereumBaseFeeV2 not set. Using default fee.")
```

web/packages/api/src/toEthereum_v3.ts
```diff
@@ -0,0 +1,856 @@
...
+        assetHubExecutionFeeNative = assetHubExecutionFeeNativeRes
+        returnToSenderExecutionFeeNative = returnToSenderExecutionFeeNativeRes
+    }
+
+    // Calculate execution cost on ethereum
+    let ethereumChain = registry.ethereumChains[registry.ethChainId.toString()]
+    let feeData = await ethereum.getFeeData()
+    let ethereumExecutionFee =
+        (feeData.gasPrice ?? 2_000_000_000n) *
+        ((tokenErcMetadata.deliveryGas ?? 100_000n) + (ethereumChain.baseDeliveryGas ?? 180_000n))
```
claravanstaden, line 340: Are these sensible defaults? Would these values ever not be set?

web/packages/api/src/toEthereum_v3.ts
```diff
@@ -0,0 +1,856 @@
...
+            registry.ethChainId,
+            tokenAddress,
+            sourceAssetMetadata
+        )
+        isNativeBalance =
+            sourceAssetMetadata.decimals === source.info.tokenDecimals &&
+            sourceAssetMetadata.symbol == source.info.tokenSymbols
+    }
+    let nativeBalanceCheckFailed = false
+    if (isNativeBalance && fee.totalFeeInNative) {
```
claravanstaden, line 405: Could you explain this check, please? Perhaps with a comment? It is not clear to me why this is an error condition?

web/packages/api/src/toEthereum_v3.ts
```diff
@@ -0,0 +1,856 @@
...
+        }
+    } else {
+        logs.push({
+            kind: ValidationKind.Warning,
+            reason: ValidationReason.DryRunApiNotAvailable,
+            message: "Source parachain can not dry run call. Cannot verify success.",
+        })
+        if (sourceParaId !== registry.assetHubParaId) {
+            let dryRunResultAssetHub: any
+            if (sourceAssetMetadata.location) {
```
claravanstaden, line 490: Some comments to explain the different cases here, and on line 509 and 531 would be great. Or make it readable, like `ifFeeTokenIsNotDot()`.

claravanstaden (commented):
Still busy viewing, but here are some comments so far. :)

web/packages/api/src/toEthereum_v3.ts
yrong: It's inherited from the V2 API without any change. TBH, I’m not entirely sure either. I'd also assume it's not a error condition, and the variable might be better renamed to something more neutral -  like `nativeBalanceCheckRequired`. Maybe @alistair-singh can provide more context here.

web/packages/api/src/toEthereum_v3.ts
yrong: Reverted to 33% in https://github.com/Snowfork/snowbridge/pull/1492/commits/b3cefe9a68031e60119b985748b224274b67a0d4

I vaguely recall it's more of a workaround when I tested on local setup, but it does not really make sense. I'll double- check on Westend.

web/packages/api/src/toEthereum_v3.ts
yrong: Basically, it's to distinguish PNA from ENA, since `assetMetadata` of PNA includes a location field, whereas ENA's doesn't.

web/packages/api/src/toEthereum_v3.ts
yrong: > Another note here is that we may only need to returnToSender in the event that the source parachain is a 20-byte address.

Not sure this is necessary. I checked that the returnToSender fee is quite minimal - about 0.02 DOT (less than 0.1$), so I'd prefer to keep things simple without introducing extra complexity.

web/packages/api/src/index.ts
```diff
@@ -253,3 +253,4 @@ export * as historyV2 from "./history_v2"
 export * as subsquid from "./subsquid"
 export * as forKusama from "./forKusama"
 export * as types from "./types"
+export * as toEthereumV3 from "./toEthereum_v3"
```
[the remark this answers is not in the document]
yrong: Renamed in https://github.com/Snowfork/snowbridge/pull/1492/commits/acb6f07c8958e4fde0b95f1af778c5ad68564a75 by adding a `SnowbridgeV2` suffix to avoid confusion.

PR #1498, opened 2025-06-13
Adds min balance to Kusama fee

claravanstaden:
When a user transfers from Polkadot AH to Kusama AH (and in the other direction), they might not have the source native token (DOT for P->K and KSM for K->P) on the destination chain. Any leftover fees deposited on the destination might be less than the minimum token balance, which then causes DepositAsset to fail.

This PR adds the min balance to the fee, so that there will at least min balance on the dest.

Issue: https://github.com/Snowfork/snowbridge-app/issues/116

### DIFF
--- a/web/packages/api/src/forKusama.ts
+++ b/web/packages/api/src/forKusama.ts
@@ -15,6 +15,7 @@ import {
     dotLocationOnKusamaAssetHub,
     ksmLocationOnPolkadotAssetHub,
     DOT_LOCATION,
+    matchesConsensusSystem,
 } from "./xcmBuilder"
 import {
     buildKusamaToPolkadotDestAssetHubXCM,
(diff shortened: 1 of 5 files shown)


web/packages/api/src/forKusama.ts
```diff
@@ -226,6 +230,11 @@ export async function getDeliveryFee(
     )
     // pad destination XCM fee
     destinationFee = destinationFee + (destinationFee * 33n) / 100n
+
+    // add minimum balance to the dest fee, to avoid not being able to deposit leftover fees
+    console.log("adding min balance", minBalanceFeeDest)
```
alistair-singh, line 235: remove this log.

alistair-singh (approved):

PR #1523, opened 2025-07-07
Upgrade to V2 contracts

yrong:
This is required for upgrading the contract from V1 to V2 - we need to move the [immutable parameters](https://github.com/Snowfork/snowbridge/blob/7b274b3b799cae540899bf23c8b876090723571c/contracts/src/Gateway.sol#L84-L86) in V1 to [storage](https://github.com/Snowfork/snowbridge/blob/78bba702e9aa73839ae0e1d120b23052efe05499/contracts/src/storage/AssetsStorage.sol#L22-L28) in V2.


contracts/src/upgrade/GatewayWithFeeInitializer.sol
```diff
@@ -0,0 +1,37 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2023 Snowfork <hello@snowfork.com>
+pragma solidity 0.8.28;
+
+import "../Gateway.sol";
+
+// New Gateway logic contract with an fee initializer
+contract GatewayWithFeeInitializer is Gateway {
```
vgeddes, line 8: I'm trying for us to have a naming style like `Gateway202508`, based on the month in which we intend the upgrade to to be deployed.

```suggestion
contract Gateway202508 is Gateway {
```

contracts/scripts/DeployGatewayWithFeeInitializer.sol
```diff
@@ -0,0 +1,54 @@
...
+pragma solidity 0.8.28;
+
+import {AgentExecutor} from "../src/AgentExecutor.sol";
+import {GatewayWithFeeInitializer} from "../src/upgrade/GatewayWithFeeInitializer.sol";
+import {ParaID} from "../src/Types.sol";
+import {Script} from "forge-std/Script.sol";
+import {stdJson} from "forge-std/StdJson.sol";
+import {console} from "forge-std/console.sol";
+
+contract DeployGatewayWithFeeInitializer is Script {
```
vgeddes, line 12: See my comment below about the naming style that uses the month of intended deployment. What do you think?

```suggestion
contract DeployGateway202508 is Script {
```

PR #1524, opened 2025-07-08
API Updates

alistair-singh:
* Remove TRAC token from Asset Hub.
* Adds support for JAMTON, but leaves it as disabled until subsquid support
* Fixed transfer bug in forKusama: Skip account checks if dry run passes.
* Simplify context setup for user examples.
  * Replace alchemy with public nodes for simpler set up and user examples.
  * Remove Ethereum Key from context config for simpler set up.
  * Added two simplified examples using helpers.
  * Removed graphql environment variable that users may need to set and grab it from the context.
  * Made graphql mandatory in config
  * Allow users to specify connections manually or pass in the context and let code our code figure it out.


web/packages/operations/src/examples/send_ether_from_eth_to_assethub.ts
yrong, line 1: It seems all token transfers follow the same pattern, so maybe we can just demonstrate the logic using a single entry point to avoid duplication.

https://github.com/Snowfork/snowbridge/blob/78bba702e9aa73839ae0e1d120b23052efe05499/web/packages/operations/package.json#L28-L37

claravanstaden: These are example scripts that we want to use in the documentation, as examples for developers who wants to use the API.

PR #1526, opened 2025-07-16
API Updates

yrong:
To coordinate with https://github.com/Snowfork/snowbridge/pull/1524 with some changes for V2 scripts.


claravanstaden (approved):
Nice!

PR #1527, opened 2025-07-17
Ethereum to Polkadot Gas Estimator

claravanstaden:
Adds a Rust project to estimate and dry run messages from Ethereum to Polkadot.

- [ ] CLI parameters for Ethereum command
- [ ] Send message v2
- [ ] Register token v2
- [ ] InboundQueue submit extrinsic fee
- [ ] AssetHub execution fee
- [ ] Delivery fee from BridgeHub to AssetHub
- [ ] Create Ether/WND pool on Westend
- [ ] Dry run on AssetHub
- [ ] Dry run on destination Parachain
- [ ] Execution fee on destination
- [ ] Switch network based on input


gas-estimator/src/estimator.rs
```diff
@@ -0,0 +1,679 @@
...
+        .map_err(|e| {
+            EstimatorError::InvalidCommand(format!("Failed to get latest block: {:?}", e))
+        })?
+        .call(payment_info_call)
+        .await
+        .map_err(|e| {
+            EstimatorError::InvalidCommand(format!("Failed to query delivery fees: {:?}", e))
+        })?;
+
+    Ok(info.partial_fee)
```
claravanstaden, line 556: The partial fee is 0, I am not sure why. Both using `.partial_fee_estimate()` and `.query_info()`

claravanstaden:
Send Ether to AH command: `/target/debug/snowbridge-gas-estimator v2-send-message --xcm 0x0508140d01020400010100460411e07f93dc4bc2b3a6cb67dad89ca26e8a54054d13916f74c982595c2e0e --claimer 0x000101004dda95fb19af4c3e9b7b4b3eb1f8801dcec183e0bd34af7bceb7c659e1fa9567 --assets "[{\"kind\": \"native\",\"token\": \"0xA0b86a33E6E73a7B4d8F0A0f17C9DfB5F10e8A12\",\"amount\": \"1000000000000000000\"}]" --execution-fee 10000000000000000 --relayer-fee 10000000000000000 --origin 0xa8467374a4288582CA894EF0127f48B35D0F35d0`

Send token to Penpal: `/target/debug/snowbridge-gas-estimator v2-send-message --xcm 0x050c31010100451f01010004020109079edaa802000b0030ef7dba020104010004020109079edaa802000b00204aa9d1010c140d01020c000101007a79e6194ddfb1854616120bd4b38476f4341a7b05a4a1776c32558e927095ae2c401fbf1756172bb6961fd4e26bfa4b2e39f2f653b17292ff3eda3d991cdd7011140d0101020109079edaa80200000101007a79e6194ddfb1854616120bd4b38476f4341a7b05a4a1776c32558e927095ae --claimer 0x000101004dda95fb19af4c3e9b7b4b3eb1f8801dcec183e0bd34af7bceb7c659e1fa9567 --assets "[{\"kind\": \"native\",\"token\": \"0xA0b86a33E6E73a7B4d8F0A0f17C9DfB5F10e8A12\",\"amount\": \"1000000000000000000\"}]" --execution-fee 10000000000000000 --relayer-fee 10000000000000000 --origin 0xa8467374a4288582CA894EF0127f48B35D0F35d0`

gas-estimator/src/estimator.rs
yrong: @claravanstaden Made a fix in https://github.com/Snowfork/snowbridge/pull/1529

gas-estimator/src/estimator.rs
```diff
@@ -0,0 +1,1222 @@
...
+    })
+}
+
+fn validate(value: u128, execution_fee: u128, relayer_fee: u128) -> Result<(), EstimatorError> {
+    if execution_fee.saturating_add(relayer_fee) > value {
+        return Err(EstimatorError::ValueTooLow);
+    }
+
+    Ok(())
+}
```
yrong, line 412: Is this validation nessessary or correct? I'd assume the value in the payload [here](https://github.com/Snowfork/snowbridge/blob/c8c2e9399360ebf7f9b9d4765cb77976b15e3334/contracts/src/v2/Calls.sol#L111) is the original Ether amount minus (execution_fee + relayer_fee), so it's likely to be less than the total fee required.

gas-estimator/src/estimator.rs
```diff
@@ -0,0 +1,1222 @@
...
+                clients,
+                dot_asset.clone(),
+                ether_asset.clone(),
+                fee_dot,
+                true,
+            )
+            .await?;
+
+            Ok((fee_dot, fee_ether))
+        }
```
yrong, line 863: I’d suggest not using another dry run to estimate the cost on the destination side. IMHO, maintaining different parachain runtimes adds complexity.

To keep things simple for now, could we just maintain a map with the fee for each parachain instead? It could potentially come from the @snowbridge/registry package which I assume won't change that often.

gas-estimator/src/estimator.rs
claravanstaden: Good catch! Removed :)

gas-estimator/src/estimator.rs
yrong: Yeah, the dry-running is definitely useful. But I assume we will be using dry-running in the Web-SDK anyway? 

In the case of transfers, I assume the XCM should follow a fixed format based on the transfer type (PNA or ENA), so the fee can be calculated via dry-run in advance (e.g. in @snowbridge/registry), then be reused here.

Or is there any way that this binary can be reused on web?

claravanstaden: @yrong I would also prefer not to duplicate the logic. My concern with using the SDK was that we shouldn't need to host an API for relayers to call. If we can package it as a binary i think this is an option.

PR #1536, opened 2025-07-28
Add local transfers

alistair-singh:
Resolves: SNO-1556

### DIFF
--- a/web/packages/api/src/parachains/parachainBase.ts
+++ b/web/packages/api/src/parachains/parachainBase.ts
@@ -137,14 +137,12 @@ export abstract class ParachainBase {
             }
         }
         if (!dotAsset) {
-            console.warn(
-                "Could not find DOT in result",
-                result,
-                "using 0 as delivery fee. Dry run will fail if this is incorrect."
-            )
-            return 0n
+            console.info("Could not find DOT in result", result)
+            throw Error(`Can not query XCM Weight.`)
         }
+
         const deliveryFee = BigInt(dotAsset.fun.fungible.toString())
+
         return deliveryFee
     }
 
--- a/web/packages/api/src/parachains/hydration.ts
+++ b/web/packages/api/src/parachains/hydration.ts
@@ -51,4 +51,33 @@ export class HydrationParachain extends ParachainBase {
         }
         return assets
     }
+
+    async calculateDeliveryFeeInDOT(destParachainId: number, xcm: any): Promise<bigint> {
+        const result = (
+            await this.provider.call.xcmPaymentApi.queryDeliveryFees(
+                { v4: { parents: 1, interior: { x1: [{ parachain: destParachainId }] } } },
+                xcm
+            )
+        ).toPrimitive() as any
+        if (!result.ok) {
+            throw Error(`Can not query XCM Weight.`)
+        }
+        let dotAsset = undefined
+        const assets = result.ok.v4 || result.ok.v5
+        for (const asset of assets) {
+            if (asset.id.parents === 1 && asset.id.interior.here === null) {
+                dotAsset = asset
+            }
+        }
+        if (!dotAsset) {
+            console.warn(
+                "Could not find DOT in result",
+                result,
+                "using 0 as delivery fee. Dry run will fail if this is incorrect."
+            )
+            return 0n
+        }
+        const deliveryFee = BigInt(dotAsset.fun.fungible.toString())
+        return deliveryFee
+    }
 }


web/packages/api/src/assets_v2.ts
```diff
@@ -95,7 +99,7 @@ export type Source = {
     type: SourceType
     id: string
     key: string
-    destinations: { [destination: string]: string[] }
+    destinations: { [destination: string]: { type: SourceType; assets: string[] } }
```
claravanstaden, line 102: Since this changes the expected data structure, we would mark this release as a major version bump, right?

web/packages/api/src/assets_v2.ts
```diff
@@ -417,17 +421,20 @@ export function getTransferLocations(
...
         const commonAssets = new Set(
-            sourceAssets.filter((sa) => destinationAssets.find((da) => da === sa))
+            ethAssets.filter((sa) => destinationAssets.find((da) => da === sa))
         )
         for (const asset of commonAssets) {
             const p1: Path = {
                 type: "ethereum",
                 id: "ethereum",
                 source: ethChain.chainId,
+                destinationType: "substrate",
```
claravanstaden, line 437: Enum (or constant) would be safer here.

web/packages/api/src/toPolkadot_v2.ts
```diff
@@ -606,6 +607,28 @@ export async function getMessageReceipt(
...
+export const depositWeth = (
+    sourceAddress: string,
+    tokenAddress: string,
+    amount: bigint
+): Promise<ContractTransaction> =>
+    WETH9__factory.connect(tokenAddress).getFunction("deposit").populateTransaction({
+        from: sourceAddress,
+        value: amount,
+    })
+
```
claravanstaden, line 631: Doesn't seem to be used

web/packages/base-types/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@snowbridge/base-types",
-  "version": "0.1.69",
+  "version": "0.1.70",
```
claravanstaden, line 3: Maybe a major bump? As noted above?

claravanstaden (commented):
Nice!

web/packages/api/src/assets_v2.ts
alistair-singh: I originally used enums because I came from like a rust/csharp/golangperspective. But in the typescript world enums are generally considered bad or atleast its controversial.

https://blog.theodo.com/2023/03/are-typescript-enums-really-harmful/

Also strings are the javascript/typescript way of doing this and this is built into the typescript typesystem. Look here how I use strings to distinguish between two cases where the type can either be a context or another object.

https://github.com/Snowfork/snowbridge/blob/main/web/packages/api/src/toEthereum_v2.ts#L101-L104
This is safe in typescript and is the right way to do it. I do not think it can be done with enums.

We should actually be using this technique on all our transfer types, then we wont have to do things like casting to/from any everywhere.  Plus I have also been burnt with enums in this codebase because when we do casting to/from any, and when we go through network boundaries that require json serialization, we have to cast and enum ints can get easily re-interpreted as the wrong enum. Where as strings are more resilient to that.

claravanstaden: Okay! My backend knowledge screams against this but I hear you. 😂  What about constants?

alistair-singh: So this is why i say typescript is built for this. It is a constant, actually its one of two possible constants.

```
export type SourceType = "substrate" | "ethereum"
```

Typescript will check and enforce it at compile time. Here is the example error message if i change the string to "substrate2".

```
src/assets_v2.ts:437:17 - error TS2820: Type '"substrate2"' is not assignable to type 'SourceType'. Did you mean '"substrate"'?

437                 destinationType: "substrate2",
                    ~~~~~~~~~~~~~~~

  src/assets_v2.ts:93:5
    93     destinationType: SourceType
           ~~~~~~~~~~~~~~~
    The expected type comes from property 'destinationType' which is declared here on type 'Path'
```

web/packages/base-types/package.json
alistair-singh: Addressed.

claravanstaden (approved):

web/packages/api/src/parachains/parachainBase.ts
```diff
@@ -137,12 +137,14 @@ export abstract class ParachainBase {
...
         }
         if (!dotAsset) {
-            console.info("Could not find DOT in result", result)
-            throw Error(`Can not query XCM Weight.`)
+            console.warn(
+                "Could not find DOT in result",
+                result,
+                "using 0 as delivery fee. Dry run will fail if this is incorrect."
+            )
+            return 0n
```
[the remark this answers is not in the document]
alistair-singh: Moved this specifically to hydration in [d11744e](https://github.com/Snowfork/snowbridge/pull/1536/commits/d11744e3f152181b5a0a6947d861be7156ac78e7)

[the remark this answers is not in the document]
alistair-singh: Yeah we can move this back to base when we support chains other than hydration.

PR #1539, opened 2025-08-04
Fix connect with heartbeat

yrong:


vgeddes (approved):

claravanstaden (commented):
Good catch!

yrong:
By the way, in this [commit](https://github.com/Snowfork/snowbridge/pull/1539/commits/e4cc66b1b0fb871b7d1a88254d5c7743e57ef752),  I also added a timeout context when making the submit contract call.

After patching that to production, I did observe the following logs in the parachain relayer,  indicating that the call timed out:

```
{"@timestamp":"2025-08-05T10:04:03.011719587Z","error":"write message loop: write message: write eth gateway: context deadline exceeded","host":"relays","level":"fatal","message":"Unhandled error","unit":"snowbridge-asset-hub-parachain-relay.service"}
```

I assume it might be because multiple relayers are competing to submit the same message, and this one just lost the race.

claravanstaden (approved):

vgeddes (commented):
Can we lift these various heartbeats/timeouts into config please?

```diff
diff --git a/relayer/config/config.go b/relayer/config/config.go
index bd5358a6..b7f67a13 100644
--- a/relayer/config/config.go
+++ b/relayer/config/config.go
@@ -3,18 +3,24 @@ package config
 import "errors"

 type PolkadotConfig struct {
-       Endpoint string `mapstructure:"endpoint"`
+       Endpoint      string `mapstructure:"endpoint"`
+       HeartbeatSecs uint64 `mapstructure:"heartbeat-secs"`
+       Timeout       uint64 `mapstructure:"timeout"`
 }

 type ParachainConfig struct {
-       Endpoint string `mapstructure:"endpoint"`
+       Endpoint      string `mapstructure:"endpoint"`
+       HeartbeatSecs uint64 `mapstructure:"heartbeat-secs"`
+       Timeout       uint64 `mapstructure:"timeout"`
 }

 type EthereumConfig struct {
-       Endpoint  string `mapstructure:"endpoint"`
-       GasFeeCap uint64 `mapstructure:"gas-fee-cap"`
-       GasTipCap uint64 `mapstructure:"gas-tip-cap"`
-       GasLimit  uint64 `mapstructure:"gas-limit"`
+       Endpoint      string `mapstructure:"endpoint"`
+       GasFeeCap     uint64 `mapstructure:"gas-fee-cap"`
+       GasTipCap     uint64 `mapstructure:"gas-tip-cap"`
+       GasLimit      uint64 `mapstructure:"gas-limit"`
+       HeartbeatSecs uint64 `mapstructure:"heartbeat-secs"`
+       Timeout       uint64 `mapstructure:"timeout"`
 }
```

relayer/relays/parachain/ethereum-writer.go
```diff
@@ -143,6 +144,10 @@ func (wr *EthereumWriter) WriteChannel(
 		LeafProofOrder: new(big.Int).SetUint64(proof.MMRProof.MerkleProofOrder),
 	}
 
+	// Add a timeout when calling submit
+	ctx, cancel := context.WithTimeout(ctx, time.Duration(time.Second*180))
```
alistair-singh, line 148: I do not think we should include this timeout. We have the heartbeat that will terminate the process if it detects any errors on the connection.

If we include this timeout then it needs to be an absolute worst case, like 10-30 minutes - or - we need to use latest nonce instead of pending nonce. 

The relayers are configured to restart on failure in production. If a tx times out here, it may still be in the mempool. The relayer will then restart and issue another pending tx that will queue behind the previous, this new tx will also timeout, causing another restart and another pending tx...

So this will result in us having to manually clear these txs, or wait for it to clear. If we use latest than the relayer will try to replace the existing pending tx with a new tx, which may fail if the gas is to low but will stop the case when we continually create new pending txs. If we set the timeout to 10-30 minutes it will also minimise the amount of pending txs created.

alistair-singh (changes requested):

relayer/relays/parachain/ethereum-writer.go
yrong: > it may still be in the mempool. The relayer will then restart and issue another pending tx that will queue behind the previous, this new tx will also timeout, causing another restart and another pending tx...

From the logs, this behavior does not appear to be an issue.  You can review the fatal error logs from the past 12 hours [here](https://eu-central-1.console.aws.amazon.com/cloudwatch/home?region=eu-central-1#logsV2:logs-insights$3FqueryDetail$3D~(end~0~start~-43200~timeType~'RELATIVE~tz~'LOCAL~unit~'seconds~editorString~'filter*20*40logStream*20*3d*20*27snowbridge-asset-hub-parachain-relay.service*27*0a*20*7c*20filter*20level*20*3d*20*22fatal*22~queryId~'23cf6803-dd70-4efc-be77-20eae08bc330~source~(~'prod*2fvector*2fjournald)~lang~'CWLI)) 

There was only one restart during that period, which was triggered by a timeout event. As mentioned earlier, this occurred because another relayer successfully submitted the transaction [here](https://etherscan.io/tx/0x495a53ab4e534313cb543089ad2f0fb2678e35bbf047f948f31122dde12e94b4) before the primary relayer could complete its submission.

After the 3-minute timeout, the relayer logged the following error and restarted gracefully:
```
{"@timestamp":"2025-08-06T17:55:40.314070254Z","error":"write message loop: write message: write eth gateway: context deadline exceeded","host":"relays","level":"fatal","message":"Unhandled error","unit":"snowbridge-asset-hub-parachain-relay.service"}
```

There were no signs of additional pending transactions or unexpected restarts afterward.

So for now, the 3-minute timeout seems sufficient. I’ll continue monitoring the behavior and consider increasing the timeout if we observe repeated occurrences.

claravanstaden:
@yrong should these changes go into the `main` branch as well?

PR #1542, opened 2025-08-06
Release new api package

yrong:
Just updated the API versions for https://github.com/Snowfork/snowbridge/pull/1531/ to be enacted on the Westend APP.

By the way, we might consider https://github.com/Snowfork/snowbridge/pull/1535 to help streamline this workflow.


alistair-singh:
I published the updated registry as part of https://github.com/Snowfork/snowbridge/pull/1536

Westend has been updated with the Frequency Token and chain.

PR #1545, opened 2025-08-11
Timeout for waitForTransaction

yrong:


relayer/relays/parachain/ethereum-writer.go
```diff
@@ -151,13 +150,6 @@ func (wr *EthereumWriter) WriteChannel(
 	}
 	options.Nonce = big.NewInt(0).SetUint64(nonce)
 
-	// Add a timeout when calling submit
-	if wr.config.Ethereum.CallTimeoutSecs > 0 {
```
vgeddes, line 155: Why are we removing this timeout now?

vgeddes (approved):

PR #1546, opened 2025-08-12
Improve V2 to Ethereum API 

yrong:
- Pull up the fee estimation logic and remove redundency.
- Pull up the validateTransfer logic and remove redundency.
- Add DryRun check on BH in validateTransfer.
- Update the registry with injected gas cost, customized for the LDO token.
- Add an optinal asset claimer (resolves https://linear.app/snowfork/issue/SNO-1561)

### DIFF
--- a/web/packages/api/src/toEthereumSnowbridgeV2.ts
+++ b/web/packages/api/src/toEthereumSnowbridgeV2.ts
@@ -324,6 +324,8 @@ export const estimateFeesFromAssetHub = async (
             )
             totalFeeInDot += ethereumExecutionFeeInNative
             totalFeeInNative = totalFeeInDot
+        } else {
+            throw new Error("Unsupported fee token location")
         }
     }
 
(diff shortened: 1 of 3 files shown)


web/packages/api/src/toEthereumSnowbridgeV2.ts
```diff
@@ -185,3 +195,295 @@ export const getSnowbridgeDeliveryFee = async (assetHub: ApiPromise, defaultFee?
     }
     return snowbridgeDeliveryFeeDOT
 }
+
+export type DeliveryXcm = {
+    localXcm: any
+    forwardedXcmToBH: any
+    forwardXcmToAH?: any
+    returnToSenderXcm?: any
```
yrong, line 203: @alistair-singh Since Snowbridge V2 has `AliasOrigin` enabled by default, and we introduced an optional `assetClaimer` in this PR, it seems unessessary to build the `returnToSenderXcm` to send assets back to source chain when they get trapped on AH.

I’d prefer to remove this flow for some cleanup, which could also save users some fees.

WDYT?

PR #1549, opened 2025-08-14
Fatal error on config error

claravanstaden:
The alarms currently do not pick up some relayer failures because the "fatal" keyword is missing from the logs.

### DIFF
--- a/relayer/cmd/run/execution/command.go
+++ b/relayer/cmd/run/execution/command.go
@@ -64,6 +64,7 @@ func run(_ *cobra.Command, _ []string) error {
 
 	err = config.Validate()
 	if err != nil {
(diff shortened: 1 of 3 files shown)


relayer/cmd/run/execution/command.go
```diff
@@ -64,6 +64,7 @@ func run(_ *cobra.Command, _ []string) error {
 
 	err = config.Validate()
 	if err != nil {
+		logrus.WithError(err).Fatal("config file validation failed")
```
alistair-singh, line 67: Doesnt fatal exit the program? Meaning that the `fmt.Errrof` will never return? I think we need to see where this error is bubbled up and put the Fatal there.

PR #1551, opened 2025-08-18
Allow disabling heartbeat

yrong:
### Context

We added an ethereum connection heartbeat in https://github.com/Snowfork/snowbridge/pull/1539, but it doesn't seems nessessary based on Alchemy’s guidelines: https://www.alchemy.com/support/what-s-the-right-way-for-the-client-to-send-a-ping-to-alchemy 

Additionally, the ping loop is already included in the go-ethereum client (see https://github.com/ethereum/go-ethereum/blob/5b2fc67eeef09b76a93dab3d93b15b725aaf1259/rpc/websocket.go#L348).

In this PR, we’ve made the heartbeat optional—disabled by default and only enabled when explicitly configured. I also deployed a parachain relayer without the heartbeat, and the relay appears to work fine.

@vgeddes - was this the reason you initially didn’t add a heartbeat for Ethereum connections?

### DIFF
--- a/relayer/chain/ethereum/connection.go
+++ b/relayer/chain/ethereum/connection.go
@@ -67,24 +67,26 @@ func (co *Connection) ConnectWithHeartBeat(ctx context.Context, eg *errgroup.Gro
 	co.client = client
 	co.chainID = chainID
 
-	ticker := time.NewTicker(heartBeat)
+	if heartBeat.Abs() > 0 {
+		ticker := time.NewTicker(heartBeat)
 
-	eg.Go(func() error {
-		defer ticker.Stop()
-		for {
-			select {
-			case <-ctx.Done():
-				return ctx.Err()
-			case <-ticker.C:
-				_, err := client.NetworkID(ctx)
-				if err != nil {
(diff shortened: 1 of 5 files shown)


relayer/chain/ethereum/connection.go
```diff
@@ -67,24 +67,26 @@ func (co *Connection) ConnectWithHeartBeat(ctx context.Context, eg *errgroup.Gro
...
-		for {
-			select {
-			case <-ctx.Done():
-				return ctx.Err()
-			case <-ticker.C:
-				_, err := client.NetworkID(ctx)
-				if err != nil {
-					log.WithField("endpoint", co.endpoint).Error("Connection heartbeat failed")
-					return err
+	if heartBeat.Abs() > 0 {
```
vgeddes, line 70: So if gsrpc has its own ping loop then this one is unnecessary and we can remove it?

yrong: I assume that the ping loop is specific to WebSocket connections, and there’s no guarantee all service providers will implement it. We can still configure a none-zero heartbeat for HTTP(S) connections or for providers other than Alchemy/Dwellir if necessary.

PR #1552, opened 2025-08-20
Adds NeuroWeb to Paseo

claravanstaden:


web/packages/api/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@snowbridge/api",
-  "version": "0.2.0",
+  "version": "0.2.1",
```
yrong, line 3: With https://github.com/Snowfork/snowbridge/pull/1535, we can now publish packages or trigger the release flow without manually updating the version number.

See https://github.com/Snowfork/snowbridge/actions/runs/17366136623 which auto-released and bumped `@snowbridge/api` to [0.2.2](https://github.com/Snowfork/snowbridge/releases/tag/%40snowbridge%2Fapi%400.2.2.).

So there is no need to manually edit the version number here. If we do a manual release, we should bump the version to 0.2.3.

web/packages/api/src/parachains/neuroweb.ts
```diff
@@ -0,0 +1,134 @@
+import { ParachainBase } from "./parachainBase"
+import { DOT_LOCATION, erc20Location } from "../xcmBuilder"
+import { PNAMap } from "../assets_v2"
+import { AssetMap } from "@snowbridge/base-types"
+import { ApiPromise } from "@polkadot/api"
+import { SubmittableExtrinsic } from "@polkadot/api/types"
+import { ISubmittableResult } from "@polkadot/types/types"
+
+export const NEUROWEB_TEST_CHAIN_ID = 11155111 // Sepolia
```
yrong, line 9: Maybe inject `ethChainId` (already in the registry) into the `ParachainBase` constructor, or move this const into a common class?

web/packages/api/src/parachains/neuroweb.ts
```diff
@@ -0,0 +1,134 @@
...
+
+    async calculateXcmFee(destinationXcm: any, asset: any): Promise<bigint> {
+        if (JSON.stringify(asset) == JSON.stringify(DOT_LOCATION)) {
+            console.warn(
+                `${this.specName} does not support calculating fee for asset '${JSON.stringify(
+                    asset
+                )}'. Using default.`
+            )
+            return 10_000_000_000n // TODO
+        }
```
yrong, line 75: It seems NeuroWeb does not support DOT? Just curious how to determine the fee here.

web/packages/operations/src/wrap_snow_trac.ts
```diff
@@ -0,0 +1,78 @@
+import { Keyring } from "@polkadot/keyring"
+import { Context, contextConfigFor } from "@snowbridge/api"
+import { cryptoWaitReady } from "@polkadot/util-crypto"
+import { assetRegistryFor } from "@snowbridge/registry"
+import { NeurowebParachain } from "@snowbridge/api/dist/parachains/neuroweb"
+
+const wrapSnowTRAC = async () => {
```
yrong, line 7: Is this wrap/unwrap extrinsic used to switch an asset between native and foreign mode?

It doesn't seem to be used in the API - or do we want to add it to the UI?

web/packages/api/src/parachains/neuroweb.ts
claravanstaden: Its just relevant to Neuroweb, and Mythos config is defined the same way: https://github.com/Snowfork/snowbridge/blob/main/web/packages/api/src/parachains/mythos.ts#L6-L9 Since it is unlikely to change and this is a small class already, it think its better to leave it here.

web/packages/operations/src/wrap_snow_trac.ts
claravanstaden: It will be added to the UI yeah.

alistair-singh (approved):

PR #1562, opened 2025-09-10
Delivery Acceleration

yrong:
### Context 

This allows users to pay additional fees to speed up transfers using the FiatShamir approach. The gas cost is benchmarked [here](https://github.com/Snowfork/snowbridge/pull/1462#issuecomment-2846445781).


yrong:
Since we don't run on-demand relayer and the parachain relayer is currently profitibility-driven,  delivery latency has increased recently. 

To reduce latency, this PR includes the cost of the two-phase consensus update submission in the Ethereum execution fee by default.

Given that the average gas price over the last six months has been around 0.5 Gwei, the cost is effectively very close to legacy V1 (1.5–2 DOT). If gas prices increase significantly, however, the fee will be much higher.

web/packages/base-types/src/index.ts
```diff
@@ -40,8 +40,14 @@ export type EthereumChain = ChainId & {
...
-  // The gas cost of v2_submit excludes command execution, mainly covers the verification.
-  baseDeliveryGas?: bigint;
+  // Gas cost of `v2_submit`, excluding command execution; primarily covers verification.
+  baseVerificationGas?: bigint;
+  // Gas cost of `v2_submit`, including dispatch overhead such as multicall and internal command dispatch.
+  baseDispatchGas?: bigint;
+  // Gas cost of the two-phase submit flow (`submitInitial` + `commitPrevRandao` + `submitFinal`)
+  twoPhaseSubmitGas?: bigint;
+  // Gas cost of `submitFiatShamir`
+  submitFiatShamirGas?: bigint;
```
yrong, line 50: Add static gas estimation for each execution stage involved in message delivery, including the consensus update.

web/packages/api/package.json
```diff
@@ -10,6 +10,14 @@
...
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
+  "exports": {
+    ".": {
+      "import": "./dist/index.js",
+      "require": "./dist/index.js",
+      "types": "./dist/index.d.ts"
+    },
+    "./*": "./dist/*"
+  },
```
alistair-singh, line 20: Why this addition?

alistair-singh (commented):
Looks good. Just one comment.

web/packages/registry/scripts/buildRegistry.ts
```diff
@@ -1013,7 +1029,10 @@ async function indexEthChain(
             name,
             assets,
             key: `ethereum_${networkChainId}`,
-            baseDeliveryGas: 120_000n,
+            baseVerificationGas: 120_000n,
+            baseDispatchGas: 80_000n,
+            twoPhaseSubmitGas: 1_000_000n,
+            submitFiatShamirGas: 2_000_000n,
```
yrong, line 1035: A conservative gas estimate for the full call would be:
- (submitFiatShamirGas + baseVerificationGas + baseDispatchGas) when accelerated mode is enabled, or
- (twoPhaseSubmitGas + baseVerificationGas + baseDispatchGas) when it is not enabled.

web/packages/api/package.json
yrong: preserve deep imports like [@snowbridge/api/dist/fees] and [@snowbridge/api/dist/xcmBuilder], already uses those subpath imports in a few places.

PR #1565, opened 2025-09-14
Submit Fiat-Shamir (Relay) 

yrong:


relayer/relays/beefy/ethereum-writer.go
```diff
@@ -140,6 +140,21 @@ func (wr *EthereumWriter) submit(ctx context.Context, task *Request) error {
...
+	// Ignore beefy block already synced
+	if uint64(task.SignedCommitment.Commitment.BlockNumber) <= state.LatestBeefyBlock {
+		log.WithFields(log.Fields{
+			"validatorSetID": state.CurrentValidatorSetID,
+			"beefyBlock":     state.LatestBeefyBlock,
+			"relayBlock":     task.SignedCommitment.Commitment.BlockNumber,
+		}).Info("Beefy block already synced, just ignore")
+		return nil
+	}
+
```
yrong: Check with the light client whether this update is outdated.

yrong:
> Would we ever relay fiat shamir from the relayers given that it is 3x more expensive? I do not think we need any relayer changes here. 

This PR is mainly intended to address a specific use case where the end user is willing to bear the full cost of both the consensus update and the message execution for time-sensitive scenarios (as mentioned by the Hydration team in the channel). In such cases, the message relay will submit the Fiat-Shamir consensus update together with the message update.

> Maybe a one shot command, that we could use in a playbook to unstuck the beefy client.

Yes, we add a one shot command for submit Fiat-Shamir in this PR.

> A more important relayer change is to make sure the beefy and on-demand-beefy relayer abandons a ticket if someone has successfully submitted a fiat shamir with a block greater than your tickets block.

Yes, we've added this check, as mentioned in https://github.com/Snowfork/snowbridge/pull/1565/files#r2563477732 before `commitRandao` and `submitFinal` in the previous two-phase flow.

PR #1570, opened 2025-09-18
Neuroweb mainnet config

claravanstaden:
- Adds Neuroweb as a parachain
- Disables TRAC transfers back to Ethereum temporarily


claravanstaden:
@alistair-singh confirmed changes by testing it with the Snowbridge UI.

PR #1574, opened 2025-09-22
Fix upgrade & add more tests

yrong:


contracts/test/ForkUpgrade202509.t.sol
```diff
@@ -68,30 +85,85 @@ contract ForkUpgradeTest is Test {
...
+       fixture.message.nonce = nonce + 1;
 
-        vm.prank(relayer);
-        IGateway(address(GATEWAY_PROXY)).submitV1(
-            fixture.message,
-            fixture.leafProof,
-            fixture.headerProof
+        // Mock call to Verification.verifyCommitment to bypass BEEFY verification.
+       address VERIFICATION_ADDR_V2 = address(Verification);
+       vm.mockCall(VERIFICATION_ADDR_V2, abi.encodeWithSelector(Verification.verifyCommitment.selector), abi.encode(true));
```
vgeddes, line 122: Verification should not be bypassed here. This is important. You've deployed the new BeefyClient, so it should be able to verify messages.

This allows to test end-to-end verification flow after contract upgrades.

To construct this test you need to:

1. Select a recent ethereum block in the past few days. Suppose you select block `K`.
2. Take the last `submitV1` payload before block K and dump into `/test/data/mainnet-gateway-submitv1.json`
2. Set the tenderly fork to block `K`
3. Upgrade the gateway in the test
4. Then submit the payload. No need to adjust nonces.

vgeddes (changes requested):

PR #1575, opened 2025-09-22
Upgrade Polkadot dependencies

claravanstaden:
Kusama -> Polkadot transfers no longer work - packages should be updated.


alistair-singh:
There was another supply chain attack on npm, Shai-Hulud attack. We may need to check our packages before committing and publishing.
https://www.truesec.com/hub/blog/500-npm-packages-compromised-in-ongoing-supply-chain-attack-shai-hulud

PR #1602, opened 2025-10-14
Improve monitor with sync status from any parachain

yrong:
To monitor the sync status of indexer service for high-TVL parachains such as NeuroWeb, Hydration, and Mythos.

Requires: https://github.com/Snowfork/snowbridge-subsquid/pull/46


claravanstaden (approved):
Great addition. 🏅

PR #1606, opened 2025-10-19
add vTAO to AssetHub

camfairchild:
Adds vTAO from Ethereum to AssetHub.

vTAO is a TAO LST bridged to Eth via LayerZero on Bittensor's frontier EVM

Etherscan: https://etherscan.io/token/0xe9f6d9898f9269b519e1435e6ebaff766c7f46bf
Project Website: https://www.tao.app/bridge

### DIFF
--- a/web/packages/registry/src/polkadot_mainnet.registry.json
+++ b/web/packages/registry/src/polkadot_mainnet.registry.json
@@ -1,5 +1,5 @@
 {
-  "timestamp": "2025-10-07T21:35:27.436Z",
+  "timestamp": "2025-10-19T20:14:50.232Z",
   "environment": "polkadot_mainnet",
   "ethChainId": 1,
   "gatewayAddress": "0x27ca963c279c93801941e1eb8799c23f407d68e7",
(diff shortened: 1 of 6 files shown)


web/packages/registry/src/polkadot_mainnet.registry.json
```diff
@@ -581,6 +588,14 @@
...
           "isSufficient": false
         },
+        "0xe9f6d9898f9269b519e1435e6ebaff766c7f46bf": {
+          "token": "0xe9f6d9898f9269b519e1435e6ebaff766c7f46bf",
+          "name": "vTAO",
+          "minimumBalance": "bigint:1",
+          "symbol": "vTAO",
+          "decimals": 18,
+          "isSufficient": false
+        },
```
claravanstaden: If you use script `NODE_ENV=polkadot_mainnet npx ts-node build.ts` to generate the registry, the metadata actually looks like this:
```suggestion
        "0xe9f6d9898f9269b519e1435e6ebaff766c7f46bf": {
          "token": "0xe9f6d9898f9269b519e1435e6ebaff766c7f46bf",
          "name": "",
          "minimumBalance": "bigint:1",
          "symbol": "",
          "decimals": 0,
          "isSufficient": false
        },
```

That is because the metadata fetched here is from AssetHub. Do you have plans to open a ref on OpenGov to set the metadata for the token?

claravanstaden (commented):
Thanks for adding :)

PR #1612, opened 2025-10-23
Fix alarm config for Undelivered Timeout

alistair-singh:
Only require one datapoint and alarm in 5 minutes. (instead of 3 datapoints in 45 min)
Dont average the results, use the maximum in a 5 minute window.

https://snowfork.eu.pagerduty.com/incidents/Q19I37LBILY9JL

### DIFF
--- a/web/packages/operations/src/alarm.ts
+++ b/web/packages/operations/src/alarm.ts
@@ -420,7 +420,8 @@ export const initializeAlarms = async () => {
             ],
             AlarmDescription: AlarmReason.IndexServiceStale.toString(),
             AlarmActions: [BRIDGE_STALE_SNS_TOPIC],
-            ...alarmCommandSharedInput,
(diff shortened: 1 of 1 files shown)


web/packages/operations/src/alarm.ts
```diff
@@ -314,7 +331,8 @@ export const initializeAlarms = async () => {
             ],
             AlarmDescription: LatencyDashboard,
             AlarmActions: [BRIDGE_STALE_SNS_TOPIC],
-            ...alarmCommandSharedInput,
+            ComparisonOperator: "GreaterThanThreshold",
+            ...absoluteValueBreachingAlarmConfig,
```
yrong: I assume `GreaterThanThreshold` applies to most scenarios, If that’s the case, it could be included in `absoluteValueBreachingAlarmConfig` as the default option.

PR #1614, opened 2025-10-26
Transact on AssetHub demo

claravanstaden:
Sends Ether to AH, swap ETH to DOT, DOT to USDC, USDC back to Ethereum.

If the InitiateTransfer `preserveOrigin = true`, it fails on [BH here](https://github.com/paritytech/polkadot-sdk/blob/master/bridges/snowbridge/primitives/outbound-queue/src/v2/converter/convert.rs#L245). The origin at that point is `AliasOrigin: { parents: '0', interior: { X1: [ { AccountKey20: { network: null, key: '0xc189de708158e75e5c88c0abfa5f9a26c71f54d1' } } ] }`, while in other normal AH->Ethereum transfers it is `AliasOrigin(Location { parents: 0, interior: Here })`. The different alias origin comes from the first XCM message from [Ethereum -> AH](https://github.com/paritytech/polkadot-sdk/blob/master/bridges/snowbridge/primitives/inbound-queue/src/v2/converter.rs#L397-L399) (I think). Adding a new agent would solve this, but I it needs a governance proposal which is not ideal its just a demo.

If I change the `preserveOrigin = false`, the router routes through the v1 router, since the message contains a `ClearOrigin` instead of an `AliasOrigin` instruction.

Tests demonstrating the router failures: https://github.com/claravanstaden/polkadot-sdk/pull/26/files#diff-39dff7c86ff299a1b2d15d478e4e3f309123c82ae8c5c4f476580c33011e44efR1107

I want to check if I'm not missing anything, and if this could work.


yrong:
> Tests demonstrating the router failures: https://github.com/claravanstaden/polkadot-sdk/pull/26/files#diff-39dff7c86ff299a1b2d15d478e4e3f309123c82ae8c5c4f476580c33011e44efR1255

```
      // Build the XCM to send to BridgeHub, which will then export to Ethereum
		let xcm = VersionedXcm::from(Xcm(vec![
			UnpaidExecution { weight_limit: Unlimited, check_origin: None },
			ExportMessage {
				network: Ethereum { chain_id: 11155111 },
				destination: Here,
        ....
	// Set origin to Ethereum beneficiary
	AliasOrigin(Location::new(
		0,
		[AccountKey20 {
			network: None,
			key: hex!("c189de708158e75e5c88c0abfa5f9a26c71f54d1"),
		}],
	)),
...
assert_ok!(<AssetHubWestend as AssetHubWestendPallet>::PolkadotXcm::send(
			RuntimeOrigin::root(),
			bx!(destination.into()),
			bx!(xcm),
		));
```

I don’t think that should be possible — allowing an end user to explicitly set `AliasOrigin` for someone and send XCM directly to BridgeHub would be dangerous. If that were allowed, an attacker could bypass burning or depositing reserve assets on the source chain and compromise the bridge.

PR #1617, opened 2025-10-30
Custom Ethereum Transact

yrong:


web/packages/api/src/transfers/toEthereum/pnaFromAH.ts
```diff
@@ -36,6 +38,7 @@ export class PNAFromAH implements TransferInterface {
             slippagePadPercentage?: bigint
             defaultFee?: bigint
             feeTokenLocation?: any
+            contractCall?: ContractCall
```
claravanstaden: Perhaps we should validate the inputs here, e.g. make sure the `target` is a valid Ethereum address, check the `calldata` is a valid hex, etc. Can we also check that the agent exists on Ethereum, as part of validation?

yrong: https://github.com/Snowfork/snowbridge/pull/1617/commits/013cf5a84b46dc50e37d4c4c6e9d15c9a63b93df

> Can we also check that the agent exists on Ethereum, as part of validation?

Definitely — but as you noted in [#1620](https://github.com/Snowfork/snowbridge/pull/1620#issue-3574415714), we might need to wait until the missing runtime API is included first. 

In the meantime, if a parachain team wants to test from a specific origin, we can manually create it for them as a temporary workaround.

PR #1623, opened 2025-11-04
Update import beacon state helper command

claravanstaden:
Needs to use Electra to unmarshal SSZs.

### DIFF
--- a/relayer/cmd/import_beacon_state.go
+++ b/relayer/cmd/import_beacon_state.go
@@ -92,13 +92,16 @@ func importBeaconState(cmd *cobra.Command, _ []string) error {
 		return fmt.Errorf("read finalized state data from file: %w", err)
 	}
 
-	afterElectraFork := (conf.Source.Beacon.Spec.ForkVersions.Electra + 1) * 32
+	headerToDetermineFork, err := syncer.Client.GetHeaderAtHead()
+	if err != nil {
+		return fmt.Errorf("fetch header: %w", err)
+	}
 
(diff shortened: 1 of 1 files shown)


relayer/cmd/import_beacon_state.go
```diff
@@ -92,13 +92,13 @@ func importBeaconState(cmd *cobra.Command, _ []string) error {
 		return fmt.Errorf("read finalized state data from file: %w", err)
 	}
 
-	afterDenebFork := (conf.Source.Beacon.Spec.ForkVersions.Deneb + 1) * 32
+	afterElectraFork := (conf.Source.Beacon.Spec.ForkVersions.Electra + 1) * 32
 
-	attestedState, err := syncer.UnmarshalBeaconState(afterDenebFork, attestedData)
+	attestedState, err := syncer.UnmarshalBeaconState(afterElectraFork, attestedData)
```
yrong: So for the testnet, should it be `afterFuluFork`? Could we just use the latest block on chain instead?

claravanstaden: Good point, added in https://github.com/Snowfork/snowbridge/pull/1623/commits/15939edbae94be6abcefba8db4e6daca14362d3b

claravanstaden:
@yrong it doesn't run into production at the moment, because as you suggested long ago, having the beacon store run as part of the beacon relayer has limited value when the beacon relayer is turned off (like this week). So I think I should probably enable it, but as a separate service like you suggested long ago.

PR #1626, opened 2025-11-07
Relayer: Cross-chain equivocation fishing 

yrong:
Based on [#1493](https://github.com/Snowfork/snowbridge/pull/1493), this includes the latest changes and resolves existing conflicts, plus some improvement including:

- Monitor submitFinal and report equivocation for all colluding validators.

- Adds a one-shot report command as a fallback in case the relay fails to report a previous equivocation commitment during downtime, as mentioned [here](https://github.com/Snowfork/snowbridge/pull/1493#issuecomment-3454478114).


alistair-singh (commented):
Does this lookup past commitments?

yrong:
> Does this lookup past commitments?

Yes — it’s based on Rob's PR (https://github.com/Snowfork/snowbridge/pull/1493), with some additional refactoring and improvements.

PR #1634, opened 2025-11-27
Add tip API and tip check in relayer profitability

claravanstaden:
Adds:

- SDK methods to add a tip to both Inbound and Outbound messages
- Add checking if there is a tip in the gas estimator
- Add the tip to the relayer profitability check in the relayer

### DIFF
--- a/web/packages/operations/src/add_tip.ts
+++ b/web/packages/operations/src/add_tip.ts
@@ -73,7 +73,9 @@ export const addTipToMessage = async () => {
 
     // Step 4: Sign and send if not a dry run
     if (process.env.DRY_RUN !== "true") {
-        const response = await addTip.signAndSend(assetHub, tipTx, userAccount)
+        const response = await addTip.signAndSend(assetHub, tipTx, userAccount, {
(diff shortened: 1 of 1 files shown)


web/packages/operations/src/add_tip.ts
```diff
@@ -0,0 +1,113 @@
...
+    const keyring = new Keyring({ type: "sr25519" })
+    const userAccount = keyring.addFromUri(
+        process.env.SUBSTRATE_KEY ?? "//Alice"
+    )
+    console.log("User account:", userAccount.address)
+
+    const assetHub = await context.assetHub()
+
+    const tipParams = {
+        direction: direction === "Inbound"
```
alistair-singh: If you just use type unions, you can avoid enum conversions.

claravanstaden:
@alistair-singh good point! Done in https://github.com/Snowfork/snowbridge/pull/1634/commits/ced993ac92e3ddb39748f22cb177d36353785c24 (and formatted)

alistair-singh (approved):

PR #1636, opened 2025-11-27
Fix for tokens that charge fees.

alistair-singh:
Resolves: SNO-1650

### DIFF
--- a/contracts/test/GatewayV1.t.sol
+++ b/contracts/test/GatewayV1.t.sol
@@ -1154,7 +1154,7 @@ contract GatewayV1Test is Test {
         MockGateway(address(gateway)).v1_handleUnlockNativeToken_public(encodedParams);
     }
 
-    function testRegisterForeignToken() public {
+    function testRegisterForeignTokenHappyPath() public {
         RegisterForeignTokenParams memory params = RegisterForeignTokenParams({
             foreignTokenID: dotTokenID, name: "DOT", symbol: "DOT", decimals: 10
         });


alistair-singh:
I noticed the event:
```
    event AgentFundsWithdrawn(bytes32 indexed agentID, address indexed recipient, uint256 amount);
```

Is unused through out our code. I am wondering if we should not emit it in the `Handler.sol` for v1 and v2. We follow this for other events.

contracts/src/Functions.sol
```diff
@@ -65,7 +66,20 @@ library Functions {
...
+        if (balanceAfter < balanceBefore) {
+            revert InvalidAmount();
+        }
+
+        uint256 delta = balanceAfter - balanceBefore;
+        if (delta == 0 || delta > type(uint128).max) {
+            revert InvalidAmount();
+        }
+
+        transferredAmount = uint128(delta);
```
yrong: Another option would be to simply revert if the delta does not match the expected amount. Notably, DeFi protocols like Uniswap do not support these edge cases either. https://docs.uniswap.org/concepts/protocol/integration-issues

contracts/src/Functions.sol
```diff
@@ -65,7 +66,20 @@ library Functions {
...
+
+        uint256 balanceAfter = IERC20(token).balanceOf(agent);
+        if (balanceAfter < balanceBefore) {
+            revert InvalidAmount();
+        }
+
+        uint256 delta = balanceAfter - balanceBefore;
+        if (delta == 0 || delta > type(uint128).max) {
+            revert InvalidAmount();
+        }
```
claravanstaden: Is there a possibility of race conditions here? Ie the balance is altered between the before and after balance check, by some other transaction?

contracts/test/GatewayV1.t.sol
```diff
@@ -1068,6 +1114,26 @@ contract GatewayV1Test is Test {
         MockGateway(address(gateway)).v1_handleUnlockNativeToken_public(encodedParams);
     }
 
+    function testUnlockNativeTokenWithFeeOnTransferEmitsActualAmount() public {
```
claravanstaden:
```suggestion
    function testUnlockNativeTokenWithFeeOnTransferReceivesReducedAmount() public {
```

Since it doesn't seem to emit anything?

claravanstaden (commented):
Nice! 🪙

contracts/src/Functions.sol
alistair-singh: Transactions are executed in order and single threaded so I do not think this can apply. There could be an issue of reentrancy within the token implementation or something like that that could do funky things but the before and after should pick that up.

contracts/src/Functions.sol
alistair-singh: I like this, just revert and not support these kinds of tokens at all. @vgeddes ?

The only bad thing with this route, since we have already accepted PAXG, is that if they enabled fees on transfer and it reverted if we detected that. Then all PAXG would essentially never be allowed to be unlocked.

alistair-singh: So I have analyzed a large portion of bridged transfers from mainet(10k, ~80%), and we have not run this case. So I think it would be safe to what you suggest and revert @yrong.

vgeddes: Agreed, just revert if delta != amount

alistair-singh: Resolved in [49bdba9](https://github.com/Snowfork/snowbridge/pull/1636/commits/49bdba92f353afeba5b5f5667dad6c25d3645bcf)

yrong (approved):
+1

contracts/src/Functions.sol
```diff
@@ -65,7 +66,13 @@ library Functions {
             revert InvalidAmount();
         }
 
+        uint256 balanceBefore = IERC20(token).balanceOf(agent);
         IERC20(token).safeTransferFrom(sender, agent, amount);
+
+        uint256 balanceAfter = IERC20(token).balanceOf(agent);
+        if (balanceAfter != balanceBefore + amount) {
+            revert TransferAmountMismatch();
```
vgeddes: simplification:
```suggestion
        if (IERC20(token).balanceOf(agent) != balanceBefore + amount) {
            // Tokens with Fee-On-Transfer behaviour are not supported
            revert InvalidToken();
```

claravanstaden (approved):

contracts/src/Functions.sol
```diff
@@ -87,8 +94,14 @@ library Functions {
...
         uint128 amount
     ) internal {
+        uint256 balanceBefore = IERC20(token).balanceOf(recipient);
         bytes memory call = abi.encodeCall(AgentExecutor.transferToken, (token, recipient, amount));
         invokeOnAgent(agent, executor, call);
+
+        uint256 balanceAfter = IERC20(token).balanceOf(recipient);
+        if (balanceAfter != balanceBefore + amount) {
+            revert TransferAmountMismatch();
+        }
```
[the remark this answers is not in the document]
alistair-singh: Resolved in [f3b4caa](https://github.com/Snowfork/snowbridge/pull/1636/commits/f3b4caa7d36be5f4f01c23979a55b433a6802b33)

contracts/src/Functions.sol
alistair-singh: Resolved in [f3b4caa](https://github.com/Snowfork/snowbridge/pull/1636/commits/f3b4caa7d36be5f4f01c23979a55b433a6802b33)

PR #1637, opened 2025-11-27
MessageIds generated with pending nonce

alistair-singh:
If a user submits two tx's which get included in the same block, the message id generation does not take into account pending transactions in the TX pool and in turn generates the same id for both messages.

### DIFF
--- a/web/packages/api/src/toEthereum_v2.ts
+++ b/web/packages/api/src/toEthereum_v2.ts
@@ -1145,7 +1145,8 @@ export async function buildMessageId(
     tokenAddress: string,
     beneficiaryAccount: string,
     amount: bigint,
-) {
+    timestamp?: number,
+): Promise<string> {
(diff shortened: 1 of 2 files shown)


yrong:
Cool! Should we also consider retrieving the pending nonce for the E->P direction? We may need to revamp https://github.com/Snowfork/snowbridge/blob/206278d9497e93ff5a4b3d2a0c4232ffc40d88b3/web/packages/api/src/transfers/toPolkadot/erc20ToAH.ts#L172 a bit — I assume `Ethers` also supports using a pending tag.

Other than that, should we also consider incorporating the timestamp when constructing the Message ID?

claravanstaden (approved):

yrong (approved):
+1

yrong:
@alistair-singh  I've added https://github.com/Snowfork/snowbridge/pull/1639 for the reverse direction.

PR #1644, opened 2025-12-04
Fix: incorrect initial setup breaking contract tests in the Nix environment

yrong:


web/packages/test/scripts/set-env.sh
```diff
@@ -95,7 +95,7 @@ export ETHERSCAN_API_KEY="${ETHERSCAN_API_KEY:-0x0}"
 # for production deployment ETH_RANDAO_DELAY should be configured in a more reasonable sense
 export RANDAO_COMMIT_DELAY="${ETH_RANDAO_DELAY:-4}"
 export RANDAO_COMMIT_EXP="${ETH_RANDAO_EXP:-32}"
-export MINIMUM_REQUIRED_SIGNATURES="${MINIMUM_REQUIRED_SIGNATURES:-16}"
+export MINIMUM_REQUIRED_SIGNATURES="${MINIMUM_REQUIRED_SIGNATURES:-17}"
```
yrong: This is the only required fix to get the contract tests passing, it is initialized [here](https://github.com/Snowfork/snowbridge/blob/5fba3c793662a44f2b3a72be8f15b6c51bc9d54c/flake.nix#L90) for the Nix setup.

alistair-singh (approved):
Nice find!

PR #1646, opened 2025-12-05
Update gas estimator bindings

claravanstaden:
Update the gas estimator to work with updated runtime definitions.


gas-estimator/src/estimator.rs
```diff
@@ -295,7 +283,23 @@ async fn calculate_delivery_fee_in_dot(
...
+        bridge_hub_runtime::apis()
+            .xcm_payment_api()
+            .query_delivery_fees(
+                versioned_destination,
+                bridge_hub_xcm,
+                VersionedAssetId::V5(BridgeHubAssetId(dot_asset)),
+            )
+    };
+
+    #[cfg(not(feature = "local"))]
```
yrong: Actually, this change also applies to the Westend runtimes as well.

PR #1648, opened 2025-12-08
Audit Issue 1: Add the missing quorum check 

yrong:
### Context

This addresses Issue 1 of the Oak Audit Report. It also covers Issue 2, as the two issues are closely related and together constitute the quorum bug.

### DIFF
--- a/contracts/src/BeefyClient.sol
+++ b/contracts/src/BeefyClient.sol
@@ -586,8 +586,8 @@ contract BeefyClient {
         numRequiredSignatures += Math.log2(validatorSetLen, Math.Rounding.Ceil);
         // Add signatures based on the signature usage count.
         numRequiredSignatures += 1 + (2 * Math.log2(signatureUsageCount, Math.Rounding.Ceil));
-        // Never require more signatures than a 2/3 majority
-        return Math.min(numRequiredSignatures, computeQuorum(validatorSetLen));
+        // Never require more signatures than a 1/3 + 1 majority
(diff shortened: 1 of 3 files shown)


yrong:
By the way, my understanding is that the **initial bitfield** represents the set of validators claiming to have signed the commitment. According to the protocol, this should always require signatures from more than two-thirds of the validators, regardless of how a subset is later sampled for the light client.

Based on this security consideration, I reverted computeQuorum in this [commit](https://github.com/Snowfork/snowbridge/pull/1648/commits/c65f2aee058a0a0de94450d593aacbe06e046ff2)

contracts/src/BeefyClient.sol
```diff
@@ -580,11 +594,11 @@ contract BeefyClient {
     }
 
     /**
-     * @dev We have 2/3rd +1 honesty assumption on polkadot validators. Hence it is sufficient (for both random sampling and Fiat Shamir) to check 1/3rd +1 validator signatures to ensure at least 1 honest validator signed the payload.
+     * @dev Calculates 2/3 majority required for quorum for a given number of validators.
      * @param numValidators The number of validators in the validator set.
      */
     function computeQuorum(uint256 numValidators) internal pure returns (uint256) {
-        return numValidators / 3 + 1;
+        return numValidators - (numValidators - 1) / 3;
```
alistair-singh: The previous code handled the case of 3 validators. Do we want to restore that?
```solidity
        if (numValidators > 3) {
            return numValidators - (numValidators - 1) / 3;
        }
        return numValidators;
```

PR #1659, opened 2025-12-18
Audit Issue 2: Seperate sample quorum

yrong:
### Context

This builds on top of https://github.com/Snowfork/snowbridge/pull/1648  and is intended to separately address Issue 2 of the Oak audit report, following a recent comment:

> to ensure a clear separation of concerns between “minimum claimed signers” in the interactive path and “checked signatures” in the Fiat–Shamir (non-interactive) path

Conceptually, separating these two quorum definitions makes sense to me. One quorum represents the majority quorum for the minimum claimed signers and should always be greater than two-thirds of the validator set. The other represents the sampled quorum, which should be at most one-third plus one of the validator set.


contracts/src/BeefyClient.sol
```diff
@@ -601,6 +601,14 @@ contract BeefyClient {
         return numValidators - (numValidators - 1) / 3;
     }
 
+    /**
+     * @dev We have 2/3rd +1 honesty assumption on polkadot validators. Hence it is sufficient (for both random sampling and Fiat Shamir) to check 1/3rd +1 validator signatures to ensure at least 1 honest validator signed the payload.
+     * @param numValidators The number of validators in the validator set.
+     */
+    function computeCheckQuorum(uint256 numValidators) internal pure returns (uint256) {
```
vgeddes: Strictly speaking, in terms of meaning, this function is not computing a quorum, but the maximum number of required signatures for both algorithms. So would the following be a better name?

```suggestion
    function computeMaxRequiredSignatures(uint256 numValidators) internal pure returns (uint256) {
```

yrong: https://github.com/Snowfork/snowbridge/pull/1659/commits/a17220c7c14f8c35893b53da5a9e6e249a55509f

PR #1702, opened 2026-02-06
Execute multiple contract calls atomically

yrong:
### Summary
Adds support to execute multiple contract calls atomically. Sub-calls are run in order and the entire command reverts on the first failure.

### Behavior
- **CallContracts**: Runs sub-calls sequentially. If any sub-call fails, the execution reverts, marking the command as failed.
- **Payload format**: `CallContractsParams` contains a list of sub-calls (`calls`).

Requires: https://github.com/paritytech/polkadot-sdk/pull/12383


contracts/test/GatewayV2.t.sol
```diff
@@ -934,6 +1162,19 @@ contract GatewayV2Test is Test {
         assertTrue(!ok, "callContract with missing agent should return false");
     }
 
+    function testCallContractsAgentDoesNotExistReturnsFalse() public {
+        CallContractParams[] memory params = new CallContractParams[](1);
+        params[0] =
+            CallContractParams({target: address(0xdead), data: "", value: uint256(0)});
+        bytes memory payload = abi.encode(params);
```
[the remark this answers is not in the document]
yrong: Fixed this by wrapping the parameters in the `CallContractsParams` struct.

contracts/src/AgentExecutor.sol
```diff
@@ -29,4 +30,33 @@ contract AgentExecutor {
...
+        for (uint256 i; i < len; ++i) {
+            bool success = Call.safeCall(params[i].target, params[i].data, params[i].value);
+            if (!success) {
+                revert();
+            }
+        }
+    }
+
+    // Sweep remaining assets when specified
+    function sweep(address recipient, address[] calldata tokens) external {
```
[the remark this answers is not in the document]
yrong: Thanks for the feedback, Alistair. 

While sticking closer to the XCM spec for claims/sweeps is a good architectural direction for protocol-level asset recovery, `CallContracts` is actually a prerequisite for a safer, revamped contract interaction pattern we are moving towards.

Specifically, the target pattern is the standard DeFi "Approve-then-Call" flow executed atomically by the agent in a single transaction:
1. **Approve**: The user agent calls the ERC20 token contract to approve the adapter (e.g. Across SpokePool/L1 Adaptor) to transfer assets on its behalf.
2. **Call**: The user agent calls the adapter to execute the action (which pulls the approved assets).

Executing these two steps atomically via `CallContracts` resolves a major security risk: we no longer need to pre-fund shared adapters. Prefunding adapters across transaction boundaries has been the subject of multiple security reports due to the risk of front-running/drainage. 

Having arbitrary atomic multi-calls gives us the flexibility to use safe patterns like this without relying on prefunding or custom protocol-level sweeping shims.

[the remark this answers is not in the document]
alistair-singh: This latest implementation is much simpler and I am for it.

contracts/src/AgentExecutor.sol
```diff
@@ -29,4 +30,15 @@ contract AgentExecutor {
...
         }
     }
+
+    // Call multiple contracts with Ether values; reverts on the first failure
+    function callContracts(CallContractParams[] calldata params) external {
+        uint256 len = params.length;
+        for (uint256 i; i < len; ++i) {
+            bool success = Call.safeCall(params[i].target, params[i].data, params[i].value);
+            if (!success) {
+                revert();
```
alistair-singh: The revert does not forward inner contracts revert data. May help debugging and we should do the same for the singular `callContract` version.

contracts/test/GatewayV2.t.sol
```diff
@@ -595,6 +625,142 @@ contract GatewayV2Test is Test {
...
+                    topic: topic,
+                    commands: makeCallContractsCommand(params)
+                }),
+                proof,
+                makeMockProof(),
+                relayerRewardAddress
+            );
+    }
+
+    function testAgentCallContractsRevertedOnSecondFailure() public {
```
alistair-singh: Instead of emitting an event with `sayHello` and testing for its non-existance can we test an actual balance update? Like a test which covers Transfer/Approve/Send like in the case with L2s.

contracts/src/AgentExecutor.sol
yrong: Implemented and bubbled up the revert data for both callContract and callContracts using assembly returndatacopy/revert. Added unit tests to GatewayV2.t.sol for verification.
https://github.com/Snowfork/snowbridge/pull/1702/changes/f905d402f87f49da394847a17982a651a9d56513

contracts/test/GatewayV2.t.sol
yrong: Added two integration tests to GatewayV2.t.sol to cover the Approve/Transfer/Send pattern using WETH and a mock adapter (MockAdapter):

1. testAgentCallContractsTransferApproveSendSuccess: Verifies the success path, asserting that the agent's WETH balance is debited and the recipient's WETH balance is credited.
2. testAgentCallContractsTransferApproveSendFailureRevertsAll: Verifies the failure path, asserting that when a subsequent contract call fails (e.g. attempting to send more than available), the entire state is reverted (meaning both agent and recipient WETH balances remain unchanged, and the allowance granted during the first step is reverted back to 0).

https://github.com/Snowfork/snowbridge/pull/1702/commits/3e7238def9cefb20315144c7d815a6baee3e5fc6

contracts/src/v2/Types.sol
```diff
@@ -36,6 +36,8 @@ library CommandKind {
     uint8 constant MintForeignToken = 4;
     // Call an arbitrary solidity contract
     uint8 constant CallContract = 5;
+    // Call multiple arbitrary solidity contracts
+    uint8 constant CallContracts = 6;
```
vgeddes: `CallContract` and `CallContracts` are too similar and easily confused. Could lead to issues.

How about calling it`MultiCall`?

yrong: Good catch — renamed the whole plural family to \`MultiCall\` 
- `CallContracts` command kind → `MultiCall`
- `CallContractsParams` → `MultiCallParams` 
- handler/executor `callContracts` → `multiCall`

The singular `CallContract` / `CallContractParams` types are unchanged and still represent per-call execution. The on-the-wire encoding remains unchanged (uint8 = 6). 

This was done in commit 089468b, and is also synced upstream in: https://github.com/paritytech/polkadot-sdk/pull/12383/changes/533668e89ce150a7cd7ee39c96fde0f22017e8df

alistair-singh (approved):

PR #1711, opened 2026-02-11
Dry run on Ethereum

yrong:
### Dry run on Ethereum

Validates Snowbridge V2 transfers end-to-end by simulating the resulting `Gateway.v2_dispatch` against a forked Ethereum node, surfacing reverts, CommandFailed events, and L1Adapter DepositCallInvoked / DepositCallFailed outcomes before the user signs anything.

For Polkadot → Ethereum (L1 and L2) transfers, after the source / AH / BridgeHub Polkadot dry-runs succeed, the SDK now:
1. Builds a synthetic `v2_dispatch(commands, agentID, nonce)` tx whose commands mirror what the relayer would submit to mainnet (UnlockNativeToken + optional CallContract).
2. Sends it to a forked Ethereum node (`FORKED_PROVIDER_URL`) using impersonation of the gateway proxy as `from`.
3. Parses the resulting receipt for the gateway's `CommandFailed` event and the L1 Adapter's `DepositCallInvoked` / `DepositCallFailed` events, surfacing failures via the existing `ValidationLog` mechanism.

### DIFF
--- a/web/packages/api/src/index.ts
+++ b/web/packages/api/src/index.ts
@@ -91,6 +91,7 @@ export class Context<T extends EthereumProviderTypes> {
     #gatewayV2?: T["Contract"] & IGatewayV2
     #beefyClient?: T["Contract"] & BeefyClient
     #l1SwapQuoter?: T["Contract"] & ISwapQuoter
+    #forkedProvider?: T["Connection"]
 
     // Substrate
     #polkadotParachains: Record<string, Promise<ApiPromise>>
@@ -335,6 +336,20 @@ export class Context<T extends EthereumProviderTypes> {
         return this.environment.indexerGraphQlUrl
     }
 
+    forkedProvider(): T["Connection"] {
+        if (this.#forkedProvider) {
+            return this.#forkedProvider
+        }
+        const apiKey =
+            process.env.FORKED_PROVIDER_API_KEY ||
+            process.env.NEXT_PUBLIC_FORKED_PROVIDER_API_KEY
+        this.#forkedProvider = this.ethereumProvider.createProvider(
+            this.environment.forkedProviderUrl || "https://fork-mainnet.snowbridge.network",
+            apiKey ? { headers: { "X-API-Key": apiKey } } : undefined,
+        )
+        return this.#forkedProvider
+    }
+
     async destroyContext(): Promise<void> {
         // clean up contract listeners
         if (this.#beefyClient) {
@@ -345,6 +360,9 @@ export class Context<T extends EthereumProviderTypes> {
         }
         if (this.#gatewayV2) {
             await this.ethereumProvider.destroyContract(this.gatewayV2())
+        }
+        if (this.#forkedProvider) {
+            this.ethereumProvider.destroyProvider(this.#forkedProvider)
         }
 
         // clean up ethereum
--- a/web/packages/api/src/index.ts
+++ b/web/packages/api/src/index.ts
@@ -84,6 +84,7 @@ export * as governance from "./governance"
 export class Context<T extends EthereumProviderTypes> {
     readonly environment: Environment
     readonly ethereumProvider: EthereumProvider<T>
+    readonly forkedProviderApiKey?: string
 
     // Ethereum
     #ethChains: Record<string, T["Connection"]>
@@ -101,9 +102,14 @@ export class Context<T extends EthereumProviderTypes> {
     static #rpcInitTimeoutMs = 40_000
     static #wsRequestTimeoutMs = 30_000
 
-    constructor(environment: Environment, ethereumProvider: EthereumProvider<T>) {
+    constructor(
+        environment: Environment,
+        ethereumProvider: EthereumProvider<T>,
+        forkedProviderApiKey?: string,
+    ) {
         this.environment = environment
         this.ethereumProvider = ethereumProvider
+        this.forkedProviderApiKey = forkedProviderApiKey
         this.#polkadotParachains = {}
         this.#kusamaParachains = {}
         this.#ethChains = {}
@@ -341,6 +347,8 @@ export class Context<T extends EthereumProviderTypes> {
             return this.#forkedProvider
         }
         const apiKey =
+            this.forkedProviderApiKey ||
+            this.environment.forkedProviderApiKey ||
             process.env.FORKED_PROVIDER_API_KEY ||
             process.env.NEXT_PUBLIC_FORKED_PROVIDER_API_KEY
         this.#forkedProvider = this.ethereumProvider.createProvider(
@@ -406,6 +414,7 @@ export class Context<T extends EthereumProviderTypes> {
 export type ApiOptions<P extends EthereumProvider<any>> = {
     info: BridgeInfo
     ethereumProvider: P
+    forkedProviderApiKey?: string
 }
 
 export type TransferImplementation<T extends EthereumProviderTypes = EthereumProviderTypes> =
@@ -477,6 +486,7 @@ export class SnowbridgeApi<P extends EthereumProvider<any>> {
         this.context = new Context<ProviderTypesFor<P>>(
             options.info.environment,
             options.ethereumProvider as unknown as EthereumProvider<ProviderTypesFor<P>>,
+            options.forkedProviderApiKey,
         )
     }
     createAgent(): AgentCreationInterface<ProviderTypesFor<P>["ContractTransaction"]> {
--- a/web/packages/api/src/toEthereumSnowbridgeV2.ts
+++ b/web/packages/api/src/toEthereumSnowbridgeV2.ts
@@ -9,6 +9,7 @@ import {
     EthereumProviderTypes,
     Parachain,
     TransferRoute,
+    V2CommandStruct,
 } from "@snowbridge/base-types"
 import { CallDryRunEffects, XcmDryRunApiError, XcmDryRunEffects } from "@polkadot/types/interfaces"
 import { Result } from "@polkadot/types"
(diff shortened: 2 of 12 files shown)


alistair-singh:
Not sure we can use tenderly in our API, Is it free an available for public use?

alistair-singh (approved):
+1

yrong:
> > > Not sure we can use tenderly in our API, Is it free an available for public use?
> > 
> > 
> > Just create a [virtual testnet](https://docs.tenderly.co/virtual-testnets) and expose a public endpoint, which can be incrementally synced with the parent (actual) network. I assume the fee is already covered by our team subscription.
> 
> I think we can just use `eth_Call` or `eth_estimateGas`. Estimate gas will even allow us to set gas appropriately.

It depends on performing an actual dry run to obtain the **_receipt logs_** with failure details (indicating which command failed).

claravanstaden:
@yrong do you want to merge this still?

yrong:
> @yrong do you want to merge this still?

Yeah, I’ve resolved the conflicts and still want to merge it. I think it should be valuable for L2 transfers.

That said, I’m still planning to set up a local Anvil node and expose it for dry-running, since we canceled the Tenderly subscription to reduce costs, mentioned in https://github.com/Snowfork/snowbridge/pull/1711#issuecomment-4169409406

web/packages/api/src/index.ts
```diff
@@ -33,6 +33,20 @@ import type { TransferInterface as ToEthereumEvmTransferInterface } from "./tran
 import { toEthereumTransferById, toPolkadotTransferById } from "./history_v2"
 import type { ToEthereumTransferResult, ToPolkadotTransferResult } from "./history"
 
+type IGatewayProxy = {
+    getFunction(name: "v2_dispatch"): {
```
alistair-singh: This should exist on `gatewayV2: IGatewayV2`, if not you should add it there instead of creating a new interface.

web/packages/api/src/index.ts
```diff
@@ -331,6 +346,18 @@ export class Context<T extends EthereumProviderTypes> {
         return this.#beefyClient!
     }
 
+    gatewayProxy(): T["Contract"] & IGatewayProxy {
```
alistair-singh: Again get this from gatewayV2.

web/packages/api/src/toEthereumSnowbridgeV2.ts
```diff
@@ -1218,7 +1301,184 @@ export const validateTransferFromAssetHub = async <T extends EthereumProviderTyp
...
+                )
+                const txGasLimit = computeDryRunDispatchGasLimit(dryRunCommandGasBudgets(transfer))
+                try {
+                    const forkedProviderApiKey =
+                        process.env.FORKED_PROVIDER_API_KEY ||
+                        process.env.NEXT_PUBLIC_FORKED_PROVIDER_API_KEY
+                    const forkedProvider = context.ethereumProvider.createProvider(
+                        process.env.FORKED_PROVIDER_URL ||
+                            process.env.NEXT_PUBLIC_FORKED_PROVIDER_URL ||
+                            "https://fork-mainnet.snowbridge.network",
```
alistair-singh: This needs to be passed in through the registry or config like the subsquid url.

web/packages/api/src/toEthereumSnowbridgeV2.ts
```diff
@@ -1218,7 +1301,184 @@ export const validateTransferFromAssetHub = async <T extends EthereumProviderTyp
...
+                        reason: ValidationReason.DryRunFailed,
+                        message: ethereumDryRunError,
+                    })
+                }
+            } catch (e) {
+                ethereumDryRunError = "Could not estimate gas on Ethereum." + (e as Error).message
+                logs.push({
+                    kind: ValidationKind.Error,
+                    reason: ValidationReason.FeeEstimationError,
+                    message: ethereumDryRunError,
```
alistair-singh: These error messages are displayed to the user. I rather you `console.error` the full error message and return a user friendly message here.

web/packages/api/src/toEthereumSnowbridgeV2.ts
```diff
@@ -1218,7 +1301,184 @@ export const validateTransferFromAssetHub = async <T extends EthereumProviderTyp
...
+                estimatedDryRunGas = await context.ethereumProvider.estimateGas(
+                    ethereum,
+                    ethereumTx,
+                )
+                const txGasLimit = computeDryRunDispatchGasLimit(dryRunCommandGasBudgets(transfer))
+                try {
+                    const forkedProviderApiKey =
+                        process.env.FORKED_PROVIDER_API_KEY ||
+                        process.env.NEXT_PUBLIC_FORKED_PROVIDER_API_KEY
+                    const forkedProvider = context.ethereumProvider.createProvider(
```
alistair-singh: Instead of creating a provider for each transfer can we cache this in the context with all the other connections?

web/packages/api/src/toEthereumSnowbridgeV2.ts
```diff
@@ -1090,15 +1212,244 @@ export const estimateFeesFromParachains = async <T extends EthereumProviderTypes
...
+            sourceParaId,
+            sourceAccountHex,
+            transfer,
+        )
+        estimatedDryRunGas = await context.ethereumProvider.estimateGas(ethereum, ethereumTx)
+        const txGasLimit = computeDryRunDispatchGasLimit(dryRunCommandGasBudgets(transfer))
+        try {
+            const forkedProviderApiKey =
+                process.env.FORKED_PROVIDER_API_KEY ||
+                process.env.NEXT_PUBLIC_FORKED_PROVIDER_API_KEY
```
alistair-singh: This needs to be provided via the environment config, not env vars. Just like how we get the rpc node url and key.

web/packages/api/src/toEthereumSnowbridgeV2.ts
```diff
@@ -1090,15 +1212,244 @@ export const estimateFeesFromParachains = async <T extends EthereumProviderTypes
...
+            const forkedProvider = context.ethereumProvider.createProvider(
```
alistair-singh: This anvil provider should be cached and kept on the context.

web/packages/api/src/index.ts
yrong: https://github.com/Snowfork/snowbridge/pull/1711/commits/175bc8f40638ef6d1a703d93b1a8465d9ca2209a

web/packages/api/src/toEthereumSnowbridgeV2.ts
yrong: https://github.com/Snowfork/snowbridge/pull/1711/changes/69a18c3ff54732dfcb4f5acf0d25bf15a63447c4

I assume the API key is a sensitive credential and shouldn't be stored in the registry. Unlike Subsquid, which is effectively read-only, the dry-run node exposes state-mutating methods such as set_storage and set_block_time, so we should protect access to it with an API key.

web/packages/api/src/toEthereumSnowbridgeV2.ts
yrong: https://github.com/Snowfork/snowbridge/pull/1711/changes/a6b486b77cc02d68f7f7b8df390a99897feb5646

web/packages/api/src/toEthereumSnowbridgeV2.ts
yrong: https://github.com/Snowfork/snowbridge/pull/1711/commits/48cc1e9e0a71779835175fcf4082c90930e133df

web/packages/api/src/toEthereumSnowbridgeV2.ts
yrong: https://github.com/Snowfork/snowbridge/pull/1711#discussion_r3338669130

web/packages/api/src/toEthereumSnowbridgeV2.ts
yrong: This seems to be a duplicate of the comment at https://github.com/Snowfork/snowbridge/pull/1711#discussion_r3310236405, and has already been addressed.

yrong:
During the review, AI identified another issue: the L2 bridge fee shown in the fee breakdown should not be denominated in ETH, but in the transfer asset itself.

https://github.com/Snowfork/snowbridge/pull/1711/commits/b2867b065ec6933f7be56e68fb550e6587860cbf

@alistair-singh Could you take a look and let me know if that makes sense, or whether it would have any impact on how the UI displays fees?

web/packages/api/src/index.ts
```diff
@@ -335,6 +336,20 @@ export class Context<T extends EthereumProviderTypes> {
         return this.environment.indexerGraphQlUrl
     }
 
+    forkedProvider(): T["Connection"] {
+        if (this.#forkedProvider) {
+            return this.#forkedProvider
+        }
+        const apiKey =
+            process.env.FORKED_PROVIDER_API_KEY ||
+            process.env.NEXT_PUBLIC_FORKED_PROVIDER_API_KEY
```
alistair-singh, line 353: Can we add this key to the environment as well?

alistair-singh (commented):
Happy to merge, would just like the key to either be added to environment or passed in another way. We may not have `process.env` if we switch to react app running in ipfs.

In general i dont like setting SDK params via environment vars. I would prefer the calling app passing them in via params. Maybe add it to `ApiOptions` and pass it to the context.

web/packages/api/src/index.ts
yrong: https://github.com/Snowfork/snowbridge/pull/1711/commits/0e6ca8d6cb07a297c261db344e1715a8b2ea112a

PR #1714, opened 2026-02-15
Improve P2E transfer

yrong:
### Context

For transfers from a non-system parachain to Ethereum, since we perform a full dry-run check at each node/hop, the previous fallback XCM — which sent assets back to the source chain from AH when execution failed on AH — is no longer necessary.

This PR removes that fallback, slightly reduces fees, and has been tested and verified with a V2 transfer from Bifrost, which worked as expected.

https://app.snowbridge.network/activity#0x7a788c60bd72580111a5b4cbab1bed244f8f6e134d78a97310afadb7eae39d01


claravanstaden:
Does it hurt to keep "return to sender" in? I think its a nice fallback, but @alistair-singh would have the final say on this :)

PR #1730, opened 2026-03-02
Restrict adapter calls to registered agents only

yrong:


contracts/src/l2-integration/SnowbridgeL1Adaptor.sol
```diff
@@ -19,15 +21,27 @@ contract SnowbridgeL1Adaptor {
...
+    constructor(address _spokePool, address _l1weth9, address _gateway) {
         SPOKE_POOL = ISpokePool(_spokePool);
         L1_WETH9 = WETH9(payable(_l1weth9));
+        GATEWAY = _gateway;
+    }
+
+    modifier onlyRegisteredGatewayAgent() {
+        require(
+            Agent(payable(msg.sender)).GATEWAY() == GATEWAY,
+            "Caller is not a registered Gateway Agent"
```
claravanstaden:
```suggestion
            IGatewayV2(GATEWAY).isAgent(msg.sender),
           "Caller is not a registered Gateway Agent"
```

Otherwise someone can deploy a contract that returns the expected GATEWAY address from a GATEWAY() function without actually being a real Agent.

claravanstaden:
Do we still want to merge this, @yrong?

PR #1772, opened 2026-04-21
Comments on L2 integration

claravanstaden:
To hopefully reduce invalid bug bounty reports.

### DIFF
--- a/contracts/src/l2-integration/SnowbridgeL1Adaptor.sol
+++ b/contracts/src/l2-integration/SnowbridgeL1Adaptor.sol
@@ -22,6 +22,14 @@ import {DepositParams} from "./Types.sol";
 ///      zero balance simply loses their own funds to the SpokePool — there is no
 ///      residual balance outside the atomic flow for any caller to steal.
 ///
+///      Recipient requirements: `recipient` is the address any residual funds (input
+///      token or native ETH) are swept back to if the SpokePool deposit fails. It is
+///      typically an EOA. If it is a contract, it MUST be able to receive native ETH
+///      (i.e. implement a payable `receive()` / `fallback()`) for the `depositNativeEther`
+///      path — otherwise the sweep reverts and the ETH becomes trapped in this adaptor,
+///      where it is reachable by the first caller of `depositNativeEther` / `receive()`.
+///      Integrating UIs should surface this requirement to users.
+///
 ///      Bug bounty note: reports claiming that "any EOA can call these functions and
 ///      drain the contract" rely on a prefunding-across-transactions assumption that
 ///      does not occur in the production flow and are out of scope. This class is also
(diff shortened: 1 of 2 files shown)


contracts/src/l2-integration/SnowbridgeL1Adaptor.sol
```diff
@@ -7,6 +7,26 @@ import {WETH9} from "canonical-weth/WETH9.sol";
 import {ISpokePool} from "./interfaces/ISpokePool.sol";
 import {DepositParams} from "./Types.sol";
 
+/// @title L1 Across Adaptor for Snowbridge V2 outbound flows
+///
+/// @dev This contract is invoked as the `target` of a V2 `CallContract` command inside
+///      the same `Gateway.v2_submit` transaction that issues the preceding
+///      `UnlockNativeToken(recipient = L1Adaptor, amount = X)` command. The paired
+///      unlock funds this contract; the paired CallContract consumes the funds and
+///      sweeps any residual back to the BEEFY-signed `recipient`. Both commands execute
```
yrong: Maybe we should add a note about the recipient: it’s typically an EOA, but if it’s a contract, it must be able to receive ETH. Otherwise, the ETH could become trapped in the adapter and potentially be exploited. Some kind of disclaimer—perhaps surfaced in the UI—would be helpful.

yrong (approved):
+1

yrong:
Once this is merged, we should update the scope of the HackenProof reports.

There’s been some confusion—for example:
https://dashboard.hackenproof.com/manager/companies/snowbridge/snowbridge-on-chain-code/reports/SNOWBSC-454 — reporters are reviewing outdated code and raising issues based on it.

contracts/src/l2-integration/SnowbridgeL2Adaptor.sol
```diff
@@ -8,6 +8,24 @@ import {ISpokePool, IMessageHandler} from "./interfaces/ISpokePool.sol";
...
 import {DepositParams, Instructions, Call, SendParams, SwapParams} from "./Types.sol";
 
+/// @title L2 Across Adaptor for Snowbridge V2 inbound flows
+///
+/// @dev End users on the L2 call `sendTokenAndCall` and `sendEtherAndCall` directly to
+///      bridge ERC20 / native ETH back to Ethereum and onward to Polkadot. These
+///      functions pull funds from the caller in the same call — via
+///      `safeTransferFrom(msg.sender, address(this), inputAmount)` for ERC20 or
+///      `require(msg.value == inputAmount)` for native ETH — and forward the pulled
+///      amount straight to the SpokePool deposit. No function in this contract moves
```
yrong, line 18: It might also be worth adding a note about the recipient. This address is used to receive any funds that may be trapped if a call fails on mainnet. Additionally, if fees are not profitable, assets will be returned to this address on the original L2.

It’s usually an EOA, but if it’s a contract, it must be able to receive assets on both L2 and mainnet.

contracts/src/l2-integration/SnowbridgeL1Adaptor.sol
```diff
@@ -7,6 +7,34 @@ import {WETH9} from "canonical-weth/WETH9.sol";
...
```
yrong: It should be 'BEEFY-verified' rather than 'BEEFY-signed'.

claravanstaden: Addressed in https://github.com/Snowfork/snowbridge/pull/1772/commits/a70a0edc92e88fa73430c11518e8ca9ed0f20017.

PR #1787, opened 2026-05-11
Halt bridge SDK

claravanstaden:
Adds bridge halt commands to the SDK, to be used in Snowbridge app page: https://github.com/Snowfork/snowbridge-app/pull/250

### DIFF
--- a/web/packages/api/src/governance/halt_bridge.ts
+++ b/web/packages/api/src/governance/halt_bridge.ts
@@ -220,10 +220,14 @@ const MAX_U128 = (1n << 128n) - 1n
 // `DefaultBridgeHubEthereumBaseFee = Balance::MAX`). The source defaults are
 // intentionally `u128::MAX` to ship the bridge in a halted state; the live
 // values were set by past governance via `system.setStorage` and are what we
-// want to restore on resume. If the fee policy is ever updated on-chain,
-// re-query these storage keys and refresh the constants below.
-const PROD_BASE_FEE_V1 = 14_929_540_998n
-const PROD_BASE_FEE_V2 = 1_000_000_000n
+// want to restore on resume.
+//
+// Exported so callers (e.g. the snowbridge-app governance UI) can prefill
+// fee inputs and let the operator confirm or override them at submission
+// time. If the fee policy is ever updated on-chain, refresh these constants
+// from the same storage keys.
+export const PROD_BASE_FEE_V1 = 14_929_540_998n
+export const PROD_BASE_FEE_V2 = 1_000_000_000n
 
 export interface ResumeBridgeOptions {
     gateway?: boolean
@@ -239,6 +243,18 @@ export interface ResumeBridgeOptions {
     assethubBaseFee?: boolean
     assethubBaseFeeV1?: boolean
     assethubBaseFeeV2?: boolean
+    /**
+     * Override the V1 base-fee value written when the assethubBaseFee or
+     * assethubBaseFeeV1 (or `all`) levers are active. Defaults to
+     * {@link PROD_BASE_FEE_V1}.
+     */
+    baseFeeV1?: bigint
+    /**
+     * Override the V2 base-fee value written when the assethubBaseFee or
+     * assethubBaseFeeV2 (or `all`) levers are active. Defaults to
+     * {@link PROD_BASE_FEE_V2}.
+     */
+    baseFeeV2?: bigint
     all?: boolean
 }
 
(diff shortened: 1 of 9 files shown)


web/packages/api/src/governance/halt_bridge.ts
```diff
@@ -0,0 +1,243 @@
...
+
+    if (opts.outboundQueue || haltAll) {
+        bhCalls.push(bridgeHub.tx.ethereumOutboundQueue.setOperatingMode("Halted"))
+        ahCalls.push(assetHub.tx.snowbridgeSystemFrontend.setOperatingMode("Halted"))
+        summary.push("Halt V1 outbound-queue on BridgeHub + AssetHub system-frontend (router-layer block for both V1 and V2).")
+    } else if (opts.systemFrontend) {
+        ahCalls.push(assetHub.tx.snowbridgeSystemFrontend.setOperatingMode("Halted"))
+        summary.push("Halt AssetHub system-frontend only (V2 router-layer P→E block; V1 continues).")
+    }
+
```
yrong: This is incorrect, as the snowbridgeSystemFrontend pausable control applies to both the V1 and V2 routes.

web/packages/api/src/governance/index.ts
```diff
@@ -0,0 +1 @@
+export * as haltBridge from "./halt_bridge"
```
alistair-singh: This file is uneeded. You can reference halt_bridge directly from  src/index.ts

web/packages/api/src/index.ts
```diff
@@ -58,24 +58,14 @@ export * as toEthereumV2 from "./types/toEthereum"
...
-    V1ToPolkadotFeeKey,
-} from "./types/fee"
 export * as utils from "./utils"
 export * as status from "./status"
 export * as assetsV2 from "./assets_v2"
 export * as history from "./history"
 export * as historyV2 from "./history_v2"
 export { TransferStatus } from "./history_v2"
 export * as subsquidV2 from "./subsquid_v2"
+export * as governance from "./governance"
```
alistair-singh:
```suggestion
export * as governance from "./governance/halt_bridge"
```

web/packages/api/src/governance/halt_bridge.ts
claravanstaden: Good catch, you're right — `snowbridgeSystemFrontend` halts the `PausableExporter` for both V1 and V2 P→E. Fixed the summary text ("blocks both V1 and V2 P→E at the router layer; V1 BH outbound-queue keeps draining in-flight messages"). Also flowed the correction back into the snowbridge-app form: dropped `systemFrontend` from the "V2 P→E only" preset and noted that there's no true V2-only operating-mode halt for P→E — the only V2-only P→E lever is bumping `BridgeHubEthereumBaseFeeV2` (fee deterrent).

web/packages/api/src/governance/index.ts
claravanstaden: Done, deleted `governance/index.ts` and changed `src/index.ts` to `export * as governance from "./governance/halt_bridge"` per your suggestion below. Consumers now use `governance.buildHaltBridgePreimage(...)` directly.

web/packages/api/src/index.ts
claravanstaden: Applied, thanks.

yrong:
I'd suggest using simulations on forked instances of both Polkadot and Ethereum mainnet, leveraging Chopsticks and Foundry for Polkadot and Ethereum respectively.

For example, in https://polkadot.subsquare.io/referenda/1869 I use opengov-cli and [referenda-tester](https://github.com/karolk91/polkadot-referenda-tester) to verify that expected events were emitted.

claravanstaden:
@yrong how does that fit into this PR?

yrong:
> @yrong how does that fit into this PR?

Well, IMHO we currently have too many options and operating modes that can achieve the same effect. While this does provide flexibility at different levels of granularity, it also complicates governance calls. In the end, we still need to verify everything against a forked live chain anyway, right?

I’d prefer to keep things simple. For example, is disabling the intermediate hop on BH still necessary, considering we can simply disable transfers at the source?

For example:

For P→E, setting the delivery fee for V1 or V2 respectively to the maximum on AH would effectively disable transfers for each version.
For E→P, calling ethereumSystem.setOperatingMode or ethereumSystemV2.setOperatingMode would disable transfers for each version respectively.

For global emergency stops, I think it’s reasonable to have a dedicated switch, but beyond that I’d avoid adding too many overlapping controls.

claravanstaden:
@yrong I added scrips (and ran them) for chopsticks + anvil simulation of halting and resuming the full bridge. Halting and resuming work as expected.

yrong (approved):
+1

PR #1788, opened 2026-05-12
Restrict Asset Hub agent from executing contract calls in V2

yrong:


contracts/test/GatewayV2.t.sol
```diff
@@ -579,6 +579,32 @@ contract GatewayV2Test is Test {
         vm.expectEmit(true, false, false, true);
         emit IGatewayV2.InboundMessageDispatched(1, topic, true, relayerRewardAddress);
 
+        address bridgeHubAgent = IGatewayV2(address(gateway)).agentOf(Constants.BRIDGE_HUB_AGENT_ID);
```
claravanstaden, line 583: Kind of a weird test to allow BH agent to make arbitrary contract calls, since this is privileged as well, right? Even if the BH agent doesn't hold funds.

contracts/src/Functions.sol
```diff
@@ -35,6 +34,14 @@ library Functions {
         }
     }
 
+    /// Looks up an agent contract address, failing if no such mapping exists or if the agent is the primary asset hub agent
+    function ensureNonPrivilegedAgent(bytes32 agentID) internal view returns (address agent) {
```
claravanstaden:
```suggestion
    function ensureNotAssetHubAgent(bytes32 agentID) internal view returns (address agent) {
```

Since it only checks AH, not other privileged origins too.

claravanstaden (commented):
I think this okay in principle. AH is the only agent that contains token transfer funds, right?

contracts/src/Functions.sol
yrong: https://github.com/Snowfork/snowbridge/pull/1788/changes/e98e560a1ce97b918f207115d278d90705ac5b55

contracts/test/GatewayV2.t.sol
yrong: Allowing BH is expected. Yes, it’s not privilege-based — it only disables the AH agent that holds the funds.

claravanstaden:
Added some more tests here: https://github.com/Snowfork/snowbridge/pull/1793

claravanstaden (approved):

vgeddes:
I'd like us to go a step further and remove support for CallContract entirely.

alistair-singh (approved):

contracts/src/interfaces/IGatewayBase.sol
```diff
@@ -13,6 +13,7 @@ interface IGatewayBase {
     error Unsupported();
     error InvalidDestinationFee();
     error AgentDoesNotExist();
+    error UnauthorizedPrivilegedAgent();
```
vgeddes: The existing `Unauthorized` error is good enough.

contracts/src/v2/Handlers.sol
```diff
@@ -65,7 +65,7 @@ library HandlersV2 {
 
     function callContract(bytes32 origin, address executor, bytes calldata data) external {
         CallContractParams memory params = abi.decode(data, (CallContractParams));
-        address agent = Functions.ensureAgent(origin);
+        address agent = Functions.ensureNotAssetHubAgent(origin);
```
vgeddes: I prefer this, its easier to audit.
```suggestion
        if (agentID == Constants.ASSET_HUB_AGENT_ID) {
            revert IGatewayBase.Unauthorized();
        }
        address agent = Functions.ensureAgent(origin);
```

yrong: https://github.com/Snowfork/snowbridge/pull/1788/changes/cce7ab8704f91a3fdc49fdd2077542633b3a8742

vgeddes (approved):

PR #1791, opened 2026-05-14
fix(fees): P→E fee dust fixes

alistair-singh:
1. Stretch gasPad quadratic decay to k=2 — pad stays ~8% at tip==rawGas
2. Use static slippage pad for DOT→ETH swap sizing — removes tip-scaled slippage collapse
3. Add AH DOT ED (0.01 DOT) unconditionally to PayFees estimation — prevents sub-ED refund traps
4. Subtract tip slippage budget from XCM want/remoteFees — relayer absorbs tip-portion drift, raises pool tolerance to ~30%

Addresses https://github.com/Snowfork/snowbridge-app/issues/255


yrong (commented):
### review: Native-Fee Swap Paths under Volume Tips (Slippage & Dust Analysis)

While reviewing the change to use `scaledSlippagePad` for the aggregate swap on the native-fee path (under volume tips), there are two main issues regarding **Slippage Vulnerability** and **Dust Dynamics** that we should address.

---

### 1. Slippage Vulnerability (Underfunded Swap Risk)
* **The Mechanism**: Under a high volume tip ($T_{\text{ETH}} \ge 2 \times C_{\text{ETH}}$), the `scaledSlippagePad` collapses to **0%**. This means the estimated native fee ($N_{\text{total}}$) is calculated using a 0% slippage pad on the swaps.
* **The Issue**: Gas padding decays because the tip itself acts as a gas buffer on Ethereum. However, the volume tip in ETH is part of the swap target and **does not protect against exchange rate fluctuations of HDX/DOT or DOT/ETH on Asset Hub**.
* **The Impact**: If the HDX/DOT pool price moves against HDX by even a fraction of a percent during execution, the first swap will yield less DOT than expected. Because we are tight with a 0% pad, this absolute DOT deficit will cause either the subsequent DOT &rarr; ETH swap (`maximal: false`) or the DOT delivery fee payment (which is charged from the remaining DOT in holding) to fail, aborting the entire transfer.

---

### 2. Dust Dynamics & Refund Traps (Positive Slippage Risk)
* **The Mechanism**: PR #1791 introduces `AH_DOT_ED` (0.01 DOT) to oversized `PayFees` estimation. This successfully prevents sub-ED refund traps arising from **PayFees overpayment** by guaranteeing that `RefundSurplus` returns $\ge \text{ED}$ to holding.
* **The Issue**: `AH_DOT_ED` does not directly pad the `exchangeAsset` DOT budget. The positive-slippage exchange dust path is a pre-existing pattern wherever `exchangeAsset` + appendix `depositAsset` is used. However, by collapsing the aggregate swap pad to 0% under volume tips, the PR **increases exposure** to this gap. 
* **The Impact**: If a favorable swap or minor positive price movement leaves a small leftover (e.g., $0.005$ DOT) in holding, the XCM appendix `depositAsset` step (which deposits to the source account on the parachain or Asset Hub) will attempt to deposit a sub-minimum amount. This will fail with a `Token::BelowMinimum` (or chain-specific ED) error, aborting the entire XCM transaction.

---

### Recommendations

1. **Keep the Aggregate Swap Pad Robust**:
   Do not use the tip-scaled slippage pad for the aggregate Native &rarr; DOT swap in `toEthereumSnowbridgeV2.ts`. Keep it at the static 20% pad (`feeSlippagePadPercentage`), or enforce a minimum floor (e.g., 10–20%) so slippage protection never collapses to 0%:
   ```diff
               totalFeeInNative = await assetHubImpl.getAssetHubConversionPalletSwap(
                   feeLocation,
                   DOT_LOCATION,
   -               padFeeByPercentage(totalFeeInDot, scaledSlippagePad),
   +               padFeeByPercentage(totalFeeInDot, feeSlippagePadPercentage),
               )
   ```

2. **Clean up the Dust Refund Leg**:
   To fully resolve the exchange-slippage dust trap (mirroring E &rarr; P ether handling), we should:
   - Ensure the DOT side of the swap budget also carries an ED floor, or
   - Implement a split deposit pattern (similar to `buildSplitDepositAsset` for Ether on E &rarr; P) so a failing DOT dust deposit leg does not block or fail the rest of the XCM execution.

PR #1801, opened 2026-06-28
Fix: filter unsupported Hydration assets

yrong:
Only include Hydration assets with an explicit Unlocked circuit breaker state, preventing locked assets like WETH from appearing in generated routes.


web/packages/api/src/parachains/hydration.ts
```diff
@@ -40,6 +40,12 @@ export class HydrationParachain extends ParachainBase {
             }
 
             const assetId = Number(id.args[0]?.toString())
+            const lockdownState: any =
+                await this.provider.query.circuitBreaker.assetLockdownState(assetId)
+            if (!lockdownState.isSome || !lockdownState.unwrap().isUnlocked) {
+                continue
+            }
+
```
claravanstaden, line 48: What is this circuit breaker now?

yrong: It's Hydration's `pallet-circuit-breaker`. `assetLockdownState` tracks whether an asset is locked down (a runtime safety mechanism that restricts transfers/trading for that asset). We only include assets that are explicitly present and `Unlocked`, so locked-down tokens like WETH don't show up in generated routes where transfers would fail.

PR #1809, opened 2026-08-26
Adds service fee

claravanstaden:
Removes the volume relayer tip, and deposits the fee into a specified recipient account. Removes the scaled fee padding.


alistair-singh (approved):
Looks good, there are some public API surface breaking changes, nothing that people actually use. Just update the minor version in the package.json.

